PR #2, opened 2015-06-30
Fix incompatibility with <GL/glext.h>

nwnk:
New extensions in glext.h follow the pattern:

    #ifndef GL_ARB_texture_rectangle
    #define GL_ARB_texture_rectangle 1
    #define GL_TEXTURE_RECTANGLE_ARB          0x84F5
    #define GL_TEXTURE_BINDING_RECTANGLE_ARB  0x84F6
    #define GL_PROXY_TEXTURE_RECTANGLE_ARB    0x84F7
    #define GL_MAX_RECTANGLE_TEXTURE_SIZE_ARB 0x84F8
    #endif /* GL_ARB_texture_rectangle */

Versions.h tries to declare:

    const char* const GL_ARB_texture_rectangle = "GL_ARB_texture_rectangle";

Which means, if you've included glext.h before Versions.h, that the
compiler will see "const char* const 1 = ...", and rightly refuse to
continue.

The ham-fisted approach taken here is to rename the variables in
Versions.h with a leading underscore.  This does sort of undermine the
comment about "better to have the compiler do spelling checks", but.

Signed-off-by: Adam Jackson <ajax@redhat.com>

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -2856,7 +2856,7 @@ void IdentifyBuiltIns(int version, EProfile profile, EShLanguage language, TSymb
         BuiltInVariable("gl_TessCoord",       EbvTessCoord,      symbolTable);
 
         if (version < 410)
-            symbolTable.setVariableExtensions("gl_ViewportIndex", 1, &GL_ARB_viewport_array);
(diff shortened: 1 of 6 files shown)


johnkslang:
I definitely see the need here, and such issues are typically handled by "namespace", which happens to not work with preexisting macros. I'm not a fan of leading underscores, and wonder if there are any other alternatives to consider.  Worse case is maybe to use an "E_" prefix, or a "_s" suffix, S meaning "string".

PR #4, opened 2015-07-08
include by relative path, so root doesn't need to be in the search path

baldurk:
I'm not sure if this is desired, but I've made some changes so that glslang doesn't depend any directories being in the include search path (which is default for building the project, but isn't so desirable when including glslang in a larger project).

---

On a related but separate note - the same include path requirement is true for osinclude.h. The include search path needs to be set up to point to either glslang/OSDependent/Linux or glslang/OSDependent/Windows on their respective platforms. I've [changed this locally](https://github.com/baldurk/glslang/commit/c0384798d86774f6ae7c7227ef8a3e25aa9fd4ef) so that osinclude.h is included relatively like so:

```cpp
#include "../glslang/OSDependent/osinclude.h"
```

and that file just contains a simple redirect based on platform macros:

```cpp
#if defined(WIN32)
#include "Windows/osinclude.h"
#elif defined(__linux__)
#include "Linux/osinclude.h"
#else
#error "Unknown platform"
#endif
```

If that's desirable to have upstream then let me know and I can make a pull request, but otherwise I'll just keep the change locally.

### DIFF
--- a/OGLCompilersDLL/InitializeDll.cpp
+++ b/OGLCompilersDLL/InitializeDll.cpp
@@ -37,9 +37,9 @@
 #include <assert.h>
 
 #include "InitializeDll.h"
-#include "Include/InitializeGlobals.h"
+#include "../glslang/Include/InitializeGlobals.h"
 
-#include "Public/ShaderLang.h"
+#include "../glslang/Public/ShaderLang.h"
 
 namespace glslang {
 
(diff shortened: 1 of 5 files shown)


johnkslang:
I think simplifying the integration into other projects is worthwhile.  (Wish the file structure was less archaic, but it's working and changing it would break some things.)  Feel free to make a pull request for the osinclude.h improvement.

Thanks

PR #10, opened 2015-07-12
common osinclude.h header file that redirects per-platform

baldurk:
As I mentioned, this now makes files that need to include osinclude.h include it from a common root, that then checks platform macros to include the right one. I didn't see any existing platform macros (and I don't know if CMake defines any by default), so I just used WIN32 & __linux__ that should come from the compiler. If you'd like to define glslang-specific macros in CMakeLists.txt somewhere I can update the PR.

I also changed bison to output to the source folder instead of to the build folder, so that relative includes will work and the include_directories lines can be removed from CMakeLists.txt. I think this makes sense but if you want to keep bison output in the build folder that's manageable without losing the rest of the change.


glslang/OSDependent/osinclude.h
```diff
@@ -0,0 +1,46 @@
...
+//ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+//POSSIBILITY OF SUCH DAMAGE.
+//
+
+#ifndef __ROOT_OSINCLUDE_H
+#define __ROOT_OSINCLUDE_H
+
+#if defined(WIN32)
+#include "Windows/osinclude.h"
+#elif defined(__linux__)
```
fkaa: Fails on OS X (`__APPLE__`)

baldurk: Good point, I hadn't thought of that. Would it make sense then to rename the Linux directory/platform to Unix? and use glslang-specific platform macros defined in CMakeLists.txt to identify the platforms, instead of the compiler built-in macros.

fkaa: I'd defer it to @johnkslang. Not sure how things should be structured, but what you describe sounds good.

baldurk:
@johnkslang what's your verdict on how to support OS X here? I can fix it whichever way is preferred, since as-is this PR isn't going to work.

xorgy:
@baldurk: My intuition is that for something self-contained like glslang, os-specific code will come down to "NT or UNIX". I don't think we need to specifically target OSX and Linux, I'm sure it would build fine on other platforms (OpenBSD, VXWorks or something).

baldurk:
Sure! I meant that at the moment people are building on OS X and we don't want to regress, it applies to any other platform. I'd agree that it seems like the nicest fix is to rename the linux/ directory to a catch-all unix/ and have matching macro defines. I'd still prefer having an explicit case per-os and an #else that's an error, otherwise it's really nasty for potential future porting:

```cpp
#if OS_A
// ...
#elif OS_B
// ...
#else
#error "Unknown"
#endif
```

baldurk:
I've rebased this branch against master and I also took the liberty of applying what seemed to be the consensus for a reasonable solution for the problems noted above - I renamed the Linux/ subdirectory to Unix/ and used CMake macros to determine which one to build, which should work for OS X now.

xorgy:
@baldurk Cool, one thing I'm noticing though is that these files generated by the build process are missing from the `.gitignore` file, so they dirty the tree (so you can't just do a `git add .`).

```
glslang/MachineIndependent/glslang_tab.cpp
glslang/MachineIndependent/glslang_tab.cpp.h
```

baldurk:
That's a fair point, I didn't really think about this since I have those files committed on my local fork. I moved them there since it was the only way to make all of the includes truly relative, otherwise it would have to at least rely on some kind of define provided by CMake which is ugly for embedding glslang in other programs as a library.

I can update the .gitignore and/or move the files elsewhere, but I'll leave that to @johnkslang to make a call. I was also very tempted to actually remove the old build system files since they're now kind of broken, but I didn't want to mess up the PR with that. Again I can do that here though if @johnkslang prefers.

johnkslang:
I agree the old build system should be removed, feel free to include that.

If general consensus is to make the OS-dependency changes you proposed, I'm fine with merging them in.  Let's just give a couple days here for feedback.

Thanks!

phaulos:
Unfortunately this broke our integration of glslang that relied on custom osinclude.h. This PR would fix it: https://github.com/KhronosGroup/glslang/pull/104

PR #15, opened 2015-07-14
Preprocessor error output

antiagainst:
This patch distinguishes preprocessing errors with normal parsing
errors and gives `glslangValidator` the ability to output preprocessing
errors.


AWoloszyn:
Sorry about losing the conversation history. I squashed a second commit into this one, and the conversation got lost. I try and point out the changes.

PR #19, opened 2015-07-16
Generate correctly structured do-while loops.

dneto0:
The loop test is always emitted before the loop body.

For do-while loops, use a phi node to track whether we're
on the first loop iteration, and only check the loop test
on the second and subsequent iterations.


Test/baseResults/spv.do-simple.vert.out
```diff
@@ -13,45 +13,49 @@ Linked vertex stage:
...
-              22:               Label
-                                Branch 11
+                              Branch 16
+              16:             Label
+                              SelectionMerge 17 None
+                              BranchConditional 13 17 18 
+              18:               Label
+              19:      7(int)   Load 9(i) 
+              21:    14(bool)   SLessThan 19 20
+                                BranchConditional 21 17 12
```
dneto0: Does this conditional branch require a SelectionMerge before it?  If it does, it would target block 12 as the merge point, I think. 
If that fix is required, I could provide it in another pull request

johnkslang: Yes (just starting to look at this now), all shader selections need a merge.

johnkslang:
"Any structured control-flow structure must have ... its own unique merge block identified"

johnkslang:
.... "(its merge block cannot appear in any other merge instruction for any other control-flow structure)"

dneto0:
Ok.  I'll rework the patch to meet those constraints.

dneto0:
Closing this pull request.  Need to rebase first.

Issue #37, opened 2015-07-23
C-style #include and #line directive support in glslang


antiagainst:
We plan to implement the `#include` and `#line` directive support via GLSL extensions, so the default glslang behaviour remains as today. We believe it will benefit the community to put the extension implementation into glslang.

Specifically, we plan to introduce two extensions: `GL_GOOGLE_cpp_style_line_directive` and `GL_GOOGLE_include_directive`. The first one allows `#line` to accept a constant string (for a filename) as its second parameter and substitute `__FILE__` accordingly. The second one enables `#include` support.

There will be a couple of commits for this issue, falling into the following three categories:

1. preliminary work to make the preprocessor handle line numbers correctly
2. add support for `GL_GOOGLE_cpp_style_line_direcitve`
3. add support for `GL_GOOGLE_include_directive`

xorgy:
`#include` seems like a wonderful addition, especially with the SPIR-V codegen use case.

baldurk:
Sounds good to me, both of these will definitely be very useful for glslang. Though can't [GL_ARB_shading_language_include] (https://www.opengl.org/registry/specs/ARB/shading_language_include.txt) be used for the #include support part of this? I imagine it will still need glslang support to pass the named strings along though.

dmircevski:
@baldurk: GL_GOOGLE_include_directive will be similar to GL_ARB_shading_language_include regarding #include in GLSL.  But it will be much more file-oriented, in contrast to GL_ARB_shading_language_include, which focuses primarily on shader strings (not files) and consequently extends the OpenGL API in ways that are out of scope for us.

baldurk:
That makes sense, I just figured I'd ask :).

fkaa:
@dmircevski any info on how the searching will work? I think the c++ standard leaves it up to the implementer entirely how it's interpreted.

dmircevski:
@fkaa our thinking right now is that there will be an implementation-defined search path, and the #include filename will be interpreted relative to that search path (plus relative to the currently parsed file if the filename is in quotes rather than angle brackets).  This is identical to the C standard, I believe.

The actual implementation in glslang will use a flexible mechanism to allow a wide variety of possible implementations by other contributors.  Watch the upcoming commits for details.

johnkslang:
Is there still a part of this to keep as an issue?

antiagainst:
Sorry forgot to close this. We already have the necessary bits in. Thanks!

PR #43, opened 2015-07-25
Extend the syntax of #line and __FILE__ to support filename strings

antiagainst:
According to the GLSL spec, the second parameter to `#line` should be
an integer source string number and `__FILE__` will be substituted
with the integer source string number currently processed. This
patch extends the syntax of `#line` and `__FILE__`. Now `#line` accepts
as the second parameter a filename string quoted by double quotation
marks. And if such a `#line` is set, `__FILE__` will be substituted with
the currently set filename string. The implementation is done via
introducing a new extension `GL_GOOGLE_cpp_style_line_directive` using
the extension framework.

The purpose is to support `cpp`-style `#line` directives, which is
required by `#include`.

This commit falls into the second step listed in Issue #37.

### DIFF
--- a/Test/baseResults/preprocessor.cpp_style___FILE__.vert.out
+++ b/Test/baseResults/preprocessor.cpp_style___FILE__.vert.out
@@ -1,26 +1,37 @@
 #extension GL_GOOGLE_cpp_style_line_directive : enable
 
 0
-#line 150 a.h
+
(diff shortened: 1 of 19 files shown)


antiagainst:
Of course. On it.

antiagainst:
Added two more commits to address the issues.

PR #46, opened 2015-08-02
#include directive support

antiagainst:
These commits add `#include` directive support in glslang. We
use a flexible mechanism so that different contributors can
have different implementation for file searching and `#include`
directive substitution. (A `ForbidInclude` is supplied though.)

Please refer to individual commit messages for more details.
This fulfills the third step mentioned in Issue #37.


johnkslang:
Cool stuff.

One simple concern:  It seems easy to accidentally not use the preprocessor version of requireExtensions() et. al.  There is now ppError() and error().  Would it be safer if there was also ppRequireExtensions() and requireExtensions().  Seems that would be better, due to consistency and clarify, assuming it could be done without too much code duplication of the internals of requireExtensions().

Thoughts?

Thanks.

antiagainst:
@johnkslang: Yeah, agreed that a separate `ppRequireExtensions` is more coherent with the rest of the code. I've pushed out a new commit for that.

PR #52, opened 2015-08-05
Compile fixes for gcc -Wall

baldurk:
Some of these fixes are safe & sensible like #ifdef'ing out windows only #pragmas, re-ordering constructor arguments.

The casts between unsigned and signed ints are less clear-cut, so this pull request isn't necessarily to take as-is if you don't want to bandage over things, but to raise these as issues. If there's a refactor that would be better let me know and I can look into doing it, or I can drop them from this PR if you'd rather take just the other fixes.

I also removed an unused variable (fragOutHasLocation) but that's maybe a warning you'd want to ignore.


xorgy:
I like the way it looks over here. Bandage or not, this is probably a better starting point for post-bandage changes. (Especially WRT that int where there should've been a size_t!)

PR #54, opened 2015-08-11
SPIR-V: Return undefined values from implicit returns instead of dummy

mknejp:
Previously if a non-void function implictly returned, a dummy variable
was created as return value. Now instead it returns the result of the
OpUndef instruction. This better conveys the presence of undefined
behavior to SPIR-V consuming tools (and humans).

It also saves one ID per occurrence...


johnkslang:
Looks good to me.

PR #66, opened 2015-09-02
cmake: Don't override CMAKE_INSTALL_PREFIX

chadversary:
Stop forcing CMAKE_INSTALL_PREFIX="install". If the user manually set
CMAKE_INSTALL_PREFIX, then trust that he knows what he's doing.

This patch does NOT change the project's default value ("install") of
CMAKE_INSTALL_PREFIX.

Change-Id: I81b46dd1986427b498fe6316bed03f01689987d4


johnkslang:
"Forcing" was unintentional, pull request accepted.

PR #68, opened 2015-09-05
SPIRV: Decorate matrices and arrays with their strides

jekstrand:


SPIRV/GlslangToSpv.cpp
```diff
@@ -1459,11 +1465,29 @@ spv::Id TGlslangToSpvTraverser::convertGlslangToSpvType(const glslang::TType& ty
...
     }
 
     return spvType;
 }
 
+// Given an array type, returns the integer stride required for that array
+int TGlslangToSpvTraverser::getArrayStride(const glslang::TType& arrayType)
+{
+    glslang::TType derefType(arrayType, 0);
+    int size;
```
xorgy: It's generally good style to place declarations at the top of the block.

johnkslang:
@jekstrand, looks good, except perhaps missing test result updates.  I'll look at adding those myself.

johnkslang:
Added test changes: 358dd560c516b2f71ae3c855dbd3d3f83294a18c.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1459,11 +1465,29 @@ spv::Id TGlslangToSpvTraverser::convertGlslangToSpvType(const glslang::TType& ty
         } else
             arraySize = type.getOuterArraySize();
         spvType = builder.makeArrayType(spvType, arraySize);
+        builder.addDecoration(spvType, spv::DecorationArrayStride, getArrayStride(type));
```
johnkslang: This code is executed for all arrays.  However, the decoration is only required on transparent (e.g., uniform block) arrays, and indeed not even well-defined for an array of opaque types (e.g., array of samplers).  I'm getting asserts on it, likely due to the latter.  Am looking at fixing.

johnkslang: Fixed with 31ed4830d15d9bd019869b5e04a053a665952f96.

Issue #67, opened 2015-09-05
[SPIR-V] texelFetch not yet implemented


jekstrand:
This seems like a pretty glaring omission.  I took a look at the code and it doesn't look like this should take much time for someone who knows what they're doing with it.  However, my knowledge of the glslang code-base is slim at best and I wasn't seeing exactly how it should be added.

johnkslang:
@jekstrand There are plenty of omissions, it's work in progress.  Thanks for the submission on this one, closing.

PR #69, opened 2015-09-05
SPIRV: Add support for texelFetch

jekstrand:
There's nothing really special about texelFetch other than the opcode and
some restrictions so adding it is pretty trivial.

This resolves issue #67


jekstrand:
This resolves issue #67

johnkslang:
@jekstrand, right the infrastructure was done so these could be added.  Probably a few dozen things to do now in the translator that are really nothing special, just have to get done.  Thanks for doing this one.

johnkslang:
FYI, I'm updated the testing section in the README.md; test changes should be part of any submission that changes functionality.

PR #71, opened 2015-09-06
SPIRV: Refine implementation of OpAtomicXXX instructions

amdrexu:


johnkslang:
@amdrexu, thanks.  I have two requests:
1) update to no longer have conflicts
2) include test changes in the request (see Testing section in the README.md)
Thanks

amdrexu:
@johnkslang Thanks very much for your valuable feedbacks. I will correct the issues, verify test cases, and send you the updated codes soon.

amdrexu:
@johnkslang Hi, I have updated the codes and there are only two effective commits: 04db3f5 and 6b86d49. The others are code sync-ing. In those two changes, I fix several issues relevant to my previous changes and those newly found during the development (e.g. texelFetch does not generate "lod" operand correctly). The image functions are fully supported. Besides, I add a new test "spv.image.frag" to cover all image functions mentioned in GLSL spec. I change some test results in /baseResults/ that are unrelated with my change. It is very straneg that the operand IDs generated in my local tests differ from those in gold outputs slightly. But the logic is correct.

Please forgive me to create such messy commit history. I know "rebase" is powerful to address my problem. But I need some time to get familiar and have a try. As a novice in using github, I am glad to learn any helpful technique in GIT use. Feel free to give me advice and suggestions that you believe are important and useful. Of course, I will create small changes in the future to avoid such cumbersome pull request.

Thanks you much.

SPIRV/GlslangToSpv.cpp
```diff
@@ -702,7 +754,15 @@ bool TGlslangToSpvTraverser::visitUnary(glslang::TVisit /* visit */, glslang::TI
...
-    spv::Id operand = builder.accessChainLoad(convertGlslangToSpvType(node->getOperand()->getType()));
+
+    spv::Id operand = spv::NoResult;
+
+    if (node->getOp() == glslang::EOpAtomicCounterIncrement ||
+        node->getOp() == glslang::EOpAtomicCounterDecrement ||
+        node->getOp() == glslang::EOpAtomicCounter)
+        operand = builder.accessChainGetLValue(); // Special case l-value operands
+    else
+        operand = builder.accessChainLoad(convertGlslangToSpvType(node->getOperand()->getType()));
```
johnkslang: This change assumes the atomic counter is the actual address of memory, rather than a handle.  I'm not sure that's true.  If it's a handle, you have to load it and pass what you loaded.  Compare this to texture access:  The image is not an address of where the image memory is, the image is an opaque handle.  So, we load the image, not use it's address.  It's address the address of the handle information.

johnkslang:
> I change some test results in /baseResults/ that are unrelated with my change. It is very strange that the operand IDs generated in my local tests differ from those in gold outputs slightly.

Yes, there seems to be a bug introduced by the do-while OpPhi code, where it depends if I compile as debug vs. release.  I think this code will be updated soon.

glslang/Include/intermediate.h
```diff
@@ -639,7 +639,7 @@ class TIntermOperator : public TIntermTyped {
     bool isImage()   const { return op > EOpImageGuardBegin   && op < EOpImageGuardEnd; }
 
     // Crack the op into the individual dimensions of texturing operation.
-    void crackTexture(TCrackedTextureOp& cracked) const
+    void crackTexture(TSampler sampler, TCrackedTextureOp& cracked) const
```
johnkslang: This is very specifically about decomposing the meaning of the *op*, not about logical of cross-combining the op with the type, which should happen outside the method.

amdrexu: Yes, I understand your intention. We could modify the result of "cracked" outside this method. If you think it is better and want to keep the conciseness of "crackTexture", I will modify the codes. It is OK.

johnkslang: I thought I deleted that note.  I think it is okay as is. Thanks.

SPIRV/GlslangToSpv.cpp
amdrexu: Thanks for your reminder. I indeed thought about this problem. It occurs when those opaque types act as function parameters. I will reconsider this change.

I have one question: how to understand the OpLoad (value <- *ptr) instruction. For numeric type, it could be interpreted as "load the value from the memory that the pointer points to". For opaque type, how to interpret it? What "value" stands for?

Test/baseResults/spv.image.frag.out
```diff
@@ -0,0 +1,497 @@
...
+             210:      6(int) Load 133(ic1D)
+             211:  126(fvec4) Load 128(v)
+                              ImageWrite 209 210 211
+             212:          95 Load 97(i2DMS)
+             213:   27(ivec2) Load 143(ic2D)
+             215:  126(fvec4) ImageRead 212 213 214
+             216:  126(fvec4) Load 128(v)
+             217:  126(fvec4) FAdd 216 215
+                              Store 128(v) 217
+             218:          95 Load 97(i2DMS)
```
amdrexu: Hi John,

2DMS image load/store should be comment-off in the test file "spv.image.frag" since it is in TODO list currently :)

PR #73, opened 2015-09-13
Fix reference to glslangValidator in runtests

xorgy:


dneto0:
Our Shaderc project builds around glslang.  We wanted to build the code and run the tests separately from the source tree, so we have a step in our build which clones the test source files and massages the EXE line in runtests.

See starting at line 23 of https://github.com/google/shaderc/blob/master/third_party/CMakeLists.txt and the corresponding script to do most of the work at, https://github.com/google/shaderc/blob/master/utils/copy-tests-if-necessary.py

Feel free to clone and adapt these techniques for your project.

PR #75, opened 2015-09-18
Fixed subtle issue that causes tests to fail in VS2013 in some configs.

AWoloszyn:
Depending on specific optimization settings VS2013 will sometimes
execute the operands to

new Instruction(builder.getUniqueId(), builder.makeBoolType(), OpPhi)

left-to-right, and sometimes right-to-left. Since makeBoolType can
also call getUniqueId(), the IDs to the OpPhi can sometimes be swapped.

This guarantees an explicit ordering of the Ids so that tests work
reliably.


johnkslang:
Awesome.  Been wanting to eliminate this non-determinism.

PR #76, opened 2015-09-24
Implement GL_ARB_shader_draw_parameters

amdrexu:
Support the parse of three builtins introduced by this extension. Also, modify spv.image.frag. It is unrelated with this change. Just make a trivial change to this file.


amdrexu:
@johnkslang I also have a question:

In the file "Intermediate.cpp", function "TIntermediate::addSymbolLinkageNodes", there is a code piece:

if (language == EShLangVertex) {
    addSymbolLikageNode(...)
}

For "gl_VertexID" and "gl_InstanceID", they are added. I am not sure if the three new builtins "gl_BaseVertexARB", "gl_BaseInstanceARB", "gl_DrawIDARB" should be added as well. I am not clear about the function of this code piece though I read it several times and try to comprehend its underlying intention. It is highly appreciated that you could make some explanations. Thank you.

johnkslang:
Let's start by not adding any more to the 

```
if (language == EShLangVertex) {
    addSymbolLikageNode(...)
}
```

I don't fully agree with that code, but it had a need by some consumer.  If the new variables don't trigger a similar need, that is an indication this code could go away, rather than be expanded.

johnkslang:
It seems something discussion got dropped.  I added another line comment today.  Primarily, this looks good, except it is adding new qualifiers, which we don't want to do.  Previous discussion pointed to the comments in the code that explained that.

Was there an update to the submission, and that dropped the previous line comments?

amdrexu:
@johnkslang I have merged the changes into a commit. This will be easy for you to review. Sorry for the previous messy codes since I do not use "rebase" to merge them. Thanks very much for your guidance. I learn a lot about the use of GIT.

Test/spv.image.frag
```diff
@@ -61,10 +61,10 @@ void main()
     imageStore(i2DArray, ic3D, v);
     v += imageLoad(iBuffer, ic1D);
     imageStore(iBuffer, ic1D, v);
-    v += imageLoad(i2DMS, ic2D, 1);
-    imageStore(i2DMS, ic2D, 2, v);
-    v += imageLoad(i2DMSArray, ic3D, 3);
-    imageStore(i2DMSArray, ic3D, 4, v);
+    // v += imageLoad(i2DMS, ic2D, 1);
```
xorgy: if you're commenting it, would it not be better to simply remove it altogether?

xorgy:
It may be helpful to review it with whitespace changes turned off in the diff, it makes for a clearer view of what has changed. https://github.com/KhronosGroup/glslang/pull/76/files?w=1

Test/spv.image.frag
johnkslang: It will need to be uncommented in the future.  It could stay uncommented, and continue to show and changes, even if it's not completely correct.  Given that it could go either way, and it's not related to the main point, I think this part could be skipped for this merge.

johnkslang:
Since the actual change is simple, and the discussion getting complex, I did the subset that made sense, and fixed some existing comments to be more clear.  See da581a2b955df1d61deadda30459474aa3a1ea07.  Thanks for supplying the code @amdrexu.

If there is something missing, let's deal with that in a fresh request dealing with that single issue.  We can use the next one to use more formal procedure.  Thanks.

Test/spv.image.frag
amdrexu: OK. I understand.

Issue #77, opened 2015-10-05
[SPIRV] textureLodOffset generates [non-const]Offset modifier instead of ConstOffset.


Themaister:
Instead of 0x08 ConstOffset modifier, 0x10 Offset is emitted, which is incorrect since textureLodOffset must have a constant expression offset.

johnkslang:
Fixed by d4a7a8e7212e2fc07aa4b447eeb3d27762bb4a58.

PR #81, opened 2015-10-11
Generate correct image operand mask for Offset and ConstOffset(#77)

amdrexu:


johnkslang:
Good catches, thanks.

PR #85, opened 2015-10-26
Various build fixes when using newer GCC versions with warnings enabled

mgadams:
Encountered with GCC-4.7.3 in a build environment where warnings are enabled
and treated as errors.


xorgy:
:+1:

PR #88, opened 2015-10-27
Remove obsolete unistd.h file

mgadams:
The presence of this file can cause build issues when integrating glslang
into other projects.

Note that the VS build projects were hand edited due to tool availability. However, given they've been deprecated, that shouldn't be an issue.


xorgy:
:+1:

PR #100, opened 2015-11-11
Scanner optimisations

mawww:
Hello,

Here are two commits that optimize some slow parts of the scanner, based on some profiling I did on our use case.

Cheers,

Maxime.

### DIFF
--- a/glslang/MachineIndependent/Scan.cpp
+++ b/glslang/MachineIndependent/Scan.cpp
@@ -293,10 +293,33 @@ int yylex(YYSTYPE* glslangTokenDesc, glslang::TParseContext& parseContext)
 
 namespace {
 
+struct str_eq
+{
+    bool operator()(const char* lhs, const char* rhs) const
+    {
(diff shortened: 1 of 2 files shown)


xorgy:
:+1: Could you share some of the details of the profiling?

mawww:
Cannot be specific, but on our set of shaders, with our integration, tInputString::scan went from 16% CPU time to around 7% with the second commit. Note that was not on an optimized build, in our case keeping fast shader compilation on debug builds is important.

johnkslang:
FYI, I'm planning to measure this, and if it looks good commit it.

johnkslang:
My measurements show a statistically significant 6% improvement, non-debug, on a couple of workloads.

From previous profiling, I think a small change in preprocessing can save a whole set of tokens from being put into the PP symbol table, which would be a bigger change.  Haven't had time to follow through on that yet (looked like all shader tokens were added, when only PP-related tokens needed to be).

PR #104, opened 2015-11-30
Do not use path when including osinclude.h

phaulos:
This allows using custom osinclude.h if needed.


baldurk:
I don't want to enforce my use-case on anyone else which is why I put PR #10 up for discussion expecting that it might be shot down. I can always keep those changes downstream on my branch if needs be. The reasoning behind that PR though was to remove the need for any glslang-specific include paths, which makes it much easier and cleaner to embed glslang in an existing source tree.

I wonder if there's a possible middle ground. Presumably you have downstream changes to add the new osinclude file somewhere else in the source tree - what about adding a new cmake define GLSLANG_OSINCLUDE_CUSTOM that is checked before the other two in the current osinclude.h like so:

```cpp
#if defined(GLSLANG_OSINCLUDE_CUSTOM)
#include "Custom/osinclude.h"
#elif defined(GLSLANG_OSINCLUDE_WIN32)
#include "Windows/osinclude.h"
#elif defined(GLSLANG_OSINCLUDE_UNIX)
#include "Unix/osinclude.h"
#else
#error "Unknown platform"
#endif
```

or even maybe `#include GLSLANG_OSINCLUDE_CUSTOM` if that works. Is that possible on your end?

phaulos:
What could work in our setup is to have an override to include "custom_osinclude.h" or similar. But that starts to feel quite arbitrary.

Since the build setup already has to decide between Windows/ossource.cpp and Unix/ossource.cpp would it make sense to just get rid of the common osinclude.h and set up right include path as well? Shouldn't add any complexity to the build since you have to have the if (WIN32) etc in place for choosing source already.

Another option would be to make osinclude.h completely platform-independent (use void* as handle types etc.) and then simply choose which implementation to compile+link in.

Which one would you prefer? Other options?

baldurk:
The include path could be set up in CMake the same way as ossource.cpp is included, but that would still mean a glslang-specific include path which I wanted to avoid for embedding in a project. I set up the common osinclude.h with the redirect to a subfolder so that everywhere else in the code could just use a single relative path.

I would definitely be happy with making osinclude.h platform-independent. It seems like right now that would be relatively straightforward to do, but I don't know if that would clash with future development plans.

I'm not in any way in a position to make any calls though, that would be up to @johnkslang. I'll certainly argue the use-case that suits me - ie. relative includes - but I'm just one of many votes :).

mknejp:
I always find it questionable when someone includes files without specifying a path in a header, especially if it has such a generic name as `osinclude.h`. That can potentially pull in a file from anywhere in your compiler's search paths and can screw things up big time (speaking from experience).

phaulos:
@baldurk With the current setup you still need a glslang-specific preprocessor flag (-DGLSLANG_OSINCLUDE_..). I would assume that with any build system you can limit include paths or other preprocessor flags to just specific source file(s) or build targets. Would that provide sufficient isolation?

phaulos:
Generic osinclude.h proposal up here: https://github.com/KhronosGroup/glslang/pull/105

johnkslang:
I pulled #105, fixed (next commit) a path problem with it, and it seems to work for me now on Linux and Windows.  Let's deal with other issues as incremental from that base.

PR #105, opened 2015-12-01
Use generic osinclude.h

phaulos:
Using platform-neutral osinclude.h makes it easier to substitute
implementation when necessary and eliminates some variability between
build configurations.


johnkslang:
I merged this, but it seems to have some problems with CMake, when regenerating.  Am working on fixing...

PR #106, opened 2015-12-02
Fix several issues and update corresponding test cases

amdrexu:
- There is one unnecessary operand generated for FrexpStruct. 
- Extension requirement for gl_PrimitiveID and gl_Layer is unnecessary for OpenGL.
- The related operator for unpackDouble2x32/packDouble2x32 is mistaken.
- Type promotion for <<, <<=, >>, >>= should be done in certain cases. Otherwise, the scalar smear operation will be wrong: u3(uvec3) << 2.
- std430 should be applied to buffer storage block.
- Array stride calculation for std140 uniform block is incorrect. And layout packing info is not involved in this calculation: layout(std140) uniform DATA { float f1[9]; }.
- Entry-point interfaces checking should be applied to non-constants: const bool cond = false (global constant).


xorgy:
Interested to see how high `n` gets.

SPIRV/GlslangToSpv.cpp
```diff
@@ -598,10 +598,12 @@ void TGlslangToSpvTraverser::visitSymbol(glslang::TIntermSymbol* symbol)
         else
             builder.setAccessChainLValue(id);
     } else {
-        // finish off the entry-point SPV instruction by adding the Input/Output <id>
-        spv::StorageClass sc = builder.getStorageClass(id);
-        if (sc == spv::StorageClassInput || sc == spv::StorageClassOutput)
-            entryPoint->addIdOperand(id);
+        if (symbol->getQualifier().storage != glslang::EvqConst) {
```
johnkslang: There are no test results that change because of this.  Is there an example of something that is Input/Output in SPIR-V but Const in glslang, and should not be included on the OpEntryPoint instruction?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5187,7 +5187,7 @@ void TParseContext::blockStageIoCheck(const TSourceLoc& loc, const TQualifier& q
         profileRequires(loc, EEsProfile, 300, nullptr, "uniform block");
         profileRequires(loc, ENoProfile, 140, nullptr, "uniform block");
         if (currentBlockQualifier.layoutPacking == ElpStd430)
-            requireProfile(loc, ~EEsProfile, "std430 on a uniform block");
+            error(loc, "requires buffer storage qualifier", "std430", "");
```
johnkslang: Good fix.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1630,9 +1632,11 @@ bool TGlslangToSpvTraverser::requiresExplicitLayout(const glslang::TType& type)
 // Given an array type, returns the integer stride required for that array
 int TGlslangToSpvTraverser::getArrayStride(const glslang::TType& arrayType)
 {
-    glslang::TType derefType(arrayType, 0);
+    assert(arrayType.getQualifier().layoutPacking == glslang::ElpStd140 ||
```
johnkslang: The policy for when this is used belongs in requiresExplicitLayout().

SPIRV/GlslangToSpv.cpp
```diff
@@ -1630,9 +1632,11 @@ bool TGlslangToSpvTraverser::requiresExplicitLayout(const glslang::TType& type)
...
+           arrayType.getQualifier().layoutPacking == glslang::ElpStd430);
     int size;
-    glslangIntermediate->getBaseAlignment(derefType, size, true);
+    glslangIntermediate->getBaseAlignment(arrayType, size, arrayType.getQualifier().layoutPacking == glslang::ElpStd140);
+    size /= arrayType.getOuterArraySize();
```
johnkslang: The specification says "If the member is an array of scalars or vectors, the base alignment and array stride are set to match the base alignment of a single array element..."

(It does not say compute the size of the whole thing and divide.)

I think the problem is the size was returned instead of the stride (I'm making that fix manually).  Another complexity to work around is OpenGL rules flatten an array of matrices to a bigger array of just rows/cols, but SPIR-V does not.  I made the appropriate changes.

I added the std140 test too, that looks good.

SPIRV/GlslangToSpv.cpp
```diff
@@ -2884,8 +2890,14 @@ spv::Id TGlslangToSpvTraverser::createMiscOperation(glslang::TOperator op, spv::
     }
 
     spv::Id id = 0;
-    if (libCall >= 0)
-        id = builder.createBuiltinCall(precision, typeId, stdBuiltins, libCall, operands);
+    if (libCall >= 0) {
+        if (libCall != spv::GLSLstd450FrexpStruct)
```
johnkslang: I put in a generic fix here, based on 'consumedOperands'.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -3279,8 +3282,8 @@ void IdentifyBuiltIns(int version, EProfile profile, int spv, EShLanguage langua
     symbolTable.relateToOperator("packUnorm4x8",    EOpPackUnorm4x8);
     symbolTable.relateToOperator("unpackUnorm4x8",  EOpUnpackUnorm4x8);
 
-    symbolTable.relateToOperator("packDouble2x32",    EOpPackUnorm4x8);
-    symbolTable.relateToOperator("unpackDouble2x32",  EOpUnpackUnorm4x8);
+    symbolTable.relateToOperator("packDouble2x32",    EOpPackDouble2x32);
+    symbolTable.relateToOperator("unpackDouble2x32",  EOpUnpackDouble2x32);
```
johnkslang: Good catch.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -488,15 +488,25 @@ TIntermTyped* TIntermediate::addConversion(TOperator op, const TType& type, TInt
...
     case EOpRightShiftAssign:
+
         if ((type.getBasicType() == EbtInt || 
              type.getBasicType() == EbtUint) &&
             (node->getType().getBasicType() == EbtInt || 
-             node->getType().getBasicType() == EbtUint))
+             node->getType().getBasicType() == EbtUint)) {
 
-            return node;
+            if (type.getBasicType() == node->getType().getBasicType() ||
```
johnkslang: Can you submit this as a separate merge request, with a test that shows the change made?  Thanks.

johnkslang:
I made most these changes, with 4 commits.  I believe I identified 2 questions above for what might not have made it.  Feel free to submit individual requests for what needs to change, and I'll try to be more responsive as well.  Thanks.

SPIRV/GlslangToSpv.cpp
amdrexu: OK. I will resubmit the request with a demonstrative case.

glslang/MachineIndependent/Intermediate.cpp
amdrexu: OK. I indeed have this case. It is very tricky: )

amdrexu:
Thanks for your review. I will follow your suggestions and resubmit the requests soon.

PR #109, opened 2015-12-06
In addUnaryMath, check for 'child' being null before dereferencing it.

xorgy:


xorgy:
Just noticed that my editor has deleted all of the trailing spaces on the lines. 

You can look at the diff ignoring these whitespaces with. https://github.com/KhronosGroup/glslang/pull/109/files?w=1

xorgy:
It might make more sense to remove the null check altogether, since it evidently hasn't even been encountered since it was introduced some time in the early '00s.

johnkslang:
I merged the fix.

About removal, maybe:  There is other code handling error follow on that might involve a 0 expression, and maybe it should all be pulled out together, to see if any of it gets exercised.  Error follow on is more robustly handled (and typically is) by substituting an expression, not by eliminating it (with 0).

PR #112, opened 2015-12-08
Fix 3 issues related with GLSL parser and SPIRV generator

amdrexu:


amdrexu:
Hi,
I misspell the file name of new test case. "spv.ShiftOps" should be changed to "spv.shiftOps" based on naming convention. Please correct it.

johnkslang:
This is three unrelated changes in the same pull request, but the changes are in different states of being accepted.

Can you have one subject per pull request, and have the pull request named after that subject?

Thanks.

PR #115, opened 2015-12-09
SPV: Scalar smear operation should use type of the scalar to make the expected vector type when the provided vector type is incompatible with the scalar.

amdrexu:
Thanks for your previous analysis. Yes, the issue I encounter is like this:

ivec3 i3;
i3 <<= 3u;

The uint 3 will be promoted to become constant vector by OpCompositeConstruct. However, the vector type is obtained from i3, which is ivec3. Thus, there will be incorrect use of OpCompositeConstruct because the result type is ivec3 while the component type is uint. There are similar cases:

uvec3 u3;
int i1;
u3 >>= i1;

Please see my test case.


SPIRV/SpvBuilder.cpp
```diff
@@ -1190,6 +1190,10 @@ Id Builder::smearScalar(Decoration /*precision*/, Id scalar, Id vectorType)
     if (numComponents == 1)
         return scalar;
 
+    // Make new vector type if the provided one is incompatible with type of the scalar
+    if (getTypeId(scalar) != getScalarTypeId(vectorType))
```
johnkslang: This should be an assert: This function is being told what type to make, it should not silently make a different type.  If the scalar and vector types differ, in the general case someone needed to do a conversion.

I'll accept this merge request, and then modify slightly.  Thanks for catching this.

amdrexu: Thanks for correcting me. You idea is better.

Issue #103, opened 2015-11-30
Remove bison binary


kraita:
Can we remove the Bison binary from the repository, please? I believe it's generally accepted principle that binaries should not be stored in source version control and Git is particularly bad at that, as far as I know.

baldurk:
I don't strongly disagree, but to give a counterpoint/devil's advocate; without rewriting history, deleting it won't help much. Non-shallow clones will still have to download the file. Only if it's updated/changed do you pay the cost of versioning binary files bloating the history.

Keeping it around has a convenience factor for building on windows, since bison can't be assumed to be available and requiring people to go off to another site and download a binary and put it in the right place in their PATH is an annoying step just to be able to build. An alternative to that is to commit the bison generated files, then bison is only needed when you modify the grammar - but there's also an argument that 'output' files shouldn't be committed.

johnkslang:
There are valid reasons to commit 'output' files, and I think this is one of them:  Only those developers making grammar changes would need to install Bison.  Also note that so far bison.exe is small with only one copy, so I don't think we need to rewrite history to get the binary out of the normal everyday picture.

baldurk:
Oh sure, like I say I was playing devil's advocate to some degree - I commit the bison files in my local embed of glslang in a project so that I don't have to have a custom build step.

About rewriting history I just meant that my understanding was that removing the file only saves you on the working copy version - which like you say is small and I don't imagine anyone is complaining about having that. Usually the reason not to commit binary files to git is that each version of the file is stored without any delta so the history inflates. Even if the file is removed today, someone cloning tomorrow will download the same size repository because it includes the file in the history.

xorgy:
Problem with rewriting history which is that old, is that any branches based on the history after it was committed will have much messier rebases.

The issue is exacerbated by the fact that the binary has been in the repository since the very first commit, so there is no common ancestor commit for a rebase to work against. You would likely end up exporting patches and then recommitting them.

kraita:
From my perspective it's fine to skip the history rewrite and just delete the binary from the latest revision.

xorgy:
`git filter-branch --force --index-filter 'git rm --cached --ignore-unmatch tools/bison.exe' --prune-empty --tag-name-filter cat -- --all`
If you wanna ;-).

I vote for let it be.

baldurk:
For what it's worth I was definitely not in favour of rewriting the history :). The opposite - I was assuming that no-one would want to, and so the horse has already bolted so to speak.

[...]

PR #122, opened 2016-01-01
Remove bison binary

baldurk:
This removes the bison binaries and commits the generated files, as decided in #103.

* remove bison.exe (and dependent files)
* leave history alone
* commit the files built by bison, so most people don't need bison
* document where to get bison.exe, for those that really need it

I removed the custom command from cmake entirely as I don't see a way to make it only fire if the glslang.y file is newer than the output glslang_tab.cpp, and if bison isn't found then it will fail. Possibly the command could be made to never fail no matter what but that sounds dangerous to me. If someone who knows cmake can tell me how to do that, let me know!


baldurk:
I see that there's a minor conflict with latest code on master now. Let me know if you want to rebase onto master, I'll leave this branch alone otherwise.

johnkslang:
Yes, we need this change, I want more systems to test it on when I make it. Feel free to update, thanks.

baldurk:
I've rebased onto head code and updated the generated `glslang_tab.cpp`/`glslang_tab.cpp.h` files.

johnkslang:
Looks good.  (I tweaked the glslang/CMakelists.)

baldurk:
Thanks! :smile:

PR #127, opened 2016-01-11
Rework loop code generation

dekimir:
Now conforming to the latest spec... :)


SPIRV/GlslangToSpv.cpp
```diff
@@ -1390,33 +1389,63 @@ void TGlslangToSpvTraverser::visitConstantUnion(glslang::TIntermConstantUnion* n
...
+        builder.setBuildPoint(&blocks.continue_target);
+        if (node->getTerminal())
+            node->getTerminal()->traverse(this);
+        builder.createBranch(&blocks.head);
     } else {
-        builder.createBranchToBody();
-    }
+        // Spec requires back edges to target header blocks, and every header
+        // block must dominate its merge block.  Create an empty header block
+        // here to ensure these conditions are met even when body contains
```
dneto0: For me the salient point that isn't quite explicit in the comment is that the header block must contain the OpLoopMerge, but the body might have nontrivial control flow an in particular the body's first block might have its own OpLoopMerge or OpSelectionMerge.
That's why we need to make room for that extra Op*Merge instruction here.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1390,33 +1389,63 @@ void TGlslangToSpvTraverser::visitConstantUnion(glslang::TIntermConstantUnion* n
...
+        if (node->getTest()) {
+            node->getTest()->traverse(this);
+            spv::Id condition =
+                builder.accessChainLoad(convertGlslangToSpvType(node->getTest()->getType()));
+            builder.createConditionalBranch(condition, &blocks.head, &blocks.merge);
+        } else {
+            // TODO: unless there was a break instruction somewhere in the body,
+            // this is an infinite loop, so we should abort code generation with
+            // a warning.  As it stands now, nothing will jump to the merge
+            // block, and it may be dropped as unreachable by the SPIR-V dumper.
```
dneto0: I'd argue that an unreachable merge block is valid, and the bug would be if the dumper dropped it. 
If a block is mentioned by a Op*Merge instruction then it should be considered live.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1390,33 +1389,63 @@ void TGlslangToSpvTraverser::visitConstantUnion(glslang::TIntermConstantUnion* n
...
+            node->getTerminal()->traverse(this);
+        if (node->getTest()) {
+            node->getTest()->traverse(this);
+            spv::Id condition =
+                builder.accessChainLoad(convertGlslangToSpvType(node->getTest()->getType()));
+            builder.createConditionalBranch(condition, &blocks.head, &blocks.merge);
+        } else {
+            // TODO: unless there was a break instruction somewhere in the body,
+            // this is an infinite loop, so we should abort code generation with
+            // a warning.  As it stands now, nothing will jump to the merge
```
dneto0: There are other ways to exit a loop: OpReturn, or OpKill.
So this comment is overly broad.
At best, there could be a caution indicating that, prima facie, this is an infinite loop.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1390,33 +1389,63 @@ void TGlslangToSpvTraverser::visitConstantUnion(glslang::TIntermConstantUnion* n
 
 bool TGlslangToSpvTraverser::visitLoop(glslang::TVisit /* visit */, glslang::TIntermLoop* node)
 {
-    // body emission needs to know what the for-loop terminal is when it sees a "continue"
-    loopTerminal.push(node->getTerminal());
+    auto blocks = builder.makeNewLoop();
+    builder.createBranch(&blocks.head);
+    if (node->testFirst() && node->getTest()) {
+        builder.setBuildPoint(&blocks.head);
```
dneto0: Coder's choice: DRY the setBuildPoint in the head.
Actually, I prefer it the way you have it already. That makes the two alternatives conceptually decomposed.

dekimir: Yes, this was a deliberate choice.  I find it more readable to always `setBuildPoint` right before emitting code; that way it's easier for the reader to envision the generated code.

SPIRV/SpvBuilder.cpp
```diff
@@ -1751,150 +1751,39 @@ void Builder::endSwitch(std::vector<Block*>& /*segmentBlock*/)
...
 {
-    const Loop& loop = loops.top();
-    assert(loop.body);
-
-    // This is a reconvergence of control flow, so no merge instruction
-    // is required.
-    createBranch(loop.body);
-    loop.function->addBlock(loop.body);
-    setBuildPoint(loop.body);
+    loops.push({makeNewBlock(), makeNewBlock(), makeNewBlock(), makeNewBlock()});
```
dneto0: C++11 8.5.4 para 4 guarantees a left-to-right order of evaluation of these.
(Collaborative confirmation by @dekimir @AWoloszyn and myself.)

SPIRV/SpvBuilder.h
```diff
@@ -382,28 +382,24 @@ class Builder {
...
-
-    // Generate an unconditional branch to the loop body.  The builder insertion
-    // point will be placed at the start of the body.  Use this when there is
-    // no loop test.
-    void createBranchToBody();
-
-    // Add a branch to the test of the current (innermost) loop.
-    // The way we generate code, that's also the loop header.
+    struct LoopBlocks {
+        Block &head, &body, &merge, &continue_target;
```
dneto0: Recommend swapping the order of "merge" and "continue_target".
That will lead to more natural code generation.  Why?  C++11 8.5.4 para 4 guarantees the left-to-right order of the evaluation of the initialization values.  You want the merge block to be added to the function's block vector last, since the dumper dumps the IR in vector order.
That will make the IR easier to read since blocks will appear in natural order; it will also play with glslangValidator's human-readable output indentation scheme. (It indents nested control flow).

dneto0: I realize this will perturb most of the test results yet again. :-(
Perhaps it should be done in a subsequent commit on this MR.

Test/baseResults/spv.dataOutIndirect.vert.out
```diff
@@ -8,64 +8,66 @@ Linked vertex stage:
...
+              28:   19(fvec4)   Load 27(color)
+              30:     29(ptr)   AccessChain 24(colorOut) 25
+                                Store 30 28
+                                Branch 13
+              12:             Label
+              35:     29(ptr) AccessChain 24(colorOut) 34
+              36:   19(fvec4) Load 35
+                              Store 33(gl_Position) 36
                               Return
+              13:             Label
```
dneto0: FYI.  This is the continue block for the loop. But it appears after the merge block for the loop (ID 12), which itself contains the remainder of the function's instructions.

Test/spv.for-notest.vert
```diff
@@ -0,0 +1,9 @@
+#version 450
+layout(location=0) out highp int r;
+void main() {
+  int i;
+  // This infinite loop results in bad SPIR-V generated, since the merge block
```
dneto0: I think it's a bug that the merge block is dropped as unreachable.  Mentioning in a loop-merge (or selection-merge) should prevent elimination.
Since it's an infinite loop, I'd be happy to defer that to a new bug and subsequent MR.

dneto0:
The code looks good to me, except for the ordering of the merge and continue_target blocks in the loop struct.  Those should be swapped.
I've reviewed only some of the deltas to the tests. They quickly become very hard to read given the weird ordering of the IR dump.
Thanks for the new corner case tests, including for-nobody. Those are good additions.

I'll review again once you put up another commit with the swapped block ordering.  It should be easier to review the test deltas.

dneto0:
@dekimir tried the reordering, but you still get unexpected IR output order. That's because the IR blocks for the loop body are created after the merge and continue_target blocks are created.  That's fundamental to the traversal.  So no quick hack will fix it; reodering the declarations of the merge and continue_target is a minor improvement.

@dekimir proposed that the best way to prettify the IR dump is to make the dumper dump in topological sort order rather than just creation-time-order.  I agree.  But that's for another CL.

I've inspected about 5 of the key cases in the set of test changes.  I suggest @johnkslang take a look at the current version of the MR.

+1

dneto0:
For the record, @dekimir plans to submit a pull request that tries to ensure the IR dumper emits blocks in a sane order.  (E.g. defer a merge block until other blocks in the construct have been  emitted.)

PR #128, opened 2016-01-11
Add .clang-format.

dekimir:
This allows automatic source-file formatting with clang-format, freeing
contributors from worrying about formatting while keeping the style
consistent everywhere.


dekimir:
The auto-format was crafted to be as close as possible to the existing source formatting.  Inevitably, there will be some small differences, which I've tried to illustrate by running clang-format on two source files (one .h and one .cpp).  Please take a quick look at them before accepting this PR.  I may be able to tweak the format some more, though there are diminishing returns.

SPIRV/SpvBuilder.cpp
```diff
@@ -1,11 +1,11 @@
 //
-//Copyright (C) 2014 LunarG, Inc.
+// Copyright (C) 2014 LunarG, Inc.
```
johnkslang: Why is an extra space here, for just this file?  The official copyright to use doesn't have the space, and it's that way (no space) in all the files, so something seems odd about making this change, just in this one place.  Why is it needed?

SPIRV/SpvBuilder.cpp
```diff
@@ -1742,7 +1735,7 @@ void Builder::nextSwitchSegment(std::vector<Block*>& segmentBlock, int nextSegme
 void Builder::endSwitch(std::vector<Block*>& /*segmentBlock*/)
 {
     // Close out previous segment by jumping, if necessary, to next segment
-    if (! buildPoint->isTerminated())
+    if (!buildPoint->isTerminated())
```
johnkslang: I really liked the visual of seeing a space here to make the ! more clear.

Totally understand different conventions/styles have their own trade-offs/preferences, but I'm not following why we there is a request to partially change the local getting used.  Possibly, I'm missing the bigger picture, but going from consistent to inconsistent style doesn't quite seem the thing to do either.

SPIRV/SpvBuilder.cpp
xorgy: "The auto-format was crafted to be as close as possible to the existing source formatting. Inevitably, there will be some small differences, which I've tried to illustrate by running clang-format on two source files (one .h and one .cpp)."
 He said this, I don't think he ran it on the whole codebase.

One other problem is that some big multi-line statements and comments will need to be reformatted because they get completely trashed by `clang-format`.

johnkslang:
I just don't understand yet how this applies to subsets.  I don't want to get religious about the convention, and am open to improvements or changes that make tools better, but I'd like to not change the convention on a subset of the project, and don't yet understand what the choices are (leave all source the same, change all the source, do it to just some files, etc.).

xorgy:
@johnkslang I'm not sure why he left out the "change all the files" step (maybe to make it easier to rebase the change). But usually you would decide to change all the files at once, or standardize the format every time you touch a file in a patch.

dekimir:
There's no need to format every file at this time if a small sample suffices to discuss the style and modify it to match the stakeholders' preferences (or abandon it if it's unworkable).

In fact, there's no requirement to ever auto-reformat everything.  I've just found having a .clang-format useful when writing new code to make it blend in better.

SPIRV/SpvBuilder.cpp
dekimir: @xorgy are these statements/comments visible in the sample files here?  There may be `.clang-format` tweaks that can minimize the impact, which I'd like to try.

johnkslang:
@dekimir Are you interested in updating this? With one main branch at the moment (GitHub master), it is a good time.

dekimir:
I'm definitely still interested.  I'll rebase against master and fiddle just a bit more to address some of the feedback.

SPIRV/SpvBuilder.cpp
dekimir: The only way I could find to avoid inserting this space was to turn off comment reflow altogether.  It must be said, though, that other projects (such as https://github.com/KhronosGroup/SPIRV-Tools/) have this space in copyright and seem fine with it.  Perhaps glslang can tolerate it, too?

SPIRV/SpvBuilder.cpp
dekimir: So I can't find a way to make `clang-format` insert a space after the ! operator (or even make it not delete the space when it exists).  This is one of those things we'll have to concede as the price of using a formatting tool.  OTOH, most of the C++ code I've ever seen doesn't insert a space after ! and reads just fine.

dekimir:
Addressed feedback, PTAL.

SPIRV/SpvBuilder.cpp
```diff
@@ -394,10 +389,10 @@ Id Builder::makeImageType(Id sampledType, Dim dim, bool depth, bool arrayed, boo
     // not found, make it
     type = new Instruction(getUniqueId(), NoType, OpTypeImage);
     type->addIdOperand(sampledType);
-    type->addImmediateOperand(   dim);
-    type->addImmediateOperand(  depth ? 1 : 0);
+    type->addImmediateOperand(dim);
+    type->addImmediateOperand(depth ? 1 : 0);
```
johnkslang: This really does make it harder to read.

Is our only choice a coding convention that requires overriding local alignment that make it easier to see what's going on?  If so, that seems like sacrificing the spirit of the law to satisfy the letter of the law.

(Maybe that's the cost for something, but I want to be clear first.  A lot of other changes appear as improvements, would be cool to pick the clarifying subset and not all of them. :) )

SPIRV/SpvBuilder.cpp
```diff
@@ -518,8 +511,7 @@ int Builder::getNumTypeConstituents(Id typeId) const
     case OpTypeVector:
     case OpTypeMatrix:
         return instr->getImmediateOperand(1);
-    case OpTypeArray:
-    {
+    case OpTypeArray: {
         Id lengthId = instr->getImmediateOperand(1);
         return module.getInstruction(lengthId)->getImmediateOperand(0);
     }
```
johnkslang: Hmm, That really looks like the close of the switch now.

SPIRV/SpvBuilder.cpp
```diff
@@ -610,10 +595,8 @@ Id Builder::findScalarConstant(Op typeClass, Op opcode, Id typeId, unsigned v1,
     Instruction* constant;
     for (int i = 0; i < (int)groupedConstants[typeClass].size(); ++i) {
         constant = groupedConstants[typeClass][i];
-        if (constant->getOpCode() == opcode &&
-            constant->getTypeId() == typeId &&
-            constant->getImmediateOperand(0) == v1 &&
-            constant->getImmediateOperand(1) == v2)
+        if (constant->getOpCode() == opcode && constant->getTypeId() == typeId &&
+            constant->getImmediateOperand(0) == v1 && constant->getImmediateOperand(1) == v2)
```
johnkslang: This lost visual clues that make it easier to see what is going on when reading the code.

I doubt there are standards that know how to add these clues, as they come from a source knowing what's going on and how to see more clearly it is getting done.

Is it necessary to lose the easier visual verify-ability of what is being compared here?

SPIRV/SpvBuilder.cpp
```diff
@@ -1273,9 +1269,11 @@ void Builder::promoteScalar(Decoration precision, Id& left, Id& right)
     int direction = getNumComponents(right) - getNumComponents(left);
 
     if (direction > 0)
-        left = smearScalar(precision, left, makeVectorType(getTypeId(left), getNumComponents(right)));
+        left =
+            smearScalar(precision, left, makeVectorType(getTypeId(left), getNumComponents(right)));
     else if (direction < 0)
-        right = smearScalar(precision, right, makeVectorType(getTypeId(right), getNumComponents(left)));
+        right =
+            smearScalar(precision, right, makeVectorType(getTypeId(right), getNumComponents(left)));
```
johnkslang: Why is this change important?

SPIRV/SpvBuilder.cpp
```diff
@@ -1718,7 +1717,7 @@ Id Builder::createMatrixConstructor(Decoration precision, const std::vector<Id>&
 
     // initialize the array to the identity matrix
     Id ids[maxMatrixSize][maxMatrixSize];
-    Id  one = makeFloatConstant(1.0);
+    Id one = makeFloatConstant(1.0);
```
johnkslang: This made it harder to see the point of the original alignment.

johnkslang:
Is it still the case that actual changes would be much broader, and the current PR is a subset for discussion?

I made a few comments, that I think I avoiding religious-type style conventions, and simply saying the new way obscured some things the existing way had tried to make clearer by how the code was aligned.  It's not about following a specific style, it's about making it simpler to visually see the code is correct, e.g., making sure things that are supposed to match are vertically aligned for very easy verification that they do match.  That's about ensuring quality.

Don't want to get too side tracked here, but several recent PRs of higher priority are using, within themselves, contradictory conventions, and want to address that one way or another.

SPIRV/SpvBuilder.cpp
dekimir: I couldn't find a way to make `clang-format` align the `?` operator.  I personally don't find it less readable without that alignment.

SPIRV/SpvBuilder.cpp
dekimir: Couldn't find a way to place one conjunct per line here.  If it's very important, we can use `// clang-format off`.

SPIRV/SpvBuilder.cpp
dekimir: Because of the (arbitrary) 100-column limit that I set.  I could increase it, but I wouldn't recommend going above 120.

SPIRV/SpvBuilder.cpp
dekimir: I could indent the whole `case` block, but that would produce quite a visual change.  Or I could put the `case` brace on a new line but only by doing the same for every `if`, `for`, `switch`, `do`, and `while`.  I uploaded that -- please see if you like it.

FWIW, I think `clang-format` is doing the right thing here by making the `case`-following brace consistent with the other control-flow statements.  It's interesting that you consider this a problem:

```
case FOO: {
  code = code + code;
}
```

but not this

```
if (FOO) {
  code = code + code;
}
```

dekimir:
It was never the case that "actual changes" would be much broader -- we can limit the scope of these changes to whatever we wish.  It remains true that this is just a (largish) sample to properly illustrate what `clang-format` will do on the code it does process.

It's precisely to avoid contradictory conventions that I'm proposing we use a program to help us format the code.

SPIRV/SpvBuilder.cpp
dekimir: I was playing around with `=` alignment but haven't figured out an acceptable setting yet.  I can certainly align these lines, but then it also tries to align the `=` inside some `for` statements with the surrounding declarations, and it looks quite weird. :astonished:

johnkslang:
Proposal:

1. use the clang format in the current state @dekimir got it too, which matches glslang convention good enough 
2. bump the line limit up to 120
3. don't apply to any existing code
4. do allow application to any new code, which may either match it or other existing glslang conventions like vertical alignment when that makes intent more clear

Yes?  I think the PR for that would just be the .clang-format.

dekimir:
@johnkslang: Done.

PR #130, opened 2016-01-12
linkValidate: Round up structure sizes to vec4 in std140

jekstrand:
This fixes a bug with structs of structs where it wasn't properly applying the last sentence of clause 9 of the std140 UBO packing rules.

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -959,6 +959,11 @@ int TIntermediate::getBaseAlignment(const TType& type, int& size, int& stride, b
             size += memberSize;
         }
 
+        // The structure may have padding at the end; the base offset of
+        // the member following the sub-structure is rounded up to the next
+        // multiple of the base alignment of the structure.
+        RoundToPow2(size, maxAlignment);
+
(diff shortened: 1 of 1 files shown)


jekstrand:
I just force-pushed the branch.  Re-reading the std140 layout spec in light of std430, it looks like std430 has the same rule.  The only difference is whether it's rounded up to a vec4 or not.  This fixes a bunch of SSBO tests.

johnkslang:
The ambiguity is whether this is compensated for when doing the alignment of the next member, versus the current member.  The rules are about where members are, rather than their size.

Note I added one assumption at the end the rules about size, but that's it.

Anyway, will be looking at this now, it parallel with any other input here.

jekstrand:
Right.  I thought about that.  However, I opted to go for size because it kept all of the rules in one place.  If it were implemented in terms of alignment, it would require either substantial refactoring or implementing the last sentence of clause 9 in updateMemberOffset which is in an entirely different part of the compiler.

jekstrand:
I just pushed again with one more fix, this time for matrices.  With these two commits, our driver now passes 100% of the UBO/SSBO layout tests.

PR #131, opened 2016-01-14
Addresses some of the memory leaks in glslang.

AWoloszyn:
This is mostly accomplished by switching some vectors in
SpirvBuilder to contain std::unique_ptrs so that they are
cleaned up automatically when the vectors go away.

The only remaining issues I have found seem to leak only once
per process so are lower priority.

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -77,7 +77,7 @@ Id Builder::import(const char* name)
     Instruction* import = new Instruction(getUniqueId(), NoType, OpExtInstImport);
     import->addStringOperand(name);
     
-    imports.push_back(import);
+    imports.push_back(std::unique_ptr<Instruction>(import));
     return import->getResultId();
 }
 
@@ -88,7 +88,7 @@ Id Builder::makeVoidType()
     if (groupedTypes[OpTypeVoid].size() == 0) {
         type = new Instruction(getUniqueId(), NoType, OpTypeVoid);
         groupedTypes[OpTypeVoid].push_back(type);
-        constantsTypesGlobals.push_back(type);
+        constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
         module.mapInstruction(type);
     } else
         type = groupedTypes[OpTypeVoid].back();
@@ -102,7 +102,7 @@ Id Builder::makeBoolType()
     if (groupedTypes[OpTypeBool].size() == 0) {
         type = new Instruction(getUniqueId(), NoType, OpTypeBool);
         groupedTypes[OpTypeBool].push_back(type);
-        constantsTypesGlobals.push_back(type);
+        constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
         module.mapInstruction(type);
     } else
         type = groupedTypes[OpTypeBool].back();
@@ -116,7 +116,7 @@ Id Builder::makeSamplerType()
     if (groupedTypes[OpTypeSampler].size() == 0) {
         type = new Instruction(getUniqueId(), NoType, OpTypeSampler);
         groupedTypes[OpTypeSampler].push_back(type);
-        constantsTypesGlobals.push_back(type);
+        constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
         module.mapInstruction(type);
     } else
         type = groupedTypes[OpTypeSampler].back();
@@ -140,7 +140,7 @@ Id Builder::makePointer(StorageClass storageClass, Id pointee)
     type->addImmediateOperand(storageClass);
     type->addIdOperand(pointee);
     groupedTypes[OpTypePointer].push_back(type);
-    constantsTypesGlobals.push_back(type);
+    constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
     module.mapInstruction(type);
 
     return type->getResultId();
@@ -162,7 +162,7 @@ Id Builder::makeIntegerType(int width, bool hasSign)
     type->addImmediateOperand(width);
     type->addImmediateOperand(hasSign ? 1 : 0);
     groupedTypes[OpTypeInt].push_back(type);
-    constantsTypesGlobals.push_back(type);
+    constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
     module.mapInstruction(type);
 
     return type->getResultId();
@@ -182,7 +182,7 @@ Id Builder::makeFloatType(int width)
     type = new Instruction(getUniqueId(), NoType, OpTypeFloat);
     type->addImmediateOperand(width);
     groupedTypes[OpTypeFloat].push_back(type);
-    constantsTypesGlobals.push_back(type);
+    constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
     module.mapInstruction(type);
 
     return type->getResultId();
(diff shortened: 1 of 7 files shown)


SPIRV/spvIR.h
```diff
@@ -155,15 +156,16 @@ class Block {
...
     virtual ~Block()
     {
-        // TODO: free instructions
     }
     
     Id getId() { return instructions.front()->getResultId(); }
 
     Function& getParent() const { return parent; }
+    // addInstruction takes ownership of the instruction once it has
+    // been passed in. It will be destroyed when the Block is destroyed.
```
dekimir: I presume the same goes for `addLocalVariable()`, right?  This can be implicitly communicated (and compiler-enforced) by making the argument a `unique_ptr`.

SPIRV/SpvBuilder.h
```diff
@@ -201,11 +202,13 @@ class Builder {
...
     Block* getBuildPoint() const { return buildPoint; }
 
-    // Make the main function.
+    // Make the main function. The returned pointer is only valid
+    // for the lifetime of this builder.
     Function* makeMain();
 
     // Make a shader-style function, and create its entry block if entry is non-zero.
     // Return the function, pass back the entry.
+	// The returned pointer is only valid for the lifetime of this builder.
```
dneto0: indentation.

StandAlone/StandAlone.cpp
```diff
@@ -218,6 +218,7 @@ void ProcessConfigFile()
 {
     char** configStrings = 0;
     char* config = 0;
+    bool allocated_config = false;
```
dekimir: Why not make `config` a smart pointer instead?  Even if it remains a `char*`, I doubt you need this new variable, as it's safe to `delete[] 0` and anyway, `config` is always either allocated or zero.

SPIRV/spvIR.h
AWoloszyn: Good call, that makes it much clearer.

SPIRV/SpvBuilder.cpp
```diff
@@ -896,7 +898,7 @@ Id Builder::createVariable(StorageClass storageClass, Id type, const char* name)
         break;
```
dneto0: If you follow @dekimir 's advice (which seems like good advice), then you'll need to do the std::unique_ptr construction in the call to addLocalVariable on old line 895, new line 897.  And that would make it rhyme with the pushback onto costantsTypesGlobals

SPIRV/spvIR.h
dneto0: +1

dneto0:
I'm happy once @dekimir  is happy, so +1 from me.

StandAlone/StandAlone.cpp
AWoloszyn: The problem isn't is config 0 or not, the problem here is that config is either allocated or comes from ReadFileData().

If it comes from ReadFileData we do not want to delete it.

dekimir: Hm, then don't you have to invoke `FreeFileData()` on it for such cases?

dekimir: Never mind, that seems to be handled through `configString`.

dekimir: OK, here's my updated understanding: you don't need `allocated_config` -- just tack on `else delete[] config;` after `if (configStrings)` below.

johnkslang:
Will look at the detailed comments soon.   Here are three high-level comments:

1. Thanks for doing this; SPIR-V generation was still a hole as far as memory testing.
2. The priority is on what would be "in the loop" when glslang is used programmatically (not the command line, the API).  So things like the file I/O memory that are specific to the command-line use is important only to the extent it's useful (for tools, hygiene).  E.g., the configuration stuff is passed in programmatically in the priority scenario.
3. Just background info for anyone not aware: there is the "-m" option that tests the programmatic use from the command line for functional testing of memory leaks of specific workloads.  I had been testing the AST generation/teardown with this, but not yet SPIR-V generation.

johnkslang:
I just extended the "-m" mode (for performance and memory measurement within the API) for the compile-and-link mode used by SPIR-V.  This can now be used to 

1. see from a functional level what progress is made in eliminating memory leaks
2. measure performance effects of moving something into or out of a pool

See the commit message for c57b2a97fa808970a4809c53de605bbf103fce12 (and comments) for more detail.

The main point is it isolates the part of glslang that is "in the loop" when used programmatically, and runs it 10,000 times (easy to change) to show change in memory over time.  It can also be timed to measure performance.

This is intended to complement this current merge request, not conflict or overlap.

As expected (from pending "TODO" in SPIR-V to free things), it shows several kilobytes of memory leaked per SPIR-V compile.  Most likely much of that goes away with this merge request.  However, there may be independent memory leaking with the generic glslang linker, as that's never been memory tested and was previously used only for validation (before SPIR-V).

glslang/Include/Common.h
```diff
@@ -160,14 +160,20 @@ template <class T> class TVector : public std::vector<T, pool_allocator<T> > {
 };
 
 template <class T> class TList  : public std::list<T, pool_allocator<T> > {
+ public:
+    POOL_ALLOCATOR_NEW_DELETE(GetThreadPoolAllocator())
```
johnkslang: What is this for?  TList is not being used with new/delete (is it?).  Or, if some new code does new/delete, that would be okay to not use the pool unless it was lots of small intensive new/delete per compile.

All the little allocations going on inside the TList should already be using the pool, due to the template argument.

(Since the pool is an extra complexity, for simplicity, it's nice to just use it where justified. If it's ever removed, it's also easier to remove if unneeded uses have enforced proper memory management along the way.)

Currently, I think this addition is dead code.

glslang/Include/Common.h
```diff
@@ -160,14 +160,20 @@ template <class T> class TVector : public std::vector<T, pool_allocator<T> > {
...
 };
 
 template <class K, class D, class CMP = std::less<K> > 
 class TMap : public std::map<K, D, CMP, pool_allocator<std::pair<K, D> > > {
+ public:
```
johnkslang: Similar comment as per TList, though I see a "new TPragmaTable" that would actually use this code.

glslang/Include/Common.h
```diff
@@ -160,14 +160,20 @@ template <class T> class TVector : public std::vector<T, pool_allocator<T> > {
...
+    POOL_ALLOCATOR_NEW_DELETE(GetThreadPoolAllocator())
 };
 
 template <class K, class D, class HASH = std::hash<K>, class PRED = std::equal_to<K> >
 class TUnorderedMap : public std::unordered_map<K, D, HASH, PRED, pool_allocator<std::pair<K const, D> > > {
+ public:
+    POOL_ALLOCATOR_NEW_DELETE(GetThreadPoolAllocator())
```
johnkslang: Same comment as for TList.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -1407,6 +1407,7 @@ TProgram::~TProgram()
     delete infoSink;
     delete reflection;
 
+    SetThreadPoolAllocator(*old_pool);
```
johnkslang: I don't understand this part.  Is TProgram use "stacked", such that this has some meaning?  Or, what triggered this?

glslang/Include/Common.h
AWoloszyn: That is true, I had done it for consistency, will remove though.

glslang/Include/Common.h
AWoloszyn: That is the occurrence I was investigating, I will remove this though as you are right it is not helpful.

AWoloszyn:
I have broken this out into 3 separate commits, I will leave this MR open for now, so that the comments can be easily read. I will close once the other MRs are resolved, or these comments are no longer useful.

AWoloszyn:
Split into #132 #133 #134

glslang/Include/Common.h
johnkslang: Since the pragmaTable is owned by something in the pool (TIntermAggregate, which I'm not sure makes sense), maybe this one (of the three) actually makes sense to do.

PR #132, opened 2016-01-18
Have TProgram restore the original thread-allocator.

AWoloszyn:
TProgram creates a new thread-pool allocator, and sets it.
The original allocator was created in InitializeProcess, and
can no longer be referenced or cleaned up after TProgram has been
run.

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -1407,6 +1407,7 @@ TProgram::~TProgram()
     delete infoSink;
     delete reflection;
 
+    SetThreadPoolAllocator(*old_pool);
     for (int s = 0; s < EShLangCount; ++s)
         if (newedIntermediate[s])
             delete intermediate[s];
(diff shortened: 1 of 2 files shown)


johnkslang:
When used in multi-threaded mode, there can be lots of TProgram coming and going, asynchronously, and with destruction happening in different orders than construction, and you don't know any of these orders.  So, I don't understand yet how a design pattern of save/restore could work, which kind of implies a "stack" ordering.

AWoloszyn:
So the main problem that this is trying to solve is avoiding throwing away the reference to the original allocator. I see how this can be a problem in multi-threaded mode.

So a question that this raises is: does TProgram need to allocate any new memory (in the same pool as the objects in Link()) after link is called?
As far as I can tell, anything other calls do not have to have the same lifetime as TProgram. TReflection is the only other object that I see and it does not use the pool.

If this is correct we may be able to just reset the allocator at the end of link(), but not destroy it.
This should allow the memory to live until TProgram is destroyed but reset the allocator to what it was before the call. This should be safe in a multi-threaded environment since it only modifies the thread-local data.

johnkslang:
Per-thread pools have been removed, to eliminate the need for DetachThread. Objects (shader, compiler, linker, and program) now own their own pools and do their own allocation/deallocation of their pools.

There is no longer a need to "stack" pools and restore originals to get proper clean up in DetachThread.

PR #133, opened 2016-01-18
Free memory associated with SPIR-V generation.

AWoloszyn:


glslang/Include/intermediate.h
```diff
@@ -784,7 +784,9 @@ class TIntermAggregate : public TIntermOperator {
 public:
     TIntermAggregate() : TIntermOperator(EOpNull), userDefined(false), pragmaTable(0) { }
     TIntermAggregate(TOperator o) : TIntermOperator(o), pragmaTable(0) { }
-    ~TIntermAggregate() { delete pragmaTable; }
+    // Since pragmaTable is allocated with the PoolAllocator, we
```
johnkslang: Seems the root problem here is that TIntermAggregate owns the pragmaTable to begin with. That can be an issue for another day though.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -1613,7 +1613,13 @@ TIntermTyped* TIntermediate::promoteConstantUnion(TBasicType promoteTo, TIntermC
 void TIntermAggregate::addToPragmaTable(const TPragmaTable& pTable)
 {
     assert(!pragmaTable);
-    pragmaTable = new TPragmaTable();
+
+    // We allocate this with the thread-pool allocator because the destructors
+    // for TIntermNode's are never called. When TIntermNodes are no longer
+    // needed, the pool allocator destroys all memory at once without
```
johnkslang: Yes, we need clear language to distinguish "deallocation" of memory from "destructor" execution.  (In my mind, destroy is a form of destruct.)

glslang/Include/intermediate.h
```diff
@@ -784,7 +784,9 @@ class TIntermAggregate : public TIntermOperator {
...
+    // only want to destroy it, not free the associated memory.
+    ~TIntermAggregate() { pragmaTable->~TPragmaTable(); }
```
johnkslang: Something related to this causes segmentation faults on Windows when run in multi-threaded mode.  I'm going to prune this MR down to just the SPIR-V stuff.

PR #135, opened 2016-01-19
Code-generate SPIR-V blocks in a more readable order

dekimir:


SPIRV/InReadableOrder.cpp
```diff
@@ -0,0 +1,68 @@
...
+    auto mergeInst = block->getMergeInstruction();
+    if (mergeInst) {
+      Id mergeId = mergeInst->getIdOperand(0);
+      mergeBlock =
+          block->getParent().getParent().getInstruction(mergeId)->getBlock();
+      delayed_[mergeBlock] = true;
+    }
+    for (const auto succ : block->getSuccessors())
+      if (succ != mergeBlock) visit(succ);
+    if (mergeBlock) {
```
dneto0: ah, ok.

dneto0:
+1 Looks good to me now.
Suggest @johnkslang take a look.

SPIRV/InReadableOrder.cpp
```diff
@@ -0,0 +1,103 @@
+//
+//Copyright (C) 2016 LunarG, Inc.
```
cnorthrop: This should be Google, I believe.

Test/baseResults/spv.while-continue-break.vert.out
```diff
@@ -60,9 +60,6 @@ Linked vertex stage:
               23:                 Label
                                   Store 25(B) 20
                                   Branch 10
-              26:                 Label
```
johnkslang: So, this gone instead of just rearranged because the traverser doesn't find unreachable code?

SPIRV/InReadableOrder.cpp
```diff
@@ -0,0 +1,103 @@
...
+// function.
+
+#include "spvIR.h"
+
+#include <cassert>
+#include <unordered_map>
+
+using spv::Block;
+using spv::Id;
+
```
johnkslang: I would prefer more similar coding conventions across files in the same library.  Does this interact with the clang-format request #128?  Is it important to the convention you want to hide the nested structure of bodies of if/for when you can fit them on one line?

SPIRV/spvIR.h
```diff
@@ -145,6 +149,7 @@ class Instruction {
     Op opCode;
     std::vector<Id> operands;
     std::string originalString;        // could be optimized away; convenience for getting string operand
+    Block* block;
```
johnkslang: Where is block initialized? I don't see it in the constructor.

Is it just for the last instruction of each block?

Test/baseResults/spv.while-continue-break.vert.out
dekimir: Yes, these are removed as unreachable blocks.  I think that's the right decision; result changes for unreachable blocks can hardly be considered bugs.  I presume we have other tests with live blocks that test the code removed here.

SPIRV/InReadableOrder.cpp
dekimir: This illustrates why I thought `.clang-format` was necessary.  I don't have it on this branch, so my new code is formatted differently from the existing files.  Once an agreed-upon `.clang-format` lands, new code can be made consistent much more easily.

SPIRV/spvIR.h
dekimir: It's zero-initialized in the constructor and later set by `setBlock()`.  That allows us to create instructions without worrying in advance which block they'll land in.  It's also the smallest change to the existing code I could think of.

johnkslang: Maybe I'm looking at the wrong code.  I'd expect a "block(nullptr)" at the end of both constructors, but don't see it in either:

```
     Instruction(Id resultId, Id typeId, Op opCode) : resultId(resultId), typeId(typeId), opCode(opCode) { }
     explicit Instruction(Op opCode) : resultId(NoResult), typeId(NoType), opCode(opCode) { }
```

dekimir: Hm, you're right.  It isn't actually initialized on creation as the code now stands.  I'll add an explicit `nullptr` initialization to all constructors.

dekimir:
Addressed feedback so far.

PR #136, opened 2016-01-20
Don't remove SPIR-V blocks before codegen.

dekimir:
A removed block releases its instructions, so Module::idToInstruction
suddenly contains dangling references.  The original motivation for
block removal was to skip some unreachable blocks, but that's already
achieved by InReadableOrder.cpp.

Also updated stale comments.


dekimir:
@johnkslang This ought to fix memory corruption we've been seeing.

johnkslang:
Looks good, thanks.

PR #148, opened 2016-01-25
Do not generate gl_PerVertex for GLSL 310

greg-lunarg:


glslang/MachineIndependent/Initialize.cpp
```diff
@@ -1720,7 +1720,7 @@ void TBuiltIns::initialize(int version, EProfile profile, int spv)
                 "highp int gl_VertexID;"      // needs qualifier fixed later
                 "highp int gl_InstanceID;"    // needs qualifier fixed later
                 );
-            if (version < 310)
+            if (version <= 310)
```
johnkslang: The other side of the interface would also need to change (it's a test by >= 310), but since AEP may need it the other way, I'm not sure this is the right fix.

johnkslang:
Generally, built-in variables are introduced for the earliest version that an extension has a dependency on.  The gap between that version and when it comes into core is protected by tests that the extension is present.  (There is infrastructure present for how to do that; see comments in Initialize.cpp.)

Basically, this merge request doesn't follow the existing design for how all other built-in variables are handled WRT extensions.

It will help to have an description (which is blank above) of the specific problem; there could well be a bug in how gl_PerVertex is handled between 3.1 and AEP; it may have a different fix that uses the existing methods.

Also, for SPIR-V, (this statement is a bit forward looking) we will always want to generate the gl_PerVertex block.  Probably nothing you can do about that now, just I'm sure for multiple reasons we don't want to regress on the current "< 310" test.

PR #161, opened 2016-02-05
Fix typo in an include guard.

ossy-szeged:


johnkslang:
This is already corrected in the SPIR-V specification (glslang is a consumer, not the source of this), and so should appear the next time (soon) the headers are pulled from SPIR-V into glslang.

Applying anyway, it shouldn't cause a merge conflict, as the changes appear to be the same.

PR #164, opened 2016-02-18
Error out if bison is not found on non-Windows operating systems.

antiagainst:
We cannot just default to use `tools/bison.exe` when it is not on
Windows.


xorgy:
:+1:

johnkslang:
Merging in, but also note interaction with issue #103, which will be resolved soon.

Issue #42, opened 2015-07-24
Poll:  Minimum required MSVC version?


johnkslang:
What should be the minimum version of MSVC required for glslang? 

I'd been setting the bar at 2012 to be conservative across a broad base, but many submissions require 2013, which I then fix to work on 2012.  (2012 supports a broad set of C++11, but not all those supported by 2013.)

Any input on a required minimum version?

xorgy:
Unreal Engine 4 seems to set the bar at 2013. I figure that's a decent benchmark for industry acceptance.

baldurk:
The project I embed glslang in uses 2010, and at least at the moment the subset of the glslang repo that I compile works there. I'm well aware that this is a minority opinion, but consider that one vote for keeping the bar as low as possible.

xorgy:
Yeah, that's the other thing: even if the majority of people would be fine with 2013 (not saying this is the case, it can be pricey if you don't make a living on it), they're not excluded when we lower the bar, but others are when we raise it.

I have no serious interest in C++ features that aren't supported in 2010.


It's a shame Microsoft doesn't offer their compilers for free, that would fix everything. Can't very well tell everyone to build with mingw.

baldurk:
In case anyone finds this issue through searching/browsing and also prefers to use VS2010, for now I'm maintaining a branch [here](https://github.com/baldurk/glslang/tree/vs2010) that compiles on VS2010. Hopefully I'll be able to keep it as a minimal changeset :smile:.

mknejp:
Since VS2015 is out by now you might want to check the Community edition, which is for free for small teams (up to 5 people, check the license for details). Furthermore, as showed at CppCon this year they want to make Clang a fully integrated frontend for Windows development in the next update. They call it clang/c2, which means we'd have a Microsoft supported and fully C++14 compliant compiler for Windows integrated into VS2015.

Unless you aren't eligible for the Community version there is no real reason not to upgrade to VS2015 (and if you aren't you probably can afford Professional anyway). With the clang/c2 update virtually all platforms will have full C++14 support, which could do wonders to glslang's implementation and interface.

xorgy:
@mknejp Not really sure that _can afford_ translates to _should budget_, though VS2015 looks like it'll be a pretty long-term release, so long as they make it possible to update the clang toolchain separately from VS itself.

mknejp:
[...]

PR #175, opened 2016-02-23
Fix compilation issues with MSVC 2010

rdb:
This PR contains a handful of tiny changes that allow compilation using the MSVC 2010 compiler.  We still need to support MSVC 2010 for various reasons, and therefore we need to compile glslang with MSVC 2010 in order to be able to link it in.

Most of the changes replace the use of range-based for loops, and one replaces the use of std::tie.


baldurk:
I've been maintaining a separate branch for this, so I would be quite happy to see it merged upstream.

However usually there's always something that breaks every time I update to latest glslang, and I imagine that will only get worse as people want to use newer and fancier C++11 features. It'd be good to actually settle issue #42 once and for all, to figure out if 2010 support is even desired upstream.

PR #181, opened 2016-02-26
WIP: Add precise qualifier support

### LINKED ISSUE #145: [4.0] Semantics/AST: Full support of `precise`
The `precise` keyword is properly handled in the grammar, but not pushed through the stack.

yavn:
This adds *minimal* support for `precise` keyword. There's not much code or tests to verify if usage is correct.

There's a high chance I made some stupid mistake along the way, and probably it won't work in all contexts. Current tests are rather basic. I added one SPIR-V test. Otherwise I added similar code as for `invariant` keyword.

I considered adding more strict rules for when keyword is allowed, but this caused two error messages to appear, so I dropped the idea. Here's what I mean:

```
parseContext.profileRequires($$.loc, ECoreProfile | ECompatibilityProfile, 320, 0, "precise");  // GL_ARB_gpu_shader5 is written against OpenGL 3.2
parseContext.profileRequires($$.loc, ECoreProfile | ECompatibilityProfile, 400, E_GL_ARB_gpu_shader5, "precise");
```

Fixes #145


yavn:
Is this MR any good? Should I update this or we don't want this change at the moment?

johnkslang:
It looks interesting and on the right path.  I'm unclear on how much of it it does versus how much is left unfinished. I'm interesting in seeing you follow through on it though.  I didn't quite follow the "strict rules" question.  Is ```precise``` being allowed in a version it should not be?

yavn:
Great, thanks for your feedback. I'll update it then.

What I wanted to express with `precise` are these rules:
- `precise` is baseline in OpenGL 4.0
- `precise` is available through extension GL_ARB_gpu_shader5, but requires OpenGL 3.2 or higher

If I'm not mistaken, currently the 3.2 version requirement is not expressed anywhere. I think it shouldn't be allowed to e.g. specify `#version 140` (GL 3.1) and `#extension GL_ARB_gpu_shader5` at the same time.

Now I see I used the wrong version in the original post code snippet, 320 instead of 150 (these are GLSL versions, right?).

johnkslang:
Regarding preventing an extension from being declared in a version earlier than the extension's dependencies section says it can be:  This is generally not implemented or enforced.  Currently, keywords show up in the right core versions, or when the extension is present, and built-in names show up in the right earliest allowed version, guarded by the extension being present.  However, we don't yet verify the extension was not declared in a version earlier than it should have been.  So, let's continue to treat that as a separate subject.

Yes, these are GLSL versions that need to be referred to, not API versions.  They don't line up until 330.

To guard a feature you can give the version it shows up in core, or the extension that supplies it, and don't yet worry about the earliest version the extension is allowed to be used.

yavn:
Okay, thanks for clearing this up. Yeah, I think I ended up implementing this as you describe it. I'll get back to this MR in a week or so, just have to finish some other stuff first.

Qining:
Hi @yavn, are you still working on this CL? I'd like to pick it up if you don't want to work on it.

Thanks

yavn:
Hi, you're free to pick it up if you wish. I wanted to get back into it, but I'm still busy with other assignments which I have to finish first and it will still take me a few days. If you wanted to start right now, then please go ahead.

Qining:
Thanks Maciej, I probably will pick it up later.

johnkslang:
Closing, as this has been superseded by #222.

PR #182, opened 2016-02-27
add missing atom strings

magestik:
Some atom didn't have their string associated in PpAtom.cpp.

This was causing pure preprocessing tools (like Google's shaderc with -E argument) to output invalid GLSL code.

### DIFF
--- a/glslang/MachineIndependent/preprocessor/PpAtom.cpp
+++ b/glslang/MachineIndependent/preprocessor/PpAtom.cpp
@@ -57,7 +57,7 @@ Except as expressly stated in this notice, no other rights or licenses
 express or implied, are granted by NVIDIA herein, including but not
 limited to any patent rights that may be infringed by your derivative
 works or by other works in which the NVIDIA Software may be
-incorporated. No hardware is licensed hereunder. 
+incorporated. No hardware is licensed hereunder.
 
 THE NVIDIA SOFTWARE IS BEING PROVIDED ON AN "AS IS" BASIS, WITHOUT
 WARRANTIES OR CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED,
@@ -98,6 +98,34 @@ const struct {
     int val;
     const char* str;
 } tokens[] = {
+
+    { PpAtomAdd,            "+=" },
+    { PpAtomSub,            "-=" },
+    { PpAtomMul,            "*=" },
+    { PpAtomDiv,            "/=" },
+    { PpAtomMod,            "%=" },
+
+    { PpAtomRight,          ">>" },
+    { PpAtomLeft,           "<<" },
+
+    { PpAtomAnd,            "&&" },
+    { PpAtomOr,             "||" },
+    { PpAtomXor,            "^"  },
+
+    { PpAtomRightAssign,    ">>=" },
+    { PpAtomLeftAssign,     "<<=" },
+    { PpAtomAndAssign,      "&=" },
+    { PpAtomOrAssign,       "|=" },
+    { PpAtomXorAssign,      "^=" },
+
+    { PpAtomEQ,             "==" },
+    { PpAtomNE,             "!=" },
+    { PpAtomGE,             ">=" },
+    { PpAtomLE,             "<=" },
+
+    { PpAtomDecrement,      "--" },
+    { PpAtomIncrement,      "++" },
+
     { PpAtomDefine,         "define" },
     { PpAtomDefined,        "defined" },
     { PpAtomUndef,          "undef" },


xorgy:
:+1:

magestik:
Should I remove it ?

johnkslang:
Is this still an issue?

I don't quite understand the change yet, as it seems to add some, but not all, of the multi-character tokens to the atom list.  So, haven't quite comprehended yet.

Would like to also hear from shaderc folks.  Also, does it have anything to do with enforcing proper GLSL semantics from Khronos?

A PR should normally include test result changes, which also helps communicate the actual improvement being pulled.  Can you add test results?  Or, a statement about why that doesn't make sense in this case?

Alternatively, can you open an issue that describes the functional problem or gives an example?  Then we could possibly see other suggested fixes.

Thanks.

glslang/MachineIndependent/preprocessor/PpAtom.cpp
```diff
@@ -98,6 +98,34 @@ const struct {
...
+    { PpAtomMul,            "*=" },
+    { PpAtomDiv,            "/=" },
+    { PpAtomMod,            "%=" },
+
+    { PpAtomRight,          ">>" },
+    { PpAtomLeft,           "<<" },
+
+    { PpAtomAnd,            "&&" },
+    { PpAtomOr,             "||" },
+    { PpAtomXor,            "^"  },
```
johnkslang: Should this be "^^"?

bschatt:
Thank you for your comment.

I encounter an issue when using Google's glslc in preprocessing mode. I mainly use glslc because it supports the GL_GOOGLE_include_directive extension (it allows to have include in GLSL). 

This particular issue happens when I try to preprocess a simple shader, say [test.frag](https://github.com/KhronosGroup/glslang/files/387012/test.frag.txt). The result is not valid GLSL because it contains some "\<bad token\>" instead of "+=" as you can see in [output.frag](https://github.com/KhronosGroup/glslang/files/387011/output.frag.txt). For reference, the output was produced using the following command : 

`glslc -E test.frag > output.frag`

The proposed patch fix this issue.

This is caused by the GetAtomString which "[Map an already created atom to its string.](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpAtom.cpp#L146)" Well except when the string is not in the map. This patch add almost every missing atom in the tokens array which is used to populate the stingMap. This is happening in [InitAtomTable](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpAtom.cpp#L172) using the AddAtomFixed (which is the only method inserting atoms in the map). 

As you noted, I added **most** of the missing atoms in the tokens array. Here is the list of atoms still missing :
- PpAtomPaste,
- PpAtomIdentifier,
- PpAtomConstInt,
- PpAtomConstUint,
- PpAtomConstInt64,
- PpAtomConstUint64,
- PpAtomConstFloat,
- PpAtomConstDouble,
- PpAtomConstString

Explanations :

1. I don't think PpAtomPaste really matters here. 
2. An Identifier does not have a fixed map. But they are added dynamically in the map anyway (see [PpScanner](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpScanner.cpp#L295)). 
3. All other atoms are constant values which obviously don't have a fixed string value. Unlike identifiers I don't think their string values are added in the map.

johnkslang:
@dneto0 any input on this?  Thanks.

magestik:
Thank you for your comment.

I encounter an issue when using Google's glslc in preprocessing mode. I mainly use glslc because it supports the GL_GOOGLE_include_directive extension (it allows to have include in GLSL). 

This particular issue happens when I try to preprocess a simple shader, say [test.frag](https://github.com/KhronosGroup/glslang/files/387012/test.frag.txt). The result is not valid GLSL because it contains some "\<bad token\>" instead of "+=" as you can see in [output.frag](https://github.com/KhronosGroup/glslang/files/387011/output.frag.txt). For reference, the output was produced using the following command : 

`glslc -E test.frag > output.frag`

The proposed patch fix this issue.

This is caused by the GetAtomString which "[Map an already created atom to its string.](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpAtom.cpp#L146)" Well except when the string is not in the map. This patch add almost every missing atom in the tokens array which is used to populate the stingMap. This is happening in [InitAtomTable](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpAtom.cpp#L172) using the AddAtomFixed (which is the only method inserting atoms in the map). 

As you noted, I added **most** of the missing atoms in the tokens array. Here is the list of atoms still missing :
- PpAtomPaste,
- PpAtomIdentifier,
- PpAtomConstInt,
- PpAtomConstUint,
- PpAtomConstInt64,
- PpAtomConstUint64,
- PpAtomConstFloat,
- PpAtomConstDouble,
- PpAtomConstString

Explanations :

1. I don't think PpAtomPaste really matters here. 
2. An Identifier does not have a fixed map. But they are added dynamically in the map anyway (see [PpScanner](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/preprocessor/PpScanner.cpp#L295)). 
3. All other atoms are constant values which obviously don't have a fixed string value. Unlike identifiers I don't think their string values are added in the map.

glslang/MachineIndependent/preprocessor/PpAtom.cpp
magestik: Yes. Actually I didn't know about the "logical exclusive or" operator.

magestik:
You can also reproduce it in glslangValidator : 

`glslangValidator -E test.frag`

magestik:
I just noticed that glslc is shipped in the latest Android NDK. Same issue.

```
/opt/android-ndk-r12b/shader-tools/linux-x86_64/glslc -E test.frag
```

Output :
```
#version 330

out vec3 outColor;

void main(void)
{
 outColor <bad token> vec3(0.5, 0.5, 0.5);
}
```

johnkslang:
This PR is addressed by commit 9f9bad0, which has the needed correct and test results.

PR #186, opened 2016-03-01
Do not emit SPIR-V Capability GeometryStreams if streams are not used

scyganINTEL:
The GS global out layout defaults was set to "stream = 0", what implied
that streams were always used in shader - TLayoutFormat.hasStream()
was returning TRUE. This happened even when streams qualifier was not used in GLSL.

When compiling to SPIR-V, that has undesired effect of emitting GeometryStreams
capability, which may be not supported.

This change leaves layoutStream initialized to layoutStreamEnd by default,
what causes LayoutFormat.hasStream() to return TRUE only, if stream qualifier
was present in GLSL.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -115,9 +115,6 @@ TParseContext::TParseContext(TSymbolTable& symt, TIntermediate& interm, bool pb,
         language == EShLangTessEvaluation ||
         language == EShLangGeometry)
         globalOutputDefaults.layoutXfbBuffer = 0;
-
-    if (language == EShLangGeometry)
-        globalOutputDefaults.layoutStream = 0;
 }
 
 TParseContext::~TParseContext()


johnkslang:
This looks like a good change to make, and generally tracking the difference between "default" and "not user set" is the right thing to do, and the typical pattern set by other layout qualifiers.

Is there code in place now though that, in the presence of multiple streams, gets the default of 0?  Most likely, the places consuming a stream should check for !hasStream(), and substitute the current default stream.

scyganINTEL:
All places accessing streamLayout are already guarded by `hasStream()`. Especially the SPIR-V does:
```
            if (type.getQualifier().hasStream()) {
                builder.addCapability(spv::CapabilityGeometryStreams);
                builder.addDecoration(spvType, spv::DecorationStream, type.getQualifier().layoutStream);
            }
```
so in default case, no decoration is emitted at all.

johnkslang:
Here is an example of what this breaks in the validator:

```
out outBN {
	layout(stream=0) vec4 bs0;  // member stream must match block stream
} outBI;
```

This is valid GLSL, because outBN inherited the global default of 0, but the pull requests removes the default of 0, and the above code would get an error.  This is regarding the rule that the member stream cannot contradict the block stream.

My response earlier would be correct if the change was to the individual variable streams getting discussed.  However, the PR is about the default, which is indeed 0, and has to be handled somehow.

I'll look a bit more at what I'd suggest here.

johnkslang:
This might be a bit difficult: To distinguish between

- doing multiple streams, but only using stream 0
- not doing multiple streams

In practice you can know from two clues:
- either the shader used EmitStreamVertex()
- or the shader used a stream > 0

That's probably sufficient, but non-trivial.  You just don't know from seeing a variable declaration at the beginning.

johnkslang:
I committed an implementation of the last comment.  Closing this PR, as that supersedes it.

scyganINTEL:
@johnkslang  Thanks!

Issue #191, opened 2016-03-07
Capability MultiViewport not declared when using gl_ViewportIndex 


Qining:
For the following shader source code, the generated SPIR-V code does decorate variable with **ViewportIndex**, but capability: **MultiViewport** is not declared.

```
#version 430
layout(triangles) in;
layout(triangle_strip, max_vertices = 3) out;

layout(location = 0) in vec4 in_color[];
layout(location = 0) out vec4 out_color;

void main() {
	for (int i=0; i<gl_in.length(); ++i) {
		gl_Position = gl_in[i].gl_Position;
		gl_ViewportIndex = int(round(gl_in[i].gl_Position.z * 3.0));
		out_color = in_color[i];
		EmitVertex();
	}
	EndPrimitive();
}
```

Generated SPIR-V code, `OpDecorate %39 BuiltIn ViewportIndex` is decorate operation:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 59
; Schema: 0
               OpCapability Geometry
               OpCapability GeometryPointSize
               OpCapability ClipDistance
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Geometry %4 "main" %26 %31 %39 %49 %52
               OpExecutionMode %4 Triangles
               OpExecutionMode %4 Invocations 1
               OpExecutionMode %4 OutputTriangleStrip
               OpExecutionMode %4 OutputVertices 3
               OpSource GLSL 430
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %4 "main"
               OpName %8 "i"
               OpName %24 "gl_PerVertex"
               OpMemberName %24 0 "gl_Position"
               OpMemberName %24 1 "gl_PointSize"
               OpMemberName %24 2 "gl_ClipDistance"
               OpName %26 ""
               OpName %27 "gl_PerVertex"
               OpMemberName %27 0 "gl_Position"
               OpMemberName %27 1 "gl_PointSize"
               OpMemberName %27 2 "gl_ClipDistance"
               OpName %31 "gl_in"
               OpName %39 "gl_ViewportIndex"
               OpName %49 "out_color"
               OpName %52 "in_color"
               OpMemberDecorate %24 0 BuiltIn Position
               OpMemberDecorate %24 1 BuiltIn PointSize
[...]

PR #192, opened 2016-03-08
Enable adding capability: MultiViewport

Qining:
Fix issue https://github.com/KhronosGroup/glslang/issues/191


johnkslang:
Does this have any effect on test results?

Qining:
Just fixed the issue #191. I didn't any difference on other tests. But still, I didn't run the compiled spirv code, just built the tests and validated the generated code.

johnkslang:
The regression test ```spv.420.geom``` uses ```gl_ViewportIndex``` and should show this change.  We need to make the changes together, so the test results match the code base.

Qining:
Oops, sorry, I just tested with Vulkan-CTS's tests. Just updated with commit be4f52a which adds MultiViewport declaration in spv.420.geom.out.

Issue #185, opened 2016-02-29
Forward referenced IDs undefined


Qining:
The SPIR-V code of the following shader has undefined IDs which are forward referenced with instruction: OpDecorate.

```
#version 310 es
precision mediump float;
layout(location = 0) in highp vec4 dEQP_Position;
layout(location = 1) in float in0;
layout(location = 0) flat out float out0;

float func (float a)
{
	return -a;
	a = a * -1.0;
	return 1.0;
}

void main()
{
	
	out0 = func(in0);
	gl_Position = dEQP_Position;

}
```

The compiled SPIR-V code, **%15** and **%17** are forward referenced by not defined later:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 39
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %4 "main" %22 %24 %31 %35
               OpSource ESSL 310
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %4 "main"
               OpName %10 "func(f1;"
               OpName %9 "a"
               OpName %22 "out0"
               OpName %24 "in0"
               OpName %25 "param"
               OpName %29 "gl_PerVertex"
               OpMemberName %29 0 "gl_Position"
               OpMemberName %29 1 "gl_PointSize"
               OpName %31 ""
               OpName %35 "dEQP_Position"
               OpDecorate %10 RelaxedPrecision
               OpDecorate %9 RelaxedPrecision
               OpDecorate %12 RelaxedPrecision
               OpDecorate %13 RelaxedPrecision
               OpDecorate %15 RelaxedPrecision
               OpDecorate %17 RelaxedPrecision
               OpDecorate %22 RelaxedPrecision
               OpDecorate %22 Flat
               OpDecorate %22 Location 0
               OpDecorate %24 RelaxedPrecision
               OpDecorate %24 Location 1
               OpDecorate %26 RelaxedPrecision
               OpDecorate %27 RelaxedPrecision
               OpMemberDecorate %29 0 BuiltIn Position
               OpMemberDecorate %29 1 BuiltIn PointSize
               OpDecorate %29 Block
               OpDecorate %35 Location 0
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
[...]

PR #193, opened 2016-03-08
Fix undefined forward referenced ID in decorations

Qining:
Fix issue #185 by dumping unreachable blocks (if they are terminated by
branch/return instructions).

Deadcode elimination will be implemented later to remove the unreachable code and related IDs.

### DIFF
--- a/SPIRV/spvIR.h
+++ b/SPIRV/spvIR.h
@@ -293,11 +293,11 @@ public:
         // instructions will be dumped after.
         std::unordered_set<const Block*> dumped;
         inReadableOrder(blocks[0], [&out, &dumped](const Block* b) {
-          b->dump(out);
-          dumped.insert(b);
+            b->dump(out);
+            dumped.insert(b);
         });
         for (auto b : blocks) {
-          if (!dumped.count(b) && b->isTerminated()) b->dump(out);
+            if (!dumped.count(b) && b->isTerminated()) b->dump(out);
         }
         Instruction end(0, 0, OpFunctionEnd);
         end.dump(out);
--- a/SPIRV/spvIR.h
+++ b/SPIRV/spvIR.h
@@ -57,6 +57,7 @@
 #include <functional>
 #include <iostream>
 #include <memory>
+#include <unordered_set>
 #include <vector>
 
 namespace spv {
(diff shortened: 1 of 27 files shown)


johnkslang:
This is a pretty big change to make to the emitted code.  There was some driving reason the dead code was eliminated originally, do we know what that was, and why it is now safe to emit it?

I actually think the smaller perturbation to make is to eliminate the bad decorations, and leave the generated code (which is now being translated by real drivers) the way it currently is.

Qining:
Hi John, @dekimir and I think this is a safe and quick fix of the issue #185 and enable using validator by default in Vulkan CTS.  But eventually we will have deadcode elimination.

SPIRV/spvIR.h
```diff
@@ -287,7 +288,17 @@ class Function {
             parameterInstructions[p]->dump(out);
 
         // Blocks
-        inReadableOrder(blocks[0], [&out](const Block* b) { b->dump(out); });
+        // Dump the reachable blocks in readable order first.
+        // Unreachable blocks which are terminated with branch/return
+        // instructions will be dumped after.
+        std::unordered_set<const Block*> dumped;
+        inReadableOrder(blocks[0], [&out, &dumped](const Block* b) {
+          b->dump(out);
```
dekimir: Indentation incosistent with the surrounding code.

dekimir:
IIRC, we started skipping all unreachable blocks in `dump()` in #135 because it was easier to implement the readable-order traversal that way.  I hadn't anticipated issues like #185.

@johnkslang if you'd rather we eliminated the expression decorations dangling due to non-emitted blocks, we can do that instead.  It'll just take a bit more time, delaying slightly the glslang-output validation in the Vulkan CTS.

johnkslang:
> if you'd rather we eliminated the expression decorations dangling due to non-emitted blocks, we can do that instead

Yes, I think that's the better thing to do, for a couple of reasons.  Thanks!

dekimir:
@johnkslang Just a clarification: do we want this dead-code elimination to happen

- before `dump()`, operating on `Block` and `Instruction` objects, or
- during `dump()`, as clever skipping of some `Instruction`s, or
- after `dump()` as a SPIRV-only optimization pass?

johnkslang:
This is a bit of a special case, because the bug exists, but I don't want to give the extra dead code to the drivers, so whatever is fastest to get us to: the only observed change is the bad decorations disappear.

I suspect that's your first choice:  go delete the dead code right before dumping, and take eliminate the dead decorations at the same time.

This could be later followed up by a more complete DCE pass, or even starting to look at how we want to structure optimization/transform passes, etc. as part of glslang, or independent tools, etc.  But, like I said in the 1st paragraph, this one feels like a special case to me that doesn't need to wait for this bigger answer.

Qining:
Close as the issue is resolved in 5639f3aca5b75cbe5419a623eecf5e3794fab917

PR #194, opened 2016-03-09
Fix array out of bounds bug in processing if-else-endif macros.

antiagainst:
If we are not inside an `if` macro, we cannot simply decrease
`elsetracker`.

Fixes https://github.com/KhronosGroup/glslang/issues/29.


antiagainst:
Although targeted at Issue #29, this patch is not intended to address the "exiting after the first syntax/semantic error" task.

PR #195, opened 2016-03-10
Remove decoration of undefined IDs

Qining:
Fix issue #185 by removing OpDecorate instructions whose target IDs are
defined in unreachable blocks and thus not dumped in the generated
SPIR-V code.

Alternative to PR #193

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -2137,14 +2137,21 @@ void Builder::eliminateDeadDecorations() {
     // Collect IDs defined in unreachable blocks. For each function, label the
     // reachable blocks first. Then for each unreachable block, collect the
     // result IDs of the instructions in it.
-    for (auto& f : module.getFunctions()) {
+    for (std::vector<Function*>::const_iterator fi = module.getFunctions().cbegin();
+        fi != module.getFunctions().cend(); fi++) {
+        Function* f = *fi;
         Block* entry = f->getEntryBlock();
         inReadableOrder(entry, [&reachable_blocks](const Block* b) {
             reachable_blocks.insert(b);
         });
-        for (auto& b : f->getBlocks()) {
+        for (std::vector<Block*>::const_iterator bi = f->getBlocks().cbegin();
+            bi != f->getBlocks().cend(); bi++) {
+            Block* b = *bi;
             if (!reachable_blocks.count(b)) {
-                for (auto& i : b->getInstructions()) {
+                for (std::vector<std::unique_ptr<Instruction> >::const_iterator
+                         ii = b->getInstructions().cbegin();
+                    ii != b->getInstructions().cend(); ii++) {
+                    Instruction* i = ii->get();
                     unreachable_definitions.insert(i->getResultId());
                 }
             }
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -2130,6 +2130,35 @@ Id Builder::accessChainGetInferredType()
     return type;
 }
 
+// comment in header
+void Builder::eliminateDeadDecorations() {
+    std::unordered_set<const Block*> reachable_blocks;
+    std::unordered_set<Id> unreachable_definitions;
+    // Collect IDs defined in unreachable blocks. For each function, label the
+    // reachable blocks first. Then for each unreachable block, collect the
+    // result IDs of the instructions in it.
+    for (auto& f : module.getFunctions()) {
+        Block* entry = f->getEntryBlock();
+        inReadableOrder(entry, [&reachable_blocks](const Block* b) {
+            reachable_blocks.insert(b);
+        });
+        for (auto& b : f->getBlocks()) {
+            if (!reachable_blocks.count(b)) {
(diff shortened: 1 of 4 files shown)


SPIRV/SpvBuilder.cpp
```diff
@@ -2130,6 +2130,35 @@ Id Builder::accessChainGetInferredType()
...
 }
 
+// comment in header
+void Builder::eliminateDeadDecorations() {
+    std::unordered_set<const Block*> reachable_blocks;
+    std::unordered_set<Id> unreachable_definitions;
+    // Collect IDs defined in unreachable blocks. For each function, label the
+    // reachable blocks first. Then for each unreachable block, collect the
+    // result IDs of the instructions in it.
+    for (auto& f : module.getFunctions()) {
```
johnkslang: Range-based loops don't compile on systems people are still using for gslang.  Can you make it a traditional loop?

SPIRV/SpvBuilder.cpp
```diff
@@ -2130,6 +2130,35 @@ Id Builder::accessChainGetInferredType()
...
+        Block* entry = f->getEntryBlock();
+        inReadableOrder(entry, [&reachable_blocks](const Block* b) {
+            reachable_blocks.insert(b);
+        });
+        for (auto& b : f->getBlocks()) {
```
johnkslang: Need traditional loop.

SPIRV/SpvBuilder.cpp
```diff
@@ -2130,6 +2130,35 @@ Id Builder::accessChainGetInferredType()
...
+            if (!reachable_blocks.count(b)) {
+                for (auto& i : b->getInstructions()) {
```
johnkslang: Ditto.

SPIRV/SpvBuilder.cpp
```diff
@@ -2130,6 +2130,35 @@ Id Builder::accessChainGetInferredType()
...
+                    unreachable_definitions.insert(i->getResultId());
+                }
+            }
+        }
+    }
+    decorations.erase(std::remove_if(decorations.begin(), decorations.end(),
```
johnkslang: Since it is SSA, it seems more straightforward to have the decorations be the unordered set, and just remove them directly as the instructions in unreachable blocks are visited.  (For each unreachable block, for each instruction, erase the id: Basically linear in number of IDs that have to be processed anyway, and no intermediate data store.)

The advantage here is that if there is more than one reason we'll want the list of unreachable IDs?

johnkslang:
Thanks Qining, nice clean design, except for a question about the needed complexity (and would prefer more old-fashioned loops).

Are there no test changes with this?  There are many examples of unreachable code in the existing test suite, so I would expect some decorations to disappear, though I guess it's possible none of the dead IDs had decorations?

SPIRV/SpvBuilder.cpp
Qining: One thing I'm not sure is the order of OpDecorate instructions. If we use unordered_set, will the dumping order be deterministic?

Originally I was thinking whether there is a 'Dummy Instruction' which can be skipped during dumping, and a map from ID to instructions is already there in Module. But I didn't find such a dummy instruction.

Anyway, if the order is deterministic cross all platforms, unordered_set should be find.

Qining:
Changed to traditional 'for' loop style in 95aa527.

Hi John, it is interesting that none of the glslang regression test needs to be updated for this fix. It seems like we don't have any dead decorate instruction in our test...

johnkslang:
@Qining, will you be able to use gtest framework to add a test (assuming it goes in soon), or is it better to add test to the existing suite?

SPIRV/SpvBuilder.cpp
```diff
@@ -2130,6 +2130,42 @@ Id Builder::accessChainGetInferredType()
...
+    // reachable blocks first. Then for each unreachable block, collect the
+    // result IDs of the instructions in it.
+    for (std::vector<Function*>::const_iterator fi = module.getFunctions().cbegin();
+        fi != module.getFunctions().cend(); fi++) {
+        Function* f = *fi;
+        Block* entry = f->getEntryBlock();
+        inReadableOrder(entry, [&reachable_blocks](const Block* b) {
+            reachable_blocks.insert(b);
+        });
+        for (std::vector<Block*>::const_iterator bi = f->getBlocks().cbegin();
```
johnkslang: BTW, "auto" is okay!  It's just the range-based bit that is not.

SPIRV/SpvBuilder.cpp
johnkslang: I think you're right, that you'd want an ordered set for cross platform test-result consistency.  That would change the time complexity.  It's okay as is.

Qining: I just realized we have a Type combining ‘list' and 'map', then we can minimize the complexity to O(1) for each remove operation and keep the order untouched.

Qining:
Hi John, just committed a CL to resolve the complexity problem: a801ec56e0a40744a96c220758d00a90c9f0fb9f

Will go ahead to test PR #190 and try to have a test with it.

PR #197, opened 2016-03-10
Fix the complexity issue of removing dead decorate

Qining:
Change the container type of decorate instructions from std:;vector to a
combination of std::list and std::unordered_set/map. This avoid the
copy overhead during eliminateDeadDecorations(), and keep the
instruction dumping order unchanged.

Issue: may be an overkill for now, and the new type should be a
template.


Qining:
Another much easier way is to let the 'Builder' have a set of 'dead IDs', and during dumpInstructions(), check if the 'ID' operands exist in the 'dead IDs' set.

Qining:
Close as this may not be required.

PR #198, opened 2016-03-11
Updated the includer interface to allow relative includes.

AWoloszyn:
This plumbs both the current file path and the include depth
back up to the includer. This allows the includer to properly
support relative paths.

This also replaces the string copy that was done during include
with a zero-copy method of accomplishing the same thing. This
prevents 2 extra copies of entire files.

### DIFF
--- a/glslang/MachineIndependent/Scan.h
+++ b/glslang/MachineIndependent/Scan.h
@@ -51,10 +51,10 @@ const int EndOfInput = -1;
 //
 class TInputScanner {
 public:
-    TInputScanner(int n, const char* const s[], size_t L[], const char* const* names = nullptr, int b = 0, int f = 0) :
+    TInputScanner(int n, const char* const s[], size_t L[], const char* const* names = nullptr, int b = 0, int f = 0, bool single = false) :
         numSources(n),
         sources(reinterpret_cast<const unsigned char* const *>(s)), // up to this point, common usage is "char*", but now we need positive 8-bit characters
-        lengths(L), currentSource(0), currentChar(0), stringBias(b), finale(f)
+        lengths(L), currentSource(0), currentChar(0), stringBias(b), finale(f), singleLogical(single)
     {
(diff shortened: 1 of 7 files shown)


glslang/Public/ShaderLang.h
```diff
@@ -292,25 +292,85 @@ class TShader {
...
     };
 
     // Returns an error message for any #include directive.
     class ForbidInclude : public Includer {
-    public:
-        std::pair<std::string, std::string> include(const char* /*filename*/) const override
-        {
-            return std::make_pair<std::string, std::string>("", "unexpected include directive");
-        }
+     public:
```
johnkslang: General comment: want to see consistent indentation, {}, etc. with local convention.  (It's hard to see the indentation changing size on the screen, etc.)

glslang/MachineIndependent/Scan.h
```diff
@@ -129,7 +136,7 @@ class TInputScanner {
                     }
                     --chIndex;
                 }
-                loc[currentSource].column = (int)(currentChar - chIndex);
+                logicalSourceLoc.column = loc[currentSource].column = (int)(currentChar - chIndex);
```
johnkslang: Please use separate assignments.  Especially since these are separate objects, they can be added/removed by changing the right set of lines, rather than the right set of tokens within a line.

johnkslang: Quick question:  When is the logical column not the column?

(I feel like something more natural is missing regarding having multiple objects tracking the location, but don't have anything constructive to suggest yet.)

AWoloszyn: The logical column will not be the column when you have 2 physical strings that represent one logical string that do not have a line-break between them. 

Currently the use of the logical string is restricted to a single location that happens to have newlines but if it is used elsewhere in the future, then this will continue to work.

AWoloszyn:
Updated from feedback.

PR #204, opened 2016-03-15
SPV: Implement boolean selection for mix().

amdrexu:


SPIRV/GlslangToSpv.cpp
```diff
@@ -3435,11 +3435,14 @@ spv::Id TGlslangToSpvTraverser::createMiscOperation(glslang::TOperator op, spv::
...
     case glslang::EOpMix:
-        if (isFloat)
+        if (! builder.isBoolType(builder.getScalarTypeId(builder.getTypeId(operands.back())))) {
+            assert(isFloat);
             libCall = spv::GLSLstd450FMix;
-        else {
+        } else {
             opCode = spv::OpSelect;
-            spv::MissingFunctionality("translating integer mix to OpSelect");
+            spv::Id operand  = operands.back();
```
johnkslang: Does std::swap() work here?

amdrexu: Yes, it works. I updated the codes. Thank you.

Issue #163, opened 2016-02-16
Composite specialization constants need to be implemented.


johnkslang:
Scalar bool/int/float specialization constants are basically, working, as is gl_WorkgroupSize, but general case composites of these are not yet implemented.

Most likely, first, there is a front-end feature needed to track how a composite's subset is specialized, and then second, there should be a general translation of that to SPIR-V in the function createSpvSpecConstant().

dneto0:
Here's a concrete example that I think illustrates what should compile, and how.

```
#version 450

layout(constant_id = 200) const float myfloat = 1.25;
layout(constant_id = 201) const int myint = 14;

struct structtype {
  float f;
  int i;
};

// This should have an initializer that is an OpSpecConstantComposite
const structtype outer_struct_var = { myfloat, myint };

// This should have an initializer that is an OpSpecConstantComposite
// (Similar to scVec example in https://www.khronos.org/registry/vulkan/specs/misc/GL_KHR_vulkan_glsl.txt)
const vec2 myvec = vec2(myfloat, myfloat);

void main() { }

void outer_composite() {
  const float f = outer_struct_var.f;
  const float i = outer_struct_var.i;
}

void inner_composite() {
  // This should have an initializer that is an OpSpecConstantComposite
  const structtype inner_struct_var = { myfloat, myint };
}
```

Glslang errors out with the following:
ERROR: 0:12: 'const' : non-matching or non-convertible constant type for const initializer 
ERROR: 0:16: 'const' : non-matching or non-convertible constant type for const initializer 
ERROR: 0:27: 'const' : non-matching or non-convertible constant type for const initializer

Qining:
Here is a simpler case in which 'SpecConstant' gets involved, but not yet supported.

Current `TIntermConstUnion`'s `ConstantUnionArray` doesn't support symbol-like elements, which are required so that we can have SpecConstants in the initializer list.

```
#version 450

layout(constant_id = 200) const float myfloat = 1.25;
layout(constant_id = 201) const int myint = 14;

const int myint2 = myint;

void main(){}
```

Hi @johnkslang, are you working on this or similar issues? We do have one issue filed in CTS repo, I wonder how is the progress.

Thanks,

johnkslang:
Yes, I was going to look into this late this week.  I see you already started, will try sorting through that crossed with what I was doing.  I had commented some on design intent, which I think you may have seen.

johnkslang:
These examples now work.

PR #208, opened 2016-03-17
WIP: Implement composite specialization constants

Qining:
Creating specialization constant aggregate variables with specialization
constants and OpSpecConstantComposite should has been already supported, issue #163.

Version/profile/semantics checking for the validity of using
specialization constants is not done yet.

Tests not added yet.

Approach:
1) Propagate 'SpecConstant' label in executeInitializer() to mark
variable's initializer as 'SpecConstant' if it should be marked.

2) Handle contant variables with 'SpecConstant' initializer in
executeInitializer(), propgate the 'SpecConstant' label to the variable
and cache the initializer subtree.

3) Add function to create SpvConstants from initializer subtree. This
may be able to replace the current SpvConstants creation routine which is using
ConstantUnionArray.

Other:
1) Renamed the SpvConstants creation functions:
  createSpvSpecConstant => createSpvConstant
  createSpvConstant => createSpvConstantFromConstUnionArray


Qining:
Example shader:
```
#version 450
layout(constant_id = 200) const float myfloat = 1.25;
layout(constant_id = 201) const int myint = 14;
struct structtype {
  float f;
  int i;
};
const structtype outer_struct_var = {myfloat, myint};
void main(){}
```

Generated SPIR-V code:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 12
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %4 "main"
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %4 "main"
               OpName %10 "structtype"
               OpMemberName %10 0 "f"
               OpMemberName %10 1 "i"
               OpDecorate %7 SpecId 200
               OpDecorate %9 SpecId 201
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
          %7 = OpSpecConstant %6 1.25
          %8 = OpTypeInt 32 1
          %9 = OpSpecConstant %8 14
         %10 = OpTypeStruct %6 %8
         %11 = OpSpecConstantComposite %10 %7 %9
          %4 = OpFunction %2 None %3
          %5 = OpLabel
               OpReturn
               OpFunctionEnd
```

Qining:
@dekimir @dneto0 @antiagainst @AWoloszyn @johnkslang PTAL.

const expr is not yet implemented.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4852,6 +4920,10 @@ TIntermNode* TParseContext::executeInitializer(const TSourceLoc& loc, TIntermTyp
         return nullptr;
     }
 
+    // Check if the initializer subtree is a specialization constants. Change
+    // its storage class to EvqConst and label it as 'SpecConstant' if it is.
+    propagateSpecConstant(initializer);
```
johnkslang: Normally, the tree is built bottom up, and should be built with correct typing each step along the way.

E.g., When two nodes are combined with an operator (which becomes their parent), all the correct promotions and type setting of the result are done at that point in time.  So, no multi-level tree should exist where the intermediate nodes don't already correctly reflect the combination of their children.  Yet, this seems like a recursive "patch up" that implies a multi-level tree was built and left lying around for a while, while carrying incorrect information.

Maybe that works for this, I'm still thinking it through, just want to be clear what the typical pattern is, and see, is there a reason going against it is better?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4819,6 +4825,68 @@ TVariable* TParseContext::declareNonArray(const TSourceLoc& loc, TString& identi
...
 
+namespace {
+
+// Traverse all the child nodes of a given node. propagate 'SpecConstant' label
+// from bottom up.
+// The given node will be labelled with EvqConst storage and 'SpecConstant' iff
+//  1) All of its child nodes are constants, either ConstantUnion or symbols
+//     with constant storage and 'SpecConstant' label.
+//  2) At least one of its child should be 'SpecConstant'.
+void propagateSpecConstant(TIntermTyped* node) {
```
johnkslang: Normally, the tree is built bottom up, and should be built with correct typing each step along the way.

E.g., When two nodes are combined with an operator (which becomes their parent), all the correct promotions and type setting of the result are done at that point in time. So, no multi-level tree should exist where the intermediate nodes don't already correctly reflect the combination of their children. Yet, this seems like a recursive "patch up" that implies a multi-level tree was built and left lying around for a while, while carrying incorrect information.

Maybe that works for this, I'm still thinking it through, just want to be clear what the typical pattern is, and see, is there a reason going against it is better?

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: I think this is going to matter for more than initializers.

glslang/Include/intermediate.h
```diff
@@ -621,10 +621,13 @@ class TIntermSymbol : public TIntermTyped {
     virtual const TIntermSymbol* getAsSymbolNode() const { return this; }
     void setConstArray(const TConstUnionArray& c) { unionArray = c; }
     const TConstUnionArray& getConstArray() const { return unionArray; }
+    void setConstInitializerSubTree(const TIntermTyped* subTree) { constInitializerSubTree = subTree; }
```
johnkslang: I think this is going to matter for more than TIntermSymbol.  Probably TIntermTyped in general?

I can see that for setting a ```layout(constant_id = ...)``` on something, it may seem like we just care about ```symbol = initializer```.  But, these things can crop up other places.  Like

```
layout(constant_id = ...) const int extra = 3;
int a[10 + extra];
```

The array size could be an arbitrarily big tree of constant expressions, but with no related symbol or initializer.

(Sorry if I missed motives here; it could solve a subset of the problem, but I think has to be elsewhere to solve the whole problem.  It is all good food for thought!)

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -484,7 +484,7 @@ TIntermTyped* TParseContext::handleBracketDereference(const TSourceLoc& loc, TIn
     TIntermTyped* result = nullptr;
 
     int indexValue = 0;
-    if (index->getQualifier().storage == EvqConst) {
+    if (index->getQualifier().isFrontEndConstant()) {
```
johnkslang: Not sure I agree here... it is okay for the front-end to check that a specialization constant fits within the default array size.  That would be a useful message to get out about your shader.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -495,15 +495,15 @@ TIntermTyped* TParseContext::handleBracketDereference(const TSourceLoc& loc, TIn
             error(loc, " left of '[' is not of type array, matrix, or vector ", base->getAsSymbolNode()->getName().c_str(), "");
         else
             error(loc, " left of '[' is not of type array, matrix, or vector ", "expression", "");
-    } else if (base->getType().getQualifier().storage == EvqConst && index->getQualifier().storage == EvqConst)
+    } else if (base->getType().getQualifier().isFrontEndConstant() && index->getQualifier().isFrontEndConstant())
```
johnkslang: Yes, it's folding that needs protection.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -495,15 +495,15 @@ TIntermTyped* TParseContext::handleBracketDereference(const TSourceLoc& loc, TIn
...
         return intermediate.foldDereference(base, indexValue, loc);
     else {
-        // at least one of base and index is variable...
+        // at least one of base and index is variable or specialization constant...
 
         if (base->getAsSymbolNode() && isIoResizeArray(base->getType()))
             handleIoResizeArrayAccess(loc, base);
 
-        if (index->getQualifier().storage == EvqConst) {
+        if (index->getQualifier().isFrontEndConstant()) {
```
johnkslang: Safer to have implicit array sizing also take into account known values of specialization constants.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4906,15 +4978,50 @@ TIntermNode* TParseContext::executeInitializer(const TSourceLoc& loc, TIntermTyp
...
+            // propagate to the intermediate node created from this TVariable.
+            variable->getWritableType().getQualifier().makeSpecConstant();
+            // Store the initializer subtree, the tree will be passed to the
+            // intermediate node created from this TVariable, and then be used
+            // when TGlslangToSpvTraverser is traversing intermediate symbol
+            // nodes.
+            variable->setConstInitializerSubTree(initializer);
+
+            // If this constant is not declared at global level, declareVariable()
+            // won't add this constant to AST as a symbol. This is fine for ConstantUnion-like
```
johnkslang: Neither path is adding it to the AST right now.  The nominal way is it gets used as a symbol, and at that point addSymbol() would create the usage for the AST as a symbol node.  If it never gets used, it's dead.  The exception is the external interface, which the linker needs to see.  That's the final linkage branch of the AST, and the globals symbols can end up in that list for that reason.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4906,15 +4978,50 @@ TIntermNode* TParseContext::executeInitializer(const TSourceLoc& loc, TIntermTyp
...
+            // constants, but not for SpecConstants, as SpecConstants are more symbol-like and
+            // needs a symbol node to represent itself.
+            if (!symbolTable.atGlobalLevel()) {
+                TIntermSymbol* sn = intermediate.addSymbol(*variable, loc);
+                sn->setConstInitializerSubTree(initializer);
```
johnkslang: I suspect this should not be done.  The return from this function is code to emit to initialize something at execution time.  But, a constant should not result in emitted code.  When a symbol node is translated to something like SPIR-V, the symbol node will have the information needed (the const subtree from above) to do the *declarative* instructions for the specialization constant, which aren't part of the *procedural* instruction stream.  The procedural instruction stream will just refer to the SPIR-V <id> of the specialization constant.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: The specification (KHR_vulkan_glsl) is also specific about what operations propagate "specialization constantness", so it makes sense to localize the propagation to where those operators are implemented.  (That is, a general algorithm would accept too many things as being valid.)

glslang/MachineIndependent/ParseHelper.cpp
Qining: Do you mean a situation like:

```
int array[a_spec_const_int];

void main() {
    array[4] = blabla
}
```
Here we probably need to check a_spec_const_int > 4?

glslang/MachineIndependent/ParseHelper.cpp
Qining: I think for internal variables, the call to addSymbol() at line 5025 will add the variable as a symbol node to AST. Usually constants (which I think should be considered as external interface), should not be added as symbol node here. But I found a case like:

```
layout(constant_id=100) const int a_spec_int = 5;
void main() {
    const int another_spec_int = a_spec_int + 1;
}
```

I remember in this case, when we execuateInitializer() runs upon another_spec_int, symbolTable.atGlobalLevel() return false, so that the call to intermediate.addSymbolLinkageNode() at line 4775, will not be called. In the end, Another_spec_int will not be added to AST.

My approach is actually a quick patch, I agree this is not a preferable way to solve the problem.

glslang/MachineIndependent/ParseHelper.cpp
Qining: I think it is not easy to check the index value for spec constants, like you said, spec constants can be a arbitrarily big tree. And even if we get the default value computed from the tree, that is still from the default value, I wonder how useful such a check can be.

If index is a spec constant and we don't use isFrontEndConstant() but isConstant(), we will get a seg fault as getAsConstantUnion() on the spec constant will return a nullptr.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: > In the end, Another_spec_int will not be added to AST.

Right, it should not be, it was never code and not part of any interface.

Qining: I see, I tested with:
```
out int OutInt;
void main() {
    const int another_spec_int = a_spec_int + 1;
    OutInt = another_spec_int;
}
```
And I can see the `another_spec_int` is added to the AST now. I wonder when it is added to the AST, as it is not added to the global level (so not in the linker Objects), and it is not added as a normal variable in the next `else` branch, I guess it is related to its usage: the assignment of OutInt, but I didn't find exactly the place.

Thanks,

johnkslang: > I wonder when it is added to the AST,

Should be when the "OutInt = another_spec_int" is recognized.

Fundamental design:  The AST has no symbol table, it only represents procedural statements, with the exception of the artificially added on branch for linkage objects.  So, normally, something gets in the AST when it needs to be procedurally executed.

> as it is not added to the global level (so not in the linker Objects),

Right.  And historically, the linker objects weren't in the AST either.  Typically, declarative stuff is not in the AST.

Qining:
Close as #211 replaces this CL.

Issue #144, opened 2016-01-23
[AST]  - implicitly-sized gl_ClipDistance[] 


johnkslang:
Implicitly-sized gl_ClipDistance[] (at least in tessellation shaders) with sizes greater than one are not getting sizes greater than one.

johnkslang:
Fixed by pull request #214.

PR #214, opened 2016-03-27
Parser: Update array size of gl_ClipDistance/gl_CullDistance in gl_in[].

amdrexu:


johnkslang:
I believe this addresses existing issue #144.

PR #215, opened 2016-03-29
Spec Constant Operations

Qining:
Approach:
Add a flag in `Builder` to indicate 'spec constant mode' and 'normal
mode'. When the builder is in 'normal mode', nothing changed. When the
builder is in 'spec constant mode', binary, unary and other instruction
creation rountines will be redirected to `createSpecConstantOp()` to
create instrution at module level with `OpSpecConstantOp <original
opcode> <operands>`.

'spec constant mode' should be enabled if and only if we are creating
spec constants. So a flager setter/recover guard is added when handling
binary/unary nodes in `createSpvConstantsFromConstSubTree()`.

Note when handling spec constants which are represented as ConstantUnion
Node, we should not use `OpSpecConstantOp` to initialize the composite
constant, so builder is set to 'normal mode'.

Tests:
Tests are added in Test/spv.specConstantOperations.vert, including:

1) Arithmetic, shift opeations for both scalar and composite type spec constants.
2) Size conversion from/to float and double for both scalar and vector.
3) Bitwise and/or/xor for both scalar and vector.
4) Unary negate/not for both scalar and vector.
5) Vector swizzles.
6) Comparisons for scalars.
7) == and != for composite type spec constants

Issues:
1) To implement == and != for composite type spec constants, the Spec needs
to allow OpAll, OpAny, OpFOrdEqual, OpFUnordEqual, OpOrdNotEqual,
OpFUnordNotEqual. Currently none of them are allowed in the Spec.

2) Other comparison opcode for float scalar type spec constants are not added, so that cases like
```
const bool c = spec_float_a < spec_float_b;
const bool d = spec_float_a <= spec_float_b;
const bool e = spec_float_a > spec_float_b;
const bool f = spec_float_a >= spec_float_b;
```
are not supported.
Similar to supporting vector type == and !=, supporting such these cases needs add extra float comparison opcodes to the set of allowed opcodes of OpSpecConstantOp.


Qining:
@johnkslang @dneto0 @yavn PTAL.

Probably we should split the tests into 'scalar' and 'vector' two groups?

johnkslang:
> 'spec constant mode' should be enabled if and only if we are creating spec constants.

I thinking through this... :)  Seems like on the right track, as early feedback; will have more time later to give more detailed feedback.

SPIRV/SpvBuilder.cpp
```diff
@@ -1208,6 +1229,102 @@ Id Builder::createOp(Op opCode, Id typeId, const std::vector<Id>& operands)
...
+        case OpSLessThan:
+        case OpUGreaterThan:
+        case OpSGreaterThan:
+        case OpULessThanEqual:
+        case OpSLessThanEqual:
+        case OpUGreaterThanEqual:
+        case OpSGreaterThanEqual:
+        // Added temporarily to enable compsite type spec constants comparison.
+        // Remove this comment after Spec being updated.
+        case OpAll:
```
Qining: Not yet decided if we want to add those opcodes here.

SPIRV/SpvBuilder.cpp
```diff
@@ -1208,6 +1229,102 @@ Id Builder::createOp(Op opCode, Id typeId, const std::vector<Id>& operands)
...
+        case OpFSub:
+        case OpFMul:
+        case OpFDiv:
+        case OpFRem:
+        case OpFMod:
+        case OpAccessChain:
+        case OpInBoundsAccessChain:
+        case OpPtrAccessChain:
+        case OpInBoundsPtrAccessChain:
+            addCapability(CapabilityKernel);
```
Qining: If glslang is not supposed to add Kernel capability, probably we should return NoResult instead of adding capability here?

SPIRV/GlslangToSpv.cpp
```diff
@@ -3854,15 +3854,37 @@ spv::Id TGlslangToSpvTraverser::createSpvConstantFromConstUnionArray(const glsla
     return builder.makeCompositeConstant(typeId, spvConsts);
 }
 
+namespace {
+class SpecConstOpCodeGenerationSettingGuard{
+  public:
```
johnkslang: Four space indention.

SPIRV/GlslangToSpv.cpp
```diff
@@ -3854,15 +3854,37 @@ spv::Id TGlslangToSpvTraverser::createSpvConstantFromConstUnionArray(const glsla
...
+  private:
+    spv::Builder* builder_;
+    bool previous_flag_;
+};
+}
+
 // Create constant ID from const initializer sub tree.
 spv::Id TGlslangToSpvTraverser::createSpvConstantFromConstSubTree(
-    const glslang::TIntermTyped* subTree) {
+    glslang::TIntermTyped* subTree) {
```
johnkslang: Function { on beginning of its own line.

SPIRV/SpvBuilder.cpp
johnkslang: One way or the other:  Glslang should not add the the Kernel capability.

The intention from SPIR-V was to not have to do all these operations on floating point.  You'll adhere to the SPIR-V spec. without doing these, and glslang's job isn't to add them back, so I'd defer these.  Maybe they show up in 1.1 or 2.0, or maybe there is not a compelling need for them and they'll never show up.

SPIRV/SpvBuilder.cpp
```diff
@@ -1208,6 +1229,102 @@ Id Builder::createOp(Op opCode, Id typeId, const std::vector<Id>& operands)
...
+        case OpAny:
+        case OpFOrdEqual:
+        case OpFUnordEqual:
+        case OpFOrdNotEqual:
+        case OpFUnordNotEqual:
+            break;
+
+        // OpCodes that need Shader capability.
+        case OpQuantizeToF16:
+            addCapability(CapabilityShader);
```
johnkslang: I think this is redundant, as it's a shader compiler, and it always has some capability that implies shader.

SPIRV/SpvBuilder.cpp
Qining: Agree, I'll also remove the OpAll and OpAny with all the float point equal/not equal opcodes in the block before this one.

SPIRV/SpvBuilder.cpp
Qining: Hmm, I think I prefer to keep it here to explicitly follow the spec. If you think the call to `addCapability()` definitely should be removed, I'll do that and merge the `OpQuantizeToF16` to the previous block.

SPIRV/GlslangToSpv.cpp
```diff
@@ -3854,15 +3854,37 @@ spv::Id TGlslangToSpvTraverser::createSpvConstantFromConstUnionArray(const glsla
...
 }
 
+namespace {
+class SpecConstOpCodeGenerationSettingGuard{
+  public:
+    SpecConstOpCodeGenerationSettingGuard(spv::Builder* builder,
+                                         bool shouldGeneratingForSpecConst)
+        : builder_(builder) {
+        previous_flag_ = builder->isInSpecConstCodeGenMode();
+        shouldGeneratingForSpecConst ? builder->setToSpecConstCodeGenMode()
```
johnkslang: Is this a simple command to do it, and it must then happen?  Just naming, but I think this should not be "should".  Also, the instruction to use is OpSpecConstantOp

How about "generateSpecConstantOp" instead of "shouldGeneratingForSpecConst"?

SPIRV/GlslangToSpv.cpp
```diff
@@ -3854,15 +3854,37 @@ spv::Id TGlslangToSpvTraverser::createSpvConstantFromConstUnionArray(const glsla
     return builder.makeCompositeConstant(typeId, spvConsts);
 }
 
+namespace {
+class SpecConstOpCodeGenerationSettingGuard{
+  public:
+    SpecConstOpCodeGenerationSettingGuard(spv::Builder* builder,
+                                         bool shouldGeneratingForSpecConst)
```
johnkslang: When would this be **false**?  I don't think a subtree of a constant-op subtree can be a non-constan-op subtree.  So, the only state changes that will ever happen are:

1. non-const-op -> const-op, while going down the tree
2. const-op -> non-const-op, while going up the tree

You get 2. from the nesting, so I think it might be clearer (unless I'm missing something) to drop the ability to pick the state (e.g., reader wondering what does **true** mean anyway, when you see it in use? You don't know without coming and reading this and realizing the above anyway).

SPIRV/SpvBuilder.cpp
johnkslang: The spec intends you to take advantage of the "capability tree", such that not all capabilities have to be explicitly declared, just enough to imply the tree.

SPIRV/SpvBuilder.cpp
johnkslang: Maybe we're lost in talking at the same time, so I'll make this a separate comment:

Summary summary: I don't think we need this list of opcodes. It creates an unnecessary thing to maintain.  Instead, use the validator to find places the front-end made bad requests in the AST.

SPIRV/GlslangToSpv.cpp
Qining: Yes, you are right. In my previous draft I also explicitly turn off the spec mode for constant unions, but now I realized that is unnecessary.

SPIRV/GlslangToSpv.cpp
Qining: Done.

SPIRV/GlslangToSpv.cpp
Qining: the bool parameter removed.

SPIRV/GlslangToSpv.cpp
Qining: Done

Issue #213, opened 2016-03-25
Specialization constants in expressions


yavn:
Hi, I prepared a bunch of expression tests involving specialization constants (available in internal CTS repo). I want to summarize here what I found is missing (tested against the master branch).

@johnkslang, @Qining let me know if you think something is not an error, or if I wrote a syntax that violates the spec. Thanks.

All shader examples begin with code similar to the one below, so I won't be repeating it:

```
#version 440

layout(constant_id = 1) const int sc0 = 2;
layout(constant_id = 2) const int sc1 = 3;

layout (set = 0, binding = 0, std430) writeonly buffer Output {
    int result;
} sb_out;
```

##### 1. constant expression with specialization constant

```
const int expr0 = sc0 + 1;
const int expr1 = sc0 + sc1;

void main (void)
{
    sb_out.result = expr0 + expr1;
}
```

Compilation error:

```
ERROR: 0:10: '=' : global const initializers must be constant 'const int'
```

##### 2. array size expression with specialization constant

```
void main (void)
{
    int a0[sc0 + 3];
    int a1[sc0 + sc1];

    const int size0 = sc0 + 3;
    const int size1 = sc0 + sc1;

    for (int i = 0; i < size0; ++i)
        a0[i] = 3 - i;
    for (int i = 0; i < size1; ++i)
        a1[i] = 5 - i;

    sb_out.r0 = a0[size0 - 1];
    sb_out.r1 = a1[size1 - 1];
}
```

This compiles, but I think generates incorrect SPIR-V. Resulting array is sized with an expression that is evaluated inside the main function.

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 71
; Schema: 0
OpCapability Shader
%1 = OpExtInstImport "GLSL.std.450"
OpMemoryModel Logical GLSL450
OpEntryPoint GLCompute %4 "main"
OpExecutionMode %4 LocalSize 1 1 1
OpSource GLSL 440
OpName %4 "main"
OpName %8 "size0"
OpName %12 "size1"
OpName %15 "i"
OpName %29 "a0"
OpName %37 "i"
OpName %49 "a1"
OpName %57 "Output"
OpMemberName %57 0 "r0"
OpMemberName %57 1 "r1"
OpName %59 "sb_out"
OpDecorate %9 SpecId 1
OpDecorate %13 SpecId 2
OpMemberDecorate %57 0 Offset 0
[...]

PR #216, opened 2016-04-01
WIP: Fix the wrong generated code when an array is declared with size calculated from spec constants

Qining:
fix the Case 2 in issue #213.


Qining:
@yavn @johnkslang PTAL. This should fix the case 2 in issue #213

johnkslang:
There may be a design issue here.  The specs would say if you (say) add two spec. constants together, you get a spec. constant. There are two approaches:

1. Bottom up: Always decide to use OpSpecConstantOp based on what the operands are.
2. Top down: Only decide to use OpSpecConstantOp based on where it's getting used.  E.g., initializer, array size, ...

The specs are written as if 1. is the way to do it, but I'm a bit worried the current design of using a "spec-const guard" is pursuing 2.  The question is, will there more be places?

Here's a question that might help get to the bottom of it:

```
... a; // spec-const
... b; // spec-const

if (d == a + b) ...  // is the subexpression (a + b) a spec-const?
```

(Note we had some similar discussion in the front end about an initializer being a "top down" point to do spec-const work, but to have it all work in general, it had to be done "bottom up".

Qining:
I think our design is fine. In the case of your example, `( a + b )` is not a spec-constant.

Whether a node is a spec-constant depends on both up-down and bottom-up directions of speculation.

An operation with only spec constants and constants CAN BE a spec constants, but not necessary to be one. It only becomes a spec constants when it is used in(itself is) a 'const' sink.

So, `const` is propagated up-down, `specConst` is propagated bottom-up.

This workflow actually depends on the traversing order when we parse the source, we need to stick to sort of 'post-order' traversing, like:

1) Get the request to handle an const declaration (let's call this parent node).
2) Handle the child nodes which build up the requested const declaration first, along with `const`.
3) Child nodes processed, and `specConst` should be carried from these child nodes to the const to be declared (parent node).
4) Process the parent node with ready-to-use child nodes and `specConst`.

If our parent node is not a const declaration, we won't affect the processing of child nodes, but in step 3), we don't carry `specConst` to our parent node (as it is not asked to be a const at all), so the result parent node won't be a spec constant.

In short, we should turn on SpecConstantOpMode when
1) The current node is asked to be a constant
2) There are spec constant nodes in its child nodes.

To guarantee we know both conditions, we need to process child nodes first with `const` (top-down) and then current node with processed child nodes with `specConst` determined (bottom-up). And I think this is what we have in our code.

Qining:
Oops, I read it wrong I think. Probably the Spec is asking to make spec constants whenever it is possible (my new understanding).

But to make `(a + b)` a spec constant, the only thing I think we need to change is 'aggressively make spec constants'. That is whenever we can make an operation to be a spec constants, make it. But the up-down and bottom-up flow should be the same, I think.

And this is probably an optimization, treating `(a+b)` as run-time operation generating valid code, but not optimal one.

johnkslang:
The key test for a good design in this case is:  Do the spec. constants natural work out to be spec. constants everywhere they might need to be, or do we have to keep finding all the right places to call the "guard" code.  I can imagine there is a good solution that still involves the guard mechanism.

Qining:
hmm, I don't see a good reason to keep the guard, if we aggressively creating constants.

I think the most straight-forward way is: when creating SPIRV instructions, check the specConstant'ness of all of its operands. Redirect to createSpecConstantOp() if any of them are spec constants and all of them are constants. This needs a method to check specConstant'ness with a given ID (or intermNode?), but I think we do have enough tool already to do that. Generally just change the `if(generatingSpecConstantOp)` condition in those create*() methods.

If we aggressively creating constants, I think the guard is not necessary. As the guard is used to prevent generating `OpSpecConstantOp` even when there are spec constants in its operands. And because we don't have `OpSpecConstantOp` for assignment, I think we won't have trouble dropping the guard.

johnkslang:
I agree. :)

Qining:
Hi @johnkslang , I just tried to make constants aggressively whenever possible. The approach is:
Removing `SpecConstantOpModeGuard`, use `Builder::isConstant()` to check if the operand IDs are representing constants. If so, make the result ID a constant with proper instructions.

However, I got some problems by doing this.

1) This might be a minor issue. Many `CompositeConstruct` instructions are replaced with `ConstantComposite` instructions. This is because the condition change in Builder::smearScalar().
Before, we use `generatingOpCodeForSpecConst` flag to determine if the smearing instruction should be  `SpecConstantOp` or `CompositeConstruct`.
Now, the condition is `isConstant()`, so whenever the `scalar` to be smeared is a constant, we make the result composite type variable a constant. This is a superset then SpecConstants, so it replaces `CompositeConstruct` with `ConstantComposite`.
However, confine the condition to `isSpecConstant()` (need an extra small function to do so) should be able to remove the issue and leave the tests unchanged, but this way we still have the redundant run-time instructions.

2) Operations allowed in with OpSpecConstantOp. The issue is that, aggressively making constants will make OpSpecConstantOp instructions with not allowed OpCodes in many cases. And changing the condition from `isConstant()` to `isSpecConstant()` won't fix this issue.
For example, in spv.specConstant.vert
```
out vec4 color;
layout(constant_id = 122) const uint dupScale = 2;
...
    if (spDupBool)
        color *= dupScale;
```

When we have SpecConstantOpModeGuard, the conversion of `dupScale` from uint type to float type will be translated to runtime instructions. But without the guard, the conversion will be implemented with `OpSpecConstant ConvertUToF`, but `ConvertUToF` needs kernel capability and should not be allowed here.

To solve the problem, we will need to check the opcode along with the const'ness of operand ids, this means a list of allowed opcodes, like the one we have before, and I don't think we like that design.

Otherwise, we will need to have another label for the AST nodes to identify which operations are allowed to be implemented as constant-result instructions. But this sounds like some non-trivial extra work in the front end, actually I doubt if it is worthy to do so.

Personally now I prefer a little bit more to the `SpecConstantOpModeGuard` way, as it won't touch the run-time instructions generated before. This way still have the problem of generating OpSpecConstantOp with not allowed opcodes, if the user write invalid glsl source, but it won't make old run-time code replaced with invalid OpSpecConstantOp instructions and force us to differentiate allowed opcodes and not-allowed one for now. And I think constant propagation can be added in future optimization?

Qining:
Hi @johnkslang , what is your opinion about 'aggressively making constants' ?

Thanks,

Qining

johnkslang:
> will be implemented with OpSpecConstant ConvertUToF, but ConvertUToF needs kernel capability and should not be allowed here.

Did the AST say the resulting type was a spec. constant?  If so, that's part of the change we need to make to make the front-end and spec. agree to the same subset.

That is, the "aggressively make spec. constant" should really be part of the front-end logic, and the front-end should intersect that with the right list from the spec. (which we need to resolve).  Then, the AST -> SPIR-V should just do what the AST says to do.  And, that doesn't mean "if operands are spec-const, then opcode and result are", but rather it means "if AST result node is spec-const, then opcode and result are".

Seems we are bouncing back and forth, that's okay, but I believe it is converging on the right answer, if the above sounds right.  Maybe that does mean use the "guard" design, or maybe it just means you need a richer test for when operating on spec-consts is to be done with a spec-const op (by checking AST node).

Easy to believe for the moment the front-end is doing too many nodes as spec-const, but there's a spec bug for that, and we can divide and conquer; we can do this in any order, it just won't all be correct until all components are correct.

Qining:
Hmm, I think this means, in the back-end, I should have some thing like
```
if (node->getType().getQualifier().isSpecConstant() {
    // turn on spec constant mode?
    // or redirect to createSpecConstantOp() directly, but this means adding this if-statement to nearly every return.
}
```
to methods like: `visitBinary()`, `visitUnary()`, etc.

Is this correct?

johnkslang:
Yes, that sounds right.  The guard design already required testing `generatingOpCodeForSpecConst` at the "bottom" (point of emitting the opcode) for every opcode that could be effected.  The new part is to turn on the guard mode at the top (by the test in your comment) for any node that could request it.

This should only be a few places though; not a long list:

- `visitBinary`, includes swizzle
- `visitUnary`, includes type conversion

Is that it?  What does that leave out?  That actually seems pretty clean.

Qining:
Hi @johnkslang , I just pushed (and rebased) a CL with the "guard" moved to visitBinary() and visitUnary(). The good news is that, as we expect, tests not related to spec constants are not touched.

In the spec constant related tests, we can see that many operations with float result type are wrapped with SpecConstantOp, which matches our expectation. So the next is to confine the propagation of spec-constant label according to the spec, in the front-end?

Qining:
Er, I found issue #213 is closed, but I think this CL is still blocking it...

johnkslang:
>  So the next is to confine the propagation of spec-constant label according to the spec, in the front-end?

Yes, I have that code locally, will push soon.

Test/baseResults/spv.specConstantOperations.vert.out
```diff
@@ -16,101 +16,116 @@ Linked vertex stage:
...
+              97:     25(int) Constant 2048
+              98:   68(ivec4) ConstantComposite 97 97 97 97
+              99:   68(ivec4) SpecConstantOp 198 71 98
+             100:     25(int) Constant 0
+             102:             TypeVector 6(int) 2
+             104:             TypeVector 6(int) 3
          4(main):           2 Function None 3
                5:             Label
                               Return
+             101:      6(int) CompositeExtract 67 0
```
Qining: This CL has side effect somewhere. Instruction from line 120 to line 123 should be wrapped with SpecConstantOp and moved out of function.

johnkslang: I suspect a guard setting call was removed, but does not yet have the corresponding node having its spec-constness set.  That is, after all the correct nodes are set, it would start working again.

(E.g., a constructor creating the RHS of an initializer, where the LHS is spec-const, but the RHS is not set that way, but should be.)

Qining: My bad, you are right. This is just because I forget to add a guard in visitSymbol().

johnkslang:
I'm going to get this combined with my code, all on master, and we should be able to fine tune from there.

Issue #218, opened 2016-04-02
Specialization constant built-in


yavn:
I'm well aware this is getting old by now, but there's one more (last?) issue regarding specialization constants. As per spec, it should be possible to specialize the value of a built-in. Consider the following shader:

```
layout(constant_id = 1) gl_MaxImageUnits;

layout (set = 0, binding = 0, std430) writeonly buffer Output {
    int maxImageUnits;
} sb_out;

void main (void)
{
    sb_out.maxImageUnits = gl_MaxImageUnits;
}
```

I suspect the problem is that built-in constants like this don't exist in SPIR-V and are simply optimized out if possible. In the above case, I'm just getting an ordinary constant:

```
OpCapability Shader
%1 = OpExtInstImport "GLSL.std.450"
OpMemoryModel Logical GLSL450
OpEntryPoint GLCompute %4 "main"
OpExecutionMode %4 LocalSize 1 1 1
OpSource GLSL 440
OpName %4 "main"
OpName %7 "Output"
OpMemberName %7 0 "maxImageUnits"
OpName %9 "sb_out"
OpMemberDecorate %7 0 NonReadable
OpMemberDecorate %7 0 Offset 0
OpDecorate %7 BufferBlock
OpDecorate %9 DescriptorSet 0
OpDecorate %9 Binding 0
%2 = OpTypeVoid
%3 = OpTypeFunction %2
%6 = OpTypeInt 32 1
%7 = OpTypeStruct %6
%8 = OpTypePointer Uniform %7
%9 = OpVariable %8 Uniform
%10 = OpConstant %6 0
%11 = OpConstant %6 8              ; this should've been a spec. constant
%12 = OpTypePointer Uniform %6
%4 = OpFunction %2 None %3
%5 = OpLabel
%13 = OpAccessChain %12 %9 %10
OpStore %13 %11
OpReturn
OpFunctionEnd
```

Qining:
Hi @yavn , PR #220 should fix this issue :)

PR #220, opened 2016-04-04
Built-in values declared as specialization constant

Qining:
Fix issue #218.

Support declaring built-in values as spec constants.

Refine the code in createSpvConstant().


yavn:
Hi, it didn't work for me, and I think the reason is that this instruction is not emitted:

```
OpDecorate %11 SpecId 1
```

Otherwise the code looked good to me. I'd expect GLSL builtin to be treated exactly as an ordinary specialization constant, so there should be no difference in SPIR-V. Only GLSL syntax is special.

Qining:
Hi @yavn, SpecId decoration should has been added now.

HI @johnkslang , PTAL, I'm not sure my solution here to assign a spec constant id to an existing symbol is correct and consistent with other parts.

johnkslang:
This looks good to me.

Issue #221, opened 2016-04-06
TIntermediate memory corruption


yavn:
Not sure what's going on, but adding `TString entryPoint` to `TIntermediate` causes segfaults when I run glslang from CTS.

Here's [this line](https://github.com/KhronosGroup/glslang/blob/hlsl-frontend/glslang/MachineIndependent/localintermediate.h#L347) in context.

The program crashes on `TPoolAllocator` most of the time, [here](https://github.com/KhronosGroup/glslang/blob/hlsl-frontend/glslang/Include/PoolAlloc.h#L220).

I didn't have time to debug it further, but I think the problem is with how TString's memory is being managed. Once I removed `entryPoint` (and fixed compilation erros), the issue stopped.

johnkslang:
I can fix this, probably later today.  I suspect the string memory lives in the wrong pool (one that got deallocated).

johnkslang:
I had trouble reproducing this, but changed the memory that holds the entry point.  Let me know at your convenience if this helped.

yavn:
Yeah, I expected it to be totally random, I was just lucky to get a consistent case and was able to pinpoint the bug.

This change fixes the issue for me, closing.

PR #222, opened 2016-04-07
Support precise qualifier

Qining:
Based on @yavn 's prior work:
ad62991

Goal:
Arithmetic instructions that contribute to calculating 'precise'
variables, should be decorated with 'NoContraction' decorator.

Approach:
Add a AST traversing pass at the end of shader parsing, called in
DoFullParse(), to propagate 'noContraction' qualifier to AST nodes. Then
in GlslangToSpvTraverser, add decoration for the nodes with
'noContraction' qualifier.

To propagate 'noContraction' qualifier along lvalues, we need to use
symbol node IDs to record the involved lvalues.

The pass to propagate 'noContraction' qualifier is consist of three parts:

1) Populate the 'precise' qualifiers to bottom symbol nodes as preprocessing

Some 'precise' qualifiers are not attached to symbol nodes, but their
parent nodes, which might be binary nodes, unary nodes or aggregate
nodes (not very sure here). Populate the 'precise' of these nodes to
their bottom symbol nodes so that we can use symbol IDs to propagate
'NoContraction' through lvalues.

Example shader:
```

layout(triangles, equal_spacing) in;
layout(location = 0) in highp vec2 in_te_position[];
layout(location = 0) out mediump vec4 in_f_color;
precise gl_Position;

void main(void) {
  highp vec2 pos = vec2(3.0, 4.0);
  pos += 0.04;
  gl_Position = vec4(pos, 0.0, 1.0);
}
```
In the AST, a binary node representing gl_Position is marked as 'precise', but
not its underlying symbol node:
```
move second child to first child (temp highp 4-component vector of float)
  gl_Position: direct index for structure (precise gl_Position highp 4-component vector of float Position)
    'anon@0' (out block{precise gl_Position highp 4-component vector of float Position gl_Position, gl_PointSize highp float PointSize gl_PointSize})
    Constant:
      0 (const uint)
  Construct vec4 (temp highp 4-component vector of float)
    'pos' (temp highp 2-component vector of float)
    Constant:
      0.000000
    Constant:
      1.000000
```
This first procedure will populate 'precise' from the binary node:
`gl_Position`, to its underlying symbol node: `anon@0`.

2) Collect symbol and their defining nodes (assignment nodes), and
collect a set of symbol IDs to be processed as work set with
source-specified 'precise' variables.

3) Start with the initial 'precise' variables collected. Mark arithmetic
instructions that contribute to these 'precise' variables in the
defining nodes (the assignment node subtrees) as 'noContraction', and
add the symbols involved to the work set. Remove the processed symbol
IDs from the work set. Repeat until the work set is empty.

In the back-end (GlslangToSpvTraverser), when visiting binary/unary
nodes, check if the current node is a 'noContraction' node. If so, add
'NoContraction' decoration upon the rvalue ID of the binary/unary
operation result.

Expectation:
1) In the AST, all the arithmetic operations that lead to defining
'precise' symbols are marked with 'noContraction' qualifier.
2) In the emitted SPIV code, 'NoContraction' decorations are added to
arithmetic operations. Note the 'NoContraction' decorations are not
referencing 'precise' variables (the pointers), but the arithmetic
operations.

Known issue:
The operations marked as 'noContraction' are superset of the 'optimal'
set. The problem is that we didn't build dominator tree.
Example:
```
precise int result;
result = a + b + 2;
a = d + 3;
```
Because symbol a is involved in the calculation of result, once before,
`a` will be added to the work set, then the addition `d + 3` will be
marked with `noContraction`. But it is actually not necessary.


Qining:
Lots of the changes are generated by bison.

Will fix format later.

Qining:
Some basic tests added, not very sure whether the generated code matches what we expect.

@johnkslang @dneto0 PTAL.

yavn:
Hi, just wanted to give an update that this in its current state doesn't solve my original problem. I need decorations on these, for example:

```
precise gl_TessLevelOuter;    // in tessellation control
precise gl_Position;          // in tessellation evaluation
```

The first one should generate an `OpDecorate`, the second is `OpMemberDecorate` (inside `gl_PerVertex`).

I know this is WIP, so just mentioning it for the record.

Qining:
Oops, just found the initial issue filed in gitlab. Thanks @yavn.

Qining:
Hi @yavn, could you give me a complete test shader? I found I can not get the complete testing shaders by running the CTS with branch 302 on gitlab. It shows all the tests pass, I guess they are disabled? And it seems that extracting the shader source directly from the merge request CL is not that straight-forward.

And I found running CTS with latest glslang will crash. It seems the problem is about setting 'entry point'.

yavn:
Make sure you're rebased on master (in glslang). John recently fixed this issue, see here #221 

The test you need is `dEQP-VK.tessellation.common_edge.*`. It is excluded from mustpass, but if you run directly from IDE then it should execute just fine. 6 tests pass and 6 fail.

The problem though is if you update glslang to test your change, CTS will stop compiling as there was a bunch of incompatible changes. Basically you have to modify CTS-side CMakeLists.txt files to include new sources in glslang (e.g. HLSL related).

I can paste the shaders here if you'd like to try to run them through stand-alone glslang. It should be sufficient to see if correct SPIR-V is generated.

```
#version 310 es
#extension GL_EXT_tessellation_shader : require
#extension GL_EXT_gpu_shader5 : require

layout(vertices = 3) out;

layout(location = 0) in highp vec2  in_tc_position[];
layout(location = 1) in highp float in_tc_tessParam[];

layout(location = 0) out highp vec2 in_te_position[];

precise gl_TessLevelOuter;

void main (void)
{
    in_te_position[gl_InvocationID] = in_tc_position[gl_InvocationID];

    gl_TessLevelInner[0] = 5.0;
    gl_TessLevelInner[1] = 5.0;

    gl_TessLevelOuter[0] = 1.0 + 59.0 * 0.5 * (in_tc_tessParam[1] + in_tc_tessParam[2]);
    gl_TessLevelOuter[1] = 1.0 + 59.0 * 0.5 * (in_tc_tessParam[2] + in_tc_tessParam[0]);
    gl_TessLevelOuter[2] = 1.0 + 59.0 * 0.5 * (in_tc_tessParam[0] + in_tc_tessParam[1]);
}
```

```
#version 310 es
#extension GL_EXT_tessellation_shader : require
#extension GL_EXT_gpu_shader5 : require

layout(triangles, equal_spacing) in;

layout(location = 0) in highp vec2 in_te_position[];

layout(location = 0) out mediump vec4 in_f_color;

precise gl_Position;

void main (void)
{
    highp vec2 pos = gl_TessCoord.x*in_te_position[0] + gl_TessCoord.y*in_te_position[1] + gl_TessCoord.z*in_te_position[2];

    highp float f = sqrt(3.0 * min(gl_TessCoord.x, min(gl_TessCoord.y, gl_TessCoord.z))) * 0.5 + 0.5;
    in_f_color = vec4(gl_TessCoord*f, 1.0);

    // Offset the position slightly, based on the parity of the bits in the float representation.
    // This is done to detect possible small differences in edge vertex positions between patches.
    uvec2 bits = floatBitsToUint(pos);
    uint numBits = 0u;
    for (uint i = 0u; i < 32u; i++)
        numBits += ((bits[0] >> i) & 1u) + ((bits[1] >> i) & 1u);
    pos += float(numBits&1u)*0.04;

    gl_Position = vec4(pos, 0.0, 1.0);
}
```

Qining:
Hi @yavn,

Thanks for telling me to rebase, that does fix the crash.

I think the first shader is actually processed correctly here. The `NoContraction` decoration should only be applied upon `arithmetic operations` contributing to `precise` variables. But the variables themselves, which are represented as "variables with pointer type" in SPIRV, should not be decorated with `NoDecoration`. I attached the SPIRV code of the first shader below.

@dekimir please let me know if I'm wrong somewhere.

The second shader is not handled correctly for now, I'm working on it.

The SPIRV code of the first shader, on my machine looks fine, as instructions with result ID:
52, 53, 54, 60, 61, 62, 68, 69, 70 are decorated with `NoContraction` but not 40 which is the `gl_TessLevelOuter` variable.
```
                              Capability Tessellation
                              MemoryModel Logical GLSL450
                              EntryPoint TessellationControl 4  "main" 12 15 20 30 40 45
                              ExecutionMode 4 OutputVertices 3
                              Source ESSL 310
                              SourceExtension  "GL_EXT_gpu_shader5"
                              SourceExtension  "GL_EXT_shader_io_blocks"
                              SourceExtension  "GL_EXT_tessellation_shader"
                              Name 4  "main"
                              Name 12  "in_te_position"
                              Name 15  "gl_InvocationID"
                              Name 20  "in_tc_position"
                              Name 30  "gl_TessLevelInner"
                              Name 40  "gl_TessLevelOuter"
                              Name 45  "in_tc_tessParam"
                              Decorate 12(in_te_position) Location 0
                              Decorate 15(gl_InvocationID) BuiltIn InvocationId
                              Decorate 20(in_tc_position) Location 0
                              Decorate 30(gl_TessLevelInner) Patch
                              Decorate 30(gl_TessLevelInner) BuiltIn TessLevelInner
                              Decorate 40(gl_TessLevelOuter) Patch
                              Decorate 40(gl_TessLevelOuter) BuiltIn TessLevelOuter
                              Decorate 45(in_tc_tessParam) Location 1
                              Decorate 52 NoContraction
                              Decorate 53 NoContraction
                              Decorate 54 NoContraction
                              Decorate 60 NoContraction
                              Decorate 61 NoContraction
                              Decorate 62 NoContraction
                              Decorate 68 NoContraction
                              Decorate 69 NoContraction
                              Decorate 70 NoContraction
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 2
               8:             TypeInt 32 0
               9:      8(int) Constant 3
              10:             TypeArray 7(fvec2) 9
              11:             TypePointer Output 10
12(in_te_position):     11(ptr) Variable Output
              13:             TypeInt 32 1
              14:             TypePointer Input 13(int)
15(gl_InvocationID):     14(ptr) Variable Input
              17:      8(int) Constant 32
              18:             TypeArray 7(fvec2) 17
              19:             TypePointer Input 18
20(in_tc_position):     19(ptr) Variable Input
              22:             TypePointer Input 7(fvec2)
              25:             TypePointer Output 7(fvec2)
              27:      8(int) Constant 2
              28:             TypeArray 6(float) 27
              29:             TypePointer Output 28
30(gl_TessLevelInner):     29(ptr) Variable Output
              31:     13(int) Constant 0
              32:    6(float) Constant 1084227584
              33:             TypePointer Output 6(float)
              35:     13(int) Constant 1
              37:      8(int) Constant 4
              38:             TypeArray 6(float) 37
              39:             TypePointer Output 38
40(gl_TessLevelOuter):     39(ptr) Variable Output
              41:    6(float) Constant 1065353216
              42:    6(float) Constant 1105985536
              43:             TypeArray 6(float) 17
              44:             TypePointer Input 43
45(in_tc_tessParam):     44(ptr) Variable Input
              46:             TypePointer Input 6(float)
              49:     13(int) Constant 2
               5:             Label
              16:     13(int) Load 15(gl_InvocationID)
              21:     13(int) Load 15(gl_InvocationID)
              23:     22(ptr) AccessChain 20(in_tc_position) 21
              24:    7(fvec2) Load 23
              26:     25(ptr) AccessChain 12(in_te_position) 16
                              Store 26 24
              34:     33(ptr) AccessChain 30(gl_TessLevelInner) 31
                              Store 34 32
              36:     33(ptr) AccessChain 30(gl_TessLevelInner) 35
                              Store 36 32
              47:     46(ptr) AccessChain 45(in_tc_tessParam) 35
              48:    6(float) Load 47
              50:     46(ptr) AccessChain 45(in_tc_tessParam) 49
              51:    6(float) Load 50
              52:    6(float) FAdd 48 51
              53:    6(float) FMul 42 52
              54:    6(float) FAdd 41 53
              55:     33(ptr) AccessChain 40(gl_TessLevelOuter) 31
                              Store 55 54
              56:     46(ptr) AccessChain 45(in_tc_tessParam) 49
              57:    6(float) Load 56
              58:     46(ptr) AccessChain 45(in_tc_tessParam) 31
              59:    6(float) Load 58
              60:    6(float) FAdd 57 59
              61:    6(float) FMul 42 60
              62:    6(float) FAdd 41 61
              63:     33(ptr) AccessChain 40(gl_TessLevelOuter) 35
                              Store 63 62
              64:     46(ptr) AccessChain 45(in_tc_tessParam) 31
              65:    6(float) Load 64
              66:     46(ptr) AccessChain 45(in_tc_tessParam) 35
              67:    6(float) Load 66
              68:    6(float) FAdd 65 67
              69:    6(float) FMul 42 68
              70:    6(float) FAdd 41 69
              71:     33(ptr) AccessChain 40(gl_TessLevelOuter) 49
                              Store 71 70
                              Return
                              FunctionEnd

```

Qining:
Hi @yavn, I tried this CL with the common_edge tests on my machine, I think all of them passed, as arithmetic instructions to calculate 'precise' variables are all decorated with 'NoContraction'. 

@yavn @johnkslang @dneto0 PTAL.

glslang/Include/Types.h
```diff
@@ -440,6 +442,8 @@ class TQualifier {
     bool readonly     : 1;
     bool writeonly    : 1;
     bool specConstant : 1;  // having a constant_id is not sufficient: expressions have no id, but are still specConstant
+    bool precise      : 1;
```
dneto0: You should explain precise and noContraction better.
What kinds of things are they applied to?
In GLSL precise is applied to a variable.
I assume that in the end state of the intermediate representation that noContraction will be applied to any expression which could feed into an assignment to a variable that is marked as precise.

Document your data and the code should be evident.  -- paraphrasing Fred Brooks.

dneto0:
Please rebase this against master.  Without that, the gtest-based tests fail to link.

glslang/MachineIndependent/glslang_tab.cpp.h
```diff
@@ -1,19 +1,19 @@
-/* A Bison parser, made by GNU Bison 2.7.  */
```
dneto0: Did the bison parser really change?  If so I would have expected a change to the glslang.y file.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,490 @@
...
+namespace {
+
+using SymbolID = int;
+
+// A mapping from Symbol IDs of symbol nodes, to their defining operation
+// nodes.
+using SymbolMapping = std::unordered_multimap<SymbolID, glslang::TIntermTyped *>;
+
+// A set of Symbol IDs of symbol nodes which are considered as 'precise'
+// symbols. The initial elements of this set should be the IDs of symbols
```
dneto0: Nitpick.  This is a type definition.  So it doesn't have an initial set of values.

Also, I'd call this SymboIDSet and then the variable would be named PreciseSymbolIDs and have the comment about initial values.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,490 @@
...
+        symbol_node_to_be_defined_ = nullptr;
+        return false;
+    }
+}
+
+// Traverse the AST and return a tuple consists of:
+// 1) a mapping from symbol IDs to the definition nodes of these symbols
+// 2) a set of symbol IDs of 'precise' symbol nodes.
+std::tuple<SymbolMapping, PreciseSymbolIDs>
+getSymbolToDefinitionMappingAndPreciseSymbolIDs(glslang::TIntermediate &intermediate)
```
dneto0: Can we make the intermediate parameter const?

johnkslang:
Do we really need both "noContraction" and "precise"?  I haven't studied this deeply yet, but every leaf symbol and intermediate node uses the same type system, and I think these mean the same thing.  I think you could just use "noContraction" everywhere.

Qining:
Yes, we could just use 'noContraction' only. Using both 'precise' and 'noContraction' here is just to make the code and AST output more legible. I can remove the 'precise' if you like.

johnkslang:
Sounds good to me.  I'd rather see `NoContraction` everywhere than a mix, and that reduces confusion understanding the code as well.  Thanks.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -466,8 +462,13 @@ void PropagateNoContraction(glslang::TIntermediate &intermediate)
     auto symbol_mapping_and_precise_symbol_ids =
         getSymbolToDefinitionMappingAndPreciseSymbolIDs(intermediate);
 
+    // A mapping symbol node IDs to their defining nodes enables us to get the
```
dneto0: Missing the word "of":   "A mapping of symbol node IDs ..."

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+//
+// Author: Qining Lu
+//
+// Visit the nodes in the glslang intermediate tree representation to
+// propagate 'precise' label.
+//
+
+#include "propagateNoContraction.h"
+#include "localintermediate.h"
```
dneto0: Nitpick.

Ideally include order is:
  - my interface
  - system includes
  - other library headers
  - headers from my library

For rationale, see https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,491 @@
...
+// 'precise' qualifiers specified upon binary/unary/aggregate nodes to their
+// bottom symbol nodes. An example of use case is like in tessellation
+// evaluation shader:
+//
+//  precise gl_Position;
+//  void main() { gl_Position = <some calculation>; }
+//
+// The 'precise' qualifier is specified upon a binary node with opcode:
+// EOpIndexDirectStruct, left node: an annoymous symbol node, right node: <some
+// calculation>. We populate the 'precise' qualifier from the binary node to
```
dneto0: I think "propagate" works better here than "populate".

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: addressed in dcf9c87

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: addressed in daf9c87d48e1488c8be6c6ab14413de74520d230

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: changed to "propagate" in daf9c87d48e1488c8be6c6ab14413de74520d230

Qining:
Hi @johnkslang, I've already removed 'precise' qualifier flag. Now we use 'noContraction' for both symbol nodes and arithmetic operation nodes. But only the 'noContraction' of the arithmetic nodes will be translated to 'NoContraction' decoration in the SPIRV code.

glslang/MachineIndependent/glslang_tab.cpp.h
Qining: Oop! I missed the change in glslang.y when I rebase yavn's work to master.

Already added back.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Addressed in 9b63f0d86e77b599ca17589c36c6f7e3ffb7b189

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Addressed in 9b63f0d86e77b599ca17589c36c6f7e3ffb7b189

yavn:
Hi, just confirming it works for me. Thanks!

Test/precise.vert
```diff
@@ -0,0 +1,70 @@
...
+  int a = 10;
+  int b = 20;
+  result = a = b + 4; // b + 4 should be 'noContraction'.
+}
+
+void convert() {
+  precise double result;
+  int a = 10;
+  int b = 20;
+  b = a + b; //  a + b should be 'noContraction'.
```
dneto0: Nice.

dneto0:
+1 from me
Overall, the code changes look good to me.  I'm not familiar enough to make sure the lists of enums for arithmetic and assignment operations in the new traversers are complete.
I took a close look at the precise.vert test case and it looks good.  The other test cases were a bit overwhelming, and I didn't check them closely.

Qining:
Rebased to 63b280bbbdbdfaf54634596368cbf4de602c6a00

Hi @johnkslang, any suggestions for this CL?

SPIRV/GlslangToSpv.cpp
```diff
@@ -591,6 +600,8 @@ void InheritQualifiers(glslang::TQualifier& child, const glslang::TQualifier& pa
         child.readonly = true;
     if (parent.writeonly)
         child.writeonly = true;
+    if (parent.noContraction)
```
johnkslang: I don't think this should be done, but agree it's subtle:

1. In SPIR-V, **precise** should have migrated to the arithmetic instructions that need them.
2. Did you find any discussion of **precise** inheritance in the GLSL specification? I was thinking this didn't exist.
3. It would seem to be the front-end's job to apply GLSL semantics to know what operations are  **NoContraction**, and label the nodes accordingly.  However, I can see a design where the details are skipped, as long as a superset gets labeled in SPIR-V, so I'm not yet clear on location of the smarts (still reading...).

SPIRV/GlslangToSpv.cpp
```diff
@@ -599,7 +610,8 @@ bool HasNonLayoutQualifiers(const glslang::TQualifier& qualifier)
     // - struct members can inherit from a struct declaration
     // - effect decorations on the struct members (note smooth does not, and expecting something like volatile to effect the whole object)
     // - are not part of the offset/st430/etc or row/column-major layout
-    return qualifier.invariant || qualifier.nopersp || qualifier.flat || qualifier.centroid || qualifier.patch || qualifier.sample || qualifier.hasLocation();
+    return qualifier.invariant || qualifier.nopersp || qualifier.flat || qualifier.centroid || qualifier.patch || qualifier.sample || qualifier.hasLocation() ||
+           qualifier.noContraction;
```
johnkslang: This needs to be revisited once the inheritance question is answered.

SPIRV/GlslangToSpv.cpp
johnkslang: It looks like GLSL is underspecified here.  All the focus was on the meaning to operations, not the type of data.

Test/precise.vert
```diff
@@ -0,0 +1,70 @@
+#version 450
+
+double minimal() {
+  precise double result = 5.0;
```
johnkslang: It's probably rare to do this for a double or a vertex shader. Usually, **invariant** is sufficient for this type of operation. Any test updates should move their center of mass toward the purpose, which is float operations in tessellation control output.

See section 4.9 of the specification to make sure we have tests that cover the intended purpose, which primarily is whether 

   a * b + c * d 

gets the same result under permutation of a, b, c, and d.

Test/spv.precise.vert
```diff
@@ -0,0 +1,70 @@
...
+  for (int i = 0; i < 3; ++i) {
+    // a's dereference + 2 should be 'noContraction'.
+    result += a0[i] + 2;
+    // result + 1 and 3 - rvalue should be 'noContraction'.
+    a0[i] = 3 - result++;
+  }
+}
+
+void loop_while() {
+  precise double result = 5.0;
```
johnkslang: Are there platforms that will support **NoContraction** but not support **double**?  I think it doesn't matter, just more concern about tests lining up with need.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3421,6 +3422,7 @@ void TParseContext::redeclareBuiltinBlock(const TSourceLoc& loc, TTypeList& newT
             oldType.getQualifier().smooth = newType.getQualifier().smooth;
             oldType.getQualifier().flat = newType.getQualifier().flat;
             oldType.getQualifier().nopersp = newType.getQualifier().nopersp;
+            oldType.getQualifier().noContraction = newType.getQualifier().noContraction;
```
johnkslang: Is this to support redeclaring a built-in block as being **precise**?

glslang/MachineIndependent/glslang_tab.cpp.h
```diff
@@ -26,292 +26,291 @@
...
-     MEDIUM_PRECISION = 509,
-     LOW_PRECISION = 510,
-     PRECISION = 511,
-     PACKED = 512,
-     RESOURCE = 513,
-     SUPERP = 514
-   };
+  enum yytokentype
+  {
+    ATTRIBUTE = 258,
```
johnkslang: Why are all these changes?  Would be better to see just the real changes.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,502 @@
...
+// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+// BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+//
+// Author: Qining Lu
```
johnkslang: Generally, glslang hasn't been doing this.  Is there a convention to point to for this?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,502 @@
+//
+// Copyright (C) 2014-2015 LunarG, Inc.
```
johnkslang: For a new file, we can just mention the company that is doing the work.  (No LunarG.)

glslang/MachineIndependent/propagateNoContraction.h
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright (C) 2014-2015 LunarG, Inc.
```
johnkslang: No LunarG.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+            case glslang::EOpAddCarry:
+            case glslang::EOpSubBorrow:
+            case glslang::EOpUMulExtended:
+            case glslang::EOpIMulExtended:
+
+            case glslang::EOpPostIncrement:
+            case glslang::EOpPostDecrement:
+            case glslang::EOpPreIncrement:
+            case glslang::EOpPreDecrement:
```
johnkslang: Does FMA also need to be considered?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,502 @@
...
+// The 'noContraction' qualifier (specified through the keyword 'precise') is
+// labelled upon a binary node (instead of a symbol node as usual cases) with
+// opcode: EOpIndexDirectStruct, left node: an annoymous symbol node, right
+// node: <some calculation>. We propagate the 'noContraction' qualifier from
+// the binary node to the bottom symbol node: the annoymous symbol node, so
+// that we can use the symbol ID later to identify different symbols.
+// Otherwise, when we collect initial work list with 'noContraction' symbol IDs,
+// we will miss the underlying symbol nodes then miss all the arithmetic
+// operations contributing to it.
+//
```
johnkslang: Design question:

Is this implementing the structure inheritance being implied by the other code?  I'm not 100% what inheritance should be supported, but worry a bit that I see some inheritance stuff happening after this algorithm, which brings into question what the point is.

Here's a question we should resolve that will help us nail this:  Is the following supposed to be supported, and if so, does this design implement it?

```
struct S { float f; };

precise out b1 {
    S s;
} inst1;

out b2 {
    S s;
} inst2;

...
    
    inst1.s.f = .... // nodes in computation labeled as precise
    inst2.s.f = .... // nodes in computation not labeled as precise
```

Now, I think this is far from the problem **precise** was trying to solve, and suspect no one cares, and it would be better if the spec was clear (if not already) on this.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,502 @@
...
+    }
+    return false;
+}
+
+// Visits a binary node and builds the mapping from symbol IDs to the definition
+// nodes of the symbols here.
+bool TSymbolDefinitionCollectingTraverser::visitBinary(glslang::TVisit /* visit */,
+                                                       glslang::TIntermBinary *node)
+{
+    if (isAssignOperation(node->getOp()))
```
johnkslang: Throughout, the coding standard of this project is open curly on same line as if/for/switch (just not for multi-line functions), and preference of no curlies for one-line substatement.

johnkslang:
I review much of it in detail, but not all.  I'd like to converge on the high-level questions about redeclarations and inheritance before finishing my detailed review.

The key point of **precise** was to get simple **float** multiplies and adds to be invariant under operand permutation, so a good first step is getting that working.  Ideally, it wouldn't involve all this other stuff with blocks/structs, but we do need to pin it down, one way or another.

glslang/MachineIndependent/glslang_tab.cpp.h
Qining: All of this is caused by adding a new qualifier 'precise' to be recognizable to glslang and we need to rerun bison....

johnkslang: But, these lines are all the same, and newlines should be canonical, so I don't understand yet.

johnkslang:
Late breaking news: The next revision of GLSL 4.5 will clarify that **precise** can apply in the broad sense, to a block, with recursive inheritance, which clears up a lot of what I was raising from the current specification.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: I guess not, as in GlslangToSpvBuiler.cpp, EOpFMA needs built-in functions from std450.

Test/precise.vert
Qining: Changed to tesc shader in 9a5b7968e42acc5d80c6db710a16141df987b8a8.

Also adopt a * b + c * d as the minimal test.

Test/spv.precise.vert
Qining: You are right, we should just test 'precise' and nothing else here. Changed all double to "float" in 9a5b7968e42acc5d80c6db710a16141df987b8a8.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Hmm, I remember I saw it somewhere before... Anyway, removed in 28b9c669aa2eddb0e3e54d5c2ccba3e29de35d1c.

And this commit also fix the wrong company name in the licenses.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Removed in 9a5b7968e42acc5d80c6db710a16141df987b8a8.

glslang/MachineIndependent/propagateNoContraction.h
Qining: Removed in 9a5b7968e42acc5d80c6db710a16141df987b8a8

glslang/MachineIndependent/glslang_tab.cpp.h
Qining: It turns out that "indentation" is the bad guy...Original code has 5-space indent, this CL has 4 instead.... yytokentype list actually doesn't change...

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: This CL will label `inst1` computation as `NoContraction`, and `inst2` computation won't be touched. I think this is what we expect. But I think there isn't "inheritance" in such cases.

However, if we have the `precise` on `f`:
```
struct S {precise float f; float f2;};
out b1 {S s;} inst1;
out b2 {S s;} inst2;
void main(){
  float a = 1.0;
  float b = 2.0;
  float c = 3.0;
  float d = 4.0;
  inst1.s.f = a + b;
  inst2.s.f = c + d;
  inst2.s.f2 = 1.0 + d;

```

`inst1` and `inst2` will be labelled as 'NoContration' (or say 'precise'), then all the struct members will be considered as 'NoContraction'. So the operation `1.0 + d` will be decorated with 'NoContraction‘. This is a real redundant 'NoContraction'. Still looking into this case to find the effect of inheritance and propagation.

Qining: Oh, I see, I misunderstood before.

This inheritance copy of 'NoContraction‘ from "parent" to "child" is what one of the things I stole  from @yavn's commit ad62991ee0ca92be12cfaf8cf0598f90b34550bb.

Actually I think I can just get rid of it here. As this is in the back-end, there is no need to propagate 'NoDecoration' any more, all the propagation should has been done in propagateNoDecoration pass.

In your example, that means, in the back-end, we don't care whether `inst1` or `inst1.s.f` is labelled as `NoContraction` at all, because we already marked all the computing operations that lead to it with `NoContraction` before.

But in the `propagateNoContraction` pass, we depends on the shader-specified `precise` qualifier of variables. As the `precise` is attached upon `inst1`, all the involved computing operations will be marked as `NoContraction`.

The example in my last comment is another totally new thing....

SPIRV/GlslangToSpv.cpp
Qining: Most of the code in GlslangToSpv.cpp is stolen from @yavn's commit ad62991ee0ca92be12cfaf8cf0598f90b34550bb.

But, actually there is no need to propagate 'NoContraction' in the back-end anymore (and should n't), as 'NoContraction' labels are only effective when attached to an operation node, and those nodes are marked in propagateNoContraction pass.

So, I can just remove this assignment here.

johnkslang: Makes sense, and understand about history.

glslang/MachineIndependent/propagateNoContraction.cpp
johnkslang: I see there are a few other "Author"; inconsistent, but all in the SPIRV folder.  I'm inclined to remove them for consistency.

Test/precise.vert
johnkslang: Okay, thanks.  A mix of everything is fine, I was just seeing a lack of emphasis on the main motivation.

glslang/MachineIndependent/glslang_tab.cpp.h
johnkslang: Reverting is fine, more interested in what's going on.  If it's a tool version, etc., wouldn't like to see bouncing back-and-forth between two files always saying everything changed.  But, it's not worth much effort...

glslang/MachineIndependent/propagateNoContraction.cpp
johnkslang: Does it get marked as **NoContraction**?  Maybe a hole in the spec?  It's just as bad to contract `a*b+c` as it is to uncontract `fma`.

Qining: I've tried with this example:
```
precise out float Out;

void main(){
  float a = 1.0;
  float b = 2.0;
  float c = 3.0;
  a = fma(a, b, c);
  Out = a;
}
```

As fma() is a builtin function and in AST it is translated into EOpFma opcode, it does not have `NoContraction` decoration. The generated code does not have any `NoContraction` decoration.

johnkslang:
Note that merging this in should close issue #145.

Qining:
@johnkslang @dneto0 @yavn PTAL

Re-implemented the code. It turns out that "preventing the precise'ness of some struct members spreading to other unrelated members in the same struct instance" makes the thing much more complex.

In short, assume we have `s.a.b` where `s` is a struct instance and `s.a.b` is 'precise'.

We need to be able to propagate 'precise' in all the following cases:
```
s = some_object // some_object.a.b should be 'precise'
s.a = some_object // some_object.b should be 'precise'
s.a.b = some_object // some_object should be 'precise'
s.a = struct constructor (may be recursive) // index should match (multi level)
s.a.b.c = some_object // some_object should be 'precise'
```

One problem is how to find all the defining nodes of `s, s.a, s.a.b...` with, say, `s.a.b` labelled as 'precise'.

Another problem is how to propagate the 'precise' from the left to the right.

The current code is not optimal in terms of time-complexity. But relatively simpler to implement. With an nested 'precise' object node of a symbol, this code still finds the defining node through the symbol ID, but then it traverse all these defining nodes to check weather the node to be defined is truly a 'precise' node. This will walk through the left node subtree with the accesschain info of the given nested 'precise' object node.

Once the precise'ness of this defining node is confirmed, we can traverse the right subtree and mark all the arithmetic operations as 'noContraction'. And we can use the accesschain paths of the involved object nodes and the accesschain info of the left node to obtain the accesschain paths of the new  'precise' object in the right side. Such the accesschain paths of such new 'precise' objects will be added to the worklist and will be used to find their defining nodes in future.

johnkslang:
> It turns out that "preventing the precise'ness of some struct members spreading to other unrelated members in the same struct instance" makes the thing much more complex.

I'll need some convincing to see this as much harder than keeping the "floatness" of a structure member from spreading to integer members.  Once inheritance of **precise** is done down to member level, it is a characteristic only of members, never of structures.

Just is if we allowed:

```
float struct S {
    int a;
    b;
};
```

and at struct-parsing time, turn that into

```
struct S {
    int a;
    float b;
} s;  // s is not precise, but s.b is
```

and then completely forget that **float** was at the struct level. That's now ancient history.

> s.a.b = some_object // some_object should be 'precise'

If `b` is a struct, then it is not precise, just some of its terminal members are.  `b` was only declared as precise as syntactic sugar.  That set some terminal members to precise, and not others.  So `b` is just an object that has some members as precise.  Likewise, `some_object` must be *the same* type as `b`, so it will already have declared the same members as precise, and there is nothing to even worry about or propagate with this assignment.

> accesschain info of the given nested 'precise' object node.

Note the base does not matter.  The inheritance of **precise** when the struct was parsed will put precise on the right members.  It is the type of the members that matters.  Don't start with an base that's a struct and consider it a **precise** struct.  Base structs are never precise.  When they are dereferenced up the chain, when you get to a terminal member, then you will know preciseness.

Qining:
HI @johnkslang,

I see that transferring 'precise' down to members makes the AST looks different. But I don't think it solve all the problems.

Below is an example:
```
some_object = another_object; // exp1
...
precise float result = some_object.a.b; // exp2
```

At parsing time, we will only see that `result` is precise in exp2. And yes, we can then say the `some_object.a` in exp2 is precise. But, `some_object` in exp2 and `some_object` in exp1 are different nodes in the AST (the memory address). So anything we do with the `some_object` in exp2 does not affect `some_object` in exp1 at all. Besides, the getType() upon these two nodes are also different ( I see it is good design in many cases, just not convenient here).

Note that exp1 are parsed before exp2, so when we parse exp1, we don't know `some_object` contains a precise member. That is the reason why we need to back-propagate `precise` in expressions. So we can not solve all the problem in the parsing phase.

So we need to have a map to tell us where to find the defining node of objects which have newly found 'precise' qualifiers, like how to find exp1 after we know `some_object` has 'precise' a precise member.

Currently I'm using 'symbol ID' as the key to build such a map.

Another problem I have is that, propagate the 'precise' in `some_object` in exp1 to `another_object`. Currently I'm using string-type access chain info to record the path to 'precise' node. Here, assume we do have the `some_object`'s type updated after we find `some_object`, how to copy the 'precise' in the left of exp1 to the right of exp1? Do we have the code already? My last commit does struggled quite a lot on solving this problem. But if we already have some code to do the work, I'd like to use that one and it can help a lot. Do you know what functions I should read for this?

In general, we need to do:
1) given a 'precise' in the left node (or contained by the left node), how to copy the 'precise' to the right nodes (including constructors, dereferencing operations, etc.).
2) Given a new found node that is 'precise' or containing 'precise' members, how to find its defining nodes and copy the 'precise' to the left node of its defining operation.

ratchetfreak:
It seems the simpler solution is to post process the data flow to back-propagate the precise qualifier.

In SSA form that sounds simple enough: for each operation that requires `precise` ensure that all input values are also calculated with precise (constants and inputs being precise by default already). Either by duplicating the op or decorating it with precise.

johnkslang:
It's also worth pointing out it is acceptable to somewhat superset the operations that must be **NoContraction**, as it's still correct, if it comes down to something very complex to be perfect, versus something pretty simple to be correct.

SPIRV/GlslangToSpv.cpp
johnkslang: Want to be clear: Khronos decided `precise` works under inheritance rules, so this code makes sense.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: This is cleared up: `precise` works under inheritance, and generally can be quite broadly used.

I'm working on just what semantics are required.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -861,6 +862,9 @@ struct DoFullParse{
             parseContext.infoSink.info << parseContext.getNumErrors() << " compilation errors.  No code generated.\n\n";
         }
 
+        // Propagate 'noContraction' label in backward from 'precise' variables.
+        PropagateNoContraction(intermediate);
```
johnkslang: @Qining consider putting this call inside `TIntermediate::postProcess` so the algorithm doesn't go all the way up to ShaderLang.cpp.  That's the design intent of postProcess anyway.

Qining: Yes. I will move the call to the postProcess. And the side effect of such a move is that, we won't have any precise propagation if there are errors when generating the AST. (I just need to adjust the tests a little bit.)

johnkslang:
The commits above do the basic front-end work of:

- recognizing `precise`
- doing GLSL semantic checks on where/how `precise` can be used
- propagating up the tree the `noContraction` status of struct/array/vector dereference, so that the top left AST node of an assignment to an object declared `precise` is marked `noContraction`
- minor test updates to get the semantics right

It does not do the work of:

- adding new test files that this PR adds, to avoid all overlap for those kinds of tests
- doing the implicit propagation of `precise` from the LHS of an assign to the RHS

So, we can now have a new (or updated) PR that does this last set (Qining and I arranged for this).  Just make sure you don't merge the basic parsing files such that we get two sets of `noContraction`, as I put them all next to the `invariant` work.

Qining:
New tests added for structs, precise return functions.

Not covered: precise output function parameters.

johnkslang:
> Not covered: precise output function parameters.

Sorry I didn't originally include that. It should be covered now by 5cc344d8ce09a6d35ff88b9ea9f351062d779081.

Qining:
Rebased to 5cc344d8ce09a6d35ff88b9ea9f351062d779081

precise output function parameters are covered now.

Qining:
This CL is ready for review. PTAL @johnkslang @dneto0 @yavn 
Thanks,

yavn:
Just did a quick check with tessellation tests, and precise is working for my simple use case:

```
precise gl_TessLevelOuter;
```

`NoContraction` spreads to all intermediate expressions that compute the final value of `gl_TessLevelOuter`.

SPIRV/GlslangToSpv.cpp
johnkslang: Actually, despite the inheritance rule, this still doesn't satisfy the second criteria in the comment:

> - effect decorations on the struct members

SPIR-V says:

> Apply to an arithmetic instruction to indicate the operation cannot be combined...

So, I don't think we are applying to data structures, just instructions.  That means structures can be shared even if they differ in their noContraction state.

SPIRV/GlslangToSpv.cpp
```diff
@@ -999,6 +1012,8 @@ bool TGlslangToSpvTraverser::visitBinary(glslang::TVisit /* visit */, glslang::T
         logger->missingFunctionality("unknown glslang binary operation");
         return true;  // pick up a child as the place-holder result
     } else {
+        // Decorate this instruction, if this node has 'noContraction' qualifier.
+        addDecoration(result, TranslateNoContractionDecoration(node->getType().getQualifier()));
```
johnkslang: Rather than trying to put this extra statement after every createBinary/createUnary, you could follow the precision-qualifier model, and require it to be passed in, and then done once inside the function.

If that's equally messy for other reasons, I understand it is a gray area.  Your choice.

SPIRV/GlslangToSpv.cpp
Qining: So, we should have the 'noContraction' removed here, I guess?

johnkslang: Right.

Also, while editing this file, remove my authorship from it at the top.  That's a left over from the history of how it got here, and not the standard being used in glslang.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+
+#include <string>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "localintermediate.h"
+namespace {
+
+// Use string to hold the accesschain information, as in most cases we the
```
johnkslang: No "we"?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+    case glslang::EOpMod:
+
+    case glslang::EOpVectorTimesScalar:
+    case glslang::EOpVectorTimesMatrix:
+    case glslang::EOpMatrixTimesVector:
+    case glslang::EOpMatrixTimesScalar:
+
+    case glslang::EOpDot:
+
+    case glslang::EOpAddCarry:
```
johnkslang: Is it clear how much this applies to integer operations?  It seems these shouldn't need it, but maybe it's safer to just do it.

johnkslang: Conversion operations feel the same to me.  But, they are missing.  It may depend what they get lowered too.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+    }
+    ~StateSettingGuard() { *state_ptr_ = previous_state_; }
+
+private:
+    T *state_ptr_;
+    T previous_state_;
+};
+
+// A helper function to get the front element from a given ObjectAccessChain
+ObjectAccessChain getFrontElement(const ObjectAccessChain &chain)
```
johnkslang: Is there a comment (hopefully above this point somewhere, or at the top) that explains the string syntax you are using?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+    bool visitUnary(glslang::TVisit, glslang::TIntermUnary *) override;
+    bool visitBinary(glslang::TVisit, glslang::TIntermBinary *) override;
+    void visitSymbol(glslang::TIntermSymbol *) override;
+    bool visitAggregate(glslang::TVisit, glslang::TIntermAggregate *) override;
+    bool visitBranch(glslang::TVisit, glslang::TIntermBranch *) override;
+
+protected:
+    // The mapping from symbol node IDs to their defining nodes. This should be
+    // populated along traversing the AST.
+    NodeMapping &symbol_definition_mapping_;
```
johnkslang: Mirror the syntax here for & that we use for *:

```
   NodeMapping& symbol...
```

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+    //  2) If the assignee object node is 'precise' or it contains 'precise'
+    //  objects, traverses the right side of the assignment operation
+    //  expression to mark arithmetic operations as 'noContration' and update
+    //  'precise' accesschain worklist with new found object nodes.
+    // Repeat above steps until the worklist is empty.
+    TNoContractionAssigneeCheckingTraverser checker(accesschain_mapping);
+    TNoContractionPropagator propagator(&precise_object_accesschains,
+                                        accesschain_mapping);
+
+    // We have to initial precise worklist to handle:
```
johnkslang: We have an initial list, or we have to initialize?

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Its depends on whether we consider the conversion operations as arithmetic operations. According to SPIRV spec, I didn't see how a conversion operation can be translated into arithmetic operations defined in the spec, so I didn't add them to this list.

Besides, it might be reasonable to add the conversion operations to the assignment operation list, but I'm not sure.

Here is an example:
```
void main() {
  precise double result;
  int a = 10;
  int b = 20;
  b = a + b; //  a + b should be 'noContraction'.
  double(b + 1.0); // Should be mark b + 1.0 with 'noContraction'?
  result  = b; // convert operation should not be 'noContraction'.
}
```
The problem is the line: `double(b + 1.0)`. Currently we don't consider conversion as an assignment operation, and so `b + 1.0` won't be marked as 'noContraction'.

Note that, if we change the line to: `double temp = double( b + 1.0); result = temp;`, then the current CL will successfully mark `b + 1.0` as 'noContraction', because we have the assignment there, we won't miss it.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Sorry. It should be:
We have two initial precise worklists to handle:

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+            }
+            // Only the upper-most object nodes should be visited, so do not
+            // visit children of this object node.
+            return false;
+        }
+        // If this is an arithmetic operation, marks this node as 'noContraction'.
+        if (isArithmeticOperation(node->getOp())) {
+            node->getWritableType().getQualifier().noContraction = true;
+        }
+        // As this node is not an object node, need to traverse the children nodes.
```
johnkslang: You can `return true` to have the traverser visit the children for you.  `return false` is to prevent the traverser from doing that, either because you did it yourself, or don't want them processed.  This will get the correct order, as the traverser can be asked to go left-right or right-left and pre-order vs. post-order, etc.

(The long-standing comment for that in the existing traverser isn't in the right place, I'm fixing that.)

johnkslang: The above is also true for unary, aggregate, etc. nodes.

glslang/MachineIndependent/propagateNoContraction.cpp
johnkslang: I think we did one conversion by adding zero (for spec. constants).  However, this is not a SPIR-V layer.  At this layer in the stack, you don't know what it will be lowered to, so this is a bit confusing.  As in, we are 99.9% well-defined to process at the tree instead of after lowering, but that's probably the best thing to do anyway.

In your example, did you mean `b = double(b + 1.0)`?  If as you wrote it, it's not part of the data flow.  If it is part of the data flow, precise needs to propagate *through* everything, but get labeled just on relevant operations (as I'm sure you know). So the `precise` on `result` goes back through the `b = double(b + 1.0)`, but since there is nothing there to re-associate or to contract, there is no point tagging it **NoContraction**.

Note that integer add is invariant under re-association and there aren't any contractions for it, so there is no point in having **NoContraction** there either.

The goal here is always to prevent variance under permutation of symmetric operands.

johnkslang:
Just a high-level comment: The high-level design now looks solid, and we are just fine tuning the implementation within that.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+#include <unordered_map>
+#include <unordered_set>
+
+#include "localintermediate.h"
+namespace {
+
+// Use string to hold the accesschain information, as in most cases we the
+// accesschain is short and may contain only one element, which is the symbol ID.
+using ObjectAccessChain = std::string;
+#ifndef StructAccessChainDelimiter
```
johnkslang: What if this is defined?  It might not even be a character.

Do you mean to #undef?

Or, would it work to use

```
const char StructAccessChainDelimiter = `/`;
```

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,897 @@
...
+    // This means the left node might not be an 'precise' object node, but it may contains
+    // 'precise' qualifier which should be propagated to the corresponding child node in
+    // the right. So we need the path from the left node to its nested 'precise' node to
+    // tell us how to find the corresponding 'precise' node in the right.
+    ObjectAccessChain remained_accesschain_;
+    // A map from node pointers to their accesschains.
+    const AccessChainMapping &accesschain_mapping_;
+};
+
+#undef StructAccessChainDelimiter
```
johnkslang: What benefits from this?  This is not a header.  See comment above about its declaration.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: I should use const char here. Will have this fixed in the coming CL.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: The only conversion I found being translated to `add` operation is: `spec uint -> spec int`. This is not needed to label it as precise I think.

The line `double(b + 1.0)` is not assigning to anything and it is not a part of the data flow. Actually we don't need to care it all. And whenever there is an `=` before it to make an explicit assignment operation, the `b + 1.0` will be naturally covered. 

I can add `if ()` in some places to exclude integer types from being labelled as 'precise'.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Addressed in 7bda98941d232afd5afdebe53553f79ae97cf28c

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Addressed in 7bda98941d232afd5afdebe53553f79ae97cf28c

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: addressed in 7bda98941d232afd5afdebe53553f79ae97cf28c

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: addressed in 7bda98941d232afd5afdebe53553f79ae97cf28c

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Added before the using ObjectAccessChain = ...  in 7bda98941d232afd5afdebe53553f79ae97cf28c

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Fixed in 7bda98941d232afd5afdebe53553f79ae97cf28c

SPIRV/GlslangToSpv.cpp
Qining: Sink the addDecoration() to createBinaryOperation() and createUnaryOperation() in 7bda98941d232afd5afdebe53553f79ae97cf28c

johnkslang: BTW, this could be vital (theoretically), in the case that more than one instruction was issued, as you would only have decorated the last one.

glslang/MachineIndependent/propagateNoContraction.cpp
johnkslang: Note that an algorithm that:

- has a worklist of pending items
- adds more items as things propagate backward

might work in practice O(N) if nodes are visited end to beginning (right to left) but O(N^2) if nodes are visited in program order (left to right).

SPIRV/GlslangToSpv.cpp
Qining: Yes, I just found the hole as you said when I read the code there.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -47,18 +47,27 @@
...
+// same symbol ID of their container symbol objects. This is because their
+// precise'ness is always the same as their container symbol objects.
 using ObjectAccessChain = std::string;
-#ifndef StructAccessChainDelimiter
-#define StructAccessChainDelimiter '/'
-#endif
+
+// The delimiter used in the ObjectAccessChain string to separate symbol ID and
+// different level of struct indices.
+const char OBJECT_ACCESSCHAIN_DELIMITER = '/';
```
johnkslang: Good, but...

The camel case name you had before was correct: it looked like a non-PP (non-macro) symbol, while this one looks like a PP symbol. Since neither glslang nor the google coding standard uses old PP style caps for const, just for macros, do you mind changing it to how it was before?

SPIRV/GlslangToSpv.cpp
johnkslang: Did it change test results?  Especially if (guessing) it went through a matrix op, or assignment op, or pre/post inc/dec op?

Qining: Not for now. I don't have matrix in tests (I remember I saw something in glsl spec saying that matrix won't need precise so I didn't add tests for them. Probably I should add some.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: No problem, this will be fixed in the coming CL

SPIRV/GlslangToSpv.cpp
johnkslang: I don't think we need to add tests unless it is to ensure **NoContraction** gets to the right nodes.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: fixed in ca4fe001d121bfc9d2cd049dd22b6a8fdbadfda7

SPIRV/GlslangToSpv.cpp
Qining: Added a test in ca4fe001d121bfc9d2cd049dd22b6a8fdbadfda7. I do missed the matrix times matrix case.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: If we don't consider struct cases, this algorithm should be O(N), because each node can only be added to the worklist once and be processed once.

We have a map from the symbols to their defining nodes, so we don't need to traverse the AST to find the defining node for a given 'precise' symbol which is just popped from the worklist.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,890 @@
...
+        object_to_be_defined_.clear();
+        node->getRight()->traverse(this);
+
+        return false;
+    } else if (isDereferenceOperation(node->getOp())) {
+        // If the left node is 'precise' object node, this node should also
+        // be 'precise' object node, and all the members of this node too. There
+        // is no need to append accesschain information into the object id.
+        if (isPreciseObjectNode(node->getLeft())) {
+            node->getWritableType().getQualifier().noContraction = true;
```
johnkslang: I think this is what was already done during parsing.  Generally, I'm confused to see code here looking at the left of a dereference to see if something is precise.  It should just have to look at the left side of an assignment.

Also, if for a hypothetical example, there were some members precise didn't go to, due to semantics in the spec., or we change the inheritance rules, the parser should be deciding exactly when to do it, and not getting overridden here by unconditionally doing it.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,890 @@
...
+{
+    // Traverses the left node to build the accesschain info for the object.
+    object_to_be_defined_.clear();
+    node->getLeft()->traverse(this);
+
+    if (isAssignOperation(node->getOp())) {
+        // We should always be able to get an accesschain for the left node.
+        // But we have some tests in which it is intented to have invalid left
+        // nodes, so just return false in such cases for now.
+        if (object_to_be_defined_.empty()) return false;
```
johnkslang: Make returns always have their own line (actually, all substatements should have their own line).

It's pretty important to be able to visually see all `return` statements in a function.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,890 @@
...
+bool TSymbolDefinitionCollectingTraverser::visitBinary(glslang::TVisit /* visit */,
+                                                       glslang::TIntermBinary* node)
+{
+    // Traverses the left node to build the accesschain info for the object.
+    object_to_be_defined_.clear();
+    node->getLeft()->traverse(this);
+
+    if (isAssignOperation(node->getOp())) {
+        // We should always be able to get an accesschain for the left node.
+        // But we have some tests in which it is intented to have invalid left
```
johnkslang: Tests with intentional errors will fail parsing, and hence not go through postProcess().  Does that simplify things?

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Oops, this is remnant of the implementation before we have the parser handling this.

But we still need to traverse the subtree of the left node, not only the node. Because we need to reach the bottom symbol and update the accesschain_mapping_, so that we can use the map to find the defining node of a given symbol without traversing the whole AST again.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Without the errors, I can use assert(). I think that is better than just return false.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: With the current parser, the code can be simplified here. The coming CL will have simplified code.

Qining:
Hi @johnkslang , all the comments are resolved, is this CL good to move forward?

johnkslang:
Sorry my review is trickling in.  It's a big change and I got sick and have a schedule for other things too.

Process-wise, I guess I'll say I finished, otherwise you can't tell if I'm in the middle or end, or is there a built-in mechanism for saying a review is done?

Qining:
Sorry to hear that you are sick and hope you get better soon :)

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+
+    case glslang::EOpVectorTimesScalar:
+    case glslang::EOpVectorTimesMatrix:
+    case glslang::EOpMatrixTimesVector:
+    case glslang::EOpMatrixTimesScalar:
+    case glslang::EOpMatrixTimesMatrix:
+
+    case glslang::EOpDot:
+
+    case glslang::EOpAddCarry:
```
johnkslang: These are integer only.  Should they:

A. not be here, as per the comment at the head, or

B. be here, but so should other integer operations?

A key point (and things have changed since we started this) is that the entire data flow is subject to analysis, but just a subset of it needs to be marked with NoContraction. I'm sure I'll see if I get the whole update in my head, but I'm just starting again, so easier to ask.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+    size_t pos_delimiter = chain.find(ObjectAccesschainDelimiter);
+    return pos_delimiter == std::string::npos ? "" : chain.substr(pos_delimiter + 1);
+}
+
+// A helper function to get the accesschain after removing a given prefix.
+ObjectAccessChain getSubAccessChainAfterPrefix(const ObjectAccessChain& chain,
+                                               const ObjectAccessChain& prefix)
+{
+    size_t pos = chain.find(prefix);
+    if (pos != 0) return chain;
```
johnkslang: Return needs its own line.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+        &current_function_definition_node_);
+    if (node->getOp() == glslang::EOpFunction) {
+        // This is function definition node, we need to cache this node so that we can
+        // get the precise'ness of the return value later.
+        current_function_definition_node_setting_guard.setState(node);
+    }
+    // Traverse the items in the sequence.
+    glslang::TIntermSequence& seq = node->getSequence();
+    for (int i = 0; i < (int)seq.size(); ++i) {
+        object_to_be_defined_.clear();
```
johnkslang: Just curious: Is this line preventing the traverser from traversing this for you (i.e. with return true)?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+            object_to_be_defined_.append(std::to_string(struct_dereference_index));
+        }
+        accesschain_mapping_[node] = object_to_be_defined_;
+
+        // For dereference node, there is no need to traverse the right child
+        // node as the right node should always be an integer type object.
+
+    } else {
+        // For other binary nodes, still traverse the right node.
+        object_to_be_defined_.clear();
```
johnkslang: Somehow, it seems nodes you don't care about should just be handled by the traverser.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+        node->getRight()->traverse(this);
+    }
+    return false;
+}
+
+// Traverses the AST and returns a tuple of three members:
```
johnkslang: Is it four now?

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,860 @@
...
+// 2) a mapping from object nodes in the AST to the accesschains of these objects.
+// 3) a set of accesschains of precise objects.
+std::tuple<NodeMapping, AccessChainMapping, ObjectAccesschainSet, ReturnBranchNodeSet>
+getSymbolToDefinitionMappingAndPreciseSymbolIDs(const glslang::TIntermediate& intermediate)
+{
+    auto result_tuple = std::make_tuple(NodeMapping(), AccessChainMapping(), ObjectAccesschainSet(),
+                                        ReturnBranchNodeSet());
+
+    TIntermNode* root = intermediate.getTreeRoot();
+    if (root == 0) return result_tuple;
```
johnkslang: Make all early return statements visible on their own line.

johnkslang:
@Qining I finished my review now, just a few more comments to address.  Thanks.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Yes you are right. I think I still need to clear the object accesschain string which the traverse is building.

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: You are right, I should have removed these instruction before. Removed in bb0ac5466047f8ef14dda37cb0b160682d20ee26

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: fixed in bb0ac5466047f8ef14dda37cb0b160682d20ee26

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: fixed in bb0ac5466047f8ef14dda37cb0b160682d20ee26

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Yes, it should be four here. Comments updated in bb0ac5466047f8ef14dda37cb0b160682d20ee26

glslang/MachineIndependent/propagateNoContraction.cpp
Qining: Should all be fixed in bb0ac5466047f8ef14dda37cb0b160682d20ee26

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -0,0 +1,858 @@
...
+    // populated along traversing the AST.
+    NodeMapping& symbol_definition_mapping_;
+    // The set of symbol node IDs for precise symbol nodes, the ones marked as
+    // 'noContraction'.
+    ObjectAccesschainSet& precise_objects_;
+    // The set of precise return nodes.
+    ReturnBranchNodeSet& precise_return_nodes_;
+    // A temporary cache of the symbol node whose defining node is to be found
+    // currently along traversing the AST.
+    ObjectAccessChain current_object__;
```
johnkslang: Why double underscore? C++ reserves these.

Qining: :( I really should change my "highlight current line" style in my vim.

Just be fixed in e543cc3.

PR #224, opened 2016-04-07
Spec const bool can be converted to spec-const int

Qining:
Bool-> uint/int with OpSpecConstantOp OpSelect instruction.

Refine a little bit of the code.


SPIRV/SpvBuilder.cpp
```diff
@@ -1198,9 +1199,8 @@ Id Builder::createBinOp(Op opCode, Id typeId, Id left, Id right)
     // Generate code for spec constants if in spec constant operation
     // generation mode.
     if (generatingOpCodeForSpecConst) {
-        std::vector<Id> operands(2);
-        operands[0] = left; operands[1] = right;
-        return createSpecConstantOp(opCode, typeId, operands, std::vector<Id>());
+        return createSpecConstantOp(
+            opCode, typeId, {left, right} /* operands */, {} /* literals */);
```
johnkslang: We need to stop using this initializer syntax, it always breaks others.

There are just two popular C++11 features to avoid:

1. range-based loops
2. initializer syntax used as an expression

(#2 also fails to communicate type to the reader, so it might be more readable anyway with the extra line of code that says what it is.)

johnkslang:
Lets keep refining this one (or do a big one at the end as discussed earlier) once Khronos pins down the exact set of ops to use.  Maybe conversions will go away, or bitcast will show up, etc.

Qining:
OK, understood, no wonder why I found all my initializers before were gone.

I think we can just close this PR, as it actually did more about code style than fixing a real issue.

PR #232, opened 2016-04-13
SpecOp bool->uint/int and uint<->int conversion

Qining:
Bool -> uint/int with OpSpecConstantOp OpSelect instruction.

uint <-> int conversion with OpSpecConstantOp OpIAdd instruction.

Note, implicit conversion: `const uint = an_int_spec_constant` is not
supported. Explicit type casting is required: `const uint =
uint(an_int_spec_constant)`


SPIRV/GlslangToSpv.cpp
```diff
@@ -3307,7 +3306,27 @@ spv::Id TGlslangToSpvTraverser::createConversion(glslang::TOperator op, spv::Dec
...
+        }
+        // Don't 'break' here as this case should be grouped together with
+        // EOpConvIntToUint when generating normal run-time conversion
+        // instruction.
     case glslang::EOpConvIntToUint:
+        if (builder.isInSpecConstCodeGenMode()) {
+            // Build zero scalar or vector for OpIAdd.
+            if (zero == 0) {
+                zero = builder.makeUintConstant(0);
+                zero = makeSmearedConstant(zero, vectorSize);
```
Qining: Is there a way to differentiate unsigned int type and signed int type? Such a function can make this part much nicer.

johnkslang: Not sure I understand.  Yes, we differentiate, and you chose unsigned, and so ```zero```'s type is unsigned, which is probably the best choice.  You could have chosen a signed-int type for 0, and that would also work, but unsigned is more appropriate.

SPIRV/GlslangToSpv.cpp
```diff
@@ -3307,7 +3306,27 @@ spv::Id TGlslangToSpvTraverser::createConversion(glslang::TOperator op, spv::Dec
         break;
 
     case glslang::EOpConvUintToInt:
+        if (builder.isInSpecConstCodeGenMode()) {
+            // Build zero scalar or vector for OpIAdd to do the conversion when
+            // generating for OpSpecConstantOp instruction.
+            zero = builder.makeIntConstant(0);
```
johnkslang: I think you can always fall through to do this, and just have one piece of code that uses an unsigned type for `zero`.

SPIRV/SpvBuilder.cpp
```diff
@@ -1212,6 +1212,14 @@ Id Builder::createBinOp(Op opCode, Id typeId, Id left, Id right)
 
 Id Builder::createTriOp(Op opCode, Id typeId, Id op1, Id op2, Id op3)
 {
+    // Generate code for spec constants if in spec constant operation
+    // generation mode.
+    if (generatingOpCodeForSpecConst) {
+        std::vector<Id> operands(3);
+        operands[0] = op1; operands[1] = op2; operands[2] = op3;
```
johnkslang: Would like one statement per line, thanks.

SPIRV/GlslangToSpv.cpp
Qining: Hmm, actually I hope to let the EOpConvUintToInt case fall through. Just hope to have function to tell whether a type ID is uint type or int type. I found this not that easy, as both int and uint are declared with OpTypeInt. It seems like the only way to tell is to check the OpTypeInt instruction's first operand.

johnkslang: You don't need to tell. Just always make uint for `zero`, or is it some other reason you want to know?

Qining: Just think having it different for int and uint may looks a little bit better. Anyway, already changed it to use 'uint' zero for both.

PR #233, opened 2016-04-13
WIP: Fix implicit conversion uint -> int

Qining:
Support shader code like:
```
layout(constant_id = 100) const uint an_uint_spec_const_int = 10;
const int x = an_uint_spec_const_int;
```

And extend the tests of spec constant integer type comparisons.


glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -669,7 +669,7 @@ bool TIntermediate::canImplicitlyPromote(TBasicType from, TBasicType to) const
         }
     case EbtInt:
         switch (from) {
-        case EbtInt:
+        case EbtUint:
```
Qining: @johnkslang Here is the only real change in this CL.

I think the original `EbtInt` here is a bug, because if the types of `from` and `to` are same, it will return earilier and won't reach here.

But I'm not quite sure, and this change does affect other tests.

johnkslang: Section 4.1.10 of the GLSL spec. lists the allowed implicit promotions. They form an acyclic dag, and does include `uint -> int`, so this proposed change breaks GLSL semantics.

All the top-level cases include self-implicit promotion, and in dag order that lets you see the pattern, which correctly reflects the semantics.

Qining:
Close as implicit conversion uint -> int should not be supported.

PR #236, opened 2016-04-14
Implement the extension GL_ARB_shader_ballot

amdrexu:
This change partially supports int64_t/uint64_t in that they are required by this extension. I will add complete support in the subsequent change to fully implement GL_ARB_gpu_shader_int64.

Currently, the extended instructions for 3 built-in functions and built-in decorations for 7 built-in variables are missing in SPIR-V. Once the header is defined and is added in glslang, I will finish this part of translation work. I use "GLSLstd450Bad" as placeholder to pass the compilation.


amdrexu:
I will add this header in next change to fully support this extension.

```
/*
** Copyright (c) 2014-2016 The Khronos Group Inc.
**
** Permission is hereby granted, free of charge, to any person obtaining a copy
** of this software and/or associated documentation files (the "Materials"),
** to deal in the Materials without restriction, including without limitation
** the rights to use, copy, modify, merge, publish, distribute, sublicense,
** and/or sell copies of the Materials, and to permit persons to whom the
** Materials are furnished to do so, subject to the following conditions:
**
** The above copyright notice and this permission notice shall be included in
** all copies or substantial portions of the Materials.
**
** MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS KHRONOS
** STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS SPECIFICATIONS AND
** HEADER INFORMATION ARE LOCATED AT https://www.khronos.org/registry/ 
**
** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
** OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
** THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
** FROM,OUT OF OR IN CONNECTION WITH THE MATERIALS OR THE USE OR OTHER DEALINGS
** IN THE MATERIALS.
*/

#ifndef GLSLextARB_H
#define GLSLextARB_H

static const int GLSLextARBVersion = 100;
static const int GLSLextARBRevision = 1;

enum spv::BuiltIn;

// SPV_ARB_shader_ballot
static const char* const E_SPV_ARB_shader_ballot = "SPV_ARB_shader_ballot";

static const spv::BuiltIn BuiltInSubGroupSizeARB        = static_cast<spv::BuiltIn>(4416);
static const spv::BuiltIn BuiltInSubGroupInvocationARB  = static_cast<spv::BuiltIn>(4417);
static const spv::BuiltIn BuiltInSubGroupEqMaskARB      = static_cast<spv::BuiltIn>(4418);
static const spv::BuiltIn BuiltInSubGroupGeMaskARB      = static_cast<spv::BuiltIn>(4419);
static const spv::BuiltIn BuiltInSubGroupGtMaskARB      = static_cast<spv::BuiltIn>(4420);
static const spv::BuiltIn BuiltInSubGroupLeMaskARB      = static_cast<spv::BuiltIn>(4421);
static const spv::BuiltIn BuiltInSubGroupLtMaskARB      = static_cast<spv::BuiltIn>(4422);

enum ShaderBallot {
    BallotARB = 0,
    ReadInvocationARB = 2,
    ReadFirstInvocationARB = 3,

    ShaderBallotCount
};

#endif  // #ifndef GLSLextARB_H
```

amdrexu:
@johnkslang The request of this extension is the same as that of GL_ARB_gpu_shader_int64. Please have a review of it. Any feedback is welcome. Some ISVs want it to be implemented in glslang. Thank you.

amdrexu:
@johnkslang I have updated the codes of this commit since the implementation of GL_ARB_gpu_shader_int64 is merged to master branch. Likewise, GLSLstd450Bad is used as placeholder till corresponding SPIR-V instructions are finalized for those three built-in functions defined in this extension. Please have a review. Thank you.

kholtnv:
FYI, there is issue 35 in SPIR-V to track the changes required for this.

amdrexu:
@kholtnv Thank you. It seems there is some missing items that have to be defined and to be ready in SPIR-V spec in order to fully support this extension.

amdrexu:
@johnkslang Is this commit OK to go? I have another commit to fully implement unfinished items in SPV generation (05c2c358f71129b678985b3982bcfb9fe92d8506).

johnkslang:
I'm looking at this now.

The highest-level question is whether we are supporting official SPIR-V extensions here, and if so, where is the extension spec, has it been approved by Khronos, etc.

E.g., someone can't just write a SPIR-V extension, call it a KHR extension, and put that into glslang, if the Khronos group hasn't actually agreed it is a KHR extension.

Maybe for unofficial extensions, the best we can do is keep a branch for them somewhere?  I think this needs some discussion.

If this PR only implements a GLSL extension, and not a SPIR-V extension, that simplifies this for the moment.

johnkslang:
This interacted with multiple recent PRs, so had conflicts, but I resolved them for you.

amdrexu:
@johnkslang My private implementation (05c2c358f71129b678985b3982bcfb9fe92d8506) is based on this SPIR-V extension spec (https://gitlab.khronos.org/spirv/SPIR-V/blob/master/extensions/shader_ballot.txt) to finish SPIR-V token generation for GL_ARB_shader_ballot.

Since this extension spec is old and there are changes of SPIR-V spec during this period of time, I am not sure whether this extension spec is valid as the base of my implementation. Some items (gl_SubGroupSizeARB, readInvocationARB()) are duplicate compared to current SPIR-V spec while other items (gl_SubGroupXxMaskARB, ballotARB(), readFirstInvocationARB()) are indeed missing.

kholtnv:
Here are several issues that come to mind.
1) Should GLSL.ext.ARB.h be owned by Vulkan as a new environment spec?
2) What are the minimal set of instructions needed in SPIRV (as you point out "group size" and
"read invocation" are there but others are not.)? These should be proposed and voted on in the
SPIR-V Khronos group.
3) Are shader environments ready to accept the "Group" SPIR-V capability? 
4) The "OpExtension" should be used so that validation layers can verify extensions.
(Note: I don't see that GL_ARB_shader_group_vote added this which may be a problem for
validation layers when an environment doesn't support that extension.)
5) Technically the opcode ranges 4416-4422 belong to Codeplay (though it does mention
"GL_ARB_shader_ballot"). That's just a bit confusing.
These issues should be brought up in a Khronos meeting.

amdrexu:
@kholtnv Thank you for your comments.

1) I think this header is similar to `GLSL.std.450.h` and could be placed where `GLSL.std.450.h` stays.
2) Yes, the extension SPV_ARB_shader_ballot proposes some SPIR-V extended instructions. They overlaps with current SPIR-V spec to some extent. Those instructions need to be refined and be minimized.
3) In SPIR-V spec, `group `capability is independent of `shader` capability. Shader environment could support `group `or not.
4) Yes, you are right. `OpExtension `has to specify those used SPIR-V extensions, such as SPV_ARB_shader_ballot. But for GL_ARB_shader_group, the implementation does not use instructions beyond current SPIR-V spec. It only employs native `OpGroupAny `and `OpGroupAll`. So I think there is no need of using `OpExtension`.
5) Yes, I agree with you. The spec of this extension has something to clear. So a revision is needed.

johnkslang:
@kholtnv :

>Here are several issues that come to mind.
> 1) Should GLSL.ext.ARB.h be owned by Vulkan as a new environment spec?

This is a SPIR-V or Vulkan issue, to take up with the SPIR-V group/GitHub/Khronos/bugzilla.  This repository is an implementation repository; let's discuss in the proper forum.

> 2) What are the minimal set of instructions needed in SPIRV (as you point out "group size" and
"read invocation" are there but others are not.)? These should be proposed and voted on in the
SPIR-V Khronos group.

Right. Let's discuss there, not here.

> 3) Are shader environments ready to accept the "Group" SPIR-V capability? 

Only the implementations that want to need to accept the group capability. That is basic definition of how capability and extensions work.  If you don't agree with how capabilities and extensions work, that's definitely a subject to take up with Khronos.

> 4) The "OpExtension" should be used so that validation layers can verify extensions.

If there is an extension, which in this case I don't think there is. Since there is already a capability, it is the capability that needs to be declared.  It could be that Vulkan is extended to remove the capability from the "can't use" list, but that's an issue on the Vulkan spec. repository, not this one.

> (Note: I don't see that GL_ARB_shader_group_vote added this which may be a problem for
validation layers when an environment doesn't support that extension.)

Unless, they tie it to the OpCapability, for the case where it was already in core SPIR-V 1.0. It would be great for you to bring this up first in the ongoing SPIR-V discussions.

> These issues should be brought up in a Khronos meeting.

Yes.  Follow up each issue in the right place.  We should not conflate any of the following together: Vulkan spec., SPIR-V spec, and glslang implementation.  They all have their own process and place to discuss their own issues.

Thanks.

PR #238, opened 2016-04-14
ParseHelper: do not use default precision for build-in functions

scyganINTEL:
This removes default precision qualifiers from parsing the build-ins. This is needed for functions,
so they get precision computed from their arguments (that is already handled in glslang).

Without that all FS build-in functions other than pack/unpack and sampler/image operations are performed in mediump.

Additionally, all build-in variables are checked to have explicit precision qualifier in declaration.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -70,7 +70,11 @@ TParseContext::TParseContext(TSymbolTable& symbolTable, TIntermediate& interm, b
         defaultSamplerPrecision[type] = EpqNone;
 
     // replace with real defaults for those that have them
-    if (profile == EEsProfile) {
+
+	// do not set default precision for build ins: build-in functions 
+	// precision depends on arguments, build in values precision should be set
+	// in declaration
+    if (profile == EEsProfile && !parsingBuiltins) {
         TSampler sampler;
         sampler.set(EbtFloat, Esd2D);
         defaultSamplerPrecision[computeSamplerTypeIndex(sampler)] = EpqLow;
@@ -4805,6 +4809,13 @@ TIntermNode* TParseContext::declareVariable(const TSourceLoc& loc, TString& iden
     // see if it's a linker-level object to track
     if (newDeclaration && symbolTable.atGlobalLevel())
         intermediate.addSymbolLinkageNode(linkage, *symbol);
+
+	// check if all build-in variables have precision set on ES profile
+	assert(profile != EEsProfile || !parsingBuiltins ||
+		symbol->getType().getBasicType() != EbtUint  ||
+		symbol->getType().getBasicType() != EbtInt   ||
+		symbol->getType().getBasicType() != EbtFloat ||
+		symbol->getType().getQualifier().precision != EpqNone);
 
     return initNode;
 }


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -70,7 +70,11 @@ TParseContext::TParseContext(TSymbolTable& symbolTable, TIntermediate& interm, b
         defaultSamplerPrecision[type] = EpqNone;
 
     // replace with real defaults for those that have them
-    if (profile == EEsProfile) {
+
+	// do not set default precision for build ins: build-in functions 
+	// precision depends on arguments, build in values precision should be set
+	// in declaration
+    if (profile == EEsProfile && !parsingBuiltins) {
```
johnkslang: This is the constructor, so I think you are assuming `parsingBuiltins` is `const`, but it's not.  I checked, and it does look like it could be: we don't actually switch this state while parsing, so maybe the design should evolve in that way.

As it is now, a slight change to keep the same context across built-in parsing and user parsing, or otherwise change the value of `parsingBuiltins` would make this break.

I'm trying to think through how to best solve the real problem here...

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -70,7 +70,11 @@ TParseContext::TParseContext(TSymbolTable& symbolTable, TIntermediate& interm, b
         defaultSamplerPrecision[type] = EpqNone;
 
     // replace with real defaults for those that have them
-    if (profile == EEsProfile) {
+
+	// do not set default precision for build ins: build-in functions
```
johnkslang: In general, these are "built-in ..."

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: I'm also concerned about the distinction of the three different ways of getting precision for a built-in function:

1. Some functions have predefined precisions.
2. For the texture sampling...  the precision of the sampler type.
3. For other built-in functions... the highest precision ... of an argument.

johnkslang:
This should also affect test results.

I'll work on a complete fix here; building on yours, thanks.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4806,6 +4810,13 @@ TIntermNode* TParseContext::declareVariable(const TSourceLoc& loc, TString& iden
     if (newDeclaration && symbolTable.atGlobalLevel())
         intermediate.addSymbolLinkageNode(linkage, *symbol);
 
+	// check if all build-in variables have precision set on ES profile
+	assert(profile != EEsProfile || !parsingBuiltins ||
```
johnkslang: gl_LocalInvocationIndex, for example, doesn't have a precision.  That might be okay (separate subject), but either have to change that, or not require all built-ins to have a precision.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: An unintended side effect is texture lookups propagate the precision of the first argument to the others.  This is a separate bug that was masked by this one, but fixing this one now makes more arguments go to lower precision (they weren't before), which would break things, so these bugs need to be fixed together.  Will work on that.

johnkslang:
This issue is now addressed by af459216a18a08c2e3eec880329c3a7a1244ddeb.

Issue #237, opened 2016-04-14
[SPIR-V] Incorrect OpSpecConstantComposite constituents


da3k:
It looks like OpSpecConstantComposite constituents are being directly translated from the GLSL syntax, with no additional transformations that are in many cases required to generate valid SPIR-V. 

In the example below, the code generated for OpSpecConstantComposite has a list of constituents that directly matches the flat list of arguments passed to the GLSL matrix constructor. However, according to the spec, the constituents should instead be columns of the matrix (two vec2() composites).

GLSL source:
```
#version 440

layout(constant_id = 1) const float sc0 = float(1);

layout (set = 0, binding = 0, std430) writeonly buffer Output {
    float result[4];
} sb_out;

const mat2 m0 = mat2(sc0, 2, 3, 4);
const mat2 m1 = mat2(1, sc0, 3, 4);
const mat2 m2 = mat2(1, 2, sc0, 4);
const mat2 m3 = mat2(1, 2, 3, sc0);

void main (void)
{
    float sum_m0 = float(0);
    float sum_m1 = float(0);
    float sum_m2 = float(0);
    float sum_m3 = float(0);

    for (int i = 0; i < 2; ++i)
    for (int j = 0; j < 2; ++j)
    {
        sum_m0 += float(m0[i][j]);
        sum_m1 += float(m1[i][j]);
        sum_m2 += float(m2[i][j]);
        sum_m3 += float(m3[i][j]);
    }

    sb_out.result[0] = sum_m0;
    sb_out.result[1] = sum_m1;
    sb_out.result[2] = sum_m2;
    sb_out.result[3] = sum_m3;
}

```
SPIR-V generated (first OpSpecConstantComposite):
```
...
%35 = OpTypeMatrix %34 2
%36 = OpSpecConstant %6 1
%37 = OpConstant %6 2
%38 = OpConstant %6 3
%39 = OpConstant %6 4
%40 = OpSpecConstantComposite %35 %36 %37 %38 %39
...

```
GLSL constructors like "vec4(sc0)" are also being translated directly, where OpSpecConstantComposite is incorrectly generated with only one sc0 in the constituents list instead of four.

Qining:
Thanks @da3k , I will go ahead and take care of this.

Qining:
Hi @da3k @johnkslang, PR #241 should fix this issue, pleas take a look. Thanks.

johnkslang:
I verified this is indeed working correctly now.

PR #241, opened 2016-04-14
Fix construtor for spec const matrix and vector

Qining:
Fix issue: #237

1. The code generated for matrix constructor should 1) build column
vectors first, 2) build matrix with the vectors.

2. When there is only one scalar type constituent in vector's
constructor, we should populate the constituent to fill all the slots in
the vector. As for matrix, the single constituent should be populated to
the diagonal positions (top-left to bottom-right diagonal).

remove createSpvConstantFromConstSubTree()


Test/spv.specConstantComposite.vert
```diff
@@ -42,6 +42,11 @@ const vec4 spec_vec4_all_spec =
     vec4(spec_float, spec_float, spec_float, spec_float);
 const vec4 spec_vec4_partial_spec =
     vec4(spec_float, spec_float, 300.14, 300.14);
+const vec4 spec_vec4_from_one_scalar = vec4(spec_float);
+
+// Matrix constructor
+const mat2x3 spec_mat2x3 = mat2x3(spec_float, spec_float, spec_float, 1.1, 2.2, 3.3);
+const mat2x3 spec_mat2x3_from_one_scalar = mat2x3(spec_float);
```
Qining: There are some changes in the test base results that are not caused by the added spec constants here. Such changes are only mismatches of instruction orders due to the remove of createSpvConstantFromConstSubTree().

In original createSpvConstantFromConstSubTree(), we call convertGlslangToSpvType() even though we don't use them immediately (actually the `typeId` is only used for handling aggregate nodes). But such a call will generate type declaration instruction, so removing the redundant call to convertGlslangToSpvType() will change the instruction order.

Qining:
@johnkslang @da3k PTAL

johnkslang:
Just to get better oriented first... we previously decided that the unary and binary nodes covered all the cases in KHR_vulkan_glsl, so I wonder if the fix here is to not be setting the spec-constness on these nodes in the front end?

(I'll have more time to look deeper, but want to make sure the basics are in place in the meantime.)

Qining:
The problem is that the original createSpvConstantFromConstSubTree() handles aggregate nodes without using the `traverse(this)` (different with binary and unary nodes). And bug is caused by missing functionalities in there.

This CL actually removed the createSpvConstantFromConstSubTree() and unified the handlers of all kinds of nodes with traverse(this) then getAccessChainLoad().

So, in this CL, we use the same visit*() functions to handle both run-time instruction generation and spec constant op instruction generation.

Qining:
Any suggestions for this CL?

johnkslang:
I asked the wrong way.  I think your code changes are just fine for AST -> SPIR-V, and I'm just asking about which composites GLSL -> AST should be allowed to be created as spec-const nodes.

SPIRV/SpvBuilder.cpp
```diff
@@ -1714,6 +1732,20 @@ Id Builder::createCompositeConstruct(Id typeId, std::vector<Id>& constituents)
 {
     assert(isAggregateType(typeId) || (getNumTypeConstituents(typeId) > 1 && getNumTypeConstituents(typeId) == (int)constituents.size()));
 
+    if (generatingOpCodeForSpecConst) {
+        // Sometime, even in spec-constant-op mode, the constant composite to be
+        // constructed may not be a specialization constant.
+        // e.g.:
+        //  const mat2 m2 = mat2(a_spec_const, a_front_end_const, another_front_end_const, third_front_end_const);
```
johnkslang: The AST nodes should be correct about what to create as a spec-const.  Is that not the case?

Qining: Yes, exactly!

Qining:
I didn't see anything about "spec constant matrix" in KHR_glsl_vulkan. Actually I also feel it is suspicious, as at line 226, we only mentioned scalars and vectors.

PR #246, opened 2016-04-21
Wrap Windows/Linux's mutex objects for MinGW uses.

Qining:
This is to solve the problem that the use of std::mutex in gtest won't 
compile with prebuilt MinGW.

Reimplement a wrapper of Windows/Linux's mutex objects solve the
cross-compilation problem.


Qining:
@antiagainst @johnkslang @dekimir @dneto0 @AWoloszyn PTAL, just a copy/paste work, not very sure if this is acceptable.

dneto0:
Looks good to me.    This will let us build with tests turn on in the NDK build.  The MinGW available to the NDK build currently doesn't have a working std::mutex.  :-(

Qining:
Close due to license problem.

PR #247, opened 2016-04-21
Remove use of std::mutex in gtest code.

dneto0:
Gtest runs in single-threaded mode.  So the GlslangInitializer
object in the test code doesn't have to do its own synchronization.

Credit to @AWoloszyn and @antiagainst for suggesting this approach and determining this is safe to do.

### DIFF
--- a/gtests/Initializer.h
+++ b/gtests/Initializer.h
@@ -35,8 +35,6 @@
 #ifndef GLSLANG_GTESTS_INITIALIZER_H
 #define GLSLANG_GTESTS_INITIALIZER_H
 
-#include <mutex>
-
 #include "glslang/Public/ShaderLang.h"
 
 namespace glslangtest {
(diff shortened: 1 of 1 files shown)


dneto0:
PTAL @AWoloszyn @antiagainst

antiagainst:
LGTM

dneto0:
PTAL @johnkslang

gtests/Initializer.h
```diff
@@ -63,55 +61,25 @@ class GlslangInitializer {
...
     // Re-initializes glsl state iff the previous messages and the current
-    // messages are incompatible.
+    // messages are incompatible.  We assume external synchronization, i.e.
+    // there is at most one acquired token at any one time.
     InitializationToken acquire(EShMessages new_messages)
     {
-        stateLock.lock();
-
         if ((lastMessages ^ new_messages) &
             (EShMsgVulkanRules | EShMsgSpvRules)) {
```
johnkslang: It seems this should be handled by proper symbol table caching/keying, which isn't done yet.  Not related to the PR, just noting it.

PR #248, opened 2016-04-22
Implement the extension GL_ARB_gpu_shader_int64

amdrexu:
- Add new keyword int64_t/uint64_t/i64vec/u64vec.
- Support 64-bit integer literals (dec/hex/oct).
- Support built-in operators for 64-bit integer type.
- Add implicit and explicit type conversion for 64-bit integer type.
- Add new built-in functions defined in this extension.


amdrexu:
@johnkslang This extension is required by several ISVs and it blocks GL_ARB_shader_ballot. So I fully implement it. Please have a review. Thank you.

glslang/MachineIndependent/Versions.cpp
```diff
@@ -627,6 +629,14 @@ void TParseVersions::doubleCheck(const TSourceLoc& loc, const char* op)
     profileRequires(loc, ECompatibilityProfile, 400, nullptr, op);
 }
 
+// Call for any operation needing GLSL 64-bit integer data-type support.
+void TParseVersions::int64Check(const TSourceLoc& loc, const char* op)
+{
+    requireProfile(loc, ECoreProfile | ECompatibilityProfile, op);
```
johnkslang: Doesn't this need to check for the presence of the extension?

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -498,6 +508,9 @@ TIntermTyped* TIntermediate::addConversion(TOperator op, const TType& type, TInt
     case EOpAnd:
     case EOpInclusiveOr:
     case EOpExclusiveOr:
+    case EOpAndAssign:
```
johnkslang: This looks like it would affect existing behavior.  Was it a bug?

johnkslang:
I've been going through this entire submission, and it's looking very good overall.  I think very little will need to change and we can pull it into master soon.

glslang/MachineIndependent/Versions.cpp
amdrexu: Initially, I indeed add this:

`requireExtensions(loc, 1, &E_GL_ARB_gpu_shader_int64, "shader int64");`

But it causes the compilation of predefined built-in functions, such as "int64_t abs(int64_t)", to fail. So I remove it. Do you have any solution to handle this? Maybe I miss something that could be used to resolve this problem.

glslang/MachineIndependent/Intermediate.cpp
amdrexu: Yes, there is a latent bug. See this expression:

`u |= i; // u = uint, i = int)`

The` &=` and `^=` are similar. Implicit type promotion should be allowed for them.

amdrexu:
Thank you for this quick review.

amdrexu:
@johnkslang There is any feedback? It is requested by several ISVs and is the base of GL_ARB_shader_ballot. So we need these two extensions implemented in glslang badly. Thank you.

glslang/MachineIndependent/Versions.cpp
johnkslang: I think what needs to happen is:

- int64Check does do the requireExtension check
- glslang.y does not do the checking
- scan.cpp does do the checking, if it's not `parseContext.symbolTable.atBuiltInLevel()`

See some other uses of this in scan.cpp.

amdrexu: Thank you. I understand. I will modify the change and re-send it soon.

amdrexu:
@johnkslang I make some changes and detect `builtInLevel` in the function `int64Check`. So the behavior is as expected now. This commit is updated.

PR #253, opened 2016-04-27
SPV: Handle matrix's OpFConvert vector by vector.

amdrexu:


amdrexu:
@johnkslang I update the codes and it is ready for review and merge. Thanks.

PR #256, opened 2016-04-29
Added missing headers to SetupLinux.sh and index.php.

DavidYen:
Some license information were missing for some of the files, I have
added the proper licensing information as well as author information
for both files.


DavidYen:
This should be a relatively easy pull request, the only thing I was unsure about was the year to place in the copyright header. Do they look right?

johnkslang:
I don't believe those two files need in any way a 3Dlabs' copyright or licence.  They are trivial, not derivatives of the original project, and have nothing to copyright and no IP, and can easily be independently recreated.

There are several tiny helper scripts that just copy something and have no need for this.  Curious, why did you picked these two to modify, of the several that exist?  (Not that I want more trivial "mkdir" or "cp" commands to have licenses.)

Thanks.

xorgy:
Not to mention, I don't think index.php is even used on the site anymore. A completely different version of the page is now on the Khronos Group website as far as I can tell.

DavidYen:
Ok, perhaps the best solution is to simply remove these files then.

PR #257, opened 2016-05-02
Avoid printing to stdout directly in library functions.

antiagainst:
Previously `GlslangToSpv()` reported missing/TBD functionalities
by directly writing to `stdout` using `printf`. That could cause
problems to callers of `GlslangToSpv()`. This patch cleans up
the error reporting logic in `GlslangToSpv()`, `TGlslangToSpvTraverser`,
and `spv::Builder` a little bit to use `ostringstream`.

Also fixed the usage of `GlslangToSpv()` in GTest fixtures to
capture warnings/errors reported when translating AST to SPIR-V.


SPIRV/GlslangToSpv.cpp
```diff
@@ -433,7 +437,7 @@ spv::BuiltIn TGlslangToSpvTraverser::TranslateBuiltInDecoration(glslang::TBuiltI
     case glslang::EbvBaseInstance:
     case glslang::EbvDrawId:
         // TODO: Add SPIR-V builtin ID.
-        spv::MissingFunctionality("Draw parameters");
+        spv::MissingFunctionality(warningsErrors, "Draw parameters");
```
johnkslang: Can we move 'MissingFunctionality' inside a class that has access to 'warningsErrors`, so that not every point of call has to replicate this?  The point is to encapsulate the part that's always the same.

johnkslang:
Actually, if you're interested in a more complete fix, `MissingFunctionality` should be something that gets registered, so that the user of the library can do whatever they want when this happens.  Of course, that's beyond your original scope, just if you're interested.  Would at least like the code line comment to be addressed.  Thanks.

antiagainst:
@johnkslang: revised as suggested.

antiagainst:
This PR fixes a test failure in the current `master` branch.

johnkslang:
I don't know if this was addressing my comment about registering, if so that deserves a bit more conversation.

The reason I suggested having a way to register MissingFunctionality is that it really error handling/abort. It always used to exit(1), which was troublesome, so I pulled those out as a work around, and left some right after the call.

By register, I mean the user of the library provides a callback which can do what the user wants when the shader can't be compiled.  E.g., the standalone executable test harnesses would probably register what used to happen: dump a message to standard out, and then exit.  Or, a driver use of it might have a way of cleaning up the compile without returning to it, as executing past a MissingFunctionality might cause a crash.  Or, maybe someone wants to try throwing an exception (Non-C++, some other kind, like Win32).

In most cases, you'd want the best chance of seeing the message before something goes wrong.  You would not want the messages cached and then lost before having a chance to see them.  Note a sequence like `MissingFunctionality(); exit(1);`.

For comparison, note that at a higher level, when parsing, there is an `infoLog` that does cache messages, as it's expected to be normal to have many such messages, without it being anything wrong with compiler, only with the user code.  However, at this level, after the AST, it's only exceptional conditions where there is something wrong with the compiler itself that we want to get a message out.

So, we aren't really trying to log messages, we are trying to handle an exception.

antiagainst:
@johnkslang: Ah, I misunderstood your comments previously then. Thanks for the detailed explanation, it's much more clear for me now.

I'll step back a little bit and use this PR to provide a minimal quick fix to the failing tests on `master` right now; the registration mechanism will not be a small change, and should be addressed via another PR.

johnkslang:
Thanks.  I should also say I don't know how high the priority is to do the next step.  If it is complex or time consuming, it might not be the next thing to do.  Especially, for example, compared to actually implementing the missing functionality instead. :)

PR #258, opened 2016-05-04
Add configuration for Travis to test on Linux and Mac OS X.

antiagainst:
AFAICT, Travis CI is approved by the Khronos Group.

With this configuration in, we can have pull requests built and tested before going in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,60 @@
+# Linux and Mac Build Configuration for Travis
+
+language: cpp
+
+os:
+  - linux
+  - osx
+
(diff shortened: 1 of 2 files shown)


antiagainst:
@johnkslang This needs your help to turn on Travis since I don't have the privilege.

antiagainst:
@johnkslang: And for a quick preview of what it looks like when running: https://travis-ci.org/antiagainst/glslang. (Yeah, one of the test is failing because of the `printf` issue in `MissingFunctionality`, which I already have a fix at #257.)

PR #265, opened 2016-05-05
Implement the extension GL_ARB_shader_group_vote.

amdrexu:


SPIRV/GlslangToSpv.cpp
```diff
@@ -3243,8 +3255,27 @@ spv::Id TGlslangToSpvTraverser::createUnaryOperation(glslang::TOperator op, spv:
         std::vector<spv::Id> args;
         args.push_back(operand);
         id = builder.createBuiltinCall(typeId, stdBuiltins, libCall, args);
-    } else
-        id = builder.createUnaryOp(unaryOp, typeId, operand);
+    } else {
+        if (op == glslang::EOpAnyInvocation || op == glslang::EOpAllInvocations || op == glslang::EOpAllInvocationsEqual) {
```
johnkslang: This is the generic path.

OpCode-specific processing should occur in the switch statement, and could be delegated to a function for it. E.g., in the switch, similar to how the atomic operations were handled:

```
case glslang::EOpAnyInvocation:
case glslang::EOpAllInvocations:
case glslang::EOpAllInvocationsEqual:
    return handleGroupOperation();
```

johnkslang: I did this myself, after dealing with all the merge conflicts we created for you today.

amdrexu: Thanks very much for your kind help.

PR #269, opened 2016-05-06
SPV: Do not decorate struct members with location or nested members with interpolation qualifiers

scyganINTEL:
This fixes some vulkanCTS tests that use struct arrays as a member of in/out interface blocks.

From Vulkan spec:

> If it is a not a Block, then the structure type must have a Location decoration. Its members are assigned consecutive locations in their declaration order, with the first member assigned to the location specified for the structure type. **The members, and their nested types, must not themselves have Location decorations**

From SPIR-V spec:

>When applied to structure-type members, the Decorations Noperspective, Flat, Patch, Centroid, and Sample can only be applied to the top-level members of the structure type. (Nested objects' types cannot be structures whose members are decorated with these decorations.)


johnkslang:
This has the potential of impacting any drivers that are accidentally relying on complete location information.

Please investigate.  Will leave pending for a couple days for this reason.

yavn:
Hi, this change looks good to me. Testing on my system I saw no difference in behavior (i.e. no driver problem surfaced). This also solves the problem with Location on arrayed members, as locations are now implicit.

I do have other concern though, but maybe let's discuss it in original issue.

PR #271, opened 2016-05-09
fix check for non-positive array size

greg-lunarg:
sizePair.size is unsigned, so the test <=0 was always false if the declared size was negative.


johnkslang:
Have you used gtest yet?  Testing is faster and includes unit tests. See the README to try it out. This PR only includes the script-based (runtests) tests, which has been the traditional way.

I'll accept the merge request, and sometime in the future, the plan is stop replicating individual tests between gtest and runtests, and which point I'll make sure nothing is dropped.  runtests will still be used for link-based and multithreaded testing, which gtest does not cover.

PR #274, opened 2016-05-10
WIP: Use strtoul instead of stoul for Android.

antiagainst:


antiagainst:
@Qining PTAL.

Qining:
LGTM +2 from me.

PR #278, opened 2016-05-11
fix for -Werror=non-virtual-dtor for Android system builds

greg-lunarg:


glslang/Public/ShaderLang.h
```diff
@@ -356,6 +356,9 @@ class TShader {
         // Signals that the parser will no longer use the contents of the
         // specified IncludeResult.
         virtual void releaseInclude(IncludeResult* result) = 0;
+#ifdef __ANDROID__
```
johnkslang: How harmful is it to do this unconditionally, versus just under android?

greg-lunarg: My understanding is that there is a small (verging on negligable) space and
time cost.


Greg Fischer

Project Manager and Senior Engineer

LunarG, Inc. - 3D Driver Innovations

greg@lunarg.com <alan@lunarg.com>

greg-lunarg: My understanding is there is a small (verging on negligable) space and time cost.

PR #284, opened 2016-05-16
Add support for testing file-based HLSL source code in GTest.

antiagainst:


gtests/TestFixture.h
```diff
@@ -77,13 +82,13 @@ enum class Semantics {
 // Enum for compilation target.
 enum class Target {
     AST,
-    Spirv,
+    Spv,
+    BothASTAndSpv,
```
antiagainst: It looks a little bit weird to duplicate; but I won't actually worry too much until we come up with the third target to compile to.

johnkslang: Agreed.  Neither is concievable. It's allowed by the command line, and best for getting lists of semantic errors out.  E.g., a shader that's just tests for a dozen different semantic errors.  But, not strictly needed yet.

I'm more worried about lack of orthogonality in the rest of it, but commented on the issue about it.

gtests/Hlsl.FromFile.cpp
```diff
@@ -0,0 +1,81 @@
...
+    ToSpirv, HlslCompileTest,
+    ::testing::ValuesIn(std::vector<FileNameEntryPointPair>{
+        {"hlsl.frag", "PixelShaderFunction"},
+    }),
+    FileNameAsCustomTestSuffix
+);
+// clang-format on
+
+}  // anonymous namespace
+}  // namespace glslangtest
```
johnkslang: I'd already basically written this locally, but will pull in yours, thanks.

PR #285, opened 2016-05-16
VS2010 compilation fixes

baldurk:
I'm still a little fuzzy on whether VS2010 compilation is intended to be supported upstream here or not, if it's not then I can just keep this up to date in my own fork.

Anyway for perusal here are fixes for C++11 constructs not supported on VS2010 that should not regress any functionality or anything or break on other compilers, as far as I understand the C++11 semantics anyway.


johnkslang:
I think these are worth it.

PR #288, opened 2016-05-17
Add support for querying vertex attributes in reflection API

thp:
This adds support for querying which vertex attributes are specified as inputs in the vertex shader and for querying the number of attributes, their names and types similar to querying uniform variables.

### DIFF
--- a/Test/reflection.vert
+++ b/Test/reflection.vert
@@ -94,6 +94,12 @@ struct deep3 {
     ivec3 v3;
 };
 
+in float attributeFloat;
+layout(location = 2) in vec2 attributeFloat2;
+in vec3 attributeFloat3;
+in vec4 attributeFloat4;
+in mat4 attributeMat4;
+
 uniform deep3 deepA[2], deepB[2], deepC[3], deepD[2];
 
 const bool control = true;
@@ -167,4 +173,10 @@ void main()
 
     f += arrBl[2].foo + arrBl[0].foo;
     f += arrBl2[i].foo;
+
+    f += attributeFloat;
+    f += attributeFloat2.x;
+    f += attributeFloat3.x;
+    f += attributeFloat4.x;
+    f += attributeMat4[0][1];
 }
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/reflection.cpp
```diff
@@ -671,6 +687,10 @@ void TLiveTraverser::visitSymbol(TIntermSymbol* base)
 {
     if (base->getQualifier().storage == EvqUniform)
         addUniform(*base);
+
+    // FIXME: Is this the right way to distinguish vertex inputs from fragment inputs?
```
johnkslang: No, this is just asking if the inputs had smooth, flat, or noperspective qualifiers.

You should be able to tell the stage by looking at `language`; e.g., intermediate.getStage().

glslang/MachineIndependent/reflection.cpp
```diff
@@ -671,6 +687,9 @@ void TLiveTraverser::visitSymbol(TIntermSymbol* base)
 {
     if (base->getQualifier().storage == EvqUniform)
         addUniform(*base);
+
+    if (intermediate.getStage() == EShLangVertex && base->getQualifier().isPipeInput())
```
thp: Tested this with some of my shaders and seems to work well, checking if the stage is the vertex language and then just checking if it's an input variable.

glslang/MachineIndependent/reflection.cpp
```diff
@@ -728,6 +747,11 @@ void TReflection::dump()
         indexToUniformBlock[i].dump();
     printf("\n");
 
+    printf("Vertex attribute reflection:\n");
```
thp: Also added this block here to be able to look at the generated reflection data using:

```
build/StandAlone/glslangValidator -l -q shader.vert shader.frag
```

Test/reflection.vert
```diff
@@ -94,6 +94,12 @@ struct deep3 {
     ivec3 v3;
 };
 
+in float attributeFloat;
```
thp: I tried to test different types and with/without layout qualifiers here, and with different types -- anything else that comes to mind that should be tested that isn't here already?

Issue #279, opened 2016-05-12
Generalizing option selection for gtests


johnkslang:
```
enum class Semantics {
    OpenGL,
    Vulkan,
};

// Enum for compilation target.
enum class Target {
    AST,
    Spirv,
};
```

@antiagainst I looked at adding HLSL tests here, and I think we need a different design.  Generally, all these things can be orthogonal, much like this subset of mask flags:

```
enum TOptions {
    EOptionIntermediate       = 0x0001,
    EOptionRelaxedErrors      = 0x0008,
    EOptionGiveWarnings       = 0x0010,
    EOptionLinkProgram        = 0x0020,
    EOptionMultiThreaded      = 0x0040,
    EOptionDumpConfig         = 0x0080,
    EOptionDumpReflection     = 0x0100,
    EOptionSuppressWarnings   = 0x0200,
    EOptionDumpVersions       = 0x0400,
    EOptionSpv                = 0x0800,
    EOptionHumanReadableSpv   = 0x1000,
    EOptionVulkanRules        = 0x2000,
    EOptionReadHlsl          = 0x10000,
};
```

For example, in the `Semantics` and ` Target`, I might want both AST and SPIR-V to be visible, or with HLSL *input* semantics and either `OpenGL` or `Vulkan` *output* semantics.  So, while those last two *output* semantics are mutually exclusive, most the rest are generally orthogonal.

Should I start adding more and more `enum class { }` that cause expansion of the argument lists etc. everywhere?  Or, switch to a mask design like `TOptions` has so that it's all wrapped up in a single parameter to pass around?  In the end, it'll be almost as many enum classes as there are bits in TOptions, but more cumbersome.

Interested in your take on generalizing this design, and certainly open to anything you can do here, as it doesn't quite feel right to keep expanding the number of arguments to pass around everywhere.

Thanks.

antiagainst:
Although we've implemented the bits to run file-based test cases in the googletest framework, I'd prefer to see actual unit tests for HLSL since that's the conventional way of using googletest. I'll push towards that direction and try to write some unit tests for the HLSL part.

[...]

PR #267, opened 2016-05-05
Remove duplicated tests from the runtests script.

antiagainst:
Offspring of https://github.com/KhronosGroup/glslang/pull/261.

Quoting @johnkslang, before merging this PR, the following statements need to be true:

- [x] we see Travis successfully executing gtests
- [x] we see general use of gtests with submitters
- [x] we have working Windows instructions for gtests plus runtests

PR #293, opened 2016-05-18
Also do link tests in the GTest framework

antiagainst:
This PR is another step towards deprecating `runtests`.

Related to #279: I'm leaning towards to use `TOption` like bitmask now.


johnkslang:
This is good to have the gtest path have this ability.

However, I'm not sure it is yet good to take it out of the runtest.  I suspect there are many places (including the recently added bots to this repository) that are using runtest to verify things are okay, and this dilutes that.

Can we start by having tests in both places?  Maybe with time we can reduce the amount in runtests, but it is an almost free way to do sanity checks in places that have not pulled in gtests, so it might be good to have a (smaller) mix of tests there for some time.

johnkslang:
The removal part is somewhat a duplicate of pull request #267, which it is checked off that submitters have generally switch to using gtests.  How do we know that?

antiagainst:
@johnkslang, ah, that's accidentally checked. Just wanted to remove the [WIP] tag... Will remove the tick.

I'm OK to keep the `runtests` code segment for this. Will revise.

antiagainst:
Done. ;)

PR #295, opened 2016-05-18
Parser: Redeclaration of gl_CullDistance is disallowed mistakenly.

amdrexu:
I have this fragment shader which causes the compilation fails. Redeclaration of `gl_CullDistance `should be allowed.

```
#version 450

layout(location = 0) out float o1;
layout(location = 1) out float o2;

layout(binding = 0) uniform Uniforms
{
    int i;
};

in float gl_ClipDistance[6];
in float gl_CullDistance[2];

layout(location = 1) in float in1;

void main()
{
    o1 = gl_ClipDistance[3] + gl_CullDistance[1];
    o2 = gl_ClipDistance[i] + gl_CullDistance[i] + in1;
}
```

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -392,6 +392,8 @@ void TIntermediate::finalCheck(TInfoSink& infoSink)
 
     if (inIoAccessed("gl_ClipDistance") && inIoAccessed("gl_ClipVertex"))
         error(infoSink, "Can only use one of gl_ClipDistance or gl_ClipVertex (gl_ClipDistance is preferred)");
+    if (inIoAccessed("gl_CullDistance") && inIoAccessed("gl_ClipVertex"))
+        error(infoSink, "Can only use one of gl_CullDistance or gl_ClipVertex (gl_ClipDistance is preferred)");
 
     if (userOutputUsed() && (inIoAccessed("gl_FragColor") || inIoAccessed("gl_FragData")))
         error(infoSink, "Cannot use gl_FragColor or gl_FragData when using user-defined outputs");
--- a/Test/450.frag
+++ b/Test/450.frag
@@ -47,3 +47,10 @@ void foo()
     s += imageSamples(i2dmsa);
     float f = imageAtomicExchange(i2dmsa, ivec3(in3), 2, 4.5);
 }
+
+in float gl_CullDistance[6];
+
+float cull(int i)
+{
+    return (i >= 6) ? gl_CullDistance[5] : gl_CullDistance[i];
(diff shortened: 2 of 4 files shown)


glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -390,8 +390,8 @@ void TIntermediate::finalCheck(TInfoSink& infoSink)
     if (invocations == TQualifier::layoutNotSet)
         invocations = 1;
 
-    if (inIoAccessed("gl_ClipDistance") && inIoAccessed("gl_ClipVertex"))
-        error(infoSink, "Can only use one of gl_ClipDistance or gl_ClipVertex (gl_ClipDistance is preferred)");
+    if ((inIoAccessed("gl_ClipDistance") || inIoAccessed("gl_CullDistance")) && inIoAccessed("gl_ClipVertex"))
+        error(infoSink, "Can only use one of gl_ClipDistance (gl_CullDistance) or gl_ClipVertex (gl_ClipDistance is preferred)");
```
johnkslang: This needs to be more clear.  I suspect leaving the existing test as it was, and just adding more tests, with simpler error messages will work better.

johnkslang:
This change looks in the right direction, but with this differ, it will be much easier to tell without the line number changes.  It's also failing the bot tests, but I don't know why.

Can you try again, but with two commits, one that just adds blank lines to the test to cause a line-number change only, and then one that adds the redeclaration without changing line numbers, to just show the functional difference?

antiagainst:
For the buildbot failure, I did a local check out and it turns out the problem is trailing whitespace.

For fixing this kind of issue, I'd suggest to run the `glslangtests` binary directly with `--update-mode`, which will flush the expected result file with the actual output. Then with `git status`, we can see clearly what's changed and expected to be fixed.

(It's interesting that the standalone path ignores some whitespace characters. I'm not sure why that happens yet. Will dig out latter.)

glslang/MachineIndependent/linkValidate.cpp
amdrexu: It is OK. I will correct this.

amdrexu:
@antiagainst Thank you for this finding. I will use `glslangtests`.

amdrexu:
@johnkslang I update the codes and tests.

johnkslang:
> (It's interesting that the standalone path ignores some whitespace characters. I'm not sure why that happens yet. Will dig out later.)

The runtests diff uses options to ignore white space.

antiagainst:
> The runtests diff uses options to ignore white space.

Ah, that's true and obvious via the `-b` command line option. Got it.

PR #300, opened 2016-05-20
Fix the slow down in noContraction propagation

Qining:


dneto0:
I have a "before" and "after" test case.  I'll measure the runtime and stare at the profiles for both.

glslang/MachineIndependent/propagateNoContraction.cpp
```diff
@@ -401,7 +401,8 @@ bool TSymbolDefinitionCollectingTraverser::visitBinary(glslang::TVisit /* visit
     } else {
         // For other binary nodes, still traverse the right node.
         current_object_.clear();
-        return true;
+        node->getRight()->traverse(this);
+        return false;
```
dneto0: This "return false" is redundant.

dneto0:
The code runs much faster now on a test case.  Redundant traversals are now removed.
The profile looks much more normal.  Most of the the time is spent in yyparse and its children. 

+1

PR #305, opened 2016-05-23
Add base class TParseables for intrinsic / builtin generation.

steve-lunarg:
Add base class TParseableBuiltInss for language-dependent intrinsics / builtins generation.  Add stubbed HLSL derivation.  GLSL derivation is still called TBuiltIns (its original name), for historical compatibility.

### DIFF
--- a/glslang/MachineIndependent/Initialize.h
+++ b/glslang/MachineIndependent/Initialize.h
@@ -1,6 +1,6 @@
 //
 //Copyright (C) 2002-2005  3Dlabs Inc. Ltd.
-//Copyright (C) 2013 LunarG, Inc.
+//Copyright (C) 2013-2016 LunarG, Inc.
 //
 //All rights reserved.
 //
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/Initialize.h
```diff
@@ -49,20 +49,48 @@ namespace glslang {
...
+public:
+    POOL_ALLOCATOR_NEW_DELETE(GetThreadPoolAllocator())
+    TParseables();
+    virtual ~TParseables();
+    virtual void initialize(int version, EProfile, int spv, int vulkan) = 0;
+    virtual void initialize(const TBuiltInResource& resources, int version, EProfile, int spv, int vulkan, EShLanguage) = 0;
+    virtual const TString& getCommonString() const { return commonBuiltins; }
+    virtual const TString& getStageString(EShLanguage language) const { return stageBuiltins[language]; }
+
+    virtual void identifyBuiltIns(int version, EProfile profile, int spv, int vulkan, EShLanguage language, TSymbolTable& symbolTable) = 0;
```
johnkslang: These (identifyBuiltIns) were independent before; not strictly tied to the actual parseable strings that this class is encapsulating.  E.g., those strings could come into existence, get processed (parsed and into the symbol table) and go away again.  Then, later identifyBuiltIns could further amend symbol table information, regardless of whether it was tied to string parsing.

I think that independence still exists today, as these could be `static` and they would still work?  Or no?

Maybe they should be a separate parallel hierarchy, or at least `static`?

steve-lunarg: In this PR, yes, they could be static, but more consideration on my part is needed so I don't miss something important.

My intent for subsequent PRs was to preserve that lifetime separation: the strings can come into existence, be parsed into the symtab, and go away.  Later on, identifyBuiltIns can run to amend the symtab - just as today.  However, I wanted that amending to be EShSource-aware, much like it's now EProfile-aware.

I had no intent of ever requiring initialize() to be called in the same compile.  If the symtab remains populated because a scope was never popped, that'll continue to be OK.  Making it virtual on TParseables was merely a convenient way to provide an HLSL specific version.  It could instead be a static method, with the behavior selected via switch() on the EShSource.  I can do that easily.

On the other hand, if the very idea of an EShSource-aware version violates the design, then my understanding was wrong, and I need to revisit.

johnkslang: What's odd is maybe just a more generic name got used to make a more base-level class, which seems appropriate (builtIns -> parseable strings), but then its interface still includes the specific stuff anyway (identify built-ins) making it not actually more generic.

It could be that three levels are best: parseables, built-in parseables, GLSL vs. HLSL built-in parseables.

Another way of saying this: it seems inconsistent to introduce the difference between GLSL and GLSL that GLSL is "TBuiltIns" while HLSL is "TParseableHLSL", which works on saying "built-in" is a GLSL specific term, but then keep the methods saying "builtin" in HLSL anyway.

So, choices:

 - If "built-in" is okay to use for HLSL, then better to have the two derived classes for GLSL and HLSL be something like "TBuiltins" and "HslsBuiltIns", mirroring what was done with TParseContext.
 - if "built-in" is not okay to use for HLSL, need some other alternative, like built-in methods are specific to GLSL while HLSL has intrinsic methods, or so.  I think the first choice might work better.

In both choices, it seems built-in stuff doesn't belong in the most basic class about a parseable, unless the most basic class getting introduced now really is a "parseable built-in" class, in which case it could be named that, reserving for a future more basic "parseable" class.  So, I'd suggest either:

- go to three levels: 1) parseable, 2) built-in parseable, 3) GLSL vs. HLSL
- keep two levels: 1) built-in parseable, 2) GLSL vs. HLSL, reserving just "parseable" for the future

steve-lunarg: It seems better not to conflate going to three levels (separating built-ins from other parseables) with a PR to provide language dependence, so I like the second option better ("keep two levels").  It seems to me that the code already mixes the concept of "parseables" and "built-ins".  E.g, the GLSL initialize() method creates strings that hold both built-in function protos, and non-function parseables such as variables and structures.  If that separation should exist, it seems better for a different PR dedicated to it.

> keep two levels: 1) built-in parseable, 2) GLSL vs. HLSL

I think it's OK to use the "built-in" terminology for HLSL.  Like so then, given the prior desire to maintain "TBuiltIns" as the exact name of the GLSL class?

base = TBuiltInParseables
GLSL = TBuiltIns : TBuiltInParseables
HLSL = HlslBuiltInsParseables : TBuiltInParseables

johnkslang: Sounds good.

PR #308, opened 2016-05-24
Handle multiple ES profile compilation units for a single shader stage

thp:
When trying to link two ES profile shader stages, an error message `"Cannot mix ES profile with non-ES profile shaders"` is displayed and linking fails. It turns out that the new `TIntermediate` created during the merging phase doesn't have version or profile information, causing this error even when all compilation units to be linked are for the ES profile. This is because the `TIntermediate` gets created with version `0` and profile `ENoProfile`.

This pull request is three parts:

1. A test case linking two ES fragment shader compilation units (63d4794) to demonstrate the bug
2. Implementation of a fix by creating the new intermediate with version/profile infos from the first shader compilation unit to be merged (b40a6d6)
3. Fix test case results for the "empty" linker test case (0bb546f), with explanation why it changed (see commit message)


johnkslang:
It is not legal to link together two ES shaders for the same stage.

This has been working correctly for desktop shaders, yes?

thp:
You're right, just looked it up in the GL ES 2 spec: "Multiple shader objects of the same type may not be attached to a single program object." I must have mixed that up with Desktop GL.

Yes, this has been working correctly for desktop shaders.

I can update the pull request to issue an additional error message when trying to merge two ES shaders for the same stage (`Multiple ES shaders of the same type may not be attached to a single program` or something), and when trying to merge a Desktop shader with an ES shader leave the error message at `Cannot mix ES profile with non-ES profile shaders` - if you think that's a good idea.

If not, feel free to just close this PR.

johnkslang:
It looks like you found a valid bug, so a complete fix (proper profile plus the ES error check) would work.  Need either both or neither.  Thanks.

Test/runtests
```diff
@@ -96,6 +96,7 @@ function runBulkTest {
 runBulkTest mains1.frag mains2.frag noMain1.geom noMain2.geom
 runBulkTest noMain.vert mains.frag
 runBulkTest link1.frag link2.frag link3.frag
+runBulkTest es-link1.frag es-link2.frag
```
antiagainst: It would be nice if this test is migrated to use the GTest framework as shown in [this file](https://github.com/KhronosGroup/glslang/blob/master/gtests/Link.FromFile.cpp#L91).

thp: Done.

thp:
Added the check for mixing ES and non-ES shader units as well as for linking multiple ES shaders to the beginning of the linking function, and updated the test cases.

Noticed that even when the error about linking non-ES and ES shaders was issued, the linking was still carried out (albeit with an error); now it returns false early, so the test case results don't output the (bogus?) linking/merging results anymore.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -1526,14 +1526,34 @@ bool TProgram::linkStage(EShLanguage stage, EShMessages messages)
...
+        } else {
+            numNonEsShaders++;
+        }
+    }
+
+    if (numEsShaders > 0 && numNonEsShaders > 0) {
+        infoSink->info.message(EPrefixError, "Cannot mix ES profile with non-ES profile shaders");
+        return false;
+    } else if (numEsShaders > 1) {
+        infoSink->info.message(EPrefixError, "Cannot attach multiple ES shaders of the same type to a single program");
```
johnkslang: Note the semantic was a per-program one, not a per stage one.  So, this probably belongs up one call, in `TProgram::link()`.

Issue #310, opened 2016-05-25
On windows, debug and release builds are installed to the same location with the same file names


jherico:
This makes it problematic to embed this project's libraries into a larger CMake project that needs to build in both debug and release.

PR #311, opened 2016-05-25
Produce different names for files in debug builds on MSVC

jherico:
Fix for #310

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ set(CMAKE_INSTALL_PREFIX "install" CACHE STRING "prefix")
 project(glslang)
 
 if(WIN32)
+    set(CMAKE_DEBUG_POSTFIX "d")
     include(ChooseMSVCCRT.cmake)
     add_definitions(-DGLSLANG_OSINCLUDE_WIN32)
 elseif(UNIX)


johnkslang:
Which file names are effected?

jherico:
The lib file artifacts that are installed, specifically `glslang.lib`, `HLSL.lib`, `OGLCompiler.lib`, `OSDependent.lib`, and `SPIRV.lib`.  With this PR, if you do a debug build in Visual Studio, each file gets a `d` suffix, like so `glslangd.lib`, `HLSLd.lib`, etc...

jherico:
An alternative would be to modify the INSTALL targets so that the debug versions of the libraries are placed in a separate directory, but I don't know how to do that off the top of my head.  using a `d` suffix is pretty common for this kind of thing in my experience.

johnkslang:
Okay, so this will impact all downstream uses that link a debug executable based on the existing names.

jherico:
Potentially, but as it stands, anyone using cmake to create and build the MSVC version of this project would presumably already be having issues since the debug and release versions overwrite one another in the install directory.  However, if someone is already working around that issue or has another build pipeline that uses MSVC, requires the debug libs as well as the release libs and has hardcoded the names for them somewhere, yes they would have to update something.

johnkslang:
So, downstream users that want debug libraries will need to modify their build system to pick them up.  Otherwise, they'll get the non-debug libraries no matter how glslang was built.  Yes?  I think that's okay, just giving a fair warning time.

PR #314, opened 2016-05-27
Added -x option to save SPIR-V as 32-bit hexadecimal numbers to a text file.

mre4ce:
This option makes embedding SPIR-V in source code trivial without the need for additional tools. The Vulkan-Samples repository encourages embedding SPIR-V in the source code to simplify a sample and minimize the build complexity. One of the samples is already using this.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -297,12 +297,6 @@ void ProcessArguments(int argc, char* argv[])
                 break;
             case 'x':
                 Options |= EOptionOutputHexadecimal;
-                binaryFileName = argv[1];
-                if (argc > 0) {
-                    argc--;
-                    argv++;
-                } else
-                    Error("no <file> provided for -x");
                 break;
             default:
                 usage();
@@ -785,7 +779,7 @@ void usage()
            "  -t          multi-threaded mode\n"
            "  -v          print version strings\n"
            "  -w          suppress warnings (except as required by #extension : warn)\n"
-           "  -x  <file>  save 32-bit hex numbers as text to <file>, requires a binary option (e.g., -V)\n"
+           "  -x          save 32-bit hexadecimal numbers as text, requires a binary option (e.g., -V)\n"
            );
 
     exit(EFailUsage);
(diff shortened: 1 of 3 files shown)


xorgy:
I personally think that `-x` should modify the behaviour of `-o` unless they can be used simultaneously. I see a problem where somebody specifies both and sees strange behaviour. They may wonder why their hex file outputs but their binary file does not.

I think it should either be possible to specify both and have them output separate files, or it should be possible to use `-x` as a flag and keep the `-o` behaviour. As it stands, I think your implementation will silently ignore the argument to `-o` when `-x` is present.

StandAlone/StandAlone.cpp
```diff
@@ -294,6 +295,15 @@ void ProcessArguments(int argc, char* argv[])
             case 'w':
                 Options |= EOptionSuppressWarnings;
                 break;
+            case 'x':
+                Options |= EOptionOutputHexadecimal;
```
xorgy: Perhaps this should just be bitmasked instead of modifying binaryFileName, see my PR comment for more information.

mre4ce:
Thanks for the feedback! I will fix the command-line options.

I have been struggling with the same thing; a separate tool or a modification of glslangValidator. I don't like cluttering the glslangValidator with options. However, I do prefer to have a single tool to compile SPIR-V. Let me know how to proceed. I can go either way.

xorgy:
Strictly speaking I hold no official vote in this case.

I would prefer `-x` simply switch the mode, and it should be pretty easy to just remove the part where it changes the file name. If it makes a real workflow significantly easier for it to be integrated into glslangValidator, then that's complexity well spent.

mre4ce:
I've changed the -x option. Let me know if this is what you had in mind.

antiagainst:
Something related: in [shaderc](https://github.com/google/shaderc), glslc already supports [outputting SPIR-V as numbers](https://github.com/google/shaderc/tree/master/glslc#452--mfmtformat).

johnkslang:
Two things are worth considering here:

1. shaderc, as mentioned by @antiagainst 

2. the big picture should really be independent components, each doing one thing, chained together by the user to do what they want at the moment

So, for larger or more complex "side" features, we'll want independent tools to hook together.  Ideally, this also applies to take the disassembler out too.  But, for now, the disassembler still has novel features that are useful, and this PR is simple enough, I think it is okay to merge in.

PR #322, opened 2016-06-02
Initial implementation of direct-mapped subset of HLSL intrinsics

steve-lunarg:
Current limitations:

* About half of the intrinsics are in place - chiefly the ones with direct mappings.  The uncommented lines in TBuiltInParseablesHlsl::identifyBuiltIns in hlsl/hlslParseables.cpp show which are connected.

* Although the prototype generator can create almost all the fn overload signatures, some types are currently disabled until other code comes online.  This will be expanded.  It's a useful subset, though.  At the moment, 1-vectors, non-square mats, and uints are omitted.  *Edit* and doubles.


jims:
This looks great :D Is there a wiki page (or pr/any kind of doc) with the current status of the HLSL front-end? I'm inclined to start playing with this a bit in our engine :) Thanks for the effort!

hlsl/hlslParseables.cpp
```diff
@@ -67,7 +220,263 @@ TBuiltInParseablesHlsl::TBuiltInParseablesHlsl()
...
+        { "transpose",                        "^M",    nullptr,   "M",          "F",      EShLangAll },
+        { "trunc",                            nullptr, nullptr,   "SVM",        "F",      EShLangAll },
+    };
+
+    // Set this to true to avoid generating prototypes that will be invalid for the GLSL parser.
+    // TODO: turn it off (and remove the code) when the HLSL parser can be used to parse builtins.
+    static const bool skipInvalidGlsl = true;
+    
+    // Create prototypes for the intrinsics
+    for (const auto& intrinsic : hlslIntrinsics) {                                          // for each intrinsic...
```
johnkslang: Could you use a standard loop instead of range based?  "auto" is okay, but range based isn't supported yet by some active consumers.

johnkslang:
@steve-lunarg Looks good, except for one minor comment.

@jims The grammar is not yet complete enough for general trying out.

hlsl/hlslParseables.cpp
steve-lunarg: Yes.  I'll push a new version.

steve-lunarg:
> however I believe the point is they are not accessible through GLSL keywords

Yep - it's just a temporary issue until the switch-over for intrinsics parsing.

PR #326, opened 2016-06-03
remapper: do not eliminate interface variables in dce by default

greg-lunarg:
this is done by counting op_entrypoint as a use/def

in lieu of a formal regression test for the remapper, I did before/after comparison across the glslang spv test shaders as well as our entire customer shader base.


steve-lunarg:
Looks OK to me.  The code doesn't quite match the "... by default" phrase in the PR title though.  It looks like it will (currently) do this unconditionally, rather than switched by option.

We should probably build a regression suite for the remapper.

Issue #315, opened 2016-05-30
Validation Error on Spir-V for Vulkan output


TiemoJung:
Hi,

we getting the following error reported by the  vulkan debug layer for some of our shaders generated with glslang:

> SC: SPIR-V module not valid: Operand 4 of MemberDecorate requires one of these capabilities: ClipDistance 

Here is the disassembled Spir-V Code of a shader that triggers this behavior:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 81
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %4 "main" %13 %65 %69 %71 %74 %75
               OpSource GLSL 450
               OpName %4 "main"
               OpName %9 "tmpvar_1"
               OpName %13 "attr_POS"
               OpName %17 "ubo_name"
               OpMemberName %17 0 "viewport_VF0"
               OpMemberName %17 1 "gui_text_depth_VF1"
               OpName %19 ""
               OpName %44 "tmpvar_2"
               OpName %63 "gl_PerVertex"
               OpMemberName %63 0 "gl_Position"
               OpMemberName %63 1 "gl_PointSize"
               OpMemberName %63 2 "gl_ClipDistance"
               OpMemberName %63 3 "gl_CullDistance"
               OpName %65 ""
               OpName %69 "xlv_TEXCOORD0"
               OpName %71 "attr_COL0"
               OpName %74 "xlv_TEXCOORD1"
               OpName %75 "attr_TC0"
               OpDecorate %13 Location 0
               OpMemberDecorate %17 0 Offset 0
               OpMemberDecorate %17 1 Offset 16
               OpDecorate %17 Block
               OpDecorate %19 DescriptorSet 0
               OpDecorate %19 Binding 0
               OpMemberDecorate %63 0 BuiltIn Position
               OpMemberDecorate %63 1 BuiltIn PointSize
               OpMemberDecorate %63 2 BuiltIn ClipDistance
               OpMemberDecorate %63 3 BuiltIn CullDistance
               OpDecorate %63 Block
               OpDecorate %69 Location 1
               OpDecorate %71 Location 3
               OpDecorate %74 Location 2
               OpDecorate %75 Location 8
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
[...]

PR #330, opened 2016-06-07
Defer capability, member decoration until use for: PointSize, ClipDis…

dneto0:
…tance, CullDistance

SpvBuilder: Only record member decorations when they are new.

Fixes an error in the base test results:
 - spv.430.vert: add the required ClipDistance capability

Fixes https://github.com/KhronosGroup/glslang/issues/315

### DIFF
--- a/Test/baseResults/spv.150.geom.out
+++ b/Test/baseResults/spv.150.geom.out
@@ -46,15 +46,13 @@ Linked geometry stage:
                               Decorate 10 Stream 3
                               Decorate 13(fromVertex) Block
                               MemberDecorate 27(gl_PerVertex) 0 BuiltIn Position
-                              MemberDecorate 27(gl_PerVertex) 1 BuiltIn PointSize
-                              MemberDecorate 27(gl_PerVertex) 2 BuiltIn ClipDistance
                               Decorate 27(gl_PerVertex) Block
                               Decorate 27(gl_PerVertex) Stream 0
                               Decorate 29 Stream 0
                               MemberDecorate 30(gl_PerVertex) 0 BuiltIn Position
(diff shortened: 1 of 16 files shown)


johnkslang:
There are multiple objects containing a clip or cull distance, and this seems to only decorate one of them now instead of all of them.  Will line comment on an example.

Test/baseResults/spv.150.geom.out
```diff
@@ -46,15 +46,13 @@ Linked geometry stage:
                               Decorate 10 Stream 3
                               Decorate 13(fromVertex) Block
                               MemberDecorate 27(gl_PerVertex) 0 BuiltIn Position
-                              MemberDecorate 27(gl_PerVertex) 1 BuiltIn PointSize
-                              MemberDecorate 27(gl_PerVertex) 2 BuiltIn ClipDistance
```
johnkslang: So, ClipDistance no longer comes out?

(There were two objects that needed it.)

johnkslang:
Is this actually removing **ClipDistance** decoration, or just trying not to do it multiple times?

(I think we need to revisit some agreements/decisions to actually fully remove the decorations.)

Test/baseResults/spv.150.geom.out
dneto0: I agree this PR should not be merged, as other things are broken.  So this comment is just for completeness.

The only mention of gl_ClipDistance in the source shader is in a comment:
https://github.com/KhronosGroup/glslang/blob/master/Test/spv.150.geom#L22
So I think the PR works as was intended.

johnkslang: Right.  I was accidentally looking at 150.geom source.  Sorry about the red herring.

PR #335, opened 2016-06-08
Fix pointer address spaces for sampler function parameters

### LINKED ISSUE #324: Sampling from texture in function causes address space mismatch.
```
#version 310 es
precision mediump float;

layout(location = 0) in vec2 vTexCoord;
layout(binding = 0) uniform sampler2D uSampler;
layout(location = 0) out vec4 FragColor;

vec4 vtexture(sampler2D samp, vec2 uv)
{
	return texture(samp, uv);
}

void main()
{
	FragColor = vtexture(uSampler, vTexCoord);
}
```

```
                             Capability Shader
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main" 24 28
                              ExecutionMode 4 OriginUpperLeft
                              Source ESSL 310
                              Name 4  "main"
                              Name 16  "vtexture(s21;vf2;"
                              Name 14  "samp"
                              Name 15  "uv"
                              Name 24  "FragColor"
                              Name 26  "uSampler"
                              Name 28  "vTexCoord"
                              Name 29  "param"
                              Decorate 16(vtexture(s21;vf2;) RelaxedPrecision
                              Decorate 14(samp) RelaxedPrecision
                              Decorate 15(uv) RelaxedPrecision
                              Decorate 18 RelaxedPrecision
                              Decorate 19 RelaxedPrecision
                              Decorate 20 RelaxedPrecision
                              Decorate 24(FragColor) RelaxedPrecision
                              Decorate 24(FragColor) Location 0
                              Decorate 26(uSampler) RelaxedPrecision
                              Decorate 26(uSampler) DescriptorSet 0
                              Decorate 26(uSampler) Binding 0
                              Decorate 28(vTexCoord) RelaxedPrecision
                              Decorate 28(vTexCoord) Location 0
                              Decorate 30 RelaxedPrecision
                              Decorate 31 RelaxedPrecision
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeImage 6(float) 2D sampled format:Unknown
               8:             TypeSampledImage 7
               9:             TypePointer Function 8
              10:             TypeVector 6(float) 2
              11:             TypePointer Function 10(fvec2)
              12:             TypeVector 6(float) 4
              13:             TypeFunction 12(fvec4) 9(ptr) 11(ptr)
              23:             TypePointer Output 12(fvec4)
   24(FragColor):     23(ptr) Variable Output
              25:             TypePointer UniformConstant 8
    26(uSampler):     25(ptr) Variable UniformConstant
              27:             TypePointer Input 10(fvec2)
   28(vTexCoord):     27(ptr) Variable Input
               5:             Label
       29(param):     11(ptr) Variable Function
              30:   10(fvec2) Load 28(vTexCoord)
                              Store 29(param) 30
              31:   12(fvec4) FunctionCall 16(vtexture(s21;vf2;) 26(uSampler) 29(param)
                              Store 24(FragColor) 31
                              Return
                              FunctionEnd
16(vtexture(s21;vf2;):   12(fvec4) Function None 13
        14(samp):      9(ptr) FunctionParameter
          15(uv):     11(ptr) FunctionParameter
              17:             Label
              18:           8 Load 14(samp)
              19:   10(fvec2) Load 15(uv)
              20:   12(fvec4) ImageSampleImplicitLod 18 19
                              ReturnValue 20
                              FunctionEnd
```

```
error: 202: OpFunctionCall Argument <id> '26's type does not match Function <id> '9's parameter type.
```

This used to work when sampler variables went through shadow param variables, so this is a recent regression.

The problem seems to be that the function parameter has Function address space and not UniformConstant.

jekstrand:
The first commit is just a storage class cleanup that ensures that all sampler, sampled image, and image pointers end up in the right storage class.  The second fixes the storage class mismatch for sampler function parameters.

Fixes #324


SPIRV/GlslangToSpv.cpp
```diff
@@ -227,7 +227,12 @@ spv::StorageClass TranslateStorageClass(const glslang::TType& type)
         return spv::StorageClassInput;
     else if (type.getQualifier().isPipeOutput())
         return spv::StorageClassOutput;
-    else if (type.getQualifier().isUniformOrBuffer()) {
+    else if (type.getBasicType() == glslang::EbtSampler) {
+        if (type.isImage())
+            return spv::StorageClassImage;
```
johnkslang: This is a big change to make for images.

The existing model is that a GLSL image is an opaque handle, and like other such "resources", go in the **UniformConstant** class.  That is, it's a pointer to an opaque type living in the **UniformConstant** class exchanged between the API and the shader.

Now, if you actually have the image memory itself, not the opaque handle describing it, then you could have a pointer into it, and that would be **StorageClassImage** memory.  Such a pointer can be obtained from the **OpImageTexelPointer** instruction.

There is some room to argue, for such things, how direct of a pointer is being exchanged with the API, but I think for images, as with textures, they are still opaque pointers, especially given the existence of the **OpImageTexelPointer** instruction that can turn the opaque handle into actual memory.

SPIRV/GlslangToSpv.cpp
```diff
@@ -2270,7 +2275,9 @@ void TGlslangToSpvTraverser::makeFunctions(const glslang::TIntermSequence& glslF
         for (int p = 0; p < (int)parameters.size(); ++p) {
             const glslang::TType& paramType = parameters[p]->getAsTyped()->getType();
             spv::Id typeId = convertGlslangToSpvType(paramType);
-            if (paramType.getQualifier().storage != glslang::EvqConstReadOnly)
+            if (paramType.getBasicType() == glslang::EbtSampler)
```
johnkslang: I think we'd want to do this for all opaque "resuorces" being exchanged, not just samplers/images.

johnkslang:
Thanks Jason.  I think this gets to the heart of some issue here, but some subtleties I tried to point out with line comments. So, I think it needs some tweaks. I'm willing to look at doing it too, if you want, but would want to know what the critical thing is you want to get out of it.

I think the key point is that special objects being stored in unusual (non-**Function**) storage classes, and not being copied (just the pointer to them is being copied) for a call, don't forget where they really are when passed to a function.

Really, anytime a pointer is copied, the storage class should be staying the same.  The two cases are

1. An object is getting copied/placed/created to/in function-local memory, so that a pointer to it can be passed to the callee. The pointer really is to the local copy, and hence really is to the **Function** storage class.

2. An object is being left right where it is (including an image handle in the **UniformStorage** class), and a copy of the pointer pointing to it is passed to the callee.  The pointer the callee gets is still to the original object and so must have the same storage class the original pointer had.

Note that images can be viewed as three layers deep:

- bottom layer: the actual image memory, which is not allocated by shader, and lives in the **Image** storage class
- next layer up: an object that is an opaque handle describing this image, which can be passed to the shader from the API, in the **UniformConstant** storage class; this corresponds to the, say, **image2D** declaration in GLSL in the GLSL **uniform** storage class
- top layer: a pointer to this opaque object, which simply results from the **OpVariable** allocation in the shader

The **OpImageTexelPointer** gets you from the middle layer to the bottom layer in this hierarchy, because we decided atomics took pointers to actual memory, and not opaque handles + texel information indirectly describing where that memory was

SPIRV/GlslangToSpv.cpp
jekstrand: Ok, that distinction was not at all clear.  I looked in the SPIR-V spec and I couldn't find anything like that.  I think it would be good to have some sort of an explanation in the spec for what each of the different storage classes mean.  That distinction in particular is very important.

jekstrand:
Ok, I've read your comments and I think we're on the same page now.  I just force-pushed with some new commits that should clear up the problems.  I *think* that it is now strictly better than it was before.

SPIRV/GlslangToSpv.cpp
```diff
@@ -2270,7 +2272,9 @@ void TGlslangToSpvTraverser::makeFunctions(const glslang::TIntermSequence& glslF
         for (int p = 0; p < (int)parameters.size(); ++p) {
             const glslang::TType& paramType = parameters[p]->getAsTyped()->getType();
             spv::Id typeId = convertGlslangToSpvType(paramType);
-            if (paramType.getQualifier().storage != glslang::EvqConstReadOnly)
+            if (paramType.isOpaque())
+                typeId = builder.makePointer(TranslateStorageClass(paramType), typeId);
```
johnkslang: If the pointer is copied, I think the storage class should be copied from the source pointer, not re-deduced.  Note the current TranslateStorageClass would re-place a sampler to **UniformConstant** if in fact a local variable had made a copy of it (even though that might be currently disallowed).

johnkslang:
Agreed we seem to be on the same page about this, and that this hasn't been entirely clear because in part we are just now making it clear.  I think you are getting the intended results now too.

There are a couple nuances to pile on:

1) One *could* consider making a copy of an opaque handle.  Indeed GLSL has discussed multiple times having

```
uniform sampler2D global_s; // opaque handle living in uniform storage class
void foo()
{
    sampler2D s = global_s; // opaque handle living in function storage class
    bar(global_s);  // bar should get a pointer pointing into UniformConstant memory
    bar(s); // bar should get a pointer pointing into Function memory
}
```

So, this interacts with some questions about what layer in the stack call-site specialization should occur, but it also means storage class should not be derived from type for images/textures/samplers.

2)  Do we have the 2-layer or 3-layer model for atomic counters? If they are just like images, then they are opaque handles going into the **UniformConstant** storage class, and we need a way to get to actual memory.  If the shader allocation of an atomic counter is the semantic illusion of actually allocating the atomic counter, then their address is to **AtomicCounter** memory.  I think we're aiming for the latter?

I understand this has not been completely spelled out (or possibly figured out in this detail), and appreciate your participation in this.

SPIRV/GlslangToSpv.cpp
```diff
@@ -227,13 +227,15 @@ spv::StorageClass TranslateStorageClass(const glslang::TType& type)
         return spv::StorageClassInput;
     else if (type.getQualifier().isPipeOutput())
         return spv::StorageClassOutput;
+    else if (type.getBasicType() == glslang::EbtSampler)
+        return spv::StorageClassUniformConstant;
```
johnkslang: These come from GLSL in the isUniformOrBuffer(), so safer to tie SPIR-V storage class to GLSL storage qualifier than the crossing domains from type to storage class.

If pointer copies reuse the source pointer storage class, this path would not be taken later, just when declared globally.

This gets into call-site specialization, which is probably why I originally copied the opaque object to the stack. Not sure yet what's best. It may be that some other rule should change here, that allows that copy.

jekstrand:
> 1) One could consider making a copy of an opaque handle. Indeed GLSL has discussed multiple times having
> 
```
> uniform sampler2D global_s; // opaque handle living in uniform storage class
> void foo()
> {
>     sampler2D s = global_s; // opaque handle living in function storage class
>     bar(global_s);  // bar should get a pointer pointing into UniformConstant memory
>     bar(s); // bar should get a pointer pointing into Function memory
> }
```
> 
> So, this interacts with some questions about what layer in the stack call-site specialization should occur, but it also means storage class should not be derived from type for images/textures/samplers.

This also raises a number of other issues.  If you can shadow variables like that then what prevents you from doing

```
uniform sampler2D a;
uniform sampler2D b;
uniform sampler2D c

void main()
{
    sampler2D s;
    if (uniform_cond_1) {
        s = a;
    } else if (uniform_cond_2) {
        s = b;
    } else {
        s = c;
    }
    gl_FragColor = texture(s, pos);
}
```

This allows you to create a form of sampler indirect that isn't simply indirecting on an array which is something not currently allowed by any of the specs.  Our compiler *cannot* handle this at the moment.  Our hardware can do it, and I've got patches floating around somewhere that make it work in our compiler (they're a bit old at this point), but it's not something we can do right now.

SPIRV/GlslangToSpv.cpp
jekstrand: I see what you mean and, over-all, I think I agree that pulling the storage class from the source class is probably better, but it isn't a full solution.  We need to be able to guarantee that we will always use the same storage class for the uniform variable, the pointer to it (that we create in the code above) and function prototype.  Given that the function prototype is declared without any knowledge of what will be passed into it, the only way that we can guarantee that all three match is to have a direct mapping from GLSL type to storage class.  If we have such a mapping, I see no reason why we can't just use it.

johnkslang:
Agreed we are not yet trying to handle resource values set through control flow, and specs currently prevent that. Just beware code that only works if an opaque pointer is in **UniformConstant**, as that's not quite the criteria (unless it's the application set copy), and that could matter when we get to having call-site specific storage classes and hence want a copy in the function, rather than just a copy of the pointer. (Or, someone writes the code to replicate the callee for each kind of call site.)

PR #336, opened 2016-06-09
Add decompositions for some HLSL intrinsics.

steve-lunarg:
Connect new HLSL intrinsics which require simple decompositions:

clip
isfinite
log10
mul (generic mat/vec/scalar forms)
rcp
saturate
sincos

This PR adds decomposeIntrinsic(), which is either (A) a no-op to pass intrinsics through unmodified to AST opcodes for downstream native handling, or (B) a set of simple decompositions to current AST opcodes, for compatibility with existing downstream code.  Only (B) is currently implemented.

* mul is turned into one of the MatrixTimesMatrix, MatrixTimesVector, MatrixTimesScalar, VectorTimesMatrix, VectorTimesScalar, Dot, or Mul opcodes.  This is only available with square matrices presently.

* clip is turned into either a test for <0 and branch with pixel kill, or a component-wise test with any() followed by same.  The HLSL docs don't specify the behavior on vectors, but experiments with FXC showed any() is what happens.

* log10 is turned into a log2 + a multiply to achieve base 10 log.

* sincos is turned into a separate sin and cos which assign to the output parameters.

* saturate is turned into a clamp to 0,1.

* rcp is turned into a divide.

* isfinite is passed through to an existing SPIR-V opcode via a new AST opcode.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -3028,7 +3028,7 @@ spv::Id TGlslangToSpvTraverser::createBinaryMatrixOperation(spv::Op op, spv::Dec
         return builder.setPrecision(result, precision);
     }
 
-    // Handle component-wise +, -, *, and / for all combinations of type.
+    // Handle component-wise +, -, *, %, and / for all combinations of type.
     // The result type of all of them is the same type as the (a) matrix operand.
     // The algorithm is to:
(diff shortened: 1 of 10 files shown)


steve-lunarg:
Attempting to temporarily withdraw this PR, as I found some helper functions to shorten the expression of the decompositions.  Will re-submit.

PR #337, opened 2016-06-10
Add decompositions for some HLSL intrinsics.

steve-lunarg:
Connect some new HLSL intrinsics which require simple decompositions:

clip
dst
isfinite
log10
mul (generic mat/vec/scalar forms)
rcp
saturate
sincos

This PR adds decomposeIntrinsic(), which is either (A) a no-op to pass intrinsics through unmodified to AST opcodes for downstream native handling, or (B) a set of simple decompositions to current AST opcodes, for compatibility with existing downstream code. Only (B) is currently implemented.

* mul is turned into one of the MatrixTimesMatrix, MatrixTimesVector, MatrixTimesScalar, VectorTimesMatrix, VectorTimesScalar, Dot, or Mul opcodes. This is only available with square matrices presently.

* clip is turned into either a test for <0 and branch with pixel kill, or a component-wise test with any() followed by same. The HLSL docs don't specify the behavior on vectors, but experiments with FXC showed any() is what happens.

* log10 is turned into a log2 + a multiply to achieve base 10 log.

* sincos is turned into a separate sin and cos which assign to the output parameters.

* saturate is turned into a clamp to 0,1.

* rcp is turned into a divide.

* isfinite is passed through to an existing SPIR-V opcode via a new AST opcode.

* dst is turned into a construction of components from each input, per HLSL docs.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -1,5 +1,5 @@
 //
-//Copyright (C) 2014-2015 LunarG, Inc.
+//Copyright (C) 2014-2016 LunarG, Inc.
 //Copyright (C) 2015-2016 Google, Inc.
 //
 //All rights reserved.
@@ -3032,7 +3032,7 @@ spv::Id TGlslangToSpvTraverser::createBinaryMatrixOperation(spv::Op op, spv::Dec
         return builder.setPrecision(result, precision);
     }
 
-    // Handle component-wise +, -, *, and / for all combinations of type.
+    // Handle component-wise +, -, *, %, and / for all combinations of type.
     // The result type of all of them is the same type as the (a) matrix operand.
     // The algorithm is to:
     //   - break the matrix(es) into vectors
@@ -3043,6 +3043,7 @@ spv::Id TGlslangToSpvTraverser::createBinaryMatrixOperation(spv::Op op, spv::Dec
     case spv::OpFAdd:
     case spv::OpFSub:
     case spv::OpFDiv:
+    case spv::OpFMod:
     case spv::OpFMul:
     {
         // one time set up...
@@ -3208,6 +3209,9 @@ spv::Id TGlslangToSpvTraverser::createUnaryOperation(glslang::TOperator op, spv:
         break;
     case glslang::EOpIsInf:
         unaryOp = spv::OpIsInf;
+        break;
+    case glslang::EOpIsFinite:
+        unaryOp = spv::OpIsFinite;
         break;
 
     case glslang::EOpFloatBitsToInt:
--- a/Test/baseResults/hlsl.intrinsics.frag.out
+++ b/Test/baseResults/hlsl.intrinsics.frag.out
@@ -2,7 +2,7 @@ hlsl.intrinsics.frag
 Shader version: 450
 gl_FragCoord origin is upper left
 0:? Sequence
-0:62  Function Definition: PixelShaderFunction(f1;f1;f1; (temp float)
+0:66  Function Definition: PixelShaderFunction(f1;f1;f1; (temp float)
 0:2    Function Parameters: 
 0:2      'inF0' (temp float)
 0:2      'inF1' (temp float)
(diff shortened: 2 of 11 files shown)


hlsl/hlslGrammar.cpp
```diff
@@ -962,6 +963,190 @@ bool HlslGrammar::acceptConstructor(TIntermTyped*& node)
     return false;
 }
 
+// Handle intrinsics that require simple decompositions to AST
+//
+void HlslGrammar::decomposeIntrinsic(HlslToken idToken, TIntermTyped*& node, TIntermTyped* arguments)
```
johnkslang: This source file should be confined to the grammar.

Note that (at least almost entirely?) anything not the grammar was farmed out, usually to either `parseContext` method or `intermediate` method. This looks like something for `parseContext` or even a new thing that does this level of translation.

hlsl/hlslGrammar.cpp
```diff
@@ -977,6 +1162,8 @@ bool HlslGrammar::acceptFunctionCall(HlslToken idToken, TIntermTyped*& node)
 
     node = parseContext.handleFunctionCall(idToken.loc, function, arguments);
```
johnkslang: Here is where the grammar part finished, and the rest is farmed out, so I'd suggest `decomposeInstrinsic` is part of the farmed out part, triggered in `handleFunctionCall`.  It is already the thing figuring out if something is a built-in function or user function, etc.

hlsl/hlslGrammar.h
```diff
@@ -88,6 +89,8 @@ namespace glslang {
 
         bool acceptSemantic();
 
+        void decomposeIntrinsic(HlslToken, TIntermTyped*&, TIntermTyped* arguments);
```
johnkslang: And take this with when it leaves. ;)

hlsl/hlslGrammar.cpp
```diff
@@ -962,6 +963,190 @@ bool HlslGrammar::acceptConstructor(TIntermTyped*& node)
...
+    switch (op) {
+    case EOpGenMul:
+        {
+            // mul(a,b) -> MatrixTimesMatrix, MatrixTimesVector, MatrixTimesScalar, VectorTimesScalar, Dot, Mul
+            TIntermTyped* arg0 = argAggregate->getSequence()[0]->getAsTyped();
+            TIntermTyped* arg1 = argAggregate->getSequence()[1]->getAsTyped();
+
+            if ((arg0->isScalar() && arg1->isVector()) ||
+                (arg0->isScalar() && arg1->isMatrix())) {       // scalar * vec  or scalar * mat
+                node = parseContext.handleBinaryMath(loc, str, EOpMul, arg1, arg0);
```
johnkslang: Curious: All GLSL calls to `handleBinaryMath` preserve argument order.  I suspect this should as well.  Then, I think only dot-product would be special.

hlsl/hlslGrammar.cpp
steve-lunarg: OK.  I will move this to TParseContext; the HlslGrammar::acceptFunctionCall would then be as before.

hlsl/hlslGrammar.cpp
steve-lunarg: Ah, it looks like something downstream (the translation AST->SPIRV?) reverses the arguments to the SPIRV opcode itself, so you're right, this doesn't have to be done here.  I had originally done it because the S,V order was fixed in the SPIR-V opcode.

Will fix, thanks.

hlsl/hlslGrammar.cpp
```diff
@@ -962,6 +963,190 @@ bool HlslGrammar::acceptConstructor(TIntermTyped*& node)
...
+                // component-wise compare: a < 0
+                TIntermAggregate* less = new TIntermAggregate(EOpLessThan);
+                less->getSequence().push_back(arg0);
+                less->setLoc(loc);
+
+                // make vec or mat of bool matching dimensions of input
+                less->setType(TType(EbtBool, EvqTemporary, 
+                                    arg0->getType().getVectorSize(),
+                                    arg0->getType().getMatrixCols(),
+                                    arg0->getType().getMatrixRows()));
```
johnkslang: This might forget if something was a vector of size 1?  That could involve calling the type's makeVector() method.

Some other tools for copying a type without accidentally dropping something on the floor:

- use the shallowCopy() method
- you can then edit the qualifier with getQualifier().makeTemporary(), if the source wasn't already a temp

- you can also directly change the basicType

So, if you want something that is exactly the same, except it's a temp Bool, then you'd do all the above, it all the parts you weren't worrying about will be exactly right.

(This is evolving a bit, even recently with the addition of vec1, so you might see existing doing it the "harder" or less-robust way. Of course GLSL currently can pretend there are no vec1.)

steve-lunarg: > if you want something that is exactly the same, except it's a temp Bool,

Just like that, yes.  I will try your method, thanks.

Issue #341, opened 2016-06-15
requesting sanity check on HLSL memory barrier mapping to SPIR-V


steve-lunarg:
I am adding HLSL memory barrier intrinsics which map to SPIR-V, but I have no direct experience with these intrinsics in any shading language.  I have created these mappings: if someone with detailed knowledge of the space sees a problem, it would be good to hear this :).

* AllMemoryBarrier (+WithGroupSync)
    Will map to spv::OpMemoryBarrier with spv::ScopeCrossDevice and spv::MemorySemanticsAllMemory
    +WithGroupSync will be OpControlBarrier, which is *also* a memory barrier in SPIR-V, using the same scope and memory semantic.

* DeviceMemoryBarrier (+WithGroupSync):
    Will map to spv::ScopeDevice and spv::MemorySemanticsAllMemory

*  GroupMemoryBarrier (++WithGroupSync):
    Will map to spv::ScopeWorkgroup and spv::MemorySemanticsWorkgroupMemoryMask

johnkslang:
Did you find something that implies cross-device barriers?  I doubt this is implemented.

steve-lunarg:
Thanks... the HLSL docs are a bit ... unspecific in this area.  I was initially thinking cross device would be "safe", but I've since been convinced by @johnkslang that may not be the case.  So if All becomes device scoped:

AllMemoryBarrier: ScopeDevice + MemorySemanticsAllMemory (which no longer requires a new AST opcode, since it matches an existing one for both memory and control).

DeviceMemoryBarrier: ScopeDevice + CrossWorkgroupMemory (_??_)  This would match the existing AST OpGroupMemoryBarrier, but there's no AST op for the control form AFAIK).

GroupMemoryBarrier: ScopeWorkgroup + MemorySemanticsWorkgroupMemory (this does **not** match EOpGroupMemoryBarrier... but maybe it's what EOpGroupMemoryBarrier should really be?)

PR #345, opened 2016-06-17
Add HLSL barrier intrinsics, EvaluateAttributeSnapped, lit, fix dst

steve-lunarg:
This adds:

* barrier intrinsics (AllMemoryBarrier, AllMemoryBarrierWithGroupSync, DeviceMemoryBarrier, DeviceMemoryBarrierWithGroupSync,  GroupMemoryBarrier, GroupMemoryBarrierWithGroupSync)

* fixed return type on dst intrinsic

* Add lit intrinsic and (optional, enabled by default) decomposition

* Add EvaluateAttributeSnapped (with optional decomposition that performs coordinate conversion from the HLSL integer coordinate space to float pixel offsets).

The barrier intrinsics introduce new AST opcodes, which are lowered to existing SPIR-V functionality.


SPIRV/GlslangToSpv.cpp
```diff
@@ -3986,6 +3992,27 @@ spv::Id TGlslangToSpvTraverser::createNoArgOperation(glslang::TOperator op)
     case glslang::EOpGroupMemoryBarrier:
         builder.createMemoryBarrier(spv::ScopeDevice, spv::MemorySemanticsCrossWorkgroupMemoryMask);
         return 0;
+    case glslang::EOpAllMemoryBarrier:
+        builder.createMemoryBarrier(spv::ScopeCrossDevice, spv::MemorySemanticsAllMemory);
```
johnkslang: I'm surprised that shaders are doing cross-device memory operations (e.g., synchronizing between multiple GPUs).  Is this correct?  It seems more likely to be the same thing as the existing `glslang::OpMemoryBarrier`, which "waits for completion of accesses to all ... variable types". As in, the "All" part of the name is the `MemorySemanticsAllMemory` bit, not the `ScopeCrossDeviceBit`.  Do we know otherwise?

steve-lunarg:
Closing PR - will resubmit after input from [Issue #341](https://github.com/KhronosGroup/glslang/issues/341), plus a rebase.

PR #346, opened 2016-06-20
Add HLSL memory barrier intrinsics, fix dst, add lit & EvaluateAttrib…

steve-lunarg:
This PR:

- Adds the HLSL memory barrier intrinsics, with input from issue #341, so there are no more cross-device barriers.  Where possible, these are shared with existing AST opcodes, but several new ones are also introduced.  It is possible that EOpWorkgroupMemoryBarrier and EOpMemoryBarrierShared should be the same thing and could then be shared, but I'm not confident to change the existing one for now.  This can be done in another PR if it should be done.

- Adds decompositions for the lit and EvaluateAttributeSnapped intrinsics.

- Fixes the return type for the dst intrinsic, which was incorrectly always a scalar before.


SPIRV/GlslangToSpv.cpp
```diff
@@ -3986,6 +3990,21 @@ spv::Id TGlslangToSpvTraverser::createNoArgOperation(glslang::TOperator op)
     case glslang::EOpGroupMemoryBarrier:
         builder.createMemoryBarrier(spv::ScopeDevice, spv::MemorySemanticsCrossWorkgroupMemoryMask);
         return 0;
+    case glslang::EOpAllMemoryBarrierWithGroupSync:
+        // Control barrier with non-"None" semantic is also a memory barrier.
+        builder.createControlBarrier(spv::ScopeDevice, spv::ScopeDevice, spv::MemorySemanticsAllMemory);
+        return 0;
+    case glslang::EOpGroupMemoryBarrierWithGroupSync:
+        // Control barrier with non-"None" semantic is also a memory barrier.
+        builder.createControlBarrier(spv::ScopeDevice, spv::ScopeDevice, spv::MemorySemanticsCrossWorkgroupMemoryMask);
```
johnkslang: 'EOpGroupMemoryBarrier...' sounds like a workgroup, or group != workgroup?  Does group == crossworkgroup?

PR #354, opened 2016-06-27
Front-end: allow max size built-in arrays like gl_ClipDistance

yavn:
Fixed off-by-one error with gl_MaxClipDistances and similar limits.

I tried to compile the following shader and it failed:

```
#version 450

layout(constant_id = 1) const int numClipDistances = 8;

out gl_PerVertex {
    vec4  gl_Position;
    float gl_ClipDistance[numClipDistances];
};
```

Based on Vulkan and OpenGL specs array of up to ```maxClipDistances``` should be allowed.

### DIFF
--- a/Test/baseResults/120.vert.out
+++ b/Test/baseResults/120.vert.out
@@ -44,7 +44,7 @@ ERROR: 0:107: 'overloadE' : no matching overloaded function found
 ERROR: 0:108: 'overloadE' : no matching overloaded function found 
 ERROR: 0:111: 'overloadE' : no matching overloaded function found 
 ERROR: 0:117: 'overloadF' : no matching overloaded function found 
-ERROR: 0:121: 'gl_TexCoord array size' : must be less than gl_MaxTextureCoords (32)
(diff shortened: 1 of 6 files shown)


yavn:
I noticed this has an unpleasant interaction with implicitly generated gl_PerVertex block.

In such block gl_CullDistance is added automatically, but since minimum size of array is 1, we now get a violation where there are 8 gl_ClipDistances and 1 gl_CullDistance thus putting us over the limit of 8 gl_MaxCombinedClipAndCullDistances...

johnkslang:
> In such block gl_CullDistance is added automatically, but since minimum size of array is 1, we now get a violation where there are 8 gl_ClipDistances and 1 gl_CullDistance thus putting us over the limit of 8 gl_MaxCombinedClipAndCullDistances...

That seems a valid GLSL problem, but separate from this PR, which is still a valid fix to glslang, yes?

yavn:
Yeah, I think PR is still good, it just uncovered that other problem (which arguably is quite an edge case).

I think this is still an improvement, as it should be possible to use up to 8 clip/cull planes as allowed by both OpenGL and Vulkan specs.

PR #358, opened 2016-06-29
SPV: Certain decorations are missing for structure-typed in/out variables

amdrexu:
We encounter a problem from Vulkan conformance tests. This is a tessellation evaluation shader. The "patch" decoration is missing for "block[4]". After investigation, this problem seems much more generic. For instance, you could define a structure and define an input variable with that structure as fragment shader input. When you decorate the input with "flat". The qualifier is not converted to "flat" decoration in SPIR-V tokens.

```
#version 450

layout(quads, equal_spacing, ccw) in;

struct S
{
    int   i1;
    vec4  f4;
    float f1[2];
};

struct TheBlock
{
    float f1[3];
    S     s[3];
    float f1_0;
};

out float f1;

layout(location = 2) in patch TheBlock block[4];

void main()
{
    f1 = block[3].s[2].f4.x;
}
```

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -4024,7 +4024,7 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
     id = createSpvVariable(symbol);
     symbolValues[symbol->getId()] = id;
 
-    if (! symbol->getType().isStruct()) {
+    if (symbol->getBasicType() != glslang::EbtBlock) {
         addDecoration(id, TranslatePrecisionDecoration(symbol->getType()));
         addDecoration(id, TranslateInterpolationDecoration(symbol->getType().getQualifier()));
         addDecoration(id, TranslateAuxiliaryStorageDecoration(symbol->getType().getQualifier()));
--- a/Test/baseResults/spv.430.vert.out
+++ b/Test/baseResults/spv.430.vert.out
@@ -63,6 +63,7 @@ Linked vertex stage:
                               Decorate 55(sampb2) Binding 5
                               Decorate 56(sampb4) DescriptorSet 0
                               Decorate 56(sampb4) Binding 31
+                              Decorate 62(var) Flat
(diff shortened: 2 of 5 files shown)


SPIRV/GlslangToSpv.cpp
```diff
@@ -4024,7 +4024,7 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
     id = createSpvVariable(symbol);
     symbolValues[symbol->getId()] = id;
 
-    if (! symbol->getType().isStruct()) {
+    if (! symbol->getBasicType() != glslang::EbtBlock) {
```
johnkslang: Is that the same thing as `symbol->getBasicType() == glslang::EbtBlock`?

johnkslang: Blocks should not have any of these decorations.

Members should, but that's handled by the code

```
                    // Add interpolation and auxiliary storage decorations only to top-level members of Input and Output storage classes
                    if (type.getQualifier().storage == glslang::EvqVaryingIn || type.getQualifier().storage == glslang::EvqVaryingOut) {
                        if (type.getBasicType() == glslang::EbtBlock) {
                            addMemberDecoration(spvType, member, TranslateInterpolationDecoration(subQualifier));
                            addMemberDecoration(spvType, member, TranslateAuxiliaryStorageDecoration(subQualifier));
```

amdrexu: Yes, you are right.  Should be simplified as this: `symbol->getBasicType() == glslang::EbtBlock`. I will update it.

The case we encounter is: we have a structure variable instead of a interface block variable. Thus, the decoration is completely missing. For structure variable, the decoration should be specified on the variable. For block variable, the decoration should be specified on its member. Please see the tests I change. In all these tests, "flat" decoration is missing in the generated SPIR-V tokens because the "flat" qualifier is placed on a structure variable.

amdrexu:
I have updated the code. The original one is not correct.

PR #359, opened 2016-06-29
HLSL: add template style constructors for vector & matrix types

steve-lunarg:

The basic types are accepted in HlslGrammar::acceptBasicType(), which is used by both of those forms as well as acceptType().  This is because only basic types may appear, and constructs such as vector&lt;vector&lt;float, 2&gt;, 2&gt; are illegal.

A new test for these is also added.

*Edit*: fix angle brackets to be HTML style so they show up in PR comment.


steve-lunarg:
Closing this and trying again - one of the auto-builds didn't complete, but there was no error and it said "potentially a stalled build".

PR #361, opened 2016-06-30
HLSL: Add template style constructors for vector & matrix types

steve-lunarg:

The basic types are accepted in HlslGrammar::acceptBasicType(), which is used by both of those forms as well as acceptType(). This is because only basic types may appear, and constructs such as vector&lt;vector&lt;float, 2&gt;, 2&gt; are illegal.

A new test for these is also added.

*Trying to re-submit*: the first time one of the auto-builds "produced no output after 10 minutes", but gave no report about a code or compilation error.  It said "potentially stalled build", so trying again.


steve-lunarg:
This time it didn't even try: the "no output for 10 minutes" message appeared instantly.

PR #363, opened 2016-06-30
Suggest gl_[Vertex,Instance]Index when gl_[Vertex,Instance]ID is encountered in Vulkan GLSL

cdwfs:
I'm not sure if this is the correct mechanism to implement it, but having some sort of message like this would've saved me a fair amount of time just now :)


johnkslang:
What was the problem? Should a built-in not have been recognized, or glslang correctly rejected and you just didn't know why?

cdwfs:
The latter. I had some existing GLSL code I was migrating from a GL ES application to Vulkan, and spent way too long wondering why the compiler was complaining about gl_VertexID being undeclared. A helpful message indicating that gl_VertexID  was removed in Vulkan GLSL would have significantly cut down on the amount of frantic web-searching required to fix the problem.

PR #366, opened 2016-06-30
WIP: replacing C style includes with their correct C++ equivalents

slowriot:
...resolves some issues building the Vulkan samples under g++ 5.4.

Fixes the following error:
```
[ 26%] Building CXX object glslang/CMakeFiles/glslang.dir/MachineIndependent/intermOut.cpp.o
In file included from /home/slowriot/code/libraries/VulkanSDK/VulkanSamples/external/glslang/glslang/MachineIndependent/../Include/InfoSink.h:39:0,
                 from /home/slowriot/code/libraries/VulkanSDK/VulkanSamples/external/glslang/glslang/MachineIndependent/intermOut.cpp:38:
/home/slowriot/code/libraries/VulkanSDK/VulkanSamples/external/glslang/glslang/MachineIndependent/intermOut.cpp: In function ‘bool {anonymous}::is_positive_infinity(double)’:
/home/slowriot/code/libraries/VulkanSDK/VulkanSamples/external/glslang/glslang/MachineIndependent/intermOut.cpp:52:15: error: expected unqualified-id before ‘(’ token
   return std::isinf(x) && (x >= 0);
               ^
glslang/CMakeFiles/glslang.dir/build.make:374: recipe for target 'glslang/CMakeFiles/glslang.dir/MachineIndependent/intermOut.cpp.o' failed
```

### DIFF
--- a/glslang/MachineIndependent/glslang_tab.cpp
+++ b/glslang/MachineIndependent/glslang_tab.cpp
@@ -2526,7 +2526,7 @@ while (0)
 #if YYDEBUG
 
 # ifndef YYFPRINTF
-#  include <stdio.h> /* INFRINGES ON USER NAME SPACE */
+#  include <cstdio> /* INFRINGES ON USER NAME SPACE */
 #  define YYFPRINTF fprintf
 # endif
 
--- a/glslang/MachineIndependent/intermOut.cpp
+++ b/glslang/MachineIndependent/intermOut.cpp
@@ -38,11 +38,11 @@
 #include "../Include/InfoSink.h"
 
 #ifdef _MSC_VER
-#include <float.h>
+#include <cfloat>
 #elif defined __ANDROID__ || defined __linux__ || __MINGW32__ || __MINGW64__
 #include <cmath>
 #else
-#include <math.h>
+#include <cmath>
 #endif
 
 namespace {
--- a/glslang/OSDependent/Unix/ossource.cpp
+++ b/glslang/OSDependent/Unix/ossource.cpp
@@ -40,9 +40,9 @@
 
 #include <pthread.h>
 #include <semaphore.h>
-#include <assert.h>
-#include <errno.h>
-#include <stdint.h>
+#include <cassert>
+#include <cerrno>
+#include <cstdint>
 
 namespace glslang {
 
@@ -69,7 +69,7 @@ static void DetachThreadLinux(void *)
 void OS_CleanupThreadData(void)
 {
 #ifdef __ANDROID__
-	DetachThreadLinux(NULL);
+  DetachThread();
 #else
 	int old_cancel_state, old_cancel_type;
 	void *cleanupArg = NULL;
--- a/glslang/OSDependent/Windows/ossource.cpp
+++ b/glslang/OSDependent/Windows/ossource.cpp
@@ -37,11 +37,11 @@
 #define STRICT
 #define VC_EXTRALEAN 1
 #include <windows.h>
-#include <assert.h>
+#include <cassert>
 #include <process.h>
 #include <psapi.h>
-#include <stdio.h>
-#include <stdint.h>
+#include <cstdio>
+#include <cstdint>
 
 //
 // This file contains contains the Window-OS-specific functions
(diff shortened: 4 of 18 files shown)


johnkslang:
Would like some data as to the effect this has on all the other build environments that are currently working.

glslang/MachineIndependent/glslang_tab.cpp
```diff
@@ -2526,7 +2526,7 @@ while (0)
 #if YYDEBUG
 
 # ifndef YYFPRINTF
-#  include <stdio.h> /* INFRINGES ON USER NAME SPACE */
+#  include <cstdio> /* INFRINGES ON USER NAME SPACE */
```
dneto0: I think this is a generated file, so this would be an impermanent change.

glslang/OSDependent/Unix/ossource.cpp
```diff
@@ -69,7 +69,7 @@ static void DetachThreadLinux(void *)
 void OS_CleanupThreadData(void)
 {
 #ifdef __ANDROID__
-	DetachThreadLinux(NULL);
+  DetachThread();
```
dneto0: This change is unrelated to the other changes.  I recommend not doing this in this CL.

glslang/OSDependent/Windows/ossource.cpp
```diff
@@ -134,7 +134,7 @@ unsigned int __stdcall EnterGenericThread (void* entry)
 
 void* OS_CreateThread(TThreadEntrypoint entry)
 {
-    return (void*)_beginthreadex(0, 0, EnterGenericThread, (void*)entry, 0, 0);
+    return (void*)_beginthreadex(0, 0, EnterGenericThread, entry, 0, 0);
```
dneto0: This change is unrelated to the header changes.  I recommend not doing this in this CL.

glslang/OSDependent/Unix/ossource.cpp
dneto0: In fact, the comment at the definition of DetachThreadLinux explains that it's needed because of pthread_cleanup_push demands a different function signature.

dneto0:
I've tested this change successfully as follows:
 - Linux x86-64 with GCC 4.8.4
 - OSX with the XCode 7.3.1 compilers (Apple LLVM 7.3.0 (clang-703.0.31))
 - Cross compile to Android x86 and Android armeabi-v7a using the Android NDK r10e
 - Built the NDK from latest sources (and including this glslang commit), which exercises cross compilation to all currently supported ABIs.

The main hole in my testing is Windows.   :-)

So overall the change is good.  But I've commented on two changes in this MR that are unrelated to the header issue, and at least one of them might introduce a failure.

johnkslang:
I can do the Windows testing, but would like to confine it to the #include changes, if that works.

@slowriot Can we see a subsetted PR for testing and pulling? Thanks.

glslang/MachineIndependent/intermOut.cpp
```diff
@@ -38,11 +38,11 @@
 #include "../Include/InfoSink.h"
 
 #ifdef _MSC_VER
-#include <float.h>
+#include <cfloat>
 #elif defined __ANDROID__ || defined __linux__ || __MINGW32__ || __MINGW64__
 #include <cmath>
 #else
-#include <math.h>
+#include <cmath>
```
baldurk: This seems strange to me since there's an `#elif` above for including `<cmath>`. As it stands this now becomes redundant which I don't think is the intention. I'm wondering why this even has to change - if you're using g++ shouldn't it be on one of those platforms in the `#elif`?

I seem to remember there was a bit of back and forth on this bit of code in other issues so it'd be good to be sure this change won't break some other configuration/compiler. Seems like the easier fix would be to find whichever platform is breaking and add it to the `#elif`, no?

baldurk:
FWIW, compiles on VS2010 for me (which I figured might one of the lowest spec compilers & more likely than others not to support modern header syntax).

glslang/MachineIndependent/intermOut.cpp
johnkslang: I suspect the "math.h" was relegated as "platforms no one knows about", and is just hypothetical, and that the real back and forth here was getting the Windows/Non-Windows distinction correct.

I'm making a different PR to address this, and a broader stab at what this PR was attempting.

johnkslang:
Since answers and new PR have not shown up, and transferring this discussion to PR #366.

PR #367, opened 2016-07-01
HLSL: Add template style constructors for vector & matrix types

steve-lunarg:

The simple scalar types are accepted in HlslGrammar::acceptBasicType(), which is used by both of those forms as well as acceptType(). This is because only basic types may appear, and constructs such as vector&lt;vector&lt;float, 2&gt;, 2&gt; are illegal.

A new test for these type declarations is also added.

This is another try for this PR: previously there was a git error during one of the auto-build processes: "fatal: index-pack failed", which happened when the "travis CI" build attempted to check out the branch.

### DIFF
--- a/hlsl/hlslGrammar.cpp
+++ b/hlsl/hlslGrammar.cpp
@@ -1,5 +1,6 @@
 //
 //Copyright (C) 2016 Google, Inc.
+//Copyright (C) 2016 LunarG, Inc.
 //
 //All rights reserved.
 //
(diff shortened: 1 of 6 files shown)


hlsl/hlslGrammar.cpp
```diff
@@ -291,12 +292,188 @@ void HlslGrammar::acceptQualifier(TQualifier& qualifier)
...
+    }
+
+    // COMMA
+    if (! acceptTokenClass(EHTokComma)) {
+        expected(",");
+        return false;
+    }
+
+    // integer
+    if (! peekTokenClass(EHTokIntConstant)) {
```
johnkslang: So, this is not allowed?  (Believable, just checking.)

```
const int w = 4;
vector<float,w> ...;
```

hlsl/hlslGrammar.cpp
```diff
@@ -291,12 +292,188 @@ void HlslGrammar::acceptQualifier(TQualifier& qualifier)
...
+        return false;
+
+    if (! acceptTokenClass(EHTokLeftAngle)) {
+        // in HLSL, 'vector' alone means float4.
+        new(&type) TType(EbtFloat, EvqTemporary, 4);
+        return true;
+    }
+
+    TBasicType basicType;
+    const bool isBasicType = acceptBasicType(basicType);
```
johnkslang: Minor tweak:

Seems the point of pulling out the `acceptBasicType` set of stuff is that it is really the "component" types to build things from. So, actually, if you did that, logic would be simpler (e.g., leave `void` where it was), and alignment between what's expected and what the function does would be greater.  If a vector of a non-basic type was added, it also is easier to see adding that new type to whatever that set of types is to a `acceptComponentType` or so.

hlsl/hlslGrammar.cpp
steve-lunarg: Right; I verified in FXC.  It wants a literal integer.

hlsl/hlslGrammar.cpp
steve-lunarg: Agreed, that seems better.  I'll make the change and resubmit.

thanks.

steve-lunarg: OK, I added a new template type acceptance method which accepts just that subset, and then left acceptType much less perturbed since the only use is now in the vector and matrix template acceptance fns.

Issue #370, opened 2016-07-06
Cache for builtin symbol tables not sensitive to GLSL vs. HLSL source language


dneto0:
This is similar to #166 but now the difference that is not being accounted for is a switch between source languages GLSL and HLSL.

That is, SetupBuiltinSymbolTable takes a "source" parameter of type EShSource but then the check for an entry in CommonSymbolTable is not sensitive to that parameter, and the function can early-return when it should not.  In that case you can try to process an HLSL source program when you have builtins for GLSL, or vice versa.

johnkslang:
Fixed by 0a6fb8549456c0b9d016eed5dec0cff1e2029662.

Issue #166, opened 2016-02-19
Cache for builtin symbol tables not sensitive to spv or vulkan parameters


dneto0:
In ShaderLang.cpp, the function
```
void SetupBuiltinSymbolTable(int version, EProfile profile, int spv, int vulkan)
```
sets up symbol tables for builtin functions.  It remembers which symbol tables it has set up, based on version and profile, and returns early if the tables have already been created.  This can save a lot of time when doing repeated compiles.
However, this scheme is insensitive to the spv or vulkan settings.  So if you're not careful, you can get invalid symbol tables when switching between spv and not-spv, or vulkan and not-vulkan.

We worked around this in Shaderc, but that workaround is fragile.  The real fix is to make the caching scheme in SetupBuiltinSymbolTable sensitive to the spv and vulkan parameters.

johnkslang:
Yes, if a single instance of glslang will be used for multiple API's, then we need to include the vulkan version as part of the cached key.

First, need to formalize better what the vulkan and spv versions are, so there can be a formal mapping between them and the cache key.  E.g., do they look like 100, 110, etc.?

dneto0:
I believe this has already been fixed, via
 - b901ade058 which consolidated the "spv" and "vulkan" parameters into  an SpvVersion struct parameter to SetupBuiltinSymbolTable, and then
 - 10119719 to make the builtin table cache sensitive to SpvVersion
Thanks!

johnkslang:
Agreed.  I failed to see the connection to this issue when fixing it.

PR #371, opened 2016-07-06
WIP: Gtest: tear down builtin tables only when switching between HLSL and …

dneto0:
…GLSL

Issue #166 has been fixed: builtin symbol table is now sensitive to
switch between spv vs. not, and between Vulkan vs. GL.

But issue #370 remains: builtin symbol table is not sensitive to the
switch between GLSL and HLSL

### DIFF
--- a/gtests/Initializer.h
+++ b/gtests/Initializer.h
@@ -46,7 +46,7 @@ namespace glslangtest {
 //
 // TODO(antiagainst): It's a known bug that some of the internal states need to
 // be reset if semantics change:
-//   https://github.com/KhronosGroup/glslang/issues/166
+//   https://github.com/KhronosGroup/glslang/issues/370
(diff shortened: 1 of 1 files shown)


dneto0:
There's no rush on this.  In particular, once #370 is fixed, then we can simplify much more: the Initializer::acquire method can completely ignore the new_messages parameter.

johnkslang:
Looks like a good change, but since there is no hurry, will save it to indicate where to make the proper change soon.  That'll be when HLSL symbol population path is pinned down and there is no longer a table caching problem.

dneto0:
+1.  Added WIP to the title. (I can't add WIP label)

johnkslang:
This is more fully fixed by 0a6fb8549456c0b9d016eed5dec0cff1e2029662.

PR #372, opened 2016-07-06
Fix range of components used without component qualifier

chrisforbes:
When no component qualifier is provided, the behavior is as if
component=0. Previously, we'd consider all components used in
this case, and then generate bogus collisions.

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -665,10 +665,10 @@ int TIntermediate::addUsedLocation(const TQualifier& qualifier, const TType& typ
     }
 
     TRange locationRange(qualifier.layoutLocation, qualifier.layoutLocation + size - 1);
-    TRange componentRange(0, 3);
+    TRange componentRange(0, type.getVectorSize() ? type.getVectorSize() - 1 : 3);
     if (qualifier.hasComponent()) {
-        componentRange.start = qualifier.layoutComponent;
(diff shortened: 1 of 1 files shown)


chrisforbes:
@johnkslang I see I need to update the tests too; will do that shortly. Do you agree with the actual change?

johnkslang:
Couple comments:
 
- Note this from the change log of GLSL, which means this isn't just a function a vector size:

> Bug 13320: Clarify that doubles take up two component's worth of a location, when using component numbers with layout(location = L, component = C).

- `getVectorSize()` returns 1 for scalars, which seems like it might simplify your logic.  There is `isVector()` that distinguishes if you need to, but in this case, I doubt that's needed.

chrisforbes:
Right, I was concerned about non-scalar and non-vector things, where I really want all components used.

chrisforbes:
I'll adjust this to behave correctly for doubles.

johnkslang:
@chrisforbes any update on this?

Thanks.

johnkslang:
Recent commits (https://github.com/KhronosGroup/glslang/commit/c8f6903b1d1e490463bf199bf503fcf2fe388eba) attempt to superset this change.  Let me know if something is still not covered, or submit an updated PR.

chrisforbes:
OK, this is roughly the direction I was going. I'm not convinced what you've done is quite correct for dvec3/dvec4, but I'll poke around and characterize it better.

PR #379, opened 2016-07-08
HLSL: Sampler/texture declarations, method syntax, partial Sample method

steve-lunarg:
This PR adds:

- Declarations for HLSL sampler and texture types.  Some features such as immediate sampler state declarations are parsed but dropped with a warning.  Only separate declarations are currently allowed, so no DX9 style.

- Method syntax.  The calling object becomes the hidden first parameter.

- A partial implementation of the texture object Sample() method.

This PR is aimed at a narrow vertical path, which will be broadened in future PRs.  Limitations include:

- Only DX10 style syntax is in place (though DX9 will be added later).

- Only the features seen in Test/hlsl.sample.basicdx10.frag are connected.  Stray outside that, and it won't work.  For example, no arrays, no multisample, no texel offset, no other methods, etc.

### DIFF
--- a/hlsl/hlslGrammar.cpp
+++ b/hlsl/hlslGrammar.cpp
@@ -69,6 +69,11 @@ void HlslGrammar::expected(const char* syntax)
     parseContext.error(token.loc, "Expected", syntax, "");
 }
 
+void HlslGrammar::unimplemented(const char* error)
+{
+    parseContext.error(token.loc, "Unimplemented", error, "");
+}
+
 // Only process the next token if it is an identifier.
 // Return true if it was an identifier.
 bool HlslGrammar::acceptIdentifier(HlslToken& idToken)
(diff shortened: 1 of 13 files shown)


hlsl/hlslGrammar.cpp
```diff
@@ -134,8 +416,35 @@ bool HlslGrammar::acceptDeclaration(TIntermNode*& node)
     // typedef
     bool typedefDecl = acceptTokenClass(EHTokTypedef);
 
-    // fully_specified_type
     TType type;
+
+    // sampler declaration
+    if (acceptSamplerDeclaration(type)) {
```
johnkslang: Can this be merged with acceptFullySpecifiedType()?  See other line comment.  Otherwise, grammar has to be duplicated for the rest of the statement?  Maybe some tradeoff I don't see.

hlsl/hlslGrammar.cpp
```diff
@@ -105,8 +110,285 @@ bool HlslGrammar::acceptCompilationUnit()
...
+{
+    const EHlslTokenClass textureType = peek();
+
+    TSamplerDim dim = EsdNone;
+    bool array = false;
+    bool ms    = false;
+
+    switch (textureType) {
+    case EHTokBuffer:            dim = EsdBuffer;                      break;
+    case EHTokTexture1d:         dim = Esd1D;                          break;
```
johnkslang: It seems all these are types?  Can these be listed in the switch we already have for peeking at the first token in a type declaration in acceptType()?  This would be analogous to how the vector<> and matrix<> templates were recognized.

PR #380, opened 2016-07-08
HLSL: Sampler/texture declarations, method syntax, partial Sample method

steve-lunarg:
This PR adds:

- Declarations for HLSL sampler and texture types.  Some features such as immediate sampler state declarations are parsed but dropped with a warning.  Only separate declarations are currently allowed, so no DX9 style.

- Method syntax.  The calling object becomes the hidden first parameter.

- A partial implementation of the texture object Sample() method.

- Compiler warning fix from fbarchard

This PR is aimed at a narrow vertical path, which will be broadened in future PRs.  Limitations include:

- Only DX10 style syntax is in place (though DX9 will be added later).

- Only the features seen in Test/hlsl.sample.basicdx10.frag are connected.  Stray outside that, and it won't work.  For example, no arrays, no multisample, no texel offset, no other methods, etc.

### DIFF
--- a/hlsl/hlslGrammar.cpp
+++ b/hlsl/hlslGrammar.cpp
@@ -69,6 +69,11 @@ void HlslGrammar::expected(const char* syntax)
     parseContext.error(token.loc, "Expected", syntax, "");
 }
 
+void HlslGrammar::unimplemented(const char* error)
+{
+    parseContext.error(token.loc, "Unimplemented", error, "");
+}
+
 // Only process the next token if it is an identifier.
 // Return true if it was an identifier.
 bool HlslGrammar::acceptIdentifier(HlslToken& idToken)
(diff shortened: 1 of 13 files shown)


johnkslang:
See my comments on request #379, it seems they still apply here.

steve-lunarg:
In reply to the #379 comments (I closed that one due to an error not caught locally):

> Can this be merged with acceptFullySpecifiedType()? See other line comment. Otherwise, grammar has to be duplicated for the rest of the statement? Maybe some tradeoff I don't see.

I think so.  Will investigate and likely push another version.

steve-lunarg:
I moved the sampler and texture type parsing to acceptTexture(), which allows re-use of acceptDeclaration() with some minor changes.

This change doesn't yet have a generic state list acceptor; the one used for immediate sampler state is specific to that particular use.  A fully generic one could be created with the sampler state becoming a use of it.  That seems best for another PR.

hlsl/hlslGrammar.cpp
```diff
@@ -105,8 +110,117 @@ bool HlslGrammar::acceptCompilationUnit()
...
 }
 
+// sampler_state
+//      : sampler_state_assignment SEMI_COLON sampler_state_assignment SEMI_COLON ...
+bool HlslGrammar::acceptSamplerState()
+{
+    // TODO: this should be genericized to accept a list of valid tokens and
+    // return token/value pairs.  Presently it is specific to texture values.
+
+    if (peek() != EHTokLeftBrace)
```
johnkslang: This doesn't really seem to match the grammar in the comment before the function.  Can these be aligned better?

hlsl/hlslGrammar.cpp
```diff
@@ -105,8 +110,117 @@ bool HlslGrammar::acceptCompilationUnit()
...
+        return true;
+
+    advanceToken();
```
johnkslang: BTW, all the code up to here should be the same thing as

```
if (! acceptTokenClass(EHTokLeftBrace))
    return true;
```

That is, `acceptTokenClass` is the same thing as `peek` followed by `advance`; it only accepts if it's a match.

steve-lunarg: Ah, OK - will fix shortly and re-push.  Thanks.

hlsl/hlslGrammar.cpp
steve-lunarg: I think the left brace is missing in the comment - probably left over from a previous iteration of this.  I'll fix the comment and re-push.

hlsl/hlslGrammar.cpp
```diff
@@ -487,6 +621,126 @@ bool HlslGrammar::acceptMatrixTemplateType(TType& type)
 }
 
 
+// sampler_type
```
johnkslang: Would like to see consistency in the grammar of having each accept show the grammar it is accepting.

hlsl/hlslGrammar.cpp
```diff
@@ -487,6 +621,126 @@ bool HlslGrammar::acceptMatrixTemplateType(TType& type)
...
+
+    TSampler sampler;
+    sampler.setPureSampler(shadow);
+
+    type.shallowCopy(TType(sampler, EvqUniform, arraySizes));
+
+    return true;
+}
+
+// texture_type
```
johnkslang: Would like to see consistency in the grammar of having each accept show the grammar it is accepting.

hlsl/hlslGrammar.cpp
```diff
@@ -105,8 +110,117 @@ bool HlslGrammar::acceptCompilationUnit()
...
+
+// sampler_declaration_dx9
+//    : SAMPLER identifier EQUAL sampler_type LEFT_BRACE sampler_state_list RIGHT_BRACE
+bool HlslGrammar::acceptSamplerDeclarationDX9(TType& type)
+{
+    if (peek() != EHTokSampler)
+        return false;
+
+    advanceToken();  // consume the sampler type keyword
+
```
johnkslang:
```
if (! acceptTokenClass(EHTokSampler))
    return false;
```

hlsl/hlslGrammar.cpp
```diff
@@ -105,8 +110,117 @@ bool HlslGrammar::acceptCompilationUnit()
...
+                return false;
+            }
+        } else if (stateName == "bordercolor") {
+            return false;
+        } else {
+            expected("texture state");
+            return false;
+        }
+
+        // SEMICOLON
```
johnkslang: Are these semicolon separated lists tested yet?  I didn't see them in the test cases, but could have missed them too.  Just asking.

johnkslang:
Review pass done for this submission, will look for the next one.

hlsl/hlslGrammar.cpp
steve-lunarg: There is a test at the top of hlsl.sample.basicdx10.frag, yeah.  (Well, it only tests that the list is successfully parsed and that an unimplemented warning is emitted).

johnkslang: Ah, I saw that, but didn't connect the dots.  Got it now, thanks.

steve-lunarg:
Updated with fixes for above review comments.

PR #384, opened 2016-07-12
Generate an error for unsized atomic counter array uniforms.

### LINKED ISSUE #382: Assertion crash in arrays.h when binding an unsized atomic_uint array uniform with binding.
Running glslangValidator with this input:
```glsl
#version 420
layout(binding=0) uniform atomic_uint a[];
```
as a fragment shader fails this assertion:
```
glslangValidator: /glslang/MachineIndependent/../Include/arrays.h:246: int glslang::TArraySizes::getCumulativeSize() const: Assertion `sizes.getDimSize(d) != UnsizedArraySize' failed.
fish: “../build/StandAlone/glslangVali…” terminated by signal SIGABRT (Abort)
```

This assertion should maybe be replaced with an error check before getCumulativeSize is ever called.

xorgy:
This should resolve #382.

I'm not certain, but it seems that with this accounted for, the `ARB_shader_atomic_counters` layout offset rules should be working. It occurred to me while reading the spec, it would be interesting/nice to warn for exceeding minimums for `GL_MAX_ATOMIC_COUNTER_BUFFER_SIZE` and friends.

If this looks good enough, I will add a test.

### DIFF
--- a/Test/baseResults/unsizedAtomicCounterArrayUniform.frag.out
+++ b/Test/baseResults/unsizedAtomicCounterArrayUniform.frag.out
@@ -1,0 +1,6 @@
+unsizedAtomicCounterArrayUniform.frag
+Warning, version 420 is not yet complete; most version-specific features are present, but some are missing.
+ERROR: 0:2: '' : atomic counter arrays can not be unsized or implicitly sized. 
+ERROR: 1 compilation errors.  No code generated.
+
+
--- a/Test/unsizedAtomicCounterArrayUniform.frag
+++ b/Test/unsizedAtomicCounterArrayUniform.frag
@@ -1,0 +1,2 @@
+#version 420
+layout(binding=0) uniform atomic_uint a[];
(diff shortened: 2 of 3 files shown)


johnkslang:
Seems sensible.

Did you see specification language prohibiting unsized arrays of atomic_uint?  If that's not actually disallowed, we might need a more sophisticated solution (or spec. change request).

xorgy:
@johnkslang Unsized uniform arrays of any sort seem to be not specified(probably because they make no sense), and uniforms bound to the atomic counter buffer would probably need extra specification.

xorgy:
I'll try not to groan, but it does seems like we'll need a _more sophisticated solution_.

johnkslang:
Yes, to be more clear, implicitly sized arrays are supported in general, so would apply to atomic_uint unless it was specifically called out as not applying.  Some checks need to be deferred to link time, to give the array time to acquire its implicit size.  It definitely must have a size by the time it is all linked.

johnkslang:
I independently fixed issue #382, which does something similar to this PR.  Link-time functionality is missing, as is generally the case, but in this case since the specification says the implicit `offset` assignments go in compilation order, I think the specification should say something about preventing this. Anyway, being tracked now as a Khronos internal bug.

PR #390, opened 2016-07-14
HLSL: Add SampleBias and SampleGrad methods, + associated tests

steve-lunarg:
This PR adds the HLSL SampleBias and SampleGrad texture object methods, and 4 tests for each (basic, arrayed, offset, and array+offset).

Also, the filenames of the previous Sample() method tests are changed to fit the naming scheme of the newer tests.


glslang/Include/intermediate.h
```diff
@@ -526,7 +526,18 @@ enum TOperator {
     EOpTextureBias,                      // HLSL texture bias: will be lowered to EOpTexture
     EOpAsDouble,                         // slightly different from EOpUint64BitsToDouble
 
-    EOpMethodSample,
+    EOpMethodSample,                     // texture object methods
+    EOpMethodSampleBias,                 // ...
```
johnkslang: Want to sanity check: Are all these operators only used temporarily, getting replaced with previously existing ones before the AST is done, such that existing consumers don't have to process these?

If so, I think we should document that consumers won't see these.  If not, I'd like to discuss design more here.

steve-lunarg: > Are all these operators only used temporarily, getting replaced with previously existing ones before the AST is done

Correct.  They are translated in HlslParseContext::decomposeSampleMethods() to existing AST.

I'll push an updated version with a comment to annotate this fact.

steve-lunarg:
Update pushed for code review comments.  Also fixed SampleGrad stage mask, and added a new SampleGrad VS test to ensure it appears outside PS.

PR #393, opened 2016-07-18
Add g++/clang warnings to match some enabled by /W4 in MSVC

steve-lunarg:
This PR adds some warnings to the g++ build which are already enabled in the MSVC build, to prevent code triggering them from escaping notice.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,14 @@ else(WIN32)
 endif(WIN32)
 
 if(CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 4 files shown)


johnkslang:
Thanks.

Note that WS default is not /W4, but I run that way to catch some things.

Others report other problems from other build environments.  The goal here is twofold; to catch things worth catching and to not have a noisy build.
 
The main problem I see with VS /W4 is complaints about constant conditionals, like`while(true)` or `const bool mode = true; if (mode)`.  If that one was silenced, we could also bump up VS to /W4.

Other input from other build environments?

PR #404, opened 2016-07-22
Remap Scope and MemorySemantics IDs in SPIRV remapper

phaulos:
Scope and MemorySemantics operands contain IDs as well and may need
remapping.


steve-lunarg:
Yep, agreed.

PR #406, opened 2016-07-23
HLSL: enable #line extension by default for HLSL source.

steve-lunarg:
This PR enables #line for HLSL, which allows it by default with no extensions needed.

Requesting review of one aspect of this in particular: Both hlslParseContext and one of its base classes TParseVersions had a copy of the extensionBehavior map.  The version in hlslParseContext was effectively hidden, because all the extension testing functions  are in the base class and only were looking at the base class map.

I removed the hlslParseContext version and made the base class version protected.  It would also be possible to have both and override the testing functions in the derived class, but since the base class one is sufficient, this seemed cleaner.

The new hlsl.pp.line.frag test was compiled under FXC for sanity.


johnkslang:
HLSL extension behavior is as yet uncharted territory, and may deserve a bit of thought.  I don't know if the GLSL base works for it or not, though I guess it's pretty general.

Was the problem also in the GLSL inheritance path, or just HLSL?

steve-lunarg:
> Was the problem also in the GLSL inheritance path, or just HLSL?

Just HLSL.  With this PR, there's only one definition of the extension map now, which is in glslang/MachineIndependent/parseVersions.h.  That's the one GLSL has been using all along.  Previously, HLSL and GLSL were using exactly the same values.  Now they are different values in that HLSL enables E_GL_GOOGLE_cpp_style_line_directive by default.  (I haven't gone through the whole extension list though to see what others may or may not be appropriate).

I believe HLSL and GLSL under E_GL_GOOGLE_cpp_style_line_directive both have the same syntax and behavior, modelled after C++, allowing a textual string for the optional source parameter.  In HLSL the __FILE__ text could be sent to printf, although there's no intrinsics implementation of printf at the moment.

In HLSL this isn't an extension (not sure it even has the concept of extensions), which raises the question of whether it's OK to provide its stock behavior by enabling the GLSL cpp_style_line_directive extension by default when compiling HLSL.  It would act like it should, but OTOH, it does piggyback on the extension mechanism.

(Tangential topic: other preprocessor directives have differences between HLSL and GLSL, for example around when trailing () parens are required or prohibited with an empty argument list in #defines).

PR #422, opened 2016-07-29
WIP: fix for block location qualifiers

bpkowaliINTEL:
According to 14.1.4 Location Assignment paragraph:
If block has a Location qualifier then members must not have Location decorations unless they are declared explicitly.

No tests were affected.


johnkslang:
Can you add or modify a test such that the change does show an effect?  Or, was it dead code?

You could also list a test here, if that's easier.  A test change makes it easier to see and discuss what's really happening here.

bpkowaliINTEL:
As I mentioned the change which I made didn't affect any internal test.
Nonetheless, I modified spv.450.tesc case in order to examine location generation code. 
Original test code isn't executed in this case due to use of array of blocks
GlslangToSpv.cpp:2118 
if (! type.isArray())

Modified code is used by following Vulkan tests:
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_implicit.isolines
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_implicit.quads
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_implicit.triangles
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_shader_builtin.isolines
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_shader_builtin.quads
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_shader_builtin.triangles
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_spec_min.isolines
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_spec_min.quads
dEQP-VK.tessellation.user_defined_io.per_patch_block.vertex_io_array_size_spec_min.triangles

johnkslang:
Here is the complete quote from the API specification:

> If the structure type is a Block but without a Location, then each of its members must have a Location decoration. If it is a Block with a Location decoration, then its members are assigned consecutive locations in declaration order, starting from the first member which is initially assigned the location specified for the Block. Any member with its own Location decoration is assigned that location. Each remaining member is assigned the location after the immediately preceding member in declaration order.

Note: It is saying what a "location assigner" (like a driver) should do, based on what it sees in SPIR-V, not really saying how SPIR-V should generate the locations to begin with.

While it says a block without a location requires all members to have a location, it does not say members cannot have a location if the block has a location.  It says

> Any member with its own Location decoration is assigned that location.

In fact GLSL more specifically says member's location override the default inheritance of block location, and Vulkan is at least not contradicting that.  Further, this Vulkan language would apply when that override happens:

> Each remaining member is assigned the location after the immediately preceding member in declaration order.

The corresponding GLSL language is:

> When a block member is declared with a location qualifier, its location comes from that qualifier: The member's location qualifier overrides the block-level declaration. Subsequent members are again assigned consecutive locations, based on the newest location, until the next member declared with a location qualifier.

Which, again, I don't see Vulkan contradicting.  So, in this:

```
layout(location = 10) in bn {
                        vec4 v1; // location 10
  layout(location = 14) vec4 v2; // location 14
                        vec4 v3; // location 15
};
```

The current glslang generates

```
                              MemberDecorate 8(bn) 0 Location 10
                              MemberDecorate 8(bn) 1 Location 14
                              MemberDecorate 8(bn) 2 Location 15
```

Which is correct, but I'm unclear if this PR might be deleting part of the algorithm that computed that 15 (though, as the comment says, this is ideally computed in the linker part of glslang, and that might be happening, making the lines you deleted dead code anyway, rather than an algorithm change).

I'm committing your test change, plus the code above, so it will be easier to see in the future what the practical change is in a PR like this.

johnkslang:
Is there any update on whether there is still a change to make here, relative to the current state of glslang?

johnkslang:
I confirmed the reason the code change in this PR doesn't have the impact it seems it should, is that it is mostly redundant WRT to earlier work the front end did.  In fact, now, the front ends leave the location information in a state that is correct, and so no mirrored algorithms are needed in either `convertGlslangStructToSpvType()` or `decorateStructType`.

I'm doing a full fix that takes this into account, and will make a new PR for it.

johnkslang:
This PR is replaced by #917, which has a full consistent change.

PR #427, opened 2016-08-01
bump required CMake version down to 2.8.11

wesleygriffin:
There are no CMake features that require CMake > 2.8.11 and that is the version that ships on stock CentOS 7. By dropping the CMake version down to 2.8.11 we can build glslang out of the box on stock CentOS 7.


johnkslang:
This is fine with me, if it is true of other environments.

Anyone else see issues with backing up the required CMake version?

johnkslang:
Doing some research, it changed from 2.8 to 2.8.12 in PR #260 to handle target_include_directories, but it does seem that only 2.8.11 is needed for that.

PR #428, opened 2016-08-02
HLSL: Fixing some casts that warn when compiled to 64 bit (size_t is 64 bit…

dankbaker:
… rather then 32 bit)


johnkslang:
Merged.  Generally, the commit message should start with something like "HLSL: ...", or "SPV: ..." , etc.

johnkslang:
Also, it looks like you don't have your .gitconfig holding a matching email address to your GitHub account.  Setting that up enables some more functionality.

PR #429, opened 2016-08-02
HLSL: Add GatherRed/Green/Blue/Alpha methods, inc 4-offset forms

steve-lunarg:
This PR adds the GatherRed/Green/Blue/Alpha family of component-selected texture gather methods, including the 1 and 4 offset forms, but not including the residency status forms (since tiled textures are not yet implemented) or the GatherCmp* forms (since multi-component depth textures are not available in the target language).  Both of those will produce an error message.

There are new tests for all these.  For unimplemented forms, the tests are commented out, and can be uncomment if they become available in the future.


johnkslang:
This seems to add new AST opcodes, but isn't there functionality already for this is the AST? Can you fill in a bit here?

steve-lunarg:
Ah, those should fall under the comment above them about "translated to existing AST methods in HlslParseContext::decomposeSampleMethods()", but I forgot to carry the "// ..." continuation through to those lines.  I'll do that for clarity.

Basically, these have slightly different parameters (e.g, 4 separate vec2 offset parameters vs a single array of 4 vec2s), and there is a translation to existing AST opcodes in that method.  I'll push a better commented version.

steve-lunarg:
New version pushed with updated comments for clarity.

PR #430, opened 2016-08-03
WIP: Skip UTF-8 BOM and multi-byte characters.

DeanoC:
Bug fix for GLslang not accepting strings with the UTF8 BOM marker. Since 4.1 glslang is spec'ed to be utf8 but glslang doesn't handled non ascii UTF8 characters.
This fix removes all multi-byte UTF8 character from shader source, which solves the BOM issue and will allow more shaders to pass through the parser. However it doesn't solve the spec failing issue about if UTF8 characters are intentionally used as names... The will require the parser becoming UTF8 aware at all levels.


johnkslang:
Some comments:

- you would probably want to squash this; it looks like something that should come in 1 or 2 commits
- rewriting the strings is not ideal, it was important a while back to get them read-only all the way through the stack while not having copies made

It could be the best steps are  

1. be sure what rules Khronos wants in this space (I will do some more research)
2. make the real scanner that scans the user-provide strings obey the rules from 1

DeanoC:
Squash done, sorry about that mess before.

I agree, rewriting the strings and still not solving the problem 'properly' isn't great. If Khronos do want proper UTF-8 handling than rejecting this whilst a proper scanner is done might be the best option. Sadly I don't feel confident enough to write that scanner, lex, bison and its ilk are not my speciality by a long way.
If they prefer a hack until its done, this does work though with the caveat of extra allocations. Potentially it could be done inline if the strings weren't const, but that brings in a whole other questions of ROM strings etc. So allocate and clean is the safest option for this style of hack imho

johnkslang:
Thanks.  Worst case, this is a good reference for what to do, as like you say, all the knowledge has to come together, so this helps.  Will label as WIP in the mean time.

Issue #400, opened 2016-07-21
Respect mediump/lowp for desktop #version for Vulkan


johnkslang:
As per recent Khronos decisions for GLSL and Vulkan:

- [x] Stop dropping precision qualifiers on the floor for desktop versions when making SPIR-V for Vulkan.  Default will be `highp` for all types.
- [x] Update precision qualifiers for built-in functions, and the algorithm used to determine precision of operation. (Correct version for ES are the ES spec + Bugzilla 15931, correct version for non-ES are in GL_KHR_vulkan_glsl.)
- [x] Give a warning about different defaults than ES versions, with an easy way to turn it off.  (Using existing precision statements is an okay way to turn it off.) Glslang should be pretty smart about only giving the warning in situations that really need it, and possibly have a command-line option for this.

johnkslang:
Steps 1 and 3 are done in pull request #434, which will be merged soon, based on feedback.

PR #434, opened 2016-08-04
WIP: Vulkan: respect desktop precision qualifiers

johnkslang:
This is steps 1 and 3 in issue #400, available for review.

The key decision to make was about desktop always having a default precision of highp for all types, while that was not true for ES shaders that might have been ported to desktop.  There was desire to

A) catch this and make sure the coder knew about it, and

B) not generate lots of noisy warning messages

To handle this trade-off, here is what this branch does:  It generates a simple two-line warning if all the following are simultaneously true:

- it is a fragment shader (the only stage that ES has int/float defaults that are not highp)
- it is a desktop shader
- it is a Vulkan shader
- the user's shader explicitly used a precision qualifier, somewhere
- the user's shader did not first declare explicit defaults for both int and float

The warning is as follows.

```
WARNING: 0:8: '' : all default precisions are highp; use precision statements to quiet warning, e.g.:
        "precision mediump int; precision highp float;"
```


scyganINTEL:
I think this change breaks doubles:
```
#version 440
layout(location = 0) in highp dmat2 attr0;
```
gives:
```
Warning, version 440 is not yet complete; most version-specific features are present, but some are missing.
ERROR: 0:2: 'double' : type cannot have precision qualifier 
```

johnkslang:
@scyganINTEL is that not the right answer?  Doubles don't have precision levels.  I guess we have two choices:

1. disallow and give error on precision qualifier on double

2. allow but ignore precision qualifier on double

The specification says

> Any single-precision floating-point declaration, integer declaration, or opaque-type declaration can have the type preceded by one of these precision qualifiers

However, since desktop always ignored them anywhere, I could see an issue with doubles and existing content.  If that's a real problem, we could relax the spec and/or error checking.

scyganINTEL:
@johnkslang  I am not assigning precision levels. I am saying, that on this MR, following two lines is enough, to trick glslang into not compiling the shader: 
```
#version 440
layout(location = 0) in highp dmat2 attr0;
```

scyganINTEL:
Ahh, so obvious.. So GL4 globally ignored them and now all of them are in effect. 

So this came from CTS, meaning it is easy to fix, so propably (1) is good for that case. Not sure on other cases.

johnkslang:
This functionality was merged in with 4d535640 and 32c169db so closing this one.

PR #436, opened 2016-08-05
HLSL: disabling dx9 attempt to parse dx9 style samplers

dankbaker:
HLSL compiler runs in 2 modes, specifically for back compat on samplers. That is, to recognize dx9 shaders you need a master switch. Attempting to parse Dx9 style simultanouly with DX10+ won't work and causes valid D3D10 HLSL to fail. You can create typeless samplers in D3D10 just fine.   That is, the line:

sampler MySampler : register (s0);

was trying to parse down the dx9 path, but this is perfectly valid D3D10+ syntax, and in this case means a sampler object, not a combined sampler texture object as it might in dx9 mode.

Thus, I've commented out this attempt to parse the DX9 style samplers. If we want to do this, there should be a master switch that puts the front end in DX9 mode.


steve-lunarg:
Interesting.  Seems OK for now since the DX9 mode is in "unimplemented" state anyway.

@dankbaker: I don't know if the [HLSL docs](https://msdn.microsoft.com/en-us/library/windows/desktop/bb509644(v=vs.85).aspx) are complete here, but the DX9 description says:

`sampler Name = SamplerType { ...`

Given your data above, we cannot disambiguate these cases with one token of lookahead due to "sampler name..." being valid in both cases: hence your change.  Do you think they can be disambiguated with another few tokens of lookahead to detect the "="?  Or does there exist some valid DX9 syntax not described on the doc page that renders even a several token lookahead insufficient?

dankbaker:
The grammer in the docs for dx9 is misleading, it's more or less deprecated. Having sampler state there was something used in DX9 effect system.  I wasn't able to get fxc.exe to accept the = SamplerType { } syntax any more poking at it for a few minutes.  I think you'd have to target DX9 fx files, but that doesn't seem to be a target anymore.

Now, the keyword "sampler" is equivalent to the keyword "SamplerState". When you put it in backward compat mode, thenI think sampler turns into something a bit different. It's a question of how much a priority trying to compile DX9 HLSL shaders we want to do. Our shaders just use the sampler keyword instead of the SamplerState keyword, I imagine many people are similar. 

I don't know enough about how back compat mode works in HLSL (since alot of work happened on that after I left), but it a special mode you put it into which actually changes the grammer a bit. fxc has the /Ges and the /Gec flags.  D3D10 doesn't have combined samplers in DX10+, so it's not an issue to know based on your shader profile that the sampler keyword must mean a SamplerState.

johnkslang:
So, is the conclusion here that we need to add a mode, and have that protect the line of code being commented out?  Or, still up for debate?  The recursive decent approach can figure things out that bison couldn't, so it's conceivable that a mode is not needed, or is auto-detected.

dankbaker:
Well, the first thing to do is verify if this syntax is even valid anymore for current HLSL. I couldn't get it to work on any supported target. I see no reason to support things that fxc doesn't.

johnkslang:
Okay, I'll merge your request, and we can deal with this when need calls for it.  Thanks.

johnkslang:
Note this should be a tab-free project; they display differently everywhere, so glslang sticks to spaces only.  The first one is free. :)

PR #447, opened 2016-08-09
WIP: Standalone: Allowing explicit specification of shader compiltion type via -T opti…

dankbaker:
…on, rather then looking at file extension. For HLSL files, this is nice because .hlsl extension is natively udnerstood by visual studio, likely to be used with the -e option. Otherwise, it is very hard to use standalone util.


johnkslang:
Just process-wise, there are a few problems with this PR:

- it either does not build correctly or pass the built-in test suite
- it combines multiple independent subjects into a single PR

Note that updating test suites also helps demonstrate what effect the code changes have.

PR's can't be pulled if they break things; there are a large number of projects depending on not seeing regressions.  However, this shouldn't slow anything down.  It takes less that one second to run the tests, and, one way or another, changes have to be tested and have tests representing them to prevent the next change in the future from causing regressions.

dankbaker:
I'm a little confused, the PR should have just been one change from one file off the standalone branch, not those other changes.  That appeared to be what the compare tool was telling me.

I was trying this is a test to see if I could make a single change on a file on an independent branch, clearly that didn't work right.  Let me close this out and try it again.

PR #448, opened 2016-08-09
HLSL: add implicit promotions for assignments and function returns.

steve-lunarg:
This PR adds:

- Return type conversion for function returns.  Some of the underpinnings of these were already in place, but not connected.  For example, the currentReturnType was already tracked during function definitions.  This change adds a HlslParseContext::handleReturnValue() function to handle conversions.  Previously, return values were returned "blindly" regardless of the function's return type.  Tests for this are added as part of the new hlsl.promotions.frag test: see the functions beginning with Fn_R_*.  The test also ensures that conversions happen for out params (which was already correct).

- HLSL allows arbitrary basic type conversions on assignment (including, e.g, multiply-assign, etc).  For example, a bool vector can be assigned to a float vector.  hlsl.promotions.frag tests all the combinations for int/uint/float/bool/double.  Note that some involve a loss of precision, but are still valid.  Other matters such as precedence rules are unaffected.  This PR does not impact shape conversions, which were already in place and source language specific.

This PR does NOT address a known defect related to selection of function signatures under promotion situations.  That is a separate problem.

### DIFF
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -524,7 +524,7 @@ TIntermTyped* TIntermediate::addConversion(TOperator op, const TType& type, TInt
         if (type.getBasicType() == node->getType().getBasicType())
             return node;
 
-        if (canImplicitlyPromote(node->getType().getBasicType(), type.getBasicType()))
+        if (canImplicitlyPromote(node->getType().getBasicType(), type.getBasicType(), op))
             promoteTo = type.getBasicType();
         else
             return 0;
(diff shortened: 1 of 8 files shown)


glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -726,11 +726,32 @@ TIntermTyped* TIntermediate::addShapeConversion(TOperator op, const TType& type,
...
-bool TIntermediate::canImplicitlyPromote(TBasicType from, TBasicType to) const
+bool TIntermediate::canImplicitlyPromote(TBasicType from, TBasicType to, TOperator op) const
 {
     if (profile == EEsProfile || version == 110)
         return false;
 
+    // Some languages allow more general (or potentially, more specific) conversions under some conditions.
+    if (source == EShSourceHlsl) {
+        switch (op) {
+        case EOpAndAssign:               // assignments can perform arbitrary conversions
```
johnkslang: Any of bool/int/float/double/void/sampler/struct can convert to any of bool/int/float/double/void/sampler/struct?

johnkslang: Maybe not a big deal, but there may be a conflating of implicit conversion with operation-specific conversions or polymorphism (e.g.,  shift allows mixing int/uint, whether or not a conversion is available).

steve-lunarg: Samplers, structs, and voids should be excluded.  I'll make that change.

steve-lunarg: About shifts, I think they shouldn't trigger this, since they're handled separately in addConversion (similar for example to constructors).

Issue #329, opened 2016-06-05
Document file naming conventions.


ioquatix:
It's not clear how one should name files.

For example, you outline that files would normally be named e.g. `.frag` for fragment shader. But the syntax of the file is glsl, so does `.glsl` also suit? You mention that the glslValidator uses the file extension to choose the appropriate internal logic, but it's not clear whether this can be overridden or this is establishing a standard.

In addition, it's not clear how the output files should be named. For example, we've been using `$shader.frag.spv` but this isn't ideal in cases where you are trying to determine the file type based purely on the file extension. It would be good to have a discussion here and then put forward a standard naming convention for shader files, in particular both glsl related files, and the spv counterparts.

The reason for this request is two fold:

1/ Set a baseline for engine and 3D software developers so that people can work consistently (e.g. resource loading systems).
2/ Make it clear what extensions should be recognised by IDEs, tools, e.g. for syntax highlighting, file icons, etc. e.g. https://github.com/hughsk/language-glsl/blob/5edd8bd17af5913f977bd7efe8c815786e6a4adf/grammars/glsl.cson#L1-L18

Thanks for your time.

ioquatix:
I guess I'll add one more point. The file extension usually indicates the contents of the file, e.g. it's format, not the function. Ideally, I think it would make the most sense to use `.glsl` and compile that to `.spv` (e.g. how `.c` compiles to `.o`). But, that's counter the existing recommendations so I guess having a discussion here would be good.

Ideally, one could put multiple functions into the `.glsl` file, e.g. both the vertex and fragment functions, etc.

johnkslang:
GLSL internally has no declaration of what stage a shader is meant for, so the information is external.  Hence, a .glsl suffix doesn't help much for a single shader staying with GLSL syntax.

There is also no syntax for bundling multiple shaders into a single file. This could be layered on top of GLSL, in a tool like glslang or a wrapper around  it, as GLSL proper has no concept of a file either.

It does establish a convention though for stage-specific file-name suffixes, which would be fine to follow, though there is standard to force that.

What does that leave?  Just a command-line option request to explicitly set the stage and ignore the file suffix?

ioquatix:
> What does that leave? Just a command-line option request to explicitly set the stage and ignore the file suffix?

I see both pros and cons in this.

I'm also suggesting the the output file name be "suggested" or "recommended", e.g. `foo.frag` should canonically map to `foo.frag.spv` or something along those lines.

[...]

PR #450, opened 2016-08-10
Allowing explicit specification of shader compiltion type via -T opti…

dankbaker:
…on, rather then looking at file extension. For HLSL files, this is nice because .hlsl extension is natively udnerstood by visual studio, likely to be used with the -e option.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -155,6 +155,7 @@ int Options = 0;
 const char* ExecutableName = nullptr;
 const char* binaryFileName = nullptr;
 const char* entryPointName = nullptr;
+const char* shaderTargetName = nullptr;
 
 //
 // Create the default name for saving a binary if -o is not provided.
@@ -236,6 +237,15 @@ void ProcessArguments(int argc, char* argv[])
                 Options |= EOptionVulkanRules;
                 Options |= EOptionLinkProgram;
                 break;
+            case 'T':
+                shaderTargetName = argv[1];
+                if (argc > 0) {
+                    argc--;
+                    argv++;
+                }
+                else
+                    Error("no <target> specified for -T");
+                break;
             case 'G':
                 Options |= EOptionSpv;
                 Options |= EOptionLinkProgram;
@@ -260,6 +270,7 @@ void ProcessArguments(int argc, char* argv[])
             case 'e':
                 // HLSL todo: entry point handle needs much more sophistication.
                 // This is okay for one compilation unit with one entry point.
+                // dankbaker - not sure it needs to be, fxc has no more sophistication then this
                 entryPointName = argv[1];
                 if (argc > 0) {
                     argc--;
@@ -685,6 +696,9 @@ EShLanguage FindLanguage(const std::string& name)
     }
 
     std::string suffix = name.substr(ext + 1, std::string::npos);
+    if (shaderTargetName)
+        suffix = shaderTargetName;
+
     if (suffix == "vert")
         return EShLangVertex;
     else if (suffix == "tesc")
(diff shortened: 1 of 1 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -778,6 +792,8 @@ void usage()
            "  -H          print human readable form of SPIR-V; turns on -V\n"
            "  -E          print pre-processed GLSL; cannot be used with -l;\n"
            "              errors will appear on stderr.\n"
+           "  -T <target> uses explicit target specified, rather then the file extension.\n"
```
johnkslang: Can this be called "stage" instead of target?  Historically, in GLSL and initially in glslang, this was called "language" (there were pack/unpack and other now dead languages being considered).

Now, this is all migrating so that ideally language means real difference in language and stage is the term for vertex/fragment/compute/etc.

StandAlone/StandAlone.cpp
```diff
@@ -236,6 +237,15 @@ void ProcessArguments(int argc, char* argv[])
                 Options |= EOptionVulkanRules;
                 Options |= EOptionLinkProgram;
                 break;
+            case 'T':
+                shaderTargetName = argv[1];
+                if (argc > 0) {
+                    argc--;
+                    argv++;
+                }
```
johnkslang: Same line for `} else`.

johnkslang:
An HLSL *target* is a combination of a stage and a shader model or even whether an FX form is being used.

I think this change is just focused on the *stage* and not the full target.  Either way, if the information is going to turn into the stage (or in older terminology, the language), it seems better to call it the stage.

johnkslang:
See also issue #329.

Just verifying that this generalizes to handle lots of files simultaneously.  Glslang can take lots of files across multiple stages and link them all together, so we need a way of specifying this per input file to be complete, or else restrict it to single files input.

dankbaker:
Yeah, calling it stage is probably best. I will rename the wording.

johnkslang:
> calling it stage is probably best. I will rename the wording

Okay, thanks.  You can remove the "WIP" from the title when it's ready to go.

dankbaker:
Ok, changed the wording from target to stage

StandAlone/StandAlone.cpp
```diff
@@ -778,6 +792,8 @@ void usage()
            "  -H          print human readable form of SPIR-V; turns on -V\n"
            "  -E          print pre-processed GLSL; cannot be used with -l;\n"
            "              errors will appear on stderr.\n"
+           "  -S <stage>  uses explicit stage specified, rather then the file extension.\n"
+           "              valid choices are vert,tesc, tese,geom, rag,comp\n"
```
johnkslang: Should be as follows?

```
valid choices are vert, tesc, tese, geom, frag, or comp
```

StandAlone/StandAlone.cpp
```diff
@@ -260,6 +270,7 @@ void ProcessArguments(int argc, char* argv[])
             case 'e':
                 // HLSL todo: entry point handle needs much more sophistication.
                 // This is okay for one compilation unit with one entry point.
+                // dankbaker - not sure it needs to be, fxc has no more sophistication then this
```
johnkslang: This is about the thing of taking multiple files (compilation units) on the command line.  You agreed with the issue in the conversation.

Nothing has to be fixed now, but it seems to me possible that these flags apply to the files listed until another one overrides.  As the files get saved away as work to do, they can snapshot their stage and entrypoint.

StandAlone/StandAlone.cpp
```diff
@@ -236,6 +237,15 @@ void ProcessArguments(int argc, char* argv[])
                 Options |= EOptionVulkanRules;
                 Options |= EOptionLinkProgram;
                 break;
+            case 'T':
```
johnkslang: Sorry I missed this earlier.  Should the 'T' match the '-S' in the usage statement?

Note things like this will be more clear if there is a test change associated with the submission.

I'd suggest a simple addition to the runtests script, as that would be much simpler than updating the google tests.  Something like (untested, just suggesting):

```
echo Running stage selection...
$EXE -S something.hlsl > $TARGETDIR/something.hlsl.out
diff -b $BASEDIR/something.hlsl.out $TARGETDIR/something.hlsl.out || HASERROR=1
```

johnkslang: (You can use the Test/bump script to update tests modified by the Test/runtests script.)

Issue #435, opened 2016-08-05
SPV remapper: does not handle wide literals, e.g. OpSwitch on 64-bit type


dneto0:
The remapper doesn't have the machinery yet to handle literals wider than 1 word.  This becomes a problem when trying to track Ids used by an OpSwitch over a 64-bit value.

Example:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
       %long = OpTypeInt 64 1
          %3 = OpTypeFunction %void
          %4 = OpConstant %long 0
          %5 = OpFunction %void None %3
          %6 = OpLabel
               OpSelectionMerge %7 None
               OpSwitch %4 %7 0 %8 1 %9
          %8 = OpLabel
               OpBranch %7
          %9 = OpLabel
               OpBranch %7
          %7 = OpLabel
               OpReturn
               OpFunctionEnd
```

I get this internal error:
spirv-remap: SPVRemapper.cpp:219: spv::Id spv::spirvbin_t::localId(spv::Id, spv::Id): Assertion `id != spv::NoResult && newId != spv::NoResult' failed.

I suggest filing this as an enhancement, except that it results in an internal error. ?

steve-lunarg:
I can investigate this. 

@dneto0: do you have a small GLSL shader I can use to duplicate it?  I tried making this one similar to your SPIR-V, but receive an error that switch conditions must be a scalar integers.  If I cast 'x' to a plain old 32 bit int, such as switch(int(x)), then it doesn't illustrate the problem any more.
```

#version 450
#extension GL_ARB_gpu_shader_int64 : enable

void main()
{
    int64_t x = 0;

    switch (x) { case 0: break; }
}

```

dneto0:
@steve-lunarg
I don't think it can be done in GLSL.
GLSL 4.50 section 6.3 says "The type of the init-expression value in a switch statement must be a scalar int or uint. "    And GL_ARB_gpu_shader_int64 doesn't extend the switch statement semantics.

You should be able to use the SPIRV-Tools assembler to assemble the above program.  I've tried attaching it, but GitHub doesn't understand .spv files...

dneto0:
FYI.  The SPIRV-Tools handle bit widths up to 64.  The analog of what you'll need can be seen in its binary parser, around setting and using the entries in the type_id_to_number_type_info map.
E.g. see https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/binary.cpp#L207

[...]

PR #456, opened 2016-08-12
SPIRV: Remapper: allow 64 bit literals in OperandVariableLiteralId iteration

steve-lunarg:
This PR addresses issue #435, where iteration over the OperandVariableLiteralId argument list class would not cope with multiple word literals.

It also includes a small improvement to dead type elimination which picks up some cases previously dropped.  The algorithm used for this is suboptimal, but it's an offline tool, it was easy to write that way, and it can be further improved later if there is ever a performance concern.

### DIFF
--- a/SPIRV/SPVRemapper.cpp
+++ b/SPIRV/SPVRemapper.cpp
@@ -127,6 +127,37 @@ namespace spv {
         }
     }
 
+    // Return the size of a type in 32-bit words.  This currently only
+    // handles ints and floats, and is only invoked by queries which must be
+    // integer types.  If ever needed, it can be generalized.
+    unsigned spirvbin_t::typeSizeInWords(spv::Id id) const
+    {
+        const unsigned typeStart = idPos(id);
+        const spv::Op  opCode    = asOpCode(typeStart);
+
+        switch (opCode) {
+        case spv::OpTypeInt:   // fall through...
+        case spv::OpTypeFloat: return (spv[typeStart+2]+31)/32;
+        default:
+            error("unimplemented type size request");
+            return 0;
+        }
+    }
+
+    // Looks up the type of a given const or variable ID, and
+    // returns its size in 32-bit words.
+    unsigned spirvbin_t::idTypeSizeInWords(spv::Id id) const
+    {
+        const unsigned idStart = idPos(id);
+        const spv::Op  opCode  = asOpCode(idStart);
+
+        if (spv::InstructionDesc[opCode].hasType())
+            return typeSizeInWords(asId(idStart+1));
+
+        error("asked for type of typeless ID");
+        return 0;
+    }
+
     // Is this an opcode we should remove when using --strip?
     bool spirvbin_t::isStripOp(spv::Op opCode) const
     {
@@ -140,6 +171,7 @@ namespace spv {
         }
     }
 
+    // Return true if this opcode is flow control
     bool spirvbin_t::isFlowCtrl(spv::Op opCode) const
     {
         switch (opCode) {
@@ -155,6 +187,7 @@ namespace spv {
         }
     }
 
+    // Return true if this opcode defines a type
     bool spirvbin_t::isTypeOp(spv::Op opCode) const
     {
         switch (opCode) {
@@ -182,6 +215,7 @@ namespace spv {
         }
     }
 
+    // Return true if this opcode defines a constant
     bool spirvbin_t::isConstOp(spv::Op opCode) const
     {
         switch (opCode) {
@@ -324,7 +358,7 @@ namespace spv {
         fnPosDCE.clear();
         fnCalls.clear();
         typeConstPos.clear();
-        typeConstPosR.clear();
+        idPosR.clear();
         entryPoint = spv::NoResult;
         largestNewId = 0;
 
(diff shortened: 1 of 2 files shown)


SPIRV/SPVRemapper.cpp
```diff
@@ -459,13 +498,18 @@ namespace spv {
...
                 return nextInst;
 
-            case spv::OperandVariableLiteralId:
-                while (numOperands > 0) {
-                    ++word;             // immediate
-                    idFn(asId(word++)); // ID
-                    numOperands -= 2;
+            case spv::OperandVariableLiteralId: {
+                // word-2 is the position of the selector ID.
+                const unsigned literalSize = idTypeSizeInWords(asId(word-2));
```
dneto0: The "word-2" expression is strongly tied to the fact that this must be an OpSwitch instruction.  It just so happens that OpSwitch is the only case so far where we have an OperandVariableLiteralId.
 I'd probably assert that opcode == OpSwitch.

steve-lunarg: Good point.  Will change.

SPIRV/SPVRemapper.cpp
```diff
@@ -127,6 +127,36 @@ namespace spv {
...
     }
 
+    // Return the size of a type in 32-bit words.
+    unsigned spirvbin_t::typeSizeInWords(spv::Id id) const
+    {
+        const unsigned typeStart = idPos(id);
+        const spv::Op  opCode    = asOpCode(typeStart);
+
+        switch (opCode) {
+        case spv::OpTypeVoid:  return 0;
```
dneto0: It's weird to handle Void but not other types like Bool.
It turns out that this function is only called for scalar integer types, since it's only called in the OpSwitch case.   I'd suggest removing the void case (since it misleads) and stating the input constraints in the comment.

dneto0:
Works for me.

The only other place in the grammar where Ids and literals are mixed is in OpGroupMemberDecorate, but there the literals are the 32-bit index of the structure member.  So this problem won't occur in that context.

This patch works on the submitted case, and also in the following, were I made OpConstant for 64-bit integer and 64-bit float:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
       %long = OpTypeInt 64 1
     %double = OpTypeFloat 64
          %4 = OpConstant %double 3.14
          %5 = OpConstant %long 5000000000
          %6 = OpTypeFunction %void
          %7 = OpConstant %long 0
          %8 = OpFunction %void None %6
          %9 = OpLabel
               OpSelectionMerge %10 None
               OpSwitch %7 %10 0 %11 1 %12
         %11 = OpLabel
               OpBranch %10
         %12 = OpLabel
               OpBranch %10
         %10 = OpLabel
               OpReturn
               OpFunctionEnd
```

SPIRV/SPVRemapper.cpp
steve-lunarg: Agreed.  Will change.

steve-lunarg:
Cool; thanks for the review.  I've pushed a new version with your review changes.

dneto0:
Ooops.  This is causing a failure in the Vulkan CTS.
I get an abnormal exit in `vk-build-programs --deqp-case=dEQP-VK.glsl.constant_expressions.other.switch_case_fragment`

dneto0:
This is the failing test case, in GLSL form.
When I compile this with -V and then run it through the remapper, I get an internal error message:
"ID not found". 

```
#version 310 es
//;dEQP-VK.glsl.constant_expressions.other.switch_case_fragment
precision highp float;
bool isOk (int a, int b)     { return (a == b); }
layout(location = 0) out mediump vec4 dEQP_FragColor;
layout(location = 0) flat in float in0;
layout(binding = 0, std140) uniform Reference
{
        int out0;
} ref;
int out0;


void main()
{
        const int _0 = 0;
        const int _1 = 1;
        const int _2 = 2;
        const int _3 = 3;
        const int _4 = 4;

        switch(int(in0))
        {
                case _0:
                        out0 = 0;
                        break;
                case _1:
                        out0 = 1;
                        break;
                case _2:
                        out0 = 2;
                        break;
                case _3:
                        out0 = 3;
                        break;
                case _4:
                        out0 = 4;
                        break;
                case 5:
                        out0 = 10;
                        break;
                default:
                        out0 = 100;

        }
        bool RES = isOk(out0, ref.out0);
dEQP_FragColor = vec4(RES, RES, RES, 1.0);

}
```

dneto0:
I've narrowed it down to the `--map types` option

dneto0:
If I use `--map names` then I end up with "unimplemented type size request".  So there's something deeply bad going on.

dneto0:
With --dce-types I get "ID not found" as well.
I've requested that this PR be backed out.

johnkslang:
It's out.

Definitely need a better testing strategy.

johnkslang:
Or, something like this could be experimented with in a branch first, like the Vulkan precision qualifier change.

steve-lunarg:
Thanks for the reproduction case; I'll check it out.

steve-lunarg:
I see the problem; it relates to how I was obtaining the type size from the instruction stream during the remapping pass, which is actively changing all the IDs.

I'll fix.  Unless you have an urgent need, I'd like to do it next week rather than when tired at the end of this one.

steve-lunarg:
Also, I'll focus soon on creating a better testing infrastructure, since my test dataset didn't expose the problem. That might take a bit longer since it seems best coupled with other infrastructure changes if it's to use SPIRV-Tools for testing.

PR #459, opened 2016-08-12
WIP: HLSL: Buffer intrinsic fixes

dankbaker:
Allows multidimensional vector loads to parse and work, as well as uint inputs to be used for the .Load intrinsic function.

Note that though it parses multi-dimensional loads, e.g. Buffer<float3> FooBuffer;, It isn't using this information to validate. It is essentially treating it as a float4.  This is likely fine for SPIR-V generation, but it means the compiler won't catch the following cases:

1) Uses 4 components even though less have been declared will still compile. 

2) In a binary expression, the compiler may decide that the buffer load is a 4 vector, which could cause some casting rules to operate incorrectly. The casting code may need to look at this dimensionality. 

However, it will now compile cases that failed before. Since it appears that SPIR-V doesn't have 2 and 3d Loads, I think this will mostly be front end validation.

Inspection of the tests appears to indicate that this works and allows a shader which did not compile to compile, and in another case the error message changes slightly (which is a little puzzling, and makes me think there might be a small a bug somewhere else that this change has uncovered).


johnkslang:
You might need to squash or do something to get this down to a relevant 1 or 2 commits (looks like it should be 1 commit).

@steve-lunarg do you have input on this?  It might be overlapping some things you are doing.

glslang/Include/Types.h
```diff
@@ -70,6 +70,8 @@ enum TSamplerDim {
 
 struct TSampler {   // misnomer now; includes images, textures without sampler, and textures with sampler
     TBasicType type : 8;  // type returned by sampler
+    int vectorSize  : 4;  //dimensionality of return type
```
johnkslang: Expanding the type is inconsistent with previous plans to treat this as syntactic sugar.

E.g., I know for GLSL interpolateAt(), we did this transform:

```
in vec4 inVar;
interpolateAt(inVar.wxy)   -->  interpolateAt(inVar).wxy
```

This is because of how SPIR-V works, and I recall hearing similar plans for HLSL texturing.

johnkslang:
Also, will want to show test result changes in the PR. From the Test directory:

```
glslangtests --update-mode
```

or

```
runtests
bump
```

For adding new HLSL tests, see the list in gtests/Hlsl.FromFile.cpp.

dankbaker:
What does squash do? I'm still new to Git? does that collapse these into one commit?

I did run the test update, I guess I forgot to push those up onto the branch.

glslang/Include/Types.h
dankbaker: This certainly isn't an ideal change. I would like to create a reference called TemplateType inside TSampler, but this causes a circular reference. Thus, it would have to be a pointer.  If it's a pointer, that would work except then the code needs to be pulled out of the header file and places somewhere. Also, it's a pointer I needed to understand more about destruction of the object.

Thus, this was starting to be a fairly big structural change so I opted not to do it and do something simpler for the time being. Even in HLSL, it can't be an arbitrary type. If it's a structure, then it has to be < 4 components of the same type, which basically means the template type is always a vector with just name aliasing. However, I feel like the compiler should be more general then that.

johnkslang:
> What does squash do?

Changes multiple commits into a single commit.  Useful if there was back and forth or successive refinement that is irrelevant to the history.  If a PR really did two things, or is easier to see it two steps (like 1. add test cases, and 2. see the effect on them), then the history of multiple commits in a single PR can be useful.

dankbaker:
How do I perform a squash?  Also, for some strange reason it isn't showing the test changes on the files change list unless you look at that specific test update commit. Maybe squashing will fix this problem? It sorts of looks like a bug in GitHub to me.

ratchetfreak:
@dankbaker  http://stackoverflow.com/q/5189560/731620

steve-lunarg:
Cool - definitely something that needs to get done sooner or later.  A few thoughts on the details:

- The expansion of Load() to handle both int and uint in the prototypes might interact with something @johnkslang is working on around implicit conversions and function signature selection.  I'm not 100% certain how they interact.  The [HLSL documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/bb509694(v=vs.85).aspx) gives only int versions of Load(), leading me to think this should be handled via the proto selection algorithm.   At least, I've been mostly trying to avoid type explosions until that code in place.  Imagining for a moment that that algorithm was in place, would it then be necessary to have both of these listed in the proto table?

- When I was thinking about doing this I was going to try using the vector size to automatically add a swizzle to get the return types correct, so for example this would work out without any truncations or other slight bits of oddness:

```
    Buffer <float3> g_mybuffer;
    ...
    float3 mydata = g_mybuffer.Load(whatever);
```

I haven't dug into the details of what that'd take yet.  If it's not too much work, maybe it could be added into this PR as well?  If it's not straightforward, I imagine we could live without it, though it should get tracked as a TODO somewhere.

dankbaker:
I made a new change that adds type data to the sampler, which should give enough info to deal with these situations. I think I will close out this PR and active a new one. The only caveat was that it looks like in some cases we want a semi-anonymous sampler type for the intrinsic tables (the case I was running into was GetDimensions wanted an exact match, but it's not necessary)

johnkslang:
Closed, as another PR says it replaces this one.

PR #461, opened 2016-08-14
HLSL: Better error message for when HLSL translation fails

dankbaker:
This will get a better error message for when things fail in the HLSL conversion.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -412,7 +412,26 @@ void StderrIfNonEmpty(const char* str)
 struct ShaderCompUnit {
     EShLanguage stage;
     std::string fileName;
-    char** text;           // memory owned/managed externally
+    char** text;             // memory owned/managed externally
+    const char*  fileNameList[1];
+
+    //Need to have a special constructors to adjust the fileNameList, since back end needs a list of ptrs
+    ShaderCompUnit(EShLanguage istage, std::string &ifileName, char** itext)
+    {
+        stage = istage;
+        fileName = ifileName;
+        text    = itext;
+		fileNameList[0] = fileName.c_str();
+    }
+
+    ShaderCompUnit(const ShaderCompUnit &rhs)
+    {
+        stage = rhs.stage;
+        fileName = rhs.fileName;
+        text = rhs.text;
+		fileNameList[0] = fileName.c_str();
+    }
+
 };
 
 //
@@ -439,7 +458,7 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
     for (auto it = compUnits.cbegin(); it != compUnits.cend(); ++it) {
         const auto &compUnit = *it;
         glslang::TShader* shader = new glslang::TShader(compUnit.stage);
-        shader->setStrings(compUnit.text, 1);
+        shader->setStringsWithLengthsAndNames(compUnit.text, NULL, compUnit.fileNameList, 1);
         if (entryPointName) // HLSL todo: this needs to be tracked per compUnits
             shader->setEntryPoint(entryPointName);
         shaders.push_back(shader);
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -116,9 +116,12 @@ bool HlslParseContext::parseShaderStrings(TPpContext& ppContext, TInputScanner& 
 
     HlslScanContext scanContext(*this, ppContext);
     HlslGrammar grammar(scanContext, *this);
-    if (! grammar.parse())
-        printf("HLSL translation failed.\n");
-
+    if (!grammar.parse())
+    {
+        //Print out a nicer error message that should be formated such that if you click on the message it will take you right to the line through most UIs
+        const glslang::TSourceLoc& sourceLoc = input.getSourceLoc();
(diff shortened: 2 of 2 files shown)


johnkslang:
This seems to address multiple different subjects, including the one addressed by PR #466, which will conflict because #466 followed the coding style (no extra space around the cast).

dankbaker:
Squashing the commit and getting rid of the cast since someone else also noticed it.

StandAlone/StandAlone.cpp
```diff
@@ -412,7 +412,29 @@ void StderrIfNonEmpty(const char* str)
 struct ShaderCompUnit {
     EShLanguage stage;
     std::string fileName;
-    char** text;           // memory owned/managed externally
+    char** text;             // memory owned/managed externally
+	const char*  fileNamePointer;  // a pointer into file name, needed due to the way the compiler interface works
```
johnkslang: Spacing seems funky; mixing tabs and spaces?

StandAlone/StandAlone.cpp
```diff
@@ -412,7 +412,29 @@ void StderrIfNonEmpty(const char* str)
...
+
+	//Need to have a special constructors to adjust the filename pointer
+	//There is just no clean way to deal with this, the glsang back end is written as C++ style with pointers,
```
johnkslang: Note the back end should have no concept of actual files and file names, just a name to go along with the file (like __FILE__), which was a number in GLSL.

StandAlone/StandAlone.cpp
johnkslang: fileName.c_str() should be a pointer to the name.  Do we really need both a `std::string` and a `const char*` for the same data?  Maybe it can be consolidated one direction or the other?

johnkslang:
Nice idea, if it works for folks, that's fine.  Just need to get some details sorted out (in-line comments).

StandAlone/StandAlone.cpp
dankbaker: Could rewrite the standalone to be more C++ style, but it gets awkard mixing std:strings and char*s. This was the least intrusive change I could think of

johnkslang: I rechecked the code involved on the other side of the interface.  This in not a strange mismatch, it's just that the core interface takes an array of names, because it can handle multiple names. Standalone doing the mapping from "single file name" to "array of size 1 of names of the strings being compiled".

So, the basic idea in the code seems good, but the new member should be an array of size 1:

```
int numStrings = 1;
char* stringNames[1];
stringNames[0] = fileName.c_str();
getStringNames() { return stringNames; }
getNumStrings() { return numStrings; }
```

I didn't compile it, it's just that something like this makes sense, but the point is it's all consistent with the existing interfaces, and the comments (if still needed) just say they translate from single file name to set of string names.

(Key possibly confusing thing here is "string" is not talking about how the file name is stored, it's talking about glslang taking a set of strings to compile, and each string gets a name, that's a string name.)

dankbaker: Sure, could change it to an array of size 1 just for code clarity as to why there is seeming pointer to something you have a pointer to, which at first glance definitely seems silly. i'll change it

dankbaker:
Ok, I modified the code style to be more clear about what's going on and tried to conform with code conventions present.

StandAlone/StandAlone.cpp
```diff
@@ -412,7 +412,26 @@ void StderrIfNonEmpty(const char* str)
...
+    char** text;             // memory owned/managed externally
+    const char*  fileNameList[1];
+
+    //Need to have a special constructors to adjust the fileNameList, since back end needs a list of ptrs
+    ShaderCompUnit(EShLanguage istage, std::string &ifileName, char** itext)
+    {
+        stage = istage;
+        fileName = ifileName;
+        text    = itext;
+		fileNameList[0] = fileName.c_str();
```
johnkslang: Use spaces for all horizontal white space.

(This is still mixing tabs and spaces, which causes problems like seen here with indentation appearing inconsistent.)

StandAlone/StandAlone.cpp
```diff
@@ -412,7 +412,26 @@ void StderrIfNonEmpty(const char* str)
...
+    }
+
+    ShaderCompUnit(const ShaderCompUnit &rhs)
+    {
+        stage = rhs.stage;
+        fileName = rhs.fileName;
+        text = rhs.text;
+		fileNameList[0] = fileName.c_str();
```
johnkslang: Fix tabs to be spaces.

hlsl/hlslParseHelper.cpp
```diff
@@ -116,9 +116,12 @@ bool HlslParseContext::parseShaderStrings(TPpContext& ppContext, TInputScanner&
 
     HlslScanContext scanContext(*this, ppContext);
     HlslGrammar grammar(scanContext, *this);
-    if (! grammar.parse())
-        printf("HLSL translation failed.\n");
-
+    if (!grammar.parse())
+    {
+        //Print out a nicer error message that should be formated such that if you click on the message it will take you right to the line through most UIs
```
johnkslang: Nit: There is no history for the comments to apply to (nicer than ?).  It's a good commit message.  The comment should just say what the code that is present actually does.

johnkslang:
This looks great, and I'd accept it earlier if it didn't mix up tabs and spaces.  I'll fix now myself.

PR #465, opened 2016-08-15
WIP: HLSL Buffer intrinsic fixes take 2, replaces #459

dankbaker:
Here, I've rearranged the sampler object to take a full type object. this allows generalization for the return type.  This is meant to replace PR #459 

It passed all the tests and seems to work well. The 2 open questions:

1) If the dimensionality of the sampler type is actually checked to be equal, then some of the intrinsic matches fail, specifically GetDimension (which doesn't care at all about the return type of the texture object)

2) I'm not 100% sure about the news are getting cleaned up. It looked to me like there was a special allocator for news that was context specific and got garbage collected - as I did not see deletes in the destructor for the type object

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -68,8 +68,12 @@ enum TSamplerDim {
     EsdNumDims
 };
 
+class TType;
 struct TSampler {   // misnomer now; includes images, textures without sampler, and textures with sampler
-    TBasicType type : 8;  // type returned by sampler
+
+	//The return type of the object
+	TType *samplerType;
+
     TSamplerDim dim : 8;
     bool    arrayed : 1;
     bool     shadow : 1;
@@ -78,6 +82,7 @@ struct TSampler {   // misnomer now; includes images, textures without sampler, 
     bool   combined : 1;  // true means texture is combined with a sampler, false means texture with no sampler
     bool    sampler : 1;  // true means a pure sampler, other fields should be clear()
     bool   external : 1;  // GL_OES_EGL_image_external
+    
 
     bool isImage()       const { return image && dim != EsdSubpass; }
     bool isSubpass()     const { return dim == EsdSubpass; }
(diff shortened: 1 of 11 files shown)


glslang/Include/Types.h
```diff
@@ -68,8 +68,12 @@ enum TSamplerDim {
     EsdNumDims
 };
 
+class TType;
 struct TSampler {   // misnomer now; includes images, textures without sampler, and textures with sampler
-    TBasicType type : 8;  // type returned by sampler
+
+	//The return type of the object
+	TType *samplerType;
```
johnkslang: Note this takes 8 times more memory, and appears as part of everything, so I'm somewhat interested in being more conservative in what is actually needed here.

johnkslang:
@steve-lunarg do you have design input here?  It may overlap or be different than the design briefly discussed elsewhere ( changing  `op(v.wx)` -> `op(v).wx`).

Maybe I missed the resolution there, but good to get @steve-lunarg and @dankbaker on the same page regarding this one.

dankbaker:
I couldn't think of another way that didn't eat up as much memory. In order to do general dereferencing of say templated objects, you'd need a list of structure members which ends up taking up quite a bit of space. The alternative  would be to implement a subset of the type structure: but it wouldn't really help matters except add even more code that needed to be maintained.

glslang/Include/Types.h
johnkslang: This is just about whether the sampler is floating point or integer, right?  Possibly (depending on design decision with @steve-lunarg) whether there is a number of components.

It seems this PR is suggesting a sampling operation can return, say, an array of structures containing a structure containing an array of doubles, etc.  Is that functionality we need?

Maybe we need an issue to first understand the goal and design for this?

johnkslang:
@dankbaker

> I couldn't think of another way that didn't eat up as much memory. In order to do general dereferencing of say templated objects, you'd need a list of structure members which ends up taking up quite a bit of space. The alternative  would be to implement a subset of the type structure: but it wouldn't really help matters except add even more code that needed to be maintained.

If that's in reply to the inline comment, see my update to the inline comment.  (If not, I need more context.)

glslang/Include/Types.h
dankbaker: In HLSL you can actually specify structs as the return type. I need to expirement with what is valid where, as IIRC it is restricted to same base type.  However,  I think it will be needed for some structured UAV writes, e.g. exactly the situation you are talking about.

Regardless, currently we _can't_ handle even simple non vector 4 types which is pretty blocking so some change needs to be made.

johnkslang: I think this is conflating how the stack will handle what GLSL calls a buffer block with the texturing subsystem.  The readable/writable buffer blocks are the fully typed objects, and glslang has full support for them already.  

Maybe we need some higher-level design figured out here.  There is no way in SPIR-V either to represent texturing operations returning arbitrary user-defined types, so adding them to the AST, doesn't get us very far, it just pushed the hard part somewhere else.

dankbaker: There may need to be some changes then to support structured buffers in HLSL.  Oxide is currently not using structured buffers, but it is an important feature of HLSL. Currently, buffers are handled as a type of sampler object

An example from the docs:

```
struct MyStruct
{
    float4 Color;
    float4 Normal;
    bool isAwesome;
};

StructuredBuffer<MyStruct> mySB;
```

However, HLSL syntax also allows structures in sampler type so long as they <= 4 components and are the same basic type.  Right now, Oxide's just trying to unblock float2, float3 etc return types which are quite common in HLSL. But Oxide's not the power user of this functionality.

If the intention is to keep buffers as types of samplers (which is how HLSL thinks of them), then I see no way around having this change. Otherwise, the choice is to not make buffers part of the sampler but have them as a specific object type in the base type. But, you'll really just end up moving where the pointer lives and I don't think it really changes anything practically.

Part of the issue that HLSL had a more abstract AST then GLSLlang does. It was essentially a LISP structure so this enabled arbitrary types on objects pretty trivially. Validation on legality occurred after the AST was built, not while it was built.  There's no right or wrong to how to do it, but it just means that there is alot of HLSL syntax that is valid so long as the specific AST ended up mapping to something the shader model could do.

steve-lunarg:
Several thoughts:

- I'm wondering how often features like returning user structs or mats actually happen in real life shaders.  Perhaps we could get a very long ways indeed with just a few bits.  That wouldn't handle the ultimate arbitrary case, but wouldn't have much of a size impact, either.  It's the "wrong direction", but it's not very far in the wrong direction.

- I don't (yet) have a design in mind for handling totally arbitrary type returns.  Also, per info from @johnkslang above, perhaps there's nothing much to be done with them now anyway.  But a random off-the-cuff thought about doing so: if size is an issue here perhaps the types could be handled as some sort of COW reference rather than including it by value.  Presumably there are not many unique cases of weird types returned from samplers in any given shader.  A few bits could index into a table and avoid the high inclusion by value cost.

johnkslang:
It will be worth considering what the SPIR-V looks like for complex-return type.  It won't look like the PR, because that doesn't translate to SPIR-V.  So, designing from other direction might help: what does the SPIR-V look like, and what AST would make that?

Or, do you think SPIR-V can't do this?

Ideally, (I don't know enough yet) what you are doing within the sampler.type is really just a TType that has the right base type.

Anyway, agreed with keeping this practically driven for the moment.  The real decision to make right now seems to be which way do we go with sub-vec4-types:

- op(v.wx) -> op(v).wx

or

- adding the vector width to the type

Note that if you prefer the latter, it is not complete, as it makes a new form of AST, and since SPIR-V does not support v.wx, you still have to write the transform op(v.wx) -> op(v).wx in order to generate SPIR-V.  So, keep in mind making the whole stack work when selecting.

johnkslang:
This was superseded by recent PRs.

The idea of using a TType in the sampler type seems valid though, as it seems HLSL can layer a struct type on top of what an image is capable of returning.  Will cross that bridge when we have to.

PR #470, opened 2016-08-16
WIP: SPIRV remapper: allow 64 bit literals in OperandVariableLiteralI…

steve-lunarg:
Removed in favor of #471, which is the same code re-submitted to try to shake the Travis-CI build loose.

### DIFF
--- a/SPIRV/SPVRemapper.cpp
+++ b/SPIRV/SPVRemapper.cpp
@@ -127,6 +127,33 @@ namespace spv {
         }
     }
 
+    // Return the size of a type in 32-bit words.  This currently only
+    // handles ints and floats, and is only invoked by queries which must be
+    // integer types.  If ever needed, it can be generalized.
+    unsigned spirvbin_t::typeSizeInWords(spv::Id id) const
+    {
+        const unsigned typeStart = idPos(id);
+        const spv::Op  opCode    = asOpCode(typeStart);
+
+        switch (opCode) {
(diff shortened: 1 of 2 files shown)


steve-lunarg:
Hmm... the "Travis CI" build has been pending for 6 hours now.  I've seen something similar before, always with the Apple build, which seems to be the case again here.  In the details page, it looks like it hasn't even started trying yet:

> Hang tight, the log cannot be shown until the build has started.

johnkslang:
Looks like it just timed out.  This is labeled as WIP, is there something that is blocked on (getting input from folks)?

If it looks solid, we could try pulling it into master and see if the hang is random or correlated.

steve-lunarg:
Last time I saw this problem I created another branch from scratch with the same code, submitted that, and all was well.  I don't know that it was a "fix" rather than a coincidence (perhaps the build system becoming happy for some other reason), but still, I'm inclined to try for a successful Travis CI.  Unless there's some urgency, I'll submit it again tomorrow.  Maybe either another day or another git branch will make it happy.

Curiously, the build log on the details page for the failing Apple builds looks completely empty for me.  Last time this happened, it showed an internal git error of some sort.

johnkslang:
Will pull when you remove your "WIP".

dneto0:
This patch is compatible with the Vulkan CTS shaders, with SPIR-V validation turned on.  So works for me, but moreso than last time.  :-)

PR #473, opened 2016-08-18
SPV Remapper: add remapper test framework

steve-lunarg:
This checkin adds remapper testing as its own executable (not part of glslangtests), with an eye towards future separation from the glslang project.  Testing is done through the following path:

Arbitrary source language (GLSL, HLSL, SPIR-V)
-> conversion to SPIR-V
-> remapping SPIR-V with per-test remap options
-> disassembly to text
-> comparison to golden text

There are a set of new tests to exercise remapper functionality and successful traversal.  One of the new tests is a direct SPIR-V binary using 64 bit literals in switch statements, which cannot be produced via the HLSL or GLSL paths.

This also fixes two small issues in the current glslang SPIR-V disassembler related to null shaders with no IDs in them.

This currently has no dependency on SPIRV-Tools, though that has been proposed as a future direction.  This would also allow use of a validator tool.

Again with the Travis CI Mac build :-/.  Since that failure seems to be a common recent occurrence in diverse PRs, I'll assume this is OK since the other builds all passed.


steve-lunarg:
Updated to flesh out some of the remapper testing.  There are now GLSL, HLSL, and direct-from-SPV tests.

Still WIP.

steve-lunarg:
If there are no specific comments about the testing strategy, I'll make this a non-WIP branch in the next day or so.

Issue #142, opened 2016-01-23
[4.0] Implement version 400 function-overloading disambiguation algorithm


johnkslang:
Currently, only the pre-version 400 function-overloading algorithm is present.  Version 400 and above should be using the second-generation algorithm.

johnkslang:
FYI, this is specified in section 6.1 of GLSL as the following, and I know it is possible to implement this as a two-pass linear algorithm across the set of choices.

> When function calls are resolved, an exact type match for all the arguments is sought. If an exact match is found, all other functions are ignored, and the exact match is used. If no exact match is found, then the implicit conversions in section 4.1.10 “Implicit Conversions” will be applied to find a match. Mismatched types on input parameters (in or inout or default) must have a conversion from the calling
argument type to the formal parameter type. Mismatched types on output parameters (out or inout) must have a conversion from the formal parameter type to the calling argument type.
>
> If implicit conversions can be used to find more than one matching function, a single best-matching
function is sought. To determine a best match, the conversions between calling argument and formal
parameter types are compared for each function argument and pair of matching functions. After these
comparisons are performed, each pair of matching functions are compared. A function declaration A is
considered a better match than function declaration B if
>
> - for at least one function argument, the conversion for that argument in A is better than the
corresponding conversion in B; and
> - there is no function argument for which the conversion in B is better than the corresponding
conversion in A.
>
> If a single function declaration is considered a better match than every other matching function
declaration, it will be used. Otherwise, a compile-time semantic error for an ambiguous overloaded
function call occurs.
>
> To determine whether the conversion for a single argument in one match is better than that for another match, the following rules are applied, in order:
>
> 1. An exact match is better than a match involving any implicit conversion.
> 2. A match involving an implicit conversion from float to double is better than a match involving any other implicit conversion.
> 3. A match involving an implicit conversion from either int or uint to float is better than a match involving an implicit conversion from either int or uint to double.
>
> If none of the rules above apply to a particular pair of conversions, neither conversion is considered better than the other.

johnkslang:
Here is a high-level design for a generic selector, which I believe can be used by both GLSL and HLSL, and probably other more C++-like languages too.

## Generic Selector

This would live in place usable by multiple languages, e.g. `ParseContextBase`, or just on it's own, taking a parse context or symbol table as needed to operate.

### Assumptions

- There is no exact match, so a selection algorithm needs to run.  This is not inside, so a language-specific parser can first look up exact match and then decide what to do in its absence.

[...]

PR #479, opened 2016-08-24
HLSL: Adding cast support for unsigned int (uint) for Load method, also upd…

dankbaker:
…ating tests to test for it.

Problem: Load method wasn't accepting uint inputs, because no entry in the symbol tables.

Added entries in symbol tables and also updated tests to check that loads from UINT are supported.

### DIFF
--- a/Test/baseResults/hlsl.load.2dms.dx10.frag.out
+++ b/Test/baseResults/hlsl.load.2dms.dx10.frag.out
@@ -2,94 +2,160 @@ hlsl.load.2dms.dx10.frag
 Shader version: 450
 gl_FragCoord origin is upper left
 0:? Sequence
-0:57  Function Definition: main( (global structure{temp 4-component vector of float Color, temp float FragDepth Depth})
-0:28    Function Parameters: 
+0:76  Function Definition: main( (global structure{temp 4-component vector of float Color, temp float FragDepth Depth})
+0:34    Function Parameters: 
 0:?     Sequence
-0:32      textureFetch (global 4-component vector of float)
-0:32        'g_tTex2dmsf4' (uniform texture2DMS)
-0:32        'c2' (uniform 2-component vector of int)
-0:32        Constant:
-0:32          3 (const int)
-0:33      textureFetch (global 4-component vector of int)
-0:33        'g_tTex2dmsi4' (uniform itexture2DMS)
-0:33        'c2' (uniform 2-component vector of int)
-0:33        Constant:
-0:33          3 (const int)
-0:34      textureFetch (global 4-component vector of uint)
-0:34        'g_tTex2dmsu4' (uniform utexture2DMS)
-0:34        'c2' (uniform 2-component vector of int)
-0:34        Constant:
-0:34          3 (const int)
-0:37      textureFetchOffset (global 4-component vector of float)
-0:37        'g_tTex2dmsf4' (uniform texture2DMS)
-0:37        'c2' (uniform 2-component vector of int)
-0:37        Constant:
-0:37          3 (const int)
-0:37        'o2' (uniform 2-component vector of int)
-0:38      textureFetchOffset (global 4-component vector of int)
-0:38        'g_tTex2dmsi4' (uniform itexture2DMS)
+0:38      textureFetch (global 4-component vector of float)
+0:38        'g_tTex2dmsf4' (uniform texture2DMS)
 0:38        'c2' (uniform 2-component vector of int)
 0:38        Constant:
 0:38          3 (const int)
-0:38        'o2' (uniform 2-component vector of int)
-0:39      textureFetchOffset (global 4-component vector of uint)
-0:39        'g_tTex2dmsu4' (uniform utexture2DMS)
+0:39      textureFetch (global 4-component vector of int)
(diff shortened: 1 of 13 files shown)


johnkslang:
Question: Are uint this supposed to be accepted through implicit conversions or with additional intrinsics?

@steve-lunarg any input on this?

Thanks.

dankbaker:
The implicit conversion currently didn't accept it, though that would work just as well

johnkslang:
So, the generic selection algorithm for signature selection under implicit conversion went in a few days ago, and the specialization for HLSL is on its way (currently at GLSL #version 400 level, and just hours from being what we've been jointly designing in the other issues for HLSL).

In the meantime, master is at GLSL #version < 400 levels, which allow int -> uint, but not uint -> int.

Having said that, if the "right" thing is to have uint intrinsics, rather than int intrinsics under uint->int conversion, I understand the PR.  Such things can matter, as far as whether or not an ambiguous selection is found or not, so it is worthwhile knowing which way is correct.

steve-lunarg:
Documentation mentions only int, which at least for the offset parameter makes some intuitive sense (no reason to restrict offsets to non-negative values).  Without a reason to the contrary, the evidence I know of seems to point to implicit conversions.

dankbaker:
It depends what you want the SPIRV to be. If it's an implicit conversion, then you end up with a cast in the SPIRV, right? That seams a little strange to me because .Load methods are basically undefined AFAIK with negative numbers as their index, or it may clamp to 0.

HLSL itself would done an implicit conversion, however that's a bit misleading because it would also have not passed on any sort of conversion down the stream, so you can't quite compare the two.

johnkslang:
> It depends what you want the SPIRV to be. If it's an implicit conversion, then you end up with a cast in the SPIRV, right? 

A conversion from uint -> int is a bitcast, not an actual operation.

>  That seams a little strange to me because .Load methods are basically undefined AFAIK with negative numbers as their index, or it may clamp to 0.

The top bit is set whether or not a bitcast or native intrinsic was involved, so I don't follow this part.

>  it would also have not passed on any sort of conversion down the stream

I don't think there is any conversion to pass downstream, as the bitcast is just a type-changing device, it does not translate to any executable code.

johnkslang:
FYI, commit fcc0aa3b647f0943cc9d29e1426b968526e94001 was just pushed, and switches over from using the hardcoded GLSL #version 120 overloaded signature selection rules, to instead use the generic selector (as designed in issue #142).

This is currently parameterized with GLSL #version 400 rules, but is the base from which to make deltas to HLSL rules.  Adding uint->int is trivial, the smearing/truncating conversions slightly less so.  Will do these next.

steve-lunarg:
I thought out of bounds loads were generally guaranteed to produce zeros in HLSL, rather than being undefined.  This verbiage is slightly ambiguous, but my SWAG would be that out of bounds access in the negative direction would act similarly to the positive.

> Note  When one or more of the coordinates in Location exceed the u, v, or w mipmap level dimensions of the texture, Load returns zero in all components. _Direct3D guarantees to return zero for any resource that is accessed out of bounds._

Independent of the topic above, I'm not sure what sort of guarantees SPIR-V makes in this regard.  I'll try to go research that tomorrow, unless someone knows off the top of their head.

dankbaker:
It might produce 0s, I can't recall what happened if it was a clamp to bound or something else.

I don't have a strong opinion about it except that a uint intrinsic is probably slightly more in line with what HLSL would produce, because the implicit conversion would have created effectively a new Load method that took uint.  The argument against it is just going down the path of having tons of slightly different intrinsic definitions in the symbol table, which seems a big negative to me as that could get rather bloated.  If you're fine with the bit of extra SPIRV casting then it's fine to do it with implicit conversions.

However, I would like to keep the tests to make sure the conversions compile.

steve-lunarg:
> However, I would like to keep the tests to make sure the conversions compile.

Yep, agreed - it seems important to have test coverage.

Some of the existing tests have ifdefed out sections meant to be enabled when implicit conversion rules are online.  They certainly don't cover all the bases though.

johnkslang:
Okay, commit 90dd70f752645863524bb91040d5631236f0d017 now takes the next step, in allowing any basic type (bool, int, uint, float, double) to convert to any of the others.  For disambiguity, the following rules were used:

        // exact match is always better than mismatch
        // float -> double is better than any other float conversion
        // int -> uint is better than any other int conversion
        // -> float is better than -> double
        // -> int is better than -> bool
        // -> uint is better than -> int

This is a guess at HLSL, I'm totally open to having more concrete/correct rules.

I agree with @steve-lunarg about tests and @dankbaker about not losing the tests in this PR: Your choice:  You can either have me turn on (need to know where) and include (from this PR), or do your own submissions.

This change may be somewhat chaotic, as it's a pretty radical semantic change from #version 400 rules, but it must be a step down the right road, given that HLSL seems to support all directions of conversion.

johnkslang:
Closing this one.

Would make sense to see any of these replace it though:

- just the tests
- the tests and the new intrinsics, if it is incorrect to have this work through implicit conversion

I welcome either of these with tests passing, etc., so they could be pulled. Thanks.

PR #481, opened 2016-08-25
HLSL: add precise keyword

steve-lunarg:
Precise is a valid HLSL qualifier.  Most everything was there, but it was missing in the scanner's table of keywords.  Now it is successfully passed through to an AST qualifier noContraction.

New test added.

PR #483, opened 2016-08-25
HLSL: add precise keyword

steve-lunarg:
This is the same code as #481, which timed out on the Mac Travis CI build failing to start.

Precise is a valid HLSL qualifier. Most everything was there, but it was missing in the scanner's table of keywords. Now it is successfully passed through to an AST qualifier noContraction.

New test added.


johnkslang:
This is achieving the same success level as master, so will pull.  Has been happening a lot recently, to many PRs, independent of the change in master.

Issue #475, opened 2016-08-18
HLSL: output semantic handling for vertex shader


StarsX:
Hi,
I find currently there seems no valid SPIR-V code could be generated, which could be really run to replace the glsl shader. I tried to modify some codes on semantics:

	void HlslParseContext::handleSemantic(TType& type, TString& semantic)
	{
	    // TODO: need to know if it's an input or an output
	    // The following sketches what needs to be done, but can't be right
	   // without taking into account stage and input/output.
		std::transform(semantic.begin(), semantic.end(), semantic.begin(), ::toupper);
	
		if (semantic == "PSIZE")
			type.getQualifier().builtIn = EbvPointSize;
		else if (semantic == "SV_POSITION" && language != EShLangFragment)
			type.getQualifier().builtIn = EbvPosition;
		else if (semantic == "POSITION")
			type.getQualifier().builtIn = EbvPosition;
		else if (semantic == "FOG")
			type.getQualifier().builtIn = EbvFogFragCoord;
		else if (semantic == "DEPTH" || semantic == "SV_DEPTH")
			type.getQualifier().builtIn = EbvFragDepth;
		else if (semantic == "VFACE" || semantic == "SV_ISFRONTFACE")
			type.getQualifier().builtIn = EbvFace;
		else if (semantic == "VPOS" || semantic == "SV_POSITION")
			type.getQualifier().builtIn = EbvFragCoord;
		else if (semantic == "SV_CLIPDISTANCE")
			type.getQualifier().builtIn = EbvClipDistance;
		else if (semantic == "SV_CULLDISTANCE")
			type.getQualifier().builtIn = EbvCullDistance;
		else if (semantic == "SV_VERTEXID")
			type.getQualifier().builtIn = EbvVertexId;
		else if (semantic == "SV_VIEWPORTARRAYINDEX")
			type.getQualifier().builtIn = EbvViewportIndex;
		else
		{
			unsigned int location;
			int suc = sscanf(semantic.c_str(), "LOCATION%u", &location);
			if (suc)
			{
				type.getQualifier().layoutLocation = location;
				type.getQualifier().smooth = true;
			}
		}
	}

and

    if (inEntrypoint) {
        // parameters are actually shader-level inputs
		for (int i = 0; i < function.getParamCount(); i++)
		{
			if (function[i].type->getQualifier().isParamOutput())
				function[i].type->getQualifier().storage = EvqVaryingOut;
			else
[...]

Issue #304, opened 2016-05-23
Storing structs to SSBOs break validation


Themaister:
Referring to this issue:
https://github.com/KhronosGroup/SPIRV-Tools/issues/217

```
error: 196: OpStore Pointer <id> '10's type does not match Object <id> '11's type.
```

SPIR-V:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 25
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %4 "main"
               OpExecutionMode %4 LocalSize 1 1 1
               OpSource ESSL 310
               OpName %4 "main"
               OpName %8 "MyStruct"
               OpMemberName %8 0 "foo"
               OpName %10 "bar"
               OpName %11 "MyStruct"
               OpMemberName %11 0 "foo"
               OpName %12 "SSBO0"
               OpMemberName %12 0 "foo_std430"
               OpName %14 ""
               OpName %20 "SSBO1"
               OpMemberName %20 0 "foo_std430_out"
               OpName %22 ""
               OpMemberDecorate %11 0 Offset 0
               OpMemberDecorate %12 0 Offset 0
               OpDecorate %12 BufferBlock
               OpDecorate %14 DescriptorSet 0
               OpDecorate %14 Binding 0
               OpMemberDecorate %20 0 Offset 0
               OpDecorate %20 BufferBlock
               OpDecorate %22 DescriptorSet 0
               OpDecorate %22 Binding 1
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
          %7 = OpTypeVector %6 4
          %8 = OpTypeStruct %7
          %9 = OpTypePointer Function %8
         %11 = OpTypeStruct %7
         %12 = OpTypeStruct %11
         %13 = OpTypePointer Uniform %12
         %14 = OpVariable %13 Uniform
         %15 = OpTypeInt 32 1
         %16 = OpConstant %15 0
         %17 = OpTypePointer Uniform %11
         %20 = OpTypeStruct %11
         %21 = OpTypePointer Uniform %20
         %22 = OpVariable %21 Uniform
          %4 = OpFunction %2 None %3
          %5 = OpLabel
         %10 = OpVariable %9 Function
         %18 = OpAccessChain %17 %14 %16
[...]

PR #484, opened 2016-08-25
WIP: First stab at system value interpretation

dankbaker:
Addresses issues #475 and others. This is my first stab at interpreting HLSL SV values correctly. A few of them I'm not sure how to map to in Vulkan land, and I'm not 100% confident of the SV_TARGET stuff if it will end up in the right place based on the qualifier stuff. 

However, this change is probably far better then what's currently implemented (Which was pretty wrong).

A decision about parsing DX9 should be made at some point, right now I turned that off by default but open to discuss how that should be handled.

I have not updated the tests yet

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -756,18 +756,14 @@ TIntermAggregate* HlslParseContext::handleFunctionDefinition(const TSourceLoc& l
     if (inEntrypoint) {
         // parameters are actually shader-level inputs
         unsigned int inCount = 0, outCount = 0;
-        for (int i = 0; i < function.getParamCount(); i++)
-        {
+        for (int i = 0; i < function.getParamCount(); i++) {
             // Fragment programs do not use ordering for the output layout position, rather it is 
             // embedded in the semantic
-            if (function[i].type->getQualifier().isParamOutput() && language != EShLangFragment)
-            {
-                if (function[i].type->getQualifier().builtIn == EbvNone)
+            if (function[i].type->getQualifier().isParamOutput() ) {
+                if (function[i].type->getQualifier().builtIn == EbvNone && language != EShLangFragment)
                     function[i].type->getQualifier().layoutLocation = outCount++;
                 function[i].type->getQualifier().storage = EvqVaryingOut;
-            }
-            else
-            {
+            } else {
                 if (function[i].type->getQualifier().builtIn == EbvNone)
                     function[i].type->getQualifier().layoutLocation = inCount++;
                 function[i].type->getQualifier().storage = EvqVaryingIn;
@@ -2276,7 +2272,9 @@ void HlslParseContext::handleSemantic(TType& type, const TString& semantic)
     // TODO: need to know if it's an input or an output
     // The following sketches what needs to be done, but can't be right
     // without taking into account stage and input/output.
-   
+
+    TString semanticUpperCase = semantic;
+    std::transform(semanticUpperCase.begin(), semanticUpperCase.end(), semanticUpperCase.begin(), ::toupper);
     // in DX9, all outputs had to have a semantic associated with them, that was either consumed
     // by the system or was a specific register assignment
     // in DX10+, only semantics with the SV_ prefix have any meaning beyond decoration
@@ -2284,24 +2282,23 @@ void HlslParseContext::handleSemantic(TType& type, const TString& semantic)
     // Also, in DX10 if a SV value is present as the input of a stage, but isn't appropriate for that
     // stage, it would just be ignored as it is likely there as part of an output struct from one stage
     // to the next
-    bool bParseDX9 = false;
     
+    
+    bool bParseDX9 = false;
     if (bParseDX9)
     {
-        if (semantic == "PSIZE")
+        if (semanticUpperCase == "PSIZE")
             type.getQualifier().builtIn = EbvPointSize;
         else if (semantic == "FOG")
             type.getQualifier().builtIn = EbvFogFragCoord;
-        else if (semantic == "DEPTH" || semantic == "SV_Depth")
+        else if (semanticUpperCase == "DEPTH")
             type.getQualifier().builtIn = EbvFragDepth;
-        else if (semantic == "VFACE" || semantic == "SV_IsFrontFace")
+        else if (semanticUpperCase == "VFACE")
             type.getQualifier().builtIn = EbvFace;
-        else if (semantic == "VPOS" || semantic == "SV_Position")
+        else if (semanticUpperCase == "VPOS")
             type.getQualifier().builtIn = EbvFragCoord;
     }
     
-    TString semanticUpperCase = semantic;
-    std::transform(semanticUpperCase.begin(), semanticUpperCase.end(), semanticUpperCase.begin(), ::toupper);
     {
         //SV Position has a different meaning in vertex vs fragment
         if (semanticUpperCase == "SV_POSITION" && language != EShLangFragment)
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -755,8 +755,24 @@ TIntermAggregate* HlslParseContext::handleFunctionDefinition(const TSourceLoc& l
     inEntrypoint = (function.getName() == intermediate.getEntryPoint().c_str());
     if (inEntrypoint) {
         // parameters are actually shader-level inputs
+        unsigned int inCount = 0, outCount = 0;
         for (int i = 0; i < function.getParamCount(); i++)
-            function[i].type->getQualifier().storage = EvqVaryingIn;
+        {
+            // Fragment programs do not use ordering for the output layout position, rather it is 
+            // embedded in the semantic
+            if (function[i].type->getQualifier().isParamOutput() && language != EShLangFragment)
+            {
+                if (function[i].type->getQualifier().builtIn == EbvNone)
+                    function[i].type->getQualifier().layoutLocation = outCount++;
+                function[i].type->getQualifier().storage = EvqVaryingOut;
+            }
+            else
+            {
+                if (function[i].type->getQualifier().builtIn == EbvNone)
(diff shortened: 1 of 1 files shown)


hlsl/hlslParseHelper.cpp
```diff
@@ -755,8 +755,24 @@ TIntermAggregate* HlslParseContext::handleFunctionDefinition(const TSourceLoc& l
     inEntrypoint = (function.getName() == intermediate.getEntryPoint().c_str());
     if (inEntrypoint) {
         // parameters are actually shader-level inputs
+        unsigned int inCount = 0, outCount = 0;
         for (int i = 0; i < function.getParamCount(); i++)
-            function[i].type->getQualifier().storage = EvqVaryingIn;
+        {
```
johnkslang: Coding convention: curlies go on same line as if/for/do/while/else/etc.

hlsl/hlslParseHelper.cpp
```diff
@@ -755,8 +755,24 @@ TIntermAggregate* HlslParseContext::handleFunctionDefinition(const TSourceLoc& l
...
+            // Fragment programs do not use ordering for the output layout position, rather it is 
+            // embedded in the semantic
+            if (function[i].type->getQualifier().isParamOutput() && language != EShLangFragment)
+            {
+                if (function[i].type->getQualifier().builtIn == EbvNone)
+                    function[i].type->getQualifier().layoutLocation = outCount++;
+                function[i].type->getQualifier().storage = EvqVaryingOut;
+            }
+            else
```
johnkslang: coding convention: `} else {`

hlsl/hlslParseHelper.cpp
```diff
@@ -2260,27 +2276,108 @@ void HlslParseContext::handleSemantic(TType& type, const TString& semantic)
...
+            type.getQualifier().builtIn = EbvTessCoord;
+        else if (semanticUpperCase == "SV_DEPTH")
+            type.getQualifier().builtIn = EbvFragDepth;
+     
+         
+        //not supported or not trivial
+        //else if( semanticUpperCase == "SV_STENCILREF")
+        //    type.getQualifier().builtIn = gl_FragStencilRef
+        //else if( semanticUpperCase == "SV_DEPTHGREATEREQUAL")
+        //    type.getQualifier().builtIn = ??
```
johnkslang: Sounds like the TLayoutDepth stuff.  E.g., `intermediate.setDepth(EldGreater)`, but not 100% sure.  That translates to a mode for the whole stage, not really a per-variable thing.

Tangentially related, there is also the intermediate.setDepthReplacing() method.

hlsl/hlslParseHelper.cpp
```diff
@@ -2260,27 +2276,108 @@ void HlslParseContext::handleSemantic(TType& type, const TString& semantic)
...
+        //else if( semanticUpperCase == "SV_DEPTHLESSEQUAL")
+        //    type.getQualifier().builtIn = ??
```
johnkslang: ... `EldLess` ...

johnkslang:
Thanks @dankbaker, looks like a step in the right direction.  I added some line comments.

> A decision about parsing DX9 should be made at some point, right now I turned that off by default but open to discuss how that should be handled.

Would it be incorrect to keep that code live?  I'd rather not have modes unless it comes down to resolving contradictions between versions.

dankbaker:
> Would it be incorrect to keep that code live? I'd rather not have modes unless it comes down to resolving contradictions between versions

It would be technically incorrect yes, as those would ignored not in compat mode. So, if you made a variable with a semantic called VPOS it would just be ignored right now not in compat mode. What's even more complex is that it doesn't make sense to mix and match the semantics, because if you are parsing DX9 rules then the vertex to fragment linkage would be based on the additional COLOR and TEXCOORD names.

I'm not sure to what extent DX9 shaders running in compat mode are living out in the wild or to what extent it would mess things up to parse the semantics in all cases. HLSL has to run in a separate mode to be

dankbaker:
The other thing that this code doesn't do (but perhaps this happens later) is it doesn't validate the semantic is valid for this shader stage. I'm not sure to what extent we need or should be doing that here.

hlsl/hlslParseHelper.cpp
dankbaker: Should we assume that the program is setting depth if the semantic is present? Presence of the semantic just means the variable was declared, if it isn't used then it would be stripped in HLSL and would be the equiv of not having it. It's quite possible you'll see HLSL shaders with over specified output expecting unused things to get pruned.

johnkslang: For DEPTHGREATER/LESSER, it is a mode, and if this is where that comes from, then it should be passed through to the `intermediate.setDepth()` here.

For depth replacement, I believe that can wait until a write to that built-in variable is seen, which is logic that should already exist somewhere else, and can remain an independent change from this PR.

johnkslang:
> The other thing that this code doesn't do (but perhaps this happens later) is it doesn't validate the semantic is valid for this shader stage. I'm not sure to what extent we need or should be doing that here.

Validation is out of scope, accept as convenient or necessary to avoid compiler stack quality issues.  So, I don't think there is any work to do here.

This PR is almost basically done, if you address the depth-settings one way or another and add the test results.  Do you want to follow through on that @dankbaker?

dankbaker:
I'm still not sure how to address the depth settings.I can set them to an enum, but you can't call the intermediate here because there could be modules in a file that aren't part of the specific shader being compiled.

I think a phantom depth enum needs to be created. That is, it's essentially interpreted as depth but if it's present as an output in the Entrypointed function (and ONLY the entrypoint function), then it calls that function.

johnkslang:
Let's log managing depth as a separate issue then (issue #489), but tie all the depth-related annotations to EbvFragDepth, so this one can wrap up.

dankbaker:
Updated with that change, Should I update the tests? I think virtually all of the HLSL tests will change

johnkslang:
Normally, yes, test changes have to go along with the source changes, before going into master.

However, in this case, since PR #490 also has to go in, and also changed all tests, one will create a set of conflicts.  So, I'll merge your commits in locally and update tests for you this time.

johnkslang:
The interesting conflict is handling the entry-point outputs.

Question:   You numbered the parameter locations in order, while I added the return value.  Where in that output parameter ordering would the return value from the entry point go?

dankbaker:
output types of the return value would get ordered first, before any additional output arguments.  To merge the change, I think you'd want to scan the last location of the output struct before processing the output arguments and use that your start location.

johnkslang:
Okay.

That raises another issue: this whole stack does not work by having locations on structure members.

The indiscriminate addition of location for the SV_TARGET* caused the debug build to assert when it detected this.  This is getting turned off right now, just to keep the build stable.  Will have to think about what the right answer is a bit later.

johnkslang:
Okay, the "location setting" emergency is over.  Note three best practices not followed combined to make this happen: 1) the PR did multiple things instead of one thing (mapping built-ins -> built-ins and synthesizing locations), 2) the coding convention was not followed (the location setting was not a separate line of code) and 3) tests, at least with asserts, were never run (would have caught the location problem early).  This was also the cause of the code conflicts.  Partly my fault, and I'll insist more strongly on following all three of the practices in the future.

My statements about locations on struct members were overly broad, just writing fast while things were broken, but have some time now to think through what really needs to happen.

Thanks though for the mappings between HLSL semantics and GLSL built ins.

johnkslang:
>  I think you'd want to scan the last location of the output struct before processing the output arguments and use that your start location

Note that probably means a single output param could consume multiple locations, but the code change just does a `outCount++` so, probably makes overlapping outputs.

Virtually all test result changes were these locations, but I think they are almost all wrong, so will need to refine, so we'll still see all these test results changing again.

Sorry it wasn't on my radar that this was going on.

johnkslang:
> I think you'd want to scan the last location of the output struct before processing the output arguments

>  a single output param could consume multiple locations

These are fixed now by commit 830b0cc98b86b24f33af3e4c0419f67d9eb869b8.  But, generally I'm concerned about this uncovering what seems a theoretical problem in the HLSL model:  Things with location are in an I/O storage class, not local variables, which is a different storage class.  HLSL seems to conflate these.  Eliminating the parameters is working okay, but the pattern of having a local struct assigned to the return value output struct is worrisome as technically they are not the same type, as they could be laid out differently in their different storage classes.

johnkslang:
Well, will wrap this up as the last issue is actually captured in some other places.

Thanks @dankbaker ; we made forward progress together; tough when different people know different parts of the system. :)

dankbaker:
> These are fixed now by commit 830b0cc. But, generally I'm concerned about this uncovering what seems a theoretical problem in the HLSL model: Things with location are in an I/O storage class, not local variables, which is a different storage class. HLSL seems to conflate these. 

It's not conflated, it's just that HLSL isn't really C in this regard. For HLSL, a struct isn't a C++ structure with well defined offsets, it's a LISP-like data structure that has no intrinsic layout until it is 'vectorized'.  There may be cases where it has different layouts depending on the context it is used. For an output assignment, it's essentially creating an implicit conversion between two possible layouts of the struct. As long as the LISP style semantics of the struct are compatible, the appropriate conversion will happen.

I'm not sure how to elegantly resolve given the structure of the GLSLang implementation. HLSL's AST was essentially a LISP like structure with somewhat artificial C style rules to restrict it to a more C style.  Stuff like data structure layout happened during the AST -> SSA process, and if there was something ambiguous or incompatible it would error there. I think for our purposes there will just be some awkwardness and special cases where there are storage class differences.

johnkslang:
The glslang AST represents the HLSL just fine.  The local struct is assigned to the output struct with EOpAssign, no problem, so there is no elegant solution to find, it is already concisely represented.

A SPIR-V OpStore though requires same types, and a undecorated struct (e.g., no location) is not the same type as a decorated struct. Maybe it could expand to a store per member, or a change to SPIR-V to be higher level.  This is already what issue #304 is about.

PR #488, opened 2016-08-26
Disable C4996 (secure CRT) recommendation on Windows for strtok().

antiagainst:
`strtok_s()` is suggested by [MSVC](https://msdn.microsoft.com/en-us/library/ftsafwz3.aspx), but it has different signature
than the [C11 standard one](http://en.cppreference.com/w/c/string/byte/strtok). So we just turn off the recommendation
here.


johnkslang:
I think this is okay, since the strtok() are outside the library use of glslang (just for standalone).

Issue #487, opened 2016-08-26
Entry point loses correct return type [was: Can't get simple shader to compile properly]


redorav:
I'm trying to compile a very simple hlsl shader to spir-v to test and compare vs glsl. I can't get it to generate valid spir-v though.

`float4 entryPoint() : SV_TARGET0
{
	return float4(1, 0, 0, 0);
}`

Using this command line 

glslangValidator.exe example.hlsl -D -V -H -S frag -o example.hlsl.spv -e entryPoint

makes it generate the following SPIR-V

```
example.hlsl
Linked fragment stage:
// Module Version 10000
// Generated by (magic number): 80001
// Id's are bound by 12
                              Capability Shader
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "entryPoint"
                              ExecutionMode 4 OriginUpperLeft
                              Name 4  "entryPoint"
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 4
               8:    6(float) Constant 1065353216
               9:    6(float) Constant 0
              10:    7(fvec4) ConstantComposite 8 9 9 9
   4(entryPoint):           2 Function None 3
               5:             Label
                              ReturnValue 10
                              FunctionEnd
```

I'm seeing that it's declaring entryPoint as a void function, but still returning a float4. If I do the same with this very simple glsl shader,

```
#version 140
out vec4 outColor;
void main()
{
	outColor = vec4(1, 0, 0, 0);
}
```

the output SPIR-V is

```
example.glsl
Linked fragment stage:
// Module Version 10000
// Generated by (magic number): 80001
// Id's are bound by 13
                              Capability Shader
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main" 9
                              ExecutionMode 4 OriginUpperLeft
                              Source GLSL 140
                              Name 4  "main"
                              Name 9  "outColor"
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 4
               8:             TypePointer Output 7(fvec4)
     9(outColor):      8(ptr) Variable Output
              10:    6(float) Constant 1065353216
              11:    6(float) Constant 0
              12:    7(fvec4) ConstantComposite 10 11 11 11
[...]

Issue #480, opened 2016-08-25
HLSL: return types on vertex shaders don't produce correct SPIRV


dankbaker:
These should both be approximately similar code, but the second one produces what looks like wrong SPIRV (and fails the validator)

(assumed) Correct SPIRV:
```
struct VSOut
{
    float4 Pos : SV_Position;
    float2 Tex : location0;
};

VSOut main(uint vid : SV_VertexID, out VSOut Output ) 
{ 
    Output.Tex = 0; 
    Output.Pos = 1;
} 
```

(assumed failed, as fails validation) SPIRV
```
struct VSOut
{
    float4 Pos : SV_Position;
    float2 Tex : location0;
};

VSOut main(uint vid : SV_VertexID ) 
{ 
    VSOut Output;
    Output.Tex = 0; 
    Output.Pos = 1;
    return Output;
} 
```

redorav:
Right, thanks for looking into it. I'll monitor the bug and test once fixed.

johnkslang:
This is now known as a duplicate of #487 (well this one was first, but there is more up-to-date discussion in the other one).

Issue #457, opened 2016-08-12
SPV Remapper: requires automated tests


steve-lunarg:
To log the issue: the SPV remapper's testing has been non-public. It needs tests that are part of whichever project it lives in, both for basic correctness, and efficacy of compression improvement across a collection of shaders with partial commonality and different ID mappings.

dneto0:
I strongly recommend using the SPIRV-Tools assembler as the basis for making test cases.  I worked on that assembler specifically so it could be the basis of SPIRV testing and conformance.  Credit goes to @kbenzie and Codeplay for kicking off the project and taking it forward a long way.   He also had the idea that the assembler could be used to make syntacitcally *invalid* SPIR-V.

In particular, a great many of the SPIRV-Tools tests are based on a single test utility class (test fixture) for  easily managing the assembly of a module and getting its results: https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/TestFixture.h#L45
(That makes up for the fact that we never publicized a C++ API for the assembler.)

steve-lunarg:
Nod - I was thinking the same thing around use of SPIRV-Tools as a basis for testing.

johnkslang:
Also consider having the remapper live as part of SPIRV-Tools instead of part of glslang.

dneto0:
The SPIRV-Tools assembler is not ideal (today) to test the ID remapping algorithm. That's because it always assigns the next-higher ID value to each new name it encounters.  To test something like ID range compression (by removing gaps in the numbering), I'd have to extend the assembly language syntax to allow forced assignment of IDs.
E.g.
```
  %bool:500 = OpTypeBool   ;  will force the ID of %bool to be 500 instead of the assembler default
```

dneto0:
The assembler does have a "force the next words to be exactly as I say" mode, but it's pretty cumbersome to use.

steve-lunarg:
I've been thinking more about the disassembler and validator.  I can start with SPIR-V generated from canned GLSL shaders for instance, remap them, and compare the spirv-dis --raw-id --no-color post-remap version to a golden one.

In the past the idea of ID mapping invariant comparison has come up, though it's probably somewhat more involved to create that.  Anyway, I'm feeling like the remapper needs some TLC, and I plan to spend time on it soon.  Also, I have a fix coming for the defect @dneto0 found in my other fix.

steve-lunarg:
[...]

PR #490, opened 2016-08-28
HLSL: Emulate write-to-output on return-from-entry-point, for return …

johnkslang:
…value.

This fixes issue #487 and #480.
It also correctly handles output parameters from the entry point.


steve-lunarg:
The remapper test diffs look sane: they're the same semantic change as the glslangtests.  Updating the base cases should address that:

`./install/bin/remappertests --update-mode
`

hlsl/hlslParseHelper.cpp
```diff
@@ -55,7 +55,8 @@ HlslParseContext::HlslParseContext(TSymbolTable& symbolTable, TIntermediate& int
     contextPragma(true, false), loopNestingLevel(0), structNestingLevel(0), controlFlowNestingLevel(0),
     postMainReturn(false),
     limits(resources.limits),
-    afterEOF(false)
+    afterEOF(false),
```
steve-lunarg: Although the variable was already there (the trailing comma triggered the diff above), doesn't this shadow the base TParseContext variable of the same name?  I guess it might not presently be used in the HLSL path anyway: AFAICS it's initialized to false, is never changed, and maybe the only consumption is in a function called only from the bison / flex glsl files.

johnkslang:
Is there a more streamlined way?  This is going to cause lots of problems that a different executable breaks than the one being modified and tested.  Somehow I was expecting the existing test methods to now include the remapper tests, rather than it being a separate step to execute.

steve-lunarg:
Sure - they can become part of the glslangtests quite easily, so they are updated with the same step.  The rational for making them separate was in #457 and #473, but it's very easy to swap it back until such time that it can be fully separated (which is a bigger scoped change).

johnkslang:
I think there is a discrepancy between using ctest vs. glslangtests.  I've only been using the latter, and the instructions for finer control only discuss the latter.  That's probably the main issue here.  What did you have in mind for making the remapper tests part of glslangtests?

steve-lunarg:
There's a Remap.FromFile.cpp set of tests that's analogous to the Hlsl.FromFile.cpp or AST.* tests.  It's currently built into a separate exe, but that's almost entirely down to the CMakeLists.txt.  It would be trivial to have them run and updated as part of glslangtests, no differently than any of the other tests there.

When the remapper is separated from the project, the tests would be moved simultaneously.

Should I make a PR for this?

johnkslang:
> Should I make a PR for this?

Sure, if that means the glslangtests executable tests both the glslang and remapper executables.  If not, other things need to change, so that at least the build/test instructions match the bot actions.

hlsl/hlslParseHelper.cpp
johnkslang: There was no shadowing, the base class has no such variable (both GLSL and HLSL had it though).

Anyway this and a related one are in fact all dead, and the correct state can be checked by querying the scanner, now that I added that with 5e56423046b470d7cf3bdd8b58339a169d9524f6.

johnkslang: Ah... I realize naming confused you.  GLSL's subclass is still called TParseContext.  Both derive from TParseContextBase.  The places you saw afterEOF were peers.

PR #501, opened 2016-09-06
Preprocessor: UINT_MAX is translated to constant incorrectly.

amdrexu:
This is the shader:

```
#version 450

#define WRITE_VALUE 4294967295u

void main()
{
    uint u0 = WRITE_VALUE;
}
```

When translated to SPIR-V, the UINT_MAX takes incorrect value:

`9:      6(int) Constant 2147483647`

We expect:

`9:      6(int) Constant 4294967295`

The root cause is that when we define a macro, the string expansion is done in `ReadToken`. We should take type of the constant into consideration before invoking the conversion function `atoi` or `atoll`.


glslang/MachineIndependent/preprocessor/PpTokens.cpp
```diff
@@ -243,6 +251,16 @@ int TPpContext::ReadToken(TokenStream *pTok, TPpToken *ppToken)
             } else
                 ppToken->i64val = atoll(ppToken->name);
             break;
+        case PpAtomConstUint64:
+            if (len > 0 && tokenText[0] == '0') {
+                if (len > 1 && (tokenText[1] == 'x' || tokenText[1] == 'X'))
+                    ppToken->i64val = (long long)strtoull(ppToken->name, nullptr, 16);
+                else
+                    ppToken->i64val = (long long)strtoull(ppToken->name, nullptr, 8);
+            }
```
johnkslang: `} else`

glslang/MachineIndependent/preprocessor/PpTokens.cpp
```diff
@@ -233,8 +232,17 @@ int TPpContext::ReadToken(TokenStream *pTok, TPpToken *ppToken)
             } else
                 ppToken->ival = atoi(ppToken->name);
             break;
+        case PpAtomConstUint:
+            if (len > 0 && tokenText[0] == '0') {
+                if (len > 1 && (tokenText[1] == 'x' || tokenText[1] == 'X'))
+                    ppToken->ival = (int)strtoul(ppToken->name, 0, 16);
+                else
+                    ppToken->ival = (int)strtoul(ppToken->name, 0, 8);
+            }
```
johnkslang: Do the `} else` the same way it is done on line 232.

johnkslang:
Good catch.  Do you have a test change to go along with it?  Perhaps at the end of the current `numeral.frag`, or a new `cppNumeral.frag` test.

amdrexu:
Thank you. I update the codes and add two cases in `spv.uint.frag` and `spv.int64.frag`.

PR #505, opened 2016-09-09
HLSL: alias HLSL Mrc onto GLSL Mcr

steve-lunarg:
For discussion: do not merge.  This updates #485, which I'll close presently.


johnkslang:
This looks okay to me, if the test results are correct.  Are there any open questions about it? If not, I believe this is a big step in the right direction, and will merge when you remove WIP.

steve-lunarg:
I'm OK with removing the WIP, but nervous because it's had zero exposure to actual rendering.  It arguably has considerable company in that regard, but it's also a bit different from other things with straightforward mappings to well sorted AST.

I hand checked the const-folded test results for the new test (hlsl.matrixindex.frag) and believe it to be correct.  The other test diffs are coming from reversal of the arg orders in M\*V, V\*M, and M\*M, and minor diffs from setting the entry point to the main() stub in the intrinsic test.

Since other folks are convinced this approach is sound, I'll remove the WIP soon.

johnkslang:
Isn't a better place to start than not doing it though?  I thought the current approach also has no rendering and we know it's more wrong than what you propose in this PR.  Yes? No?

steve-lunarg:
Agreed.  Removed WIP.

PR #509, opened 2016-09-15
HLSL: Add location offsets per resource type

steve-lunarg:
This PR adds the ability to offset sampler, texture, and UBO bindings from provided base bindings, and to auto-number bindings that are not provided with explicit register numbers.  The mechanism works as follows:

- Offsets may be given on the command line for all stages, or individually for one or more single stages, in which case the offset will be auto-selected according to the stage being compiled.  There is also an API to set them.  The new command line options are --base-sampler-binding, --base-texture-binding, and --base-UBO-binding.

- Uniforms which are not given explicit bindings in the source code are auto-numbered if and only if they are in live code, as determined by the algorithm used to build the reflection database.  This auto-numbering avoids using any binding slots which were explicitly provided in the code, whether or not that explicit use was live.  E.g, "uniform Texture1D foo : register(t3);" with --base_texture_binding 10 will reserve binding 13, whether or not foo is used in live code.

- This code changes from a hardcoded "main" used to determine the live call tree, to the provided entry point, so should work in all cases where that is correct.

Code details: 

- The liveness algorithm present in the reflection database has been refactored out and is now used both by reflection, and IO mapping.  TLiveTraverser is the factored-out common code, TReflectionTraverser handles reflection, and TBindingTraverser and TIoMappingTraverser handle IO mapping.  This is short-circuited if no offsets are given.

- The testing infrastructure is slightly extended to allow use of the binding offset API, and a new test spv.register.autoassign.frag is added for comparing the resulting SPIR-V.


steve-lunarg:
Closing: will recreate as several smaller PRs.

Issue #292, opened 2016-05-17
Add auto-location-assigning mode to glslang, with linking


johnkslang:
## From Michał Pyrzowski:

If an explicit location is not specified for interface variables in the GLSL shader than the glslang generated SPIR-V code does not contain location decorations for those variables.

Such SPIR-V code does not follow Vulkan validation rule:

> All variables (or block members) in the Input and Output storage class must have a location decoration.

The cross-stage linking would probably be the best solution to the problem.

I suppose that if Glslang is unable to generate a valid Vulkan SPIR-V from the provided GLSL source than at least the user should be notified of it. It could be done by a warning/error message.

johnkslang:
Correct.  Glslang is not inventing location information; that currently needs to come from the shader.

scyganINTEL:
CC @scyganINTEL

johnkslang:
Note auto-numbering is required by the HLSL path.  There is the in/out interface that needs location numbers, and then the resource/buffer interfaces that need binding numbers.  Wondering about leverage/interaction between these.

ChrisGautier:
I second the emission of e.g. an error message. We have seen content broken because of the absence of locations.

arcady-lunarg:
This has been implemented, please see the `--auto-map-locations` command-line option. There are no plans to implement the cross-stage form of this at this time.

PR #514, opened 2016-09-20
HLSL: Add location offsets per resource type

steve-lunarg:
HLSL: Add location offsets per resource type

This PR adds the ability to offset sampler, texture, and UBO bindings from provided base bindings, and to auto-number bindings that are not provided with explicit register numbers. The mechanism works as follows:

- Shift offsets may be given on the command line for all stages, or individually for one or more single stages, in which case the offset will be auto-selected according to the stage being compiled. There is also an API to set them. The new command line options are --shift-sampler-binding, --shift-texture-binding, and --shift-UBO-binding.  See the --help for details.

- Uniforms which are not given explicit bindings in the source code are auto-numbered if and only if they are in live code as determined by the algorithm used to build the reflection database, and the --auto-map-bindings option is given. This auto-numbering avoids using any binding slots which were explicitly provided in the code, whether or not that explicit use was live. E.g, "uniform Texture1D foo : register(t3);" with --shift-texture-binding 10 will reserve binding 13, whether or not foo is used in live code.

- Shorter synonyms for the command line options are available.  See the --help output.

The testing infrastructure is slightly extended to allow use of the binding offset API, and two new tests spv.register.(no)autoassign.frag are added for comparing the resulting SPIR-V.

(This message is in the git commit as well).


StandAlone/StandAlone.cpp
```diff
@@ -223,6 +262,18 @@ void ProcessArguments(int argc, char* argv[])
     for (; argc >= 1; argc--, argv++) {
         if (argv[0][0] == '-') {
             switch (argv[0][1]) {
+            case '-':
+                // handle --word style options
+                if (strcmp(argv[0]+2, "base-sampler-binding") == 0) {
```
johnkslang: Probably want case insensitive compares.

gtests/TestFixture.h
```diff
@@ -234,6 +234,52 @@ class GlslangTest : public GT {
...
+
+        if (success && (controls & EShMsgSpvRules)) {
+            std::vector<uint32_t> spirv_binary;
+            glslang::GlslangToSpv(*program.getIntermediate(kind),
+                                  spirv_binary, &logger);
+
+            std::ostringstream disassembly_stream;
+            spv::Parameterize();
+            spv::Disassemble(disassembly_stream, spirv_binary);
+            return {{{shaderName, shader.getInfoLog(), shader.getInfoDebugLog()},},
```
johnkslang: I think this is the syntax that doesn't work with our defacto minimum level of C++.

johnkslang:
Check also issue #292 for any interaction.

steve-lunarg:
Code updated with:

- Renamed command line options to --shift-sampler-bindings, etc, per request.

- Command line options are now case insensitive.

- Command line options have synonyms: --ssb == --shift-sampler-bindings == --shift-sampler-binding, etc.

- Auto-numbering must be explicitly enabled, and is no longer triggered automatically if the sampler bindings are to be offset.  --auto-map-bindings or --amb.

- A new test case is added so we test both auto mapping and not, since auto-mapping is no longer tied to shifting.

- Since auto-numbering is no longer tied to offsetting bindings, and 0 is a no-op for offsetting binding numbers (new_binding = old_binding+0), the defaults are still 0.  If there's nothing for the io mapper to do, it trivially returns without traversing the AST.

gtests/TestFixture.h
steve-lunarg: This was already used in a bunch of places in the same file; I got it by copying and modifying an existing instance.  So it shouldn't be introducing anything new, but we could make a pass over the file and remove all of them.  Maybe as its own PR?

johnkslang: Ah, I see, it's only used outside glslang proper, in the test harness.  So, I think this must be okay.

PR #515, opened 2016-09-21
Flatten uniform arrays

steve-lunarg:
Flatten uniform arrays

This checkin adds a --flatten-uniform-arrays option which can break uniform arrays of samplers, textures, or UBOs up into individual scalars named (e.g) myarray[0], myarray[1], etc.  These appear as individual linkage objects.

Code notes:

- shouldFlatten is replaced with: shouldFlattenArray, shouldFlattenStruct, and shouldFlattenType (which is true if either of the others are).

- There's some error checking around unhandled situations.  E.g, flattening uniform arrays with initializer lists is not implemented.

- This piggybacks on as much of the existing mechanism for struct flattening as it can.  E.g, it uses the same flattenMap, and the same flattenAccess() method.

- handleAssign() has been generalized to cope with either structs or arrays.

- Extended test infrastructure to test flattening ability.

### DIFF
--- a/hlsl/hlslParseHelper.h
+++ b/hlsl/hlslParseHelper.h
@@ -83,9 +83,6 @@ public:
     TIntermTyped* handleBinaryMath(const TSourceLoc&, const char* str, TOperator op, TIntermTyped* left, TIntermTyped* right);
     TIntermTyped* handleUnaryMath(const TSourceLoc&, const char* str, TOperator op, TIntermTyped* childNode);
     TIntermTyped* handleDotDereference(const TSourceLoc&, TIntermTyped* base, const TString& field);
-    bool shouldFlatten(const TType&) const;
-    void flatten(const TVariable& variable);
-    TIntermTyped* flattenAccess(TIntermTyped* base, int member);
     void assignLocations(TVariable& variable);
     TFunction& handleFunctionDeclarator(const TSourceLoc&, TFunction& function, bool prototype);
     TIntermAggregate* handleFunctionDefinition(const TSourceLoc&, TFunction&);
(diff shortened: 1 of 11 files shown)


steve-lunarg:
Re-push to resolve minor conflict with 6714bcc2ca97f668c0358f3d8ce039d61b5ab25e.

hlsl/hlslParseHelper.h
```diff
@@ -83,8 +83,11 @@ class HlslParseContext : public TParseContextBase {
     TIntermTyped* handleBinaryMath(const TSourceLoc&, const char* str, TOperator op, TIntermTyped* left, TIntermTyped* right);
     TIntermTyped* handleUnaryMath(const TSourceLoc&, const char* str, TOperator op, TIntermTyped* childNode);
     TIntermTyped* handleDotDereference(const TSourceLoc&, TIntermTyped* base, const TString& field);
-    bool shouldFlatten(const TType&) const;
-    void flatten(const TVariable& variable);
+    bool shouldFlattenStruct(const TType&) const;
+    bool shouldFlattenArray(const TType&) const;
+    bool shouldFlattenType(const TType& type) const { return shouldFlattenStruct(type) || shouldFlattenArray(type); }
```
johnkslang: I'd made the existing functions be "shouldFlatten" and "flatten" to mean this already; the generic form.  So, there is the matter of preference to go one way or the other, but to cross check something important:  a single object can be both an array and a struct at the same time, in which case the arrayness is outside the struct (it's an array of struct).  So, shouldFlattenStruct might be bypassing the arrayness.  Just verifying, in case there is any unexpected bug here (and thinking the generic one would do the right thing, rather than something externally having to decide which question to ask).

steve-lunarg:
Updated as follows:

- shouldFlatten() is generic and used everywhere that used to call the more specific forms.  (It still uses more specific ones internally, but they're not directly invoked elsewhere).  The more specific versions are renamed to shouldFlattenIO and shouldFlattenUniform().

- flatten() is also generic.  However, it tests for arrayed structs and errors out, since it can't handle that case.  Internally, it invokes flattenArray or flattenStruct as appropriate, since the mechanics are different.

- Moved the methods to protected, since they shouldn't really be public.

- Added a protection for the command line option that will error out if it's not used with HLSL source, since all this flattening code only exists in the hlsl path.

PR #521, opened 2016-09-26
SPV: PrimitiveId in frag shader will emit OpCapability

yavn:
Using `PrimitiveId` in a fragment shader requires declaring an `OpCapability` with either `Geometry` or `Tessellation`.

Not sure if this is an adequate fix. Currently if a fragment shader uses a `gl_PrimitiveId` then the generated SPIR-V doesn't include a required capability, which leads to a validation failure (in SPIRV-Tools). To be honest, it isn't clear to me from the spec if either of these capabilities will do, or if both are needed.


yavn:
I'll get test failures fixed tomorrow.

johnkslang:
This PR needs to be accompanied by any test changes, or if none, then a test must be added.  See test list in the file gtests/Spv.FromFile.cpp.

PR #523, opened 2016-09-27
Parser: Implement extension GL_AMD_gpu_shader_half_float.

amdrexu:
- Add built-in types: float16_t, f16vec, f16mat.
- Add support of half float constant: hf, HF.
- Extend built-in floating-point operators: +, -, *, /, ++, --, +=, -=,
  *=, /=, ==, !=, >=, <=, >, <.
- Add support of type conversions: float16_t -> XXX, XXX -> float16_t.
- Add new built-in functions.


amdrexu:
Hi,
Some special notes are as follow:

(1) The specs are published here: 
https://www.khronos.org/registry/spir-v/extensions/AMD/SPV_AMD_gpu_shader_half_float.html
https://www.opengl.org/registry/specs/AMD/gpu_shader_half_float.txt

(2) All codes are within AMD_EXTENSIONS.

(3) I borrow two utility files from SPIRV-tools: bitutils.h and hex_float.h. I am not sure if it is acceptable. The two files is used to perform bit pattern conversion between float32 and float16.

(4) There is a test spv.float16.frag. I add it to the test folder. But I do not enable it by default in glslangtests. 

Thanks very much for your review.

amdrexu:
I modify this change and add spv.float16.frag to glslangtests. Still, it is within the macro AMD_EXTENSIONS.

DominikWitczakAMD:
@johnkslang : Can we get this merged, please? We have a bunch of dependencies blocked on this PR. Thanks!

johnkslang:
I read through it now, and it looks pretty good.

> (2) All codes are within AMD_EXTENSIONS.

Did you mean for that to also apply to all the new built-ins? It is a lot, but I guess they'll be in the symbol table anyway, because now we run with this always turned on.

This may be the real cost of having all extensions on, lots of growth here, where lots of users won't care.  But, that just might be how it is.

> (3) I borrow two utility files from SPIRV-tools: bitutils.h and hex_float.h. I am not sure if it is acceptable. The two files are used to perform bit pattern conversion between float32 and float16.

@dneto0 any input on this?

dneto0:
bitutils.h and hex_float.h utils are licensed under Apache 2.0 just like the rest of SPIRV-Tools.  There was a relicensing step a few weeks ago. Please use the Apache 2.0 version; just grab latest sources.

Also, there are a bunch of unit tests for hex_float.h in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/HexFloat.cpp using the gtest test framework.
I recommend you bring them over too.  They too are under Apache 2 license.
They should be easily transportable to Glslang's gtest-based suite.  (You can delete the #include "UnitSPIRV.h" in HexFloat.cpp as it's not needed.  I plan to remove it.)
The tests run very quickly, just a few milliseconds.

dneto0:
Ok, I've removed the useless include in the HexFloat.cpp test file.
https://github.com/KhronosGroup/SPIRV-Tools/pull/441
Now you can port it more easily (just adjust the #include of hex_float.h)

dneto0:
(Lest anyone think the testing is overkill:  hex_float.h relies on std::istream.  In somewhat related code in SPIRV-Tools, I used std::istringstream and testing revealed 3 distinct bugs in various platform libraries that I then had to work around.)

johnkslang:
So, to summarize from the whole discussion what needs to be done:

- Must: Get the most recent versions of the files copied from SPIRV-Tools.
- Desired: Include tests for them.

Thanks.

amdrexu:
Thanks for your suggestions. I have updated the codes based on the requirements.

Issue #517, opened 2016-09-22
Remove optionality of support for vendor-specific extension in glslang.


DominikWitczakAMD:
Right now, in order to be able to access AMD-specific extensions (such as GL_AMD_shader_ballot), ISVs are required to modify CMake configuration and build customized versions of glslang, as CMake's ENABLE_AMD_EXTENSIONS option is disabled by default.

We would like to have that CMake option removed, making extension support *always enabled* in glslang. We see this problem being frequently raised by ISVs who use Vulkan, and it is also our belief that customized glslang builds are not the right way for providing support for GLSL / SPIR-V extensions in glslang in the foreseeable future.

While I referred to our existing extensions here as an example, we are worried about current approach to adding support for vendor-specific extensions as a whole. We would like to see glslang include support for all vendor-specific extensions in general, in order to avoid fragmenting Vulkan ecosystem.

baldurk:
:+1: from me, especially since at runtime it's still protected by needing`#extension VK_FOO_bar_blah : enable`, so having default-on compile-time support won't affect shaders unless people opt-in anyway.

scyganINTEL:
As glslang is being distributed in Vulkan SDKs (Lunar-G SDK, Android NDK, ...) as a prebuild binary, the build-time configuration may be highly uncomfortable for the developers. 

Any run-time way for enabling extensions is better. If people are concerned in thight control of the feature set used by SPIR-V output, a commandline or config file can be used to select target SPIR-V extensions (in parallel to GLSL #extension enables in shader source).

Novum:
Yes from our side as well. I don't think a command line option is necessary because the shader already controls the enabled extensions.

kayru:
From Frostbite Engine point of view, `#extension` opt-in mechanism in shaders should be the only requirement for using an extension with vanilla VulkanSDK. We would like to see vendor extensions regularly up-streamed and available to use with minimal setup.

johnkslang:
This was done to connect the code quality of an extension implementation with the extension.  That is, accepting vendor-specific PRs with less code reading and testing when they were being "owned" by a specific vendor, making quality that vendor's responsibility, rather than Khronos's.  That is, I was ensuring the quality of what's outside the #ifdef's but not what's inside.

That would be most important for use of glslang in a online environment rather than offline.  E.g. vendor A doesn't want vendor B breaking glslang in an online situation.

[...]

PR #524, opened 2016-09-27
#517: Enable AMD extensions by default

DominikWitczakAMD:
Please see #517 for more details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.11)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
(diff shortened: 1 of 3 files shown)


DominikWitczakAMD:
@amdrexu : Thanks! Re-submitted, should be fine now.

steve-lunarg:
Small note to hopefully help someone else: since this PR changes a CMakeLists.txt variable, I had to remove CMakeCache.txt and re-run cmake for it to be reflected and properly match the new test gold files.  (There are other means to the same end, such as recreating the build dir).

Nothing wrong with the commit; I just forgot to do that at first and saw some test diffs due to the stale cache.

johnkslang:
Or, if you use GUI's you check the box next to the AMD thing and it all works.

I suppose ongoing, since we are having these all turned now, we don't need to maintain a set of check boxes.

PR #527, opened 2016-09-29
HLSL: Restrict uniform array flattening to sampler and texture arrays

steve-lunarg:
Previously the uniform array flattening feature would trigger on loose uniform arrays of any basic type (e.g, floats).  This PR restricts it to sampler and texture arrays.  Other arrays would end up in their own uniform block (anonymous or otherwise).  (Atomic counter arrays might be an exception, but those are not currently flattened).

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -736,7 +736,9 @@ bool HlslParseContext::shouldFlattenUniform(const TType& type) const
 
     return type.isArray() &&
         intermediate.getFlattenUniformArrays() &&
-        qualifier == EvqUniform;
+        qualifier == EvqUniform &&
+        // Testing the EbtSampler basic type covers samplers and textures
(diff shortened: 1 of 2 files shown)


hlsl/hlslParseHelper.cpp
```diff
@@ -736,7 +736,9 @@ bool HlslParseContext::shouldFlattenUniform(const TType& type) const
 
     return type.isArray() &&
         intermediate.getFlattenUniformArrays() &&
-        qualifier == EvqUniform;
+        qualifier == EvqUniform &&
+        // Testing the EbtSampler basic type covers samplers and textures
+        type.getBasicType() == EbtSampler;
```
johnkslang: FYI, it is a bit more accurate to call TType::isOpaque().  I think it is a don't care for this path, but generally, there is perfect correlation between what can't go into a buffer and what is opaque, which is also the distinction here, that it's things that get their own binding, because they can't be part of a buffer.

PR #532, opened 2016-10-01
Change binding auto-map to use provided offsets.

steve-lunarg:
Previously, the binding auto-mapping facility was free to use any unused binding.  This change makes auto-bindings use the same offset value as explicit bindings.


steve-lunarg:
Re-pushed to update test that was added in #531 - no code changes.

PR #533, opened 2016-10-01
add reflection queries to return a TType.  Fix minor issue with inter…

steve-lunarg:
…face names.

- Add new queries: TProgram::getUniformTType and getUniformBlockTType, which return a const TType*, or nullptr on a bad index.  These are valid for any source language.

- Interface name for HLSL cbuffers is taken from the (only) available declaration name, whereas before it was always an empty string, which caused some troubles with reflection mapping them all to the same index slot.  This also makes it appear in the SPIR-V binary instead of an empty string.

- Print the binding as part of the reflection textual dump.

- Since it's invalid to have both an instance and a block name in HLSL (e.g, cbuffer A { ... } B), and some existing HLSL tests were doing that, I have changed a few of the tests so they don't. That in effect hoists the members to the global namespace, which caused a few collisions, so I renamed those.  E.g, in hlsl.reflection.vert.  This makes them valid code.

- TType::clone becomes const.  Needed to call it from a const method, and anyway it doesn't change the object it's called on.

- Because the TObjectReflection constructor is called with a TType *reference* (not pointer) so that it's guaranteed to pass in a type, and the "badReflection" value should use a nullptr there, that now has a dedicated static method to obtain the bad value.  It uses a private constructor, so external users can't create one with a nullptr type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,6 @@ build/
 Test/localResults/
 Test/multiThread.out
 Test/singleThread.out
+Test/vert.spv
+Test/frag.spv
 External/googletest
--- a/Test/baseResults/hlsl.buffer.frag.out
+++ b/Test/baseResults/hlsl.buffer.frag.out
@@ -101,23 +101,23 @@ gl_FragCoord origin is upper left
                               Name 22  ""
                               MemberName 22 0  "v2"
                               Name 24  ""
-                              Name 28  ""
-                              MemberName 28 0  "v3"
-                              MemberName 28 1  "i3"
+                              Name 28  "cbufName"
+                              MemberName 28(cbufName) 0  "v3"
+                              MemberName 28(cbufName) 1  "i3"
                               Name 30  ""
-                              Name 35  ""
-                              MemberName 35 0  "v4"
-                              MemberName 35 1  "i4"
-                              MemberName 35 2  "f1"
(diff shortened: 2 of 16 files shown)


johnkslang:
I'm concerned that this sounds like it is fundamentally incapable of supporting a syntax that gives a scope to buffer members.  I'd like to not lose that functionality.

johnkslang:
BTW, completing my review of this, it otherwise looks good.

steve-lunarg:
Hmm... I don't think it should preclude such an expansion, if that happens in the future.  In (current) HLSL the block name is the interface name, but if (future) HLSL adds an instance name akin to what current GLSL has, it would be a one or two line change to use it, or to do whatever those rules say to do.  (E.g, it might be shader model specific).  AFAIK no underlying support for the instance name is removed - it's just down to what's actually passed in.

steve-lunarg:
Thinking about this a little more, we could do something like this:

`    const TString& interfaceName = (instanceName && !instanceName->empty()) ? *instanceName : type.getTypeName();
`

That would use the instance name if it's there, else fall back to using the block name, which should act identically to the currently-pushed code for all current shaders.  Of course that's rather speculative around what the right future behavior would actually be.

steve-lunarg:
Changed to WIP while looking something over...

PR #536, opened 2016-10-03
HLSL: fix for flattening assignments from non-symbol R-values.

steve-lunarg:
If a member-wise assignment from a non-flattened struct to a flattened struct sees a complex R-value  (not a symbol), it now creates a temporary to hold that value, to avoid repeating the R-value.  This avoids, e.g, duplicating a whole function call.  Also, it avoids re-using the AST node, making a  new one for each member inside the member loop.    

The latter (re-use of AST node) was also an issue in the GetDimensions intrinsic decomposition, so this PR fixes that one too.

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -952,10 +952,53 @@ TIntermTyped* HlslParseContext::handleAssign(const TSourceLoc& loc, TOperator op
     const TVector<TVariable*>* leftVariables = nullptr;
     const TVector<TVariable*>* rightVariables = nullptr;
 
+    // A temporary to store the right node's value, so we don't keep indirecting into it
+    // if it's not a simple symbol.
+    TVariable*     rhsTempVar   = nullptr;
+
+    // If the RHS is a simple symbol node, we'll copy it for each member.
+    TIntermSymbol* cloneSymNode = nullptr;
+
+    // Array structs are not yet handled in flattening.  (Compilation error upstream, so
+    // this should never fire).
+    assert(!(left->getType().isStruct() && left->getType().isArray()));
+
+    int memberCount = 0;
+
+    // Track how many items there are to copy.
+    if (left->getType().isStruct())
+        memberCount = left->getType().getStruct()->size();
+    if (left->getType().isArray())
+        memberCount = left->getType().getCumulativeArraySize();
+
     if (flattenLeft)
         leftVariables = &flattenMap.find(left->getAsSymbolNode()->getId())->second;
-    if (flattenRight)
+
+    if (flattenRight) {
         rightVariables = &flattenMap.find(right->getAsSymbolNode()->getId())->second;
+    } else {
+        // The RHS is not flattened.  There are several cases:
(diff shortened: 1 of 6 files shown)


johnkslang:
I think the real problem disguised itself as this.

The real problem is a single node got reused multiple times, getting multiple parents, making an DAG instead of a tree.  I think this was true whether or not it was a complex R-value.

Does the proposed fix still leave things such that if the RHS was not complex, a DAG was formed?

I'm thinking we really need a PR more along the lines "fix flattening to make a tree instead of a DAG".

steve-lunarg:
Makes sense.  Updated the PR to make a new AST node for each, and made a similar change to the decomposition of the GetDimensions intrinsic, which had the same reuse-of-AST-node issue.

hlsl/hlslParseHelper.cpp
```diff
@@ -952,10 +952,29 @@ TIntermTyped* HlslParseContext::handleAssign(const TSourceLoc& loc, TOperator op
...
+        // a temporary copy if it isn't a symbol.
+
+        // If it's a symbol, we'll use it directly below.  Otherwise, make a temp in the symtab for it.
+        if (right->getAsSymbolNode() == nullptr) {
+            rhsTempVar = makeInternalVariable("flattenTemp", right->getType());
+            rhsTempVar->getWritableType().getQualifier().makeTemporary();
+            TIntermTyped* noFlattenRHS = intermediate.addSymbol(*rhsTempVar, loc);
+
+            // Add this to the aggregate being built.
+            assignList = intermediate.growAggregate(assignList, intermediate.addAssign(op, noFlattenRHS, right, loc), loc);
```
johnkslang: It makes sense for the assignList to hold the copy operation from the complex symbol to the temporary symbol, but not just the temporary symbol.

hlsl/hlslParseHelper.cpp
```diff
@@ -978,7 +997,10 @@ TIntermTyped* HlslParseContext::handleAssign(const TSourceLoc& loc, TOperator op
         const auto& members = *left->getType().getStruct();
 
         for (int member = 0; member < (int)members.size(); ++member) {
-            TIntermTyped* subRight = getMember(flattenRight, right, *rightVariables, member,
+            // Add a new AST symbol node if we have a temp variable holding a complex RHS.
+            TIntermTyped* noFlattenRHS = rhsTempVar ? intermediate.addSymbol(*rhsTempVar, loc) : right;
```
johnkslang: Don't we always need a symbol?  Assuming we are to keep this a tree instead of a DAG.

Outside this function, a single usage of the right side was created.  If only a single use of it is needed, the one passed in can be used.  Inside here, if we decide we need more than one use of it, the extra ones need to be synthesized.

This is orthogonal to how complex the right side is, but agree with not replicating complex ones, but rather replicating a temporary symbol node holding the result.

So, final replication has to do with whether the for-loop has more than one iteration,

This seems it should impact  line 972 above as well.

steve-lunarg:
Re-pushed with code review changes.

Issue #342, opened 2016-06-16
Thread safety issue in spirvbin_t (race condition in Parameterize function)


C0lumbo:
The documentation (README-spirv-remap.txt) says: "It is safe to instantiate one spirvbin_t per thread and process a different SPIR-V in each."

However, spv::Parameterize(); is called at the start of spirvbin_t::remap. spv::Parameterize() uses the local static 'initialized' to avoid doing its initialization work multiple times, which is not a thread-safe approach. 

Calling spirvbin_t::remap on two threads at once leads to thread 1 proceeding past the 'initialized' check and doing the proper work of the Parameterize function. Meanwhile thread 2 sees that 'initialized' is now true and quickly exits the Parameterize function and proceeds to do the remapping work before thread 1 has completed initialization eventually leading to a "vector subscript out of range" assert.

Projects can work around this problem by calling spv::Parameterize before beginning any threaded spirvbin_t::remap work.

A fix could just be a modification of the documentation to inform users to call spv::Parameterize. Or the 'initialized' var could be protected by a mutex.

mknejp:
> However, spv::Parameterize(); is called at the start of spirvbin_t::remap. spv::Parameterize() uses the local static 'initialized' to avoid doing its initialization work multiple times, which is not a thread-safe approach.
Since C++11 it is required that initialization of local static variables is thread-safe.

C0lumbo:
Very interesting. I don't think it makes a difference in this case though. The issue would be the same if 'initialized' was a global var. There's nothing to stop a second thread doing remap work while a first thread is still doing the Parameterize work.

ratchetfreak:
The test-then-set is still a race, in 2 ways

1. 2 (or more) threads can pass the if check before any of them set the static to true, leading to several threads filling the maps and thrashing on each other.

2. a thread coming in later while the first thread is still busy will attempt to use partially initialized maps.

To fix this you would need to lock over the entire initialization code:

``` C++
void Parameterize()
{
    static bool initialized = false;
    if (initialized) 
        return;
    auto g = lock_guard(&mutex);
    if (initialized) 
        return; //double check

    // do init

    initialized = true;
}
```

or just use std::call_once

void Parameterize()
{
    static std::once_flag flag;
    std::call_once(flag, [](){

        // do init

    }
}
```

mknejp:
Hmm actually you're right. The easiest would be to just use a lambda for the side effects.
```cpp
void Parameterize()
{
  static const bool init = []
[...]

PR #539, opened 2016-10-06
Use pthread_mutex for global lock on Linux.

schwa423:
This allows multithreaded use of multiple TShader instances on Linux.  If I wasn't lazy, I'd also implement OS_CreateThread() and friends, but I don't need them and I already spent hours tracing my crash to the silent lack of a global lock.

### DIFF
--- a/glslang/OSDependent/Unix/ossource.cpp
+++ b/glslang/OSDependent/Unix/ossource.cpp
@@ -165,11 +165,12 @@ bool OS_FreeTLSIndex(OS_TLSIndex nIndex)
 		return false;
 }
 
-// TODO: non-windows: if we need these on linux, flesh them out
-void InitGlobalLock() { }
-void GetGlobalLock() { }
-void ReleaseGlobalLock() { }
+static pthread_mutex_t gMutex;
+void InitGlobalLock() { pthread_mutex_init(&gMutex, NULL); }
+void GetGlobalLock() { pthread_mutex_lock(&gMutex); }
+void ReleaseGlobalLock() { pthread_mutex_unlock(&gMutex); }
 
+// TODO: non-windows: if we need these on linux, flesh them out
 void* OS_CreateThread(TThreadEntrypoint /*entry*/)
 {
     return 0;
--- a/glslang/OSDependent/Unix/ossource.cpp
+++ b/glslang/OSDependent/Unix/ossource.cpp
@@ -165,10 +165,27 @@ bool OS_FreeTLSIndex(OS_TLSIndex nIndex)
 		return false;
 }
 
-static pthread_mutex_t gMutex;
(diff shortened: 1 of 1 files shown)


j9brown:
/cc @johnkslang 

Not sure why the tests are failing on the build server here, but we encountered a need to support multi-threaded shader compilation on our platform.  Any suggestions?

johnkslang:
@steve-lunarg is this the same space you were looking at a while back? Any input?

steve-lunarg:
Yeah, although for a different end.  Mine was for SPIR-V remapper thread safety, but it fleshed out the same functions in the same way.   I made a WIP branch for feedback for issue #342.  The code for that is [here](https://github.com/steve-lunarg/glslang/pull/1).  The mutex changes in that WIP are essentially identical to this, except I'd used PTHREAD_MUTEX_INITIALIZER, like so:

```
namespace {
    pthread_mutex_t GlobalLock = PTHREAD_MUTEX_INITIALIZER;
} // end anonymous namespace

```

That has the same effect as pthread_mutex_init with the default options, so this approach seems fine too, as long as InitGlobalLock gets called.

steve-lunarg:
BTW, @j9brown: you might try re-submitting the same code.  There have been cases in the past where the build bots time out for some reason.  I'm not certain that's the case here, but it could be.

johnkslang:
This is not the same hanging problem we sometimes see.

From the logs, this looks like a real hang running the reflection tests.

Did you locally test this change using the `Tests/runtests` script? It is believable there is a real lock problem here with the submission (or use of it in the code paths exercised by `runtests`).

j9brown:
Thanks!  We'll go take a closer look at the tests.

schwa423:
When running there is a call to GetGlobalLock() and ReleaseGlobalLock() before the call to InitGlobalLock() is made, so there is apparently a problem with the code paths exercised by the tests.

I'll try with the PTHREAD_MUTEX_INITIALIZER approach mentioned by @steve-lunarg, which I expect will fix the problem on Unix, although probably not the underlying problem.

schwa423:
Some more details. 

PTHREAD_MUTEX_INITIALIZER doesn't fix the problem, which results from two calls to GetGlobalLock() from the same thread without an intervening call to ReleaseGlobalLock().

I tried to use PTHREAD_RECURSIVE_MUTEX_INITIALIZER, but it doesn't compile on my machine (Ubuntu 14.04).  However, I was able to call pthread_mutexattr_settype() with PTHREAD_MUTEX_RECURSIVE.

This fixes the problem for me.  However, there still remains the issue that GetGlobalLock() and ReleaseGlobalLock() are called once before InitializeGlobalLock() is called.  I added https://github.com/KhronosGroup/glslang/issues/552 to track this.

schwa423:
Woo hoo!  Tests now pass on the server!

PR #542, opened 2016-10-08
HLSL: phase 2: add operator[]

steve-lunarg:
This PR is in four parts, and adds operator[] support to RWTextures and RWBuffers.

The PR has four separate commits as follows:

1. add r-value support
2. add l-value support
3. add HLSL specific lValueErrorCheck
4. Minor tweak to allow [] on non-rw textures.

This does not yet add non-vec4 image formats.  That will happen in phase 3.

There are a few known limitations, such as cascading unary postfix operators.  Most things work: fn outputs, read/modify/write opts like *= or unary++, and so on.

There is no image atomic support.  That will be a phase 4.


steve-lunarg:
Some small changes, new tests, comment improvements.  Removed WIP: ready for code review.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1955,32 +1955,31 @@ bool TParseContext::lValueErrorCheck(const TSourceLoc& loc, const char* op, TInt
...
+            error(loc, " l-value required", op, "", "");
+            return true;
+        }
     }
 
+    // Let the base class check errors
+    if (TParseContextBase::lValueErrorCheck(loc, op, node))
+        return true;
 
+    // GLSL specific
```
johnkslang: Remove this.  This whole file is GLSL specific.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1955,32 +1955,31 @@ bool TParseContext::lValueErrorCheck(const TSourceLoc& loc, const char* op, TInt
...
     const char* symbol = nullptr;
     TIntermSymbol* symNode = node->getAsSymbolNode();
     if (symNode != nullptr)
         symbol = symNode->getName().c_str();
 
     const char* message = nullptr;
     switch (node->getQualifier().storage) {
-    case EvqConst:          message = "can't modify a const";        break;
-    case EvqConstReadOnly:  message = "can't modify a const";        break;
     case EvqVaryingIn:      message = "can't modify shader input";   break;
```
johnkslang: Since this code is being revisited, and after it was written the method "isPipeInput" was written, I think now is the time to consider whether it should be used.

steve-lunarg:
Code review changes are in the most recent (phase 5e) commit, which adds lower level functions to create unary and binary nodes, and uses those from wherever the nodes were manually created before.  There's only one static place for each, now.  There are two flavors, one accepting a type, since many consumers of them want to provide a type.

PR #543, opened 2016-10-09
Address some compiler warnings.

slime73:
- Add explicit casts from long to int.
- Comment out method argument names that are unused.
- Always initialize a boolean variable before it's read.

### DIFF
--- a/glslang/Include/InfoSink.h
+++ b/glslang/Include/InfoSink.h
@@ -74,7 +74,6 @@ public:
     TInfoSinkBase& operator<<(const char* s)           { append(s); return *this; }
     TInfoSinkBase& operator<<(int n)                   { append(String(n)); return *this; }
     TInfoSinkBase& operator<<(unsigned int n)          { append(String(n)); return *this; }
-    TInfoSinkBase& operator<<(long unsigned int n)     { append(String((int)n)); return *this; }
     TInfoSinkBase& operator<<(float n)                 { const int size = 40; char buf[size]; 
                                                          snprintf(buf, size, (fabs(n) > 1e-8 && fabs(n) < 1e8) || n == 0.0f ? "%f" : "%g", n);
                                                          append(buf); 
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/ParseContextBase.cpp
```diff
@@ -320,8 +320,8 @@ bool TParseContextBase::insertGlobalUniformBlock()
     if (globalUniformBlock == nullptr)
         return true;
 
-    int numMembers = globalUniformBlock->getType().getStruct()->size();
-    bool inserted;
+    int numMembers = (int)globalUniformBlock->getType().getStruct()->size();
+    bool inserted = false;
```
johnkslang: This is clearly set in both paths; it seems something is wrong with your compiler if it thinks this is unset?

It carries slightly more information to leave it out, to indicate expectation of the subsequent path to set it, and in the general case, possibly saves performance (that is, I wouldn't want a general rule that all variables are initialized, but we've been slowly going down that path due to this warning). We have done many of these, but I want to stop losing these two things just because some compilers erroneously report something.

glslang/Include/InfoSink.h
```diff
@@ -74,7 +74,7 @@ class TInfoSinkBase {
     TInfoSinkBase& operator<<(const char* s)           { append(s); return *this; }
     TInfoSinkBase& operator<<(int n)                   { append(String(n)); return *this; }
     TInfoSinkBase& operator<<(unsigned int n)          { append(String(n)); return *this; }
-    TInfoSinkBase& operator<<(long unsigned int n)     { append(String(n)); return *this; }
+    TInfoSinkBase& operator<<(long unsigned int n)     { append(String((int)n)); return *this; }
```
johnkslang: It would be better to either delete this definition, or add the corresponding String() function.

glslang/MachineIndependent/ParseContextBase.cpp
slime73: > This is clearly set in both paths; it seems something is wrong with your compiler if it thinks this is unset?

There are three paths: `firstNewMember == 0`, `firstNewMember <= numMembers`, and `firstNewMember > numMembers`. The third isn't handled in the if/else if section and hopefully won't happen in practice (I haven't looked particularly hard), but modern Clang versions catch it as a warning.

PR #544, opened 2016-10-09
SPV: Implement extension SPV_KHR_shader_draw_parameters.

amdrexu:


SPIRV/spirv.hpp
```diff
@@ -1,913 +1,924 @@
...
-inline FunctionControlMask operator|(FunctionControlMask a, FunctionControlMask b) { return FunctionControlMask(unsigned(a) | unsigned(b)); }
-inline MemorySemanticsMask operator|(MemorySemanticsMask a, MemorySemanticsMask b) { return MemorySemanticsMask(unsigned(a) | unsigned(b)); }
-inline MemoryAccessMask operator|(MemoryAccessMask a, MemoryAccessMask b) { return MemoryAccessMask(unsigned(a) | unsigned(b)); }
-inline KernelProfilingInfoMask operator|(KernelProfilingInfoMask a, KernelProfilingInfoMask b) { return KernelProfilingInfoMask(unsigned(a) | unsigned(b)); }
-
-}  // end namespace spv
-
-#endif  // #ifndef spirv_HPP
-
+// Copyright (c) 2014-2016 The Khronos Group Inc.
```
johnkslang: Is there a line ending problem that makes this entire file be different? We should find root cause of that and fix it.

amdrexu: Yes, the problem is caused by line ending.  I have resolved it. Thank you.

PR #555, opened 2016-10-18
HLSL: Fix unary and binary operator type conversion issues

steve-lunarg:
This fixes defects as follows:

1. handleLvalue could be called on a non-L-value, and it shouldn't be.

2. HLSL allows unary negation on non-bool values.  TUnaryOperator::promote
   can now promote other types (e.g, int, float) to bool for this op.

3. HLSL allows binary logical operations (&&, ||) on arbitrary types, similar
   (2).

4. HLSL allows mod operation on arbitrary types, which will be promoted.
   E.g, int % float -> float % float.


glslang/Include/intermediate.h
```diff
@@ -845,7 +847,7 @@ class TIntermOperator : public TIntermTyped {
     virtual       TIntermOperator* getAsOperator()       { return this; }
     virtual const TIntermOperator* getAsOperator() const { return this; }
     TOperator getOp() const { return op; }
-    virtual bool promote() { return true; }
+    virtual bool promote(TIntermediate&) { return true; }
```
steve-lunarg: Yeah, agreed.  This design issue was also why I had left this as a WIP for now - it didn't feel quite right as it stands above.  I'll make that change and re-push.

PR #560, opened 2016-10-20
silence -Wunknown-pragmas warnings on gcc

jeremy-lunarg:
Unfortunately, suppress only has a one line budget which gets consumed by
the "#endif" so I had to resort to something a bit more bloated.

### DIFF
--- a/StandAlone/ResourceLimits.cpp
+++ b/StandAlone/ResourceLimits.cpp
@@ -35,6 +35,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <sstream>
+#include <cctype>
 
 #include "ResourceLimits.h"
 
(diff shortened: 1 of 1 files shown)


johnkslang:
I don't quite understand this. There are other places a similar #pragma is used, without a one-line limit. Are they incorrect?  There are other alternatives too, like fixing the original problem (possible?) or using the #define _CRT_SECURE_NO_WARNINGS?

Just suggesting; I haven't investigated enough to know what's best yet.  But, maybe this leads to something that works better.

steve-lunarg:
I dunno about other uses, but assuming the warnings are about the strtok, what about something [akin to this](https://github.com/steve-lunarg/glslang/tree/resource-fix)?  Might be a better phrasing possible, but that idea anyway.

johnkslang:
@steve-lunarg: sure, if that works.  Presumably the author was saving the time of doing this extra work, but now that you've done it, I think it is fine.

jeremy-lunarg:
Yeah, +1 for Steve's version. Mine was just trying to fix duck tape anyway.

steve-lunarg:
You're welcome to snag the file.  I tried it against a .conf file I made with the -c option, but it could probably use a once-over.

ChrisGautier:
Why not simply guard the three ``#pragma`` with ``#ifdef _MSC_VER`` ? That would be consistent with the rest of the code.

johnkslang:
> Why not simply guard the three ``#pragma`` with ``#ifdef _MSC_VER`` ?

Because it's the `suppress` form and not the `disable` form:

> suppress only has a one line budget which gets consumed by the "#endif"

The following form works for the rest of the file:

```
#pragma warning(disable: 4996)
```

Alternatively, the C++ standard is made for #pragma to be ignored when not recognized, so one could also argue the fix belongs with the compiler that's complaining.

So, tons of solutions here: use 'disable', fix the complaining compiler, or use the code that side steps it (current WIP).

I don't think we need further debate.

Your call.  Make sure the PR is correct, remove the WIP, and I'll merge it in.  Thanks.

PR #561, opened 2016-10-20
WIP: apply unused variable

jeremy-lunarg:
I happened upon numArgs while hunting for unused variables. I suspect
the intent was to apply it as shown in this patch. However, I am not a
compiler dude. Someone more appropriate should grok this change.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -1216,7 +1216,7 @@ void TParseContext::computeBuiltinPrecisions(TIntermTyped& node, const TFunction
                                         function.getType().getQualifier().precision;
     } else if (TIntermAggregate* agg = node.getAsAggregate()) {
         TIntermSequence& sequence = agg->getSequence();
-        int numArgs = (int)sequence.size();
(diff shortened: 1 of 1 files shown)


jeremy-lunarg:
The appveyor error is a mystery to me...

johnkslang:
It is true that the set of arguments to look at depends on the operation, so something like your suggested change is needed.  It should be done with matching "unsigned" (or signed if that's better) types though, and verified it is the right subset of arguments being checked.

The bot failures look like something other than your change. Could be if you tweak for signedness and resubmit, it will go through (probably making numArgs unsigned, but I think either way would work).

jeremy-lunarg:
Whoops. I'll fix the signed-ness tomorrow morning. Irony. :)

steve-lunarg:
FYI, I checked the numArgs values which are now used from the opcode switch.  I believe they match the described operation of the function, and also make intuitive sense e.g, in that field locations do not effect the precision of the bitfield ops.

I have not verified that this is the entire such list.

jeremy-lunarg:
John, if you're comfortable with Steve's assessment, then I'll remove the WIP.

johnkslang:
It looks correct.

PR #562, opened 2016-10-21
HLSL -> Spir-V: Resource mapping handler

TiemoJung:
Adds a resource mapping handler.

If not provided it behaves as before.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -61,177 +61,278 @@
 //    c. implicit dead bindings are left un-bound.
 //
 
-   
+
 namespace glslang {
 
-// Map of IDs to bindings
-typedef std::unordered_map<unsigned int, int> TBindingMap;
-typedef std::unordered_set<int> TUsedBindings;
+struct TVarEntryInfo
+{
+    int             id;
+    TIntermSymbol*  symbol;
+    bool            live;
+    int             newBinding;
+    int             newSet;
 
-
-// This traverses the AST to determine which bindings are used, and which are implicit
-// (for subsequent auto-numbering)
-class TBindingTraverser : public TLiveTraverser {
-public:
-    TBindingTraverser(const TIntermediate& i, TBindingMap& bindingMap, TUsedBindings& usedBindings,
-                      bool traverseDeadCode = false) :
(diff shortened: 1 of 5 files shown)


steve-lunarg:
The idea of a generic mapping handler seems good to me.

I haven't digested the algorithmic details, but would it be easy / possible to represent the existing behavior (which is in use by some people) in terms of the TIoMapResolver, rather than having it branch off down a different path?

TiemoJung:
@steve-lunarg Its pretty much done, please take a look.

gtests/TestFixture.h
```diff
@@ -246,19 +358,22 @@ class GlslangTest : public GT {
...
-        
+
         success &= program.link(controls);
-        success &= program.mapIO();
+
+        TDefaultIoResolver resolver;
+        resolver.baseSamplerBinding = baseSamplerBinding;
+        resolver.baseTextureBinding = baseTextureBinding;
+        resolver.baseUboBinding = baseUboBinding;
+        resolver.autoMapBindings = autoMapBindings;
```
steve-lunarg: The default resolver approach looks good, but can it still get the base bindings from the same place, rather than removing the setShift*Binding() APIs?  I believe there are external consumers (outside the glslang project) of that API, so it shouldn't go away.

glslang/MachineIndependent/iomapper.cpp
```diff
@@ -61,177 +61,200 @@
...
+        // 3) has no binding but set
+        // 4) has no binding and no set
+        inline bool operator()(const TVarEntryInfo& l, const TVarEntryInfo& r)
+        {
+            const TQualifier& lq = l.symbol->getQualifier();
+            const TQualifier& rq = r.symbol->getQualifier();
+
+            if (lq.hasBinding())
+            {
+                if (rq.hasBinding())
```
steve-lunarg: Code style: the rest of the code base has curleys on the same line as the if.

TiemoJung:
@steve-lunarg Addressed your remarks and restored the old interface. If no resolver is provided, then it uses an internal default resolver.

johnkslang:
@steve-lunarg do you have any pending comments/issues here?

steve-lunarg:
No; looks like the original API is preserved now, and both sets of behaviors are accomplished with mapping handlers, so I'm happy with the external interface.  Seems like a good design.

PR #563, opened 2016-10-21
SPV: Implement the extension SPV_NV_sample_mask_override_coverage

chaoc:


SPIRV/disassemble.cpp
```diff
@@ -480,6 +495,11 @@ void SpirvStream::disassembleInstruction(Id resultId, Id /*typeId*/, Op opCode,
                 } else if (extInstSet == GLSLextAMDInst) {
                     out << "(" << GLSLextAMDGetDebugNames(name, entrypoint) << ")";
 #endif
+#ifdef NV_EXTENSIONS
+                }
```
johnkslang: use `} else ... {` to match existing style

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4135,6 +4146,13 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
                 error(loc, "unknown blend equation", "blend_support", "");
             return;
         }
+#ifdef NV_EXTENSIONS 
+        if (id == "override_coverage") {
+            requireExtensions(loc, 1, &E_GL_NV_sample_mask_override_coverage, "sample mask override coverage");
+            publicType.qualifier.layoutOverrideCoverage = true;
```
kholtnv: My read is that it is "per variable":

layout(override_coverage) out int gl_SampleMask[];

SPIRV/disassemble.cpp
chaoc: Thanks for the comment. Fixed accordingly

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: "Per variable" means we need lots of bits, one per variable.

So far, my understanding is this needs just one bit, total for the stage. Like origin_upper_left, or vertex spacing.  If so, this should go into `TShaderQualifiers` and transferred to `TIntermediate`, rather than a bool for every node in the AST.

chaoc: Hi John, I have modified according to your request

chaoc:
Hi John. I have modified the implementation to per stage as requested. Could you please take another look?

Thanks

johnkslang:
Looking pretty good. I had one comment. I can merge after you address that and remove WIP.

Just asking side questions here: What about the SPIR-V tools, like the disassembler, validator, etc.?

chaoc:
Hi John

Thanks for the review. This might sound a bit stupid, but I couldn’t find the comment you suggest me to address. Do you mind pointing it out for me?

As for the SPIR-V tools. I’ll get to them ASAP.

Thanks

Chao

johnkslang:
I had to press one more button for the review to show up. Hopefully, you see the comment now.

glslang/Include/Types.h
```diff
@@ -939,6 +943,7 @@ struct TShaderQualifiers {
         earlyFragmentTests = false;
         layoutDepth = EldNone;
         blendEquation = false;
+        layoutOverrideCoverage = false;
```
johnkslang: I think if the declaration is under `#ifdef NV_EXTENSIONS`, then all uses must be too.

johnkslang:
One final request: Can we get this down to 1 or 2 (maybe 3?) commits that are actually about what point is? 26 commits for a reasonable simple addition doesn't make sense and obscures what's going on.

chaoc:
Hi John

I was trying really hard to squash the commit. But it always seem to pick up the merge remote from Khoronos. After rebase, all the merge comes out as standalone commit which made the pull request even bigger. If you are not okay with the current commit number, I think it might be best for us to close this pull request and open another one?

Thanks
Chao

chaoc:
Hi John

Could you let me know if you are okay with the current commit number or should I open another pull request?

Thanks

Chao

From: Chao Chen
Sent: Wednesday, December 14, 2016 5:21 PM
To: 'KhronosGroup/glslang'
Subject: RE: [KhronosGroup/glslang] SPV: Implement the extension SPV_NV_sample_mask_override_coverage (#563)

Hi John

I was trying really hard to squash the commit. But it always seem to pick up the merge remote from Khoronos. After rebase, all the merge comes out as standalone commit which made the pull request even bigger. If you are not okay with the current commit number, I think it might be best for us to close this pull request and open another one?

Thanks
Chao

chaoc:
Rebased all change to one commit as requested.

johnkslang:
Thanks for investing in that. However, I now see 29 commits, rather than the preferred, say, 1 to 3.

chaoc:
Let's close this pull request and open another new one to address the commit number issues

PR #565, opened 2016-10-23
Disabled MSVC 2015 compile warnings.

mre4ce:


danilaml:
Wouldn't this PR disable those warnings for all projects including those files?
Why not use `push`/`pop` https://msdn.microsoft.com/library/2c8f766e.aspx ?

PR #570, opened 2016-10-27
HLSL: Add min*{float,int,uint} types

steve-lunarg:
These HLSL types are guaranteed to have at least the given number of bits, but may have more.

min{16,10}float is mapped to EbtFloat16 with medium precision -> SPIR-V RelaxedPrecision.
min{16,12}int and min16uint are mapped to mediump -> SPIR-V RelaxedPrecision


oscarbg:
so this maps to use VK_AMD_gpu_shader_half_float which adds Float16 support on AMD cards?

johnkslang:
Today, we need to map `min16` to SPIR-V 32-bit types with **RelaxedPrecision**.

steve-lunarg:
Changed to relaxed precision float32 for min float types, removed WIP.

hlsl/hlslGrammar.cpp
```diff
@@ -1118,6 +1126,91 @@ bool HlslGrammar::acceptType(TType& type)
...
+        break;
+    case EHTokMin16float3:
+        new(&type) TType(min16float_bt, EvqTemporary, EpqMedium, 3);
+        break;
+    case EHTokMin16float4:
+        new(&type) TType(min16float_bt, EvqTemporary, EpqMedium, 4);
+        break;
+        
+    case EHTokMin10float:
+        new(&type) TType(min10float_bt, EvqTemporary, EpqLow); // equiv to Medium when generating SPIR-V:
```
johnkslang: Did you verify that **lowp** is sufficient for `min10`?

(It may be irrelevant if this is only for SPIR-V, but then, if that's the case **mediump** would work everywhere without needing to know if **lowp** is okay.)

steve-lunarg:
I didn't find an explicit precision for lowp (nor mediump, except for how it maps to the SPIR-V RelaxedPrecision decoration).  I changed the PR to use mediump for safety, for all the new types.

PR #571, opened 2016-10-28
Add explicit lambda return types, for compilers without C++14 support

baldurk:
In the same vein as #548, VS2010 in particular needs an explicit return type declared for lambdas with multiple statements, as that was added in C++14.


steve-lunarg:
Looks good to me.  Those were my lambdas - sorry about that, I hadn't known until just this moment that only the trivial return case was handled before 14.  Some compilers seem to do the deduction anyway.

PR #577, opened 2016-11-02
HLSL: add intrinsic implicit promotions

steve-lunarg:
HLSL: add intrinsic function implicit promotions

This PR handles implicit promotions for intrinsics when there is no exact match, such as for example clamp(int, bool, float).  In this case the int and bool will be promoted to a float, and the clamp(float, float, float) form used.

These promotions can be mixed with shape conversions, e.g, clamp(int, bool2, float2).

Output conversions are handled either via the existing addOutputArgumentConversion function, which this PR generalizes to handle either aggregates or unaries, or by intrinsic decomposition.  If there are methods or intrinsics to be decomposed, then decomposition is responsible for any output  onversions, which turns out to happen automatically in all current cases.  This can be revisited once inout conversions are in place.

Some cases of actual ambiguity were fixed in several tests, e.g, spv.register.autoassign.*

Some intrinsics with only uint versions were expanded to signed ints natively, where the underlying AST and SPIR-V supports that.  E.g, countbits.  This avoids extraneous conversions.

A new function promoteAggregate is added, and used by findFunction.  This is essentially a generalization of the "promote 1st or 2nd arg" algorithm in promoteBinary.

The actual selection proceeds in three steps, as described in the comments in hlslParseContext::findFunction:

1. Attempt an exact match.  If found, use it.
2. If not, obtain the operator from step 1, and promote arguments.
3. Re-select the intrinsic overload from the results of step 2.


steve-lunarg:
Repush, still WIP.  Changed to a three step resolution process for finding builtin functions, as described in comments in HlslParseContext::findFunction L4398...

steve-lunarg:
Updated, and remove WIP.  Some complex stuff, needs review.

steve-lunarg:
Re-pushed after merge with recent commits.  No functional change since last push.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -901,13 +926,17 @@ bool TIntermediate::canImplicitlyPromote(TBasicType from, TBasicType to, TOperat
             return version >= 400;
         case EbtUint:
             return true;
+        case EbtBool:
```
johnkslang: What about going the opposite direction?  Is that allowed?

hlsl/hlslParseHelper.cpp
```diff
@@ -2660,9 +2666,20 @@ void HlslParseContext::addInputArgumentConversions(const TFunction& function, TI
 //
 // Returns a node of a subtree that evaluates to the return value of the function.
 //
-TIntermTyped* HlslParseContext::addOutputArgumentConversions(const TFunction& function, TIntermAggregate& intermNode)
+TIntermTyped* HlslParseContext::addOutputArgumentConversions(const TFunction& function, TIntermOperator& intermNode)
 {
-    TIntermSequence& arguments = intermNode.getSequence();
+    if (intermNode.getAsAggregate() == nullptr && intermNode.getAsUnaryNode() == nullptr)
```
johnkslang: Shouldn't this expression be the assert?

```assert(intermNode.getAsAggregate() != nullptr || ...)```

steve-lunarg: Heh - yes; I'll change.  Still thinking about the other review comment...

glslang/MachineIndependent/Intermediate.cpp
steve-lunarg: About the other direction:

I tried some experiments, and HLSL does allow for example passing a float to an intrinsic such as countbits(uint) which has only integer overloads, which is the "wrong" direction.  We do handle such a case correctly with this PR, because it's passed through as an EOpFunctionCall in findFunction, which allows any promotion for HLSL.  The "best" signature selection then selects the uint signature, which is the closest match, and input arg conversion adds the float->uint conversions.

However, the current behavior of findFunction treating all builtins like EOpFunctionCall does lose data, denying canImplicitlyPromote the ability to behave differently based on the builtin op.  I have an in-progress change to address that, which is needed to for example avoid trying to promote the underlying type of a buffer or texture object.

steve-lunarg:
Moving back to WIP while sorting out merge conflicts.

steve-lunarg:
I have added a new test hlsl.intrinsics.promote.down.frag for e.g, calling a uint-accepting intrinsic with a float value.  This verifies that we perform a float->uint conversion on the input argument, and then select the uint function.

The algorithm is unchanged from the description above.  The promotion step (using the existing promotion facilities) seems awkward though, since the incoming arg list is different for unary builtin fns vs others.  In the unary case, it points directly to the single argument, so there are special cases.  Open to better ideas here...

zeux:
Any news on merging this? FWIW we are trying to compile ~100 KB HLSL shaders, and after this change we need no changes to successfully build the codebase to SPIR-V, but without this there are a lot of conversions like max(v, 0) (with float v) or dot(v, 1) (with float3 v) that fail to compile and require manual casts.

PR #578, opened 2016-11-04
WIP: allow renaming of shader entry point when creating SPIR-V

steve-lunarg:
Use "--rename-entrypoint altname" on the command line, or the TShader::setEntryPointAltName(char*) API.  This is API agnostic.

A new test is added for this ability.


gtests/TestFixture.h
```diff
@@ -229,6 +229,48 @@ class GlslangTest : public GT {
         }
     }
 
+    // Compiles and links the given source |code| of the given shader,
+    // and renames the entry point to entryPointNameAlt.
+    // Returns a GlslangResult instance containing all the information generated
+    // during the process. If the target includes SPIR-V, also disassembles
+    // the result and returns disassembly text.
+    GlslangResult compileLinkAndRename(
```
johnkslang: FYI, we still use `runtests` for one-off tests, as it's way easier just to add one line in `runtests` than invest in all this.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -167,9 +167,17 @@ class TIntermediate {
     EShSource getSource() const { return source; }
     void setEntryPointName(const char* ep) { entryPointName = ep; }
     void setEntryPointMangledName(const char* ep) { entryPointMangledName = ep; }
+    void renameEntryPoint(const char* newName) { newEntryPointName = newName; }
     const std::string& getEntryPointName() const { return entryPointName; }
     const std::string& getEntryPointMangledName() const { return entryPointMangledName; }
 
+    // The entry point can be renamed from the shader text to an alternate name
```
johnkslang: It seems this isn't an alternate name, but a new name.  Can we have fewer synonyms for this functionality (so far have seen rename, new name, alt name).

More seriously, I think we should be clear whether the functionality is

A) change the name of the entry point

or 

B) leave the entry-point name the same, but have the SPIR-V call it something else

Where I'd think A would have a much simpler implementation, unless there is fear of the name having conflicts with other names?

steve-lunarg: Ah, yeah, I meant to rename all of those, but one slipped through.  I'll fix and repush.  It should be as follows:  renameEntryPoint as the verb, and newEntryPointName as the noun.

There are two names tracked because (for example) detection of whether a function is the entry point must still be based on name A, as that's what's given in the shader text.  However, the SPIR-V module is produced with an entry point called B, instead of A.  The SPIR-V will have no mention of name A.

steve-lunarg: Also, to capture the info: currently this renaming happens late, during AST generation.  However, it could be made to happen early, in the grammar.  That will be simpler, but is HLSL-specific unless added explicitly for other languages.  You'd suggested that is OK, so absent other objections, I'll change the point in time for the renaming.

PR #579, opened 2016-11-07
Fix variable shadowing warnings.

null77:
These were showing up in a Visual Studio compile.

Also curious if there's a way to enable specific warnings for the Visual Studio build in CMake. These were caught using a different build system.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -1269,19 +1269,19 @@ bool TGlslangToSpvTraverser::visitUnary(glslang::TVisit /* visit */, glslang::TI
             else
                 op = glslang::EOpSub;
 
-            spv::Id result = createBinaryOperation(op, precision,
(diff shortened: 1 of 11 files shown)


johnkslang:
Can you clarify what the actual issue is?

null77:
Sure, variable shadowing: https://en.wikipedia.org/wiki/Variable_shadowing

This was showing up as a compiler warning. I don't have the exact text on hand. It's generally agreed to not be a great programming practice, and can lead to odd situations (ie re-declaring an inner loop variable "i", then needing to access the outer variable).

johnkslang:
This is a key feature of C++ to enable multiple authors to not have to know everything in order to add something.  With the proposed changes now, an author could now modify code accidentally using the more natural original name, and get the one from the wrong scope.

I could see doing something like this is some confusing cases, but would want to more carefully craft the names (e.g., the outer name should be longer and more descriptive than the local name).

null77:
I agree the names aren't great. Unfortunately I won't work on this.

Issue #581, opened 2016-11-09
barrier should synchronize across WorkGroup, not Device


ChrisGautier:
The ESSL and GLSL specs say this about `barrier()`:

> For any given static instance of barrier(), all tessellation control shader
> invocations for a single input patch must enter it before any will be
> allowed to continue beyond it, or all invocations for a single work group
> must enter it before any will continue beyond it.

The Vulkan specifications also restrict the scope for execution:
> Scope for execution must be limited to:
>    **Workgroup**
>    **Subgroup**

But glslangValidator uses **Device** when converting calls to the ``barrier`` built-in function.

```
          %7 = OpConstant %uint 1
          %8 = OpConstant %uint 0
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpControlBarrier %7 %7 %8
```

```
case glslang::EOpBarrier:
   builder.createControlBarrier(spv::ScopeDevice, spv::ScopeDevice, ...);
```

(Note: #502 discusses the memory-semantics operand)

ChrisGautier:
Closing as duplicate of updated #592.

PR #582, opened 2016-11-09
The execution scope for barriers should be Workgroup.

ChrisGautier:
See issue #581


johnkslang:
A PR like this needs to include the test changes.  You can also just use an issue to request the change for someone else to do a complete PR, if you wish.

PR #587, opened 2016-11-11
WIP: HLSL: use hlsl parser for intrinsic protos, add float1/etc intrinsics

steve-lunarg:
This PR switches HLSL intrinsic parsing over from the GLSL parser to the HLSL parser, and enables vec1 intrinsic prototypes.  It is in three commits, to isolate test changes:

step 1: Add CreateParseContext function to get source-specific parse context
step 2: throw big switch to use HLSL parser for intrinsics.
step 3: add vec1 intrinsic tests

See comments in each commit for details.


steve-lunarg:
This will likely have a minor conflict with #577.  Whichever one goes in first, I'll fix the other one and re-push.

steve-lunarg:
Changing this to WIP: it's fine by itself, but together with pending PR #577 there is an ambiguity around float fs float1 prototypes under implicit promotion.  I'll fix that at the same time as resolving the test result conflicts.

PR #596, opened 2016-11-16
HLSL: use HLSL parser for HLSL intrinsic prototypes, enable int/bool mats

steve-lunarg:
This PR adds a CreateParseContext() fn analogous to CreateBuiltInParseables(), to create a language specific built in parser.  (This code was present before but not encapsualted in a fn).  This can now be used to create a source language specific parser for builtins.

Along with this, the code creating HLSL intrinsic prototypes can now produce them in HLSL syntax, rather than GLSL syntax.  There was some existing code for this, but it had never been enabled, so had a few problems which this PR fixes.  This relaxes certain prior restrictions at the parser level.  Lower layers (e.g, SPIR-V) may still have similar restrictions: this code does not impact that.

This PR also fleshes out matrix types for bools and ints, both of which were partially in place before.   This was easier than maintaining the restrictions in the HLSL prototype generator to avoid creating protoypes with those types.

Many tests change because the result type from intrinsics moves from "global" to "temp".

Several new tests are added for the new types.


Test/hlsl.matType.bool.frag
```diff
@@ -0,0 +1,53 @@
...
+    bool2x3 b2x3;
+    bool3x3 b3x3;
+    bool4x3 b4x3;
+    
+    bool1x4 b1x4;
+    bool2x4 b2x4;
+    bool3x4 b3x4;
+    bool4x4 b4x4;
+    
+    // TODO: Currently SPIR-V disallows Nx1 or 1xN mats.
```
johnkslang: That is likely to continue (SPIR-V not having these types). What would the action to take be?  E.g., is the todo to provide appropriate lowers of these operations to native SPIR-V types?  GLSL also does not have these types.

steve-lunarg: I wasn't sure of the direction this would go. They could be lowered to existing SPIR-V if that was reasonable to do, or if not, there has been talk about an extension to SPIR-V to allow easier mapping of some HLSL features.

johnkslang: I think an HLSL extension still wouldn't add degenerate types, so these would still need to be translated.

PR #599, opened 2016-11-18
HLSL: Add GS support

steve-lunarg:
This PR adds:

- [maxvertexcount(n)] attributes

- point/line/triangle/lineadj/triangleadj qualifiers

- PointStream/LineStream/TriangleStream templatized types

- Append method on above template types

- RestartStrip method on above template types.


hlsl/hlslParseHelper.cpp
```diff
@@ -4284,6 +4348,16 @@ const TFunction* HlslParseContext::findFunction(const TSourceLoc& loc, const TFu
         return nullptr;
     }
 
+    // These are special cased because they can accept any type.
```
johnkslang: I think this can be moved at least after finding an exact match, and possibly deeper.

steve-lunarg: That seems better... moved it down, which no longer requires a string match, nor an additional symtab fetch.

Also verified the AST as far as possible.  Will remove WIP shortly.

Issue #588, opened 2016-11-11
HLSL with wrong entry point generates code that SPIRV-Cross crashes on


hrydgard:
The following HLSL:
```
float4 PixelShaderFunction(float4 input) : COLOR0   {
    return float4(0.0, 1.0, 0.0, 2.0);
}
```
with setEntryPoint("main") generates the following SPIR-V:
```
                              Capability Shader
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main"
                              ExecutionMode 4 OriginUpperLeft
                              Name 4  "main"
                              Name 11  "PixelShaderFunction(vf4;"
                              Name 10  "input"
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 4
               8:             TypePointer Function 7(fvec4)
               9:             TypeFunction 7(fvec4) 8(ptr)
              13:    6(float) Constant 0
              14:    6(float) Constant 1065353216
              15:    6(float) Constant 1073741824
              16:    7(fvec4) ConstantComposite 13 14 13 15
         4(main):           2 Function None 3
               5:             Label
                              FunctionEnd
11(PixelShaderFunction(vf4;):    7(fvec4) Function None 9
       10(input):      8(ptr) FunctionParameter
              12:             Label
                              ReturnValue 16
                              FunctionEnd
```
This is invalid code, caused by the mismatched entry point name. Surely glslang should not generate invalid SPIR-V.

(Oh yeah, in addition, the return type should be SV_Target, too, not COLOR0, but I don't think that has anything to do with the issue).

hrydgard:
Argh, never mind, the entry point name "main" didn't match "PixelShaderFunction". Doh.

Anyway, SPIRV-Cross should probably output some nicer error, but that's a different issue.

hrydgard:
Reopening this, because glslang should clearly throw an error here instead of proceeding to generate invalid SPIR-V. See https://github.com/KhronosGroup/SPIRV-Cross/issues/68 .

hrydgard:
Pared the issue report down to the interesting part.

dneto0:
I've experienced this too.

johnkslang:
> glslang should clearly throw an error here instead of proceeding to generate invalid SPIR-V

[...]

PR #600, opened 2016-11-20
Fix a wrong entry point generates code

durswd:
I fixed a bug.
https://github.com/KhronosGroup/glslang/issues/588

The cause of bag is that builder.leaveFunction is not called before calling dumpSpv.

In addition, I found that some entry points in tests don't match shaders.
So I make names of entry point match between tests and shaders.


SPIRV/GlslangToSpv.cpp
```diff
@@ -906,17 +906,18 @@ void TGlslangToSpvTraverser::dumpSpv(std::vector<unsigned int>& out)
     for (auto it = iOSet.cbegin(); it != iOSet.cend(); ++it)
         entryPoint->addIdOperand(*it);
 
+	if (!mainTerminated) {
```
johnkslang: Spaces turned to tabs, causing incorrect indentation. Glslang uses 4 spaces per indent, and never tabs.

SPIRV/GlslangToSpv.cpp
```diff
@@ -906,17 +906,18 @@ void TGlslangToSpvTraverser::dumpSpv(std::vector<unsigned int>& out)
...
+		builder.setBuildPoint(lastMainBlock);
+		builder.leaveFunction();
+	}
+
     builder.eliminateDeadDecorations();
     builder.dump(out);
 }
 
 TGlslangToSpvTraverser::~TGlslangToSpvTraverser()
 {
```
johnkslang: There is a design problem here (not yours, already existing). Completing the translation and dumping the SPIR-V should be two operations, not mixed together.

Test/baseResults/hlsl.array.implicit-size.frag.out
```diff
@@ -30,7 +30,7 @@ gl_FragCoord origin is upper left
 0:12        2.000000
 0:12        3 (const int)
 0:12        4.000000
-0:26  Function Definition: main(struct-PS_OUTPUT-vf41; (temp void)
+0:26  Function Definition: PixelShaderFunction(struct-PS_OUTPUT-vf41; (temp void)
```
johnkslang: It is possibly useful to have some tests verifying that this change works correctly, but fixing all the tests bypasses that.  Not sure what's best here, but something in the range of leaving tests alone, or having one specific test testing this, or fixing all but one test.

johnkslang:
@durswd good find, thanks. Between needed two fixes for bug #588 and the review comments here that would need to be addressed, esp. the design aspect, I did this a different way in fca826212cddbef03d69c4d7ab3a3efa5e2c81e7 plus giving the warning in e50dc536ff76310bd24fa222397cca9d0b6acacf.

Thanks.

johnkslang:
Forgot to say:  I did not change the nature of the tests, but the ones that did have a missing entry point now show the **OpReturn** getting correctly added, which has some value. I'd be happy to see a PR that more cleanly fixes the tests as you did, as long as there is at least one test that tests these changes.

PR #604, opened 2016-11-30
HLSL: allow renaming of shader entry point when creating SPIR-V

steve-lunarg:
Use "--source-entrypoint name" on the command line, or the TShader::setSourceEntryPoint(char*) API.

When the name given to the above interfaces is detected in the shader source, it will be renamed to the entry point name supplied to the -e option or the TShader::setEntryPoint() method.

The motivation for this change is that Vulkan requires the shader entry point name both at compile time and runtime, as opposed to just at compile time.  Existing shaders may be using names not known by the API.


hlsl/hlslParseHelper.cpp
```diff
@@ -5539,4 +5539,16 @@ TIntermNode* HlslParseContext::addSwitch(const TSourceLoc& loc, TIntermTyped* ex
     return switchNode;
 }
 
+// Rename entry point, if there is a new name.
+void HlslParseContext::renameEntryPoint(TString*& name) const
+{
+    // Replace the entry point name with the new entry point name, if there is one.
+    if (name->compare(intermediate.getEntryPointName().c_str())) {
```
johnkslang: Isn't compare returning an integer that means something, not a bool? Better to make it explicit.

StandAlone/StandAlone.cpp
```diff
@@ -547,6 +557,8 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
         shader->setStringsWithLengthsAndNames(compUnit.text, NULL, compUnit.fileNameList, 1);
         if (entryPointName) // HLSL todo: this needs to be tracked per compUnits
             shader->setEntryPoint(entryPointName);
+        if (newEntryPointName)
+            shader->renameEntryPoint(newEntryPointName);
```
johnkslang: I'd rather this just keep the one setEntryPoint() call that sets the already known correct entry point name, not using TIntermediate as a storage area for managing parsing options.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -168,9 +168,14 @@ class TIntermediate {
     EShSource getSource() const { return source; }
     void setEntryPointName(const char* ep) { entryPointName = ep; }
     void setEntryPointMangledName(const char* ep) { entryPointMangledName = ep; }
+    void renameEntryPoint(const char* newName) { newEntryPointName = newName; }
     const std::string& getEntryPointName() const { return entryPointName; }
     const std::string& getEntryPointMangledName() const { return entryPointMangledName; }
 
+    // The entry point can be renamed from the shader text to a new name
+    // provided in the API or command line.
+    const std::string& getNewEntryPointName() const { return newEntryPointName; }
```
johnkslang: If we must use TIntermediate to manage parsing options, call this getEntryPointRename(), so we don't have two different names floating around for the same concept.

(Or, use setEntryPointNewName() and getEntryPointNewName().  It's not like 'renameEntryPoint' actually renamed the entry point, so that's kind of misleading.)

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -1523,6 +1523,11 @@ void TShader::setEntryPoint(const char* entryPoint)
     intermediate->setEntryPointName(entryPoint);
 }
 
+void TShader::renameEntryPoint(const char* newName)
+{
+    intermediate->renameEntryPoint(newName);
```
johnkslang: Design: It makes most sense, design-wise, to keep this in the shader with the parse strings, and moved around with the parse strings, into the parse context, because it is during parsing that the parsed name will turn into this name.  TIntermediate never has to hold anything other than the correct (final) entry point name, and everything would work.

Unfortunately, I think that means adding an extra parameter to ProcessDeferred(), because it serves two interfaces, C style.  Generally, though, it is the contents of TShader being passed into ProcessDeferred(), so this all makes sense design wise.  Probably better (beyond the scope of this check in) to pass in a struct of this TShader info. Generally, this all lines up being it is more correct  to add a parameter to ProcessDeferred().  This was done somewhat recently with 'includer' to manage parse-time include policy too.

hlsl/hlslParseHelper.cpp
johnkslang: However, if the design I want to switch to is used, the new name will simply be a TString in the parse helper, and this will be ==.

hlsl/hlslParseHelper.cpp
```diff
@@ -5539,4 +5539,16 @@ TIntermNode* HlslParseContext::addSwitch(const TSourceLoc& loc, TIntermTyped* ex
...
+        if (!intermediate.getNewEntryPointName().empty()) {
+            name = new TString(intermediate.getNewEntryPointName().c_str());
+            intermediate.setEntryPointName(intermediate.getNewEntryPointName().c_str());
```
johnkslang: What happens to the intermediate new name now? Should it get unset? It's not the new name anymore. 

Note the TIntermediate gets consumed long after all this parsing is done, and should be holding the current state of the intermediate representation, so really shouldn't be tracking a new name in it's interface.

johnkslang:
Overall, I think this is a good improvement to pull earlier the rename, and be consistent it is a rename (or be consistent it is a new name).

However, design wise, I'd like to get this into the TShader and THlslParseContext, and not be part of the TIntermediate; it really does not belong there.

steve-lunarg:
Re-pushed:

- Changed API, command line, and underlying code to reverse the previous sense.  Now, provide the shader source function entry point, which will be renamed to the real entry point name as given to TShader::setEntryPoint().

- Removed all the extra gtests stuff that was overkill for one test, and replaced it with a runtest.

PR #611, opened 2016-12-03
Remove redundant output for successful link attempts

zeux:
The presence of non-warning/error output is bad since the callers have
to filter out warning/error messages with regular expressions or the
like to keep messages that are actually important.

TIntermediate::warn/error prefix their messages with the stage the issue
is with so this message appears completely redundant.


zeux:
Aaaand now I need to regenerate all test results. Please let me know if you think this change is a good idea or if there's a better way to suppress this output.

johnkslang:
It is still needed when printing the AST. fb06e9552ec81d0b9441b9c90b42d09ccdf28a99 does this and updates the tests.

zeux:
Thanks!

PR #625, opened 2016-12-09
Unbreak build on FreeBSD with GCC/libstdc++ 6.2.0

jbeich:
Probably affects Linux as well with non-standard libc that doesn't bootleg `<stdarg.h>`.
```c++
In file included from C:/Projects/glslang/glslang/MachineIndependent/glslang.y:59:0:
glslang/MachineIndependent/ParseHelper.h:276:24: error: 'va_list' has not been declared
                        va_list args);
                        ^~~~~~~
```


johnkslang:
Note this fixes issue #636. I've verified it works on VS.

PR #632, opened 2016-12-15
HLSL: inter-stage structure splitting.

steve-lunarg:
WIP: do not merge.

Note github is collapsing several of the file dfifs into an "expand" button - those are easy to accidentally overlook.
This adds structure splitting, which among other things will enable GS support where input structs are passed, and thus become input arrays of structs in the GS inputs.  That is a common GS case.

The salient points of this PR are:

* Structure splitting has been changed from "always between stages" to "only into the VS and out of   the PS".  It had previously happened between stages because it's not legal to pass a struct   containing a builtin IO variable.

* Structs passed between stages are now split into a struct containing ONLY user types, and a   collection of loose builtin IO variables, if any.  The user-part is passed as a normal struct   between stages, which is valid SPIR-V now that the builtin IO is removed.

* Internal to the shader, a sanitized struct (with IO qualifiers removed) is used, so that e.g,   functions can work unmodified.

* If a builtin IO such as Position occurs in an arrayed struct, for example as an input to a GS,   the array reference is moved to the split-off loose variable, which is given the array dimension   itself.

When passing things around inside the shader, such as over a function call, the the original type is used in a sanitized form that removes the builtIn qualifications and makes them temporaries. This means internal function calls do not have to change.  However, the type when returned from the shader will be member-wise copied from the internal sanitized one to the external type. The sanitized type is used in variable declarations.

When copying split types and unsplit, if a sub-struct contains only user variables, it is copied as a single entity to avoid more AST verbosity.

Above strategy arrived at with talks with @johnkslang.

This is a big complex change.  I'm inclined to leave it as a WIP until it can get some exposure to real world cases.


steve-lunarg:
Update with support for copying of nested split types (e.g, a struct of a struct containing a mixture of builtin interstage IO and user types).  New test added for this.

Also, when copying such types, a subtree will be copied with one assignment rather than member-wise IFF it contains only user types, to save AST verbosity.

steve-lunarg:
Another update with support for moving arrayness to builtin IO var members of structs.  For example:

```
struct PS_IN 
{ 
    float4 pos : SV_Position; 
}; 
```

with PS_IN i[3], then a loose variable called i.pos is created with a type of float4[3].  Any references to i[x].pos are changed to i.pos[x], - i.e, as the same index into the loose array variable.

glslang/Include/Types.h
```diff
@@ -1223,6 +1223,16 @@ class TType {
             typeName = NewPoolTString(copyOf.typeName->c_str());
     }
 
+    // Recursively make temporary
+    void makeTemporary()
```
johnkslang: This is a bit worrying. Storage class should go with object, not with type, and structures can be shared between objects. Is there code picking up the storage class from something "in the middle" of a type graph, rather than from the base object?

steve-lunarg: As far as I know, that should be unchanged.  The purpose of that code is to convert the IO boundary type with potentially mixed builtin and user variables to a pure temporary to pass around internally.  So this:
```
   struct {
      float4 pos : SV_Position;
      float4 foo;
   };
```

to this:

```
   struct {
      float4 pos;
      float4 foo;
   };
```

It's recursive since builtins can appear at any nesting level.  The type is shared for all variables declared from it though; that shouldn't be any different.  This is just to make the internal type from whatever the shader said, which may have been an arbitrary mixture.

PR #635, opened 2016-12-19
HLSL: allow "sample" in expressions.

steve-lunarg:
Unlike other qualifiers, HLSL allows "sample" to be either a qualifier keyword or an identifier (e.g, a variable or function name).

A fix to allow this was made a while ago, but that fix was insufficient when 'sample' was used in an expression.  The problem was around the initial ambiguity between:

 ``sample float a; // "sample" is part of a fully specified type``
and
`` sample.xyz;     // sample is a keyword in a dot expression``

Both start the same.  The "sample" was being accepted as a qualifier before enough further parsing was done to determine we were not a declaration after all.  This consumed the token, causing it to fail for its real purpose.

Now, when accepting a fully specified type, the token is pushed back onto the stack if the thing is not a fully specified type.  This leaves it available for subsequent purposes.

Changed the "hlsl.identifier.sample.frag" test to exercise this situation, distilled down from a production shaders.


hlsl/hlslGrammar.cpp
```diff
@@ -475,8 +476,15 @@ bool HlslGrammar::acceptFullySpecifiedType(TType& type)
     TSourceLoc loc = token.loc;
 
     // type_specifier
-    if (! acceptType(type))
+    if (! acceptType(type)) {
+        // If this is not a type, we may have inadvertently gone down a wrong path
+        // py parsing "sample", which can be treated like either an identifier or a
```
johnkslang: py -> by?

steve-lunarg: Of course I'd have made this error in the grammar...

Issue #255, opened 2016-04-28
 '##' : token pasting not implemented


kalmiya:
We have (many) openGL-shaders using expanding macro's. When using glslangValidator.exe to process these shaders, we get the message that "'##': token pasting not implemented". 

Are there any plans to support this ( if so, what is the timeline ) ? 

example shader:
# version 330
# define bla(aa)  Heh_##aa

void main()
{
    float bla(0);
}

kalmiya:
I started looking into the code and try and get a (hacky) short-term solution. 

The below "bla0", "bla1" and "bla2" testcases work (at least -E shows the expected output and when using -V it generates a binary). 

Testcase "bla4" is more difficult and not yet working. Not sure whether it should "loop/recurse" the code added so far ( to keep scanning for more ##), or if it should return from tMacroInput::scan and get another call to expand the second ## (that doesn't seem to be the case). Maybe someone more knowledgeable on the code could give some pointers whether this is going in the correct direction?

Testcase:

`#version 330 

#define bla0(aa)		AAA_##aa
#define bla1(aa)		aa##_BBB
#define bla2(aa,bb)		aa##bb
#define bla3()			CCC_##DDD
#define bla4(aa)		EEE_##aa##_FFF

void main()
{
	// AAA_X
	float bla0(X);
	// Y_BBB
	float bla1(Y);
	// ZW
	float bla2(Z,W);
	// CCC_DDD
//	float bla3();	// 'weird' situation ( can just avoid doing this )
	// 
//	bla4(GG);		// nasty testcase, but we actually need this one...
}  `

`

Output of glslangvalidator -E for the example is :

`#version 330
void main()
{
 float AAA_X;
 float Y_BBB;
 float ZW;
}`


[glslang-macro-doublehash.txt](https://github.com/KhronosGroup/glslang/files/242368/glslang-macro-doublehash.txt)

johnkslang:
I currently don't know of anyone else planning to do this.

justxyz:
please support it, because many shader use it.

kalmiya:
Here's an update of my hacky attempt - your mileage may vary :)

[glslang-macro-doublehash2.txt](https://github.com/KhronosGroup/glslang/files/290504/glslang-macro-doublehash2.txt)

jshofmann:
We'd like to see this supported as well.

TiemoJung:
We'd like to see this supported too

johnkslang:
[...]

PR #637, opened 2016-12-19
PP: Implement token pasting for PP identifiers.

johnkslang:
Implement token pasting as per the C++ specification, within the current
style of the PP code.
Non-identifiers (turning 12 ## 10 into the numeral 1210) is not yet covered;
they should be a simple incremental change built on this one.
Addresses issue #255.


johnkslang:
Note: Making bigger numbers out of smaller ones, or something like "*=" requires just a bit more local logic; they can be handled as simple incremental changes, but the way this is implemented will require the pasted tokens to be a valid token.  (C++ is undefined when the posted token is not valid, so requiring a valid token adheres to the specification.)

PR #647, opened 2016-12-26
HLSL: default function parameters

steve-lunarg:
This PR adds support for default function parameters in the following cases:

1. Simple constants, such as void fn(int x, float myparam = 3)
2. Expressions that can be const folded, such a ... myparam = sin(some_const)
3. Initializer lists that can be const folded, such as ... float2 myparam = {1,2}

New tests are added: hlsl.params.default.frag and hlsl.params.default.err.frag (for testing error situations, such as ambiguity or non-const-foldable).

In order to avoid sampler method ambiguity, the hlsl better() lambda now considers sampler matches.  Previously, all sampler types looked identical since only the basic type of EbtSampler was considered.

### DIFF
--- a/Test/baseResults/hlsl.intrinsics.negative.frag.out
+++ b/Test/baseResults/hlsl.intrinsics.negative.frag.out
@@ -1,6 +1,10 @@
 hlsl.intrinsics.negative.frag
 ERROR: 0:10: 'determinant' : no matching overloaded function found 
 ERROR: 0:12: 'f32tof16' : unimplemented intrinsic: handle natively 
+ERROR: 0:23: 'length' : ambiguous best function under implicit type conversion 
+ERROR: 0:25: 'normalize' : ambiguous best function under implicit type conversion 
+ERROR: 0:26: 'reflect' : ambiguous best function under implicit type conversion 
+ERROR: 0:27: 'refract' : ambiguous best function under implicit type conversion 
 ERROR: 0:28: 'refract' : no matching overloaded function found 
 ERROR: 0:30: 'transpose' : no matching overloaded function found 
 ERROR: 0:39: 'GetRenderTargetSamplePosition' : no matching overloaded function found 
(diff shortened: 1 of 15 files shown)


glslang/MachineIndependent/glslang_tab.cpp
```diff
@@ -109,316 +107,315 @@ using namespace glslang;
...
-     MEDIUM_PRECISION = 537,
-     LOW_PRECISION = 538,
-     PRECISION = 539,
-     PACKED = 540,
-     RESOURCE = 541,
-     SUPERP = 542
-   };
+  enum yytokentype
+  {
+    ATTRIBUTE = 258,
```
johnkslang: Is this all just white space differences?  The grammar files are showing massive changes for what should trivially effect only effect five productions, so I wonder if there's a more concise way. Maybe the versions are just too different?

glslang/MachineIndependent/glslang.y
```diff
@@ -361,14 +361,14 @@ function_call_header_no_parameters
 
 function_call_header_with_parameters
     : function_call_header assignment_expression {
-        TParameter param = { 0, new TType };
+        TParameter param = { 0, new TType, 0 };
```
johnkslang: I think better in this case to not add these: GLSL implementation can still operate as if there are no default parameters.  C++ makes the third parameter be `nullptr` anyway (via value initialization) when they are not provided. (If they are provided, we are moving everything to `nullptr` style instead of 0 style.) This would also avoid the minor issue with massive grammar file swings.

steve-lunarg: OK, I'll fix and repush.  That will also remove the changes to glslang_tab.cpp.  (I get a different glslang_tab.cpp than what's checked in even if I re-gen it with no changes to the .y, so I'm assuming that's down to a different version of bison, but it won't be an issue if I don't have to regenerate it).

PR #650, opened 2016-12-28
HLSL: allow destination swizzles when writing RWTexture/RWBuffer

steve-lunarg:
Reads and write syntax to UAV objects is turned into EOpImageLoad/Store operations.  This translation did not support destination swizzles, for example, "mybuffer[tc].zyx = 3;", so such statements would fail to
compile.  Now they work.

New test: hlsl.rw.swizzle.frag


hlsl/hlslParseHelper.cpp
```diff
@@ -336,26 +388,32 @@ TIntermTyped* HlslParseContext::handleLvalue(const TSourceLoc& loc, const char*
                 //   OpSequence
                 //      coordtmp = load's param1
                 //      rhsTmp = OpImageLoad(object, coordTmp)
-                //      rhsTmp op= rhs
+                //      rhsTmp op = rhs
```
johnkslang: Seems like a mistaken change.

steve-lunarg: > Studying this, I believe that's handled at a lower level, and this is correctly setting up the AST for only RMW of the requested components.

(I might be misunderstanding your concern; I'll try again if I'm not answering the question you're asking).  The intent is to set up the AST to avoid changing components that are not being written, and also, to avoid needless loads or stores, while allowing swizzled writes, which weren't handle at all before.  The swizzles are handle downstream; they're just injected here.

So for example, given a vec4 resource foo, then "foo[c].zywx = myvec4;" will only generate only a single image store on a temporary that's written with a swizzled L-value, since the code detects all channels are being updated.  So in effect it's "temp.zywx=myvec4, store(temp), temp;".  For the partial update case: "foo[x].xy = myvec2;" will load the old value of foo[x] into a vec4 temp, write two channels of data to its .xy, then store the updated value, so as to preserve the prior .zw.  In that case we do create both a load and a store.  If there's a better way to express that in AST, I can change the decomposition accordingly.

johnkslang: >  For the partial update case: "foo[x].xy = myvec2;" will load the old value of foo[x] into a vec4 temp, write two channels of data to its .xy, then store the updated value, so as to preserve the prior .zw. In that case we do create both a load and a store.

The problem is loading a full vec4 of foo[x]. Another shader invocation might be modifying foo[x].z at the same time, and injecting an explicit load/store of the z component in this invocation (by loading the full foo[x]), it can overwrite the store in the other invocation.

Main rule: We can't introduce stores of components the shader doesn't explicitly store to, unless we know the particular case is immune to the data race (which depends on the memory model of the consuming environment).

steve-lunarg: OK: so that case is generating wrong AST then, in the current code.

I have all the info to properly mask the write (and avoid the read); I just don't know the AST incantation for doing that.  I'll investigate further.

steve-lunarg:
Repush with an unimplemented error for partial update cases.

johnkslang:
Looks good, except for change `op=` to `op  =`.  I'll fix that manually.

PR #652, opened 2017-01-02
HLSL: Support for vec -> scalar conversion on assignments

TiemoJung:
Enables automatic conversion from any vector type to a scalar type on asignment


johnkslang:
FYI, 

- I (and @steve-lunarg) and trying to understand more holistically this conversion space, so still analyzing.
- It looks like the white-space changes made that are unrelated to the PR lead to merge conflicts.

Also, while I'm fine with removing trailing white space from source code, there is no project rule for doing so, and it should absolutely never be done to test files, where it is important to keep as is.

johnkslang:
d5d9ffbdfdc5a3af9eaef7b8f35bfc8f5064d225 should be a superset holistic solution to this problem.

PR #654, opened 2017-01-02
WIP: HLSL: Support for vector truncation for vector multiply

TiemoJung:
Adds support for vector truncation for vector multiply (e.g float3 = float3 * float4).

also seems to accidentally fixes a spir-v gen bug (turns imul of floats in correct fmul of floats)

Change-Id: Iad18b280c96a601e2e33f2276e66e97dd3571e3e

### DIFF
--- a/Test/baseResults/hlsl.float1.frag.out
+++ b/Test/baseResults/hlsl.float1.frag.out
@@ -18,13 +18,16 @@ gl_FragCoord origin is upper left
 0:5      'inScalar' (in float)
 0:?     Sequence
 0:6      Branch: Return with expression
-0:6        add (temp 1-component vector of float)
-0:6          vector-scale (temp 1-component vector of float)
-0:6            'f1' (global 1-component vector of float)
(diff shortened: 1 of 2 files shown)


steve-lunarg:
I looked into this issue a bit, and it seems like if we fix it in the general sense (i.e, not just for multiplies), it exposes a limitation caused by a lack of generality in selecting which available argument to an operation to convert.

So, I think there are two matters of interest: (a) unblocking @TiemoJung's shaders, and (b) improving the conversion scheme and subsequently removing the limitation on truncating vectors to scalars in the general case, since generally such truncations aren't limited to multiplies.

johnkslang:
Changing to WIP, since a more generic answer is needed, and need to know how that interacts with other changes @steve-lunarg has been making in this space.

johnkslang:
d5d9ffbdfdc5a3af9eaef7b8f35bfc8f5064d225 should be a superset holistic solution to this problem.

PR #660, opened 2017-01-05
HLSL: user defined semantic handling

TiemoJung:
Adds an interface to allow the user to control the mapping of HLSL semantics

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -37,6 +37,7 @@
 #define _COMPILER_INTERFACE_INCLUDED_
 
 #include "../Include/ResourceLimits.h"
+#include "../Include/BaseTypes.h"
 #include "../MachineIndependent/Versions.h"
 
 #include <cstring>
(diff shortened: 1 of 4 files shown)


johnkslang:
Not sure I understand yet why headers need to be refactored. Generally, a lower-level interface needs things like type information which it gets by including type files. In the meantime, the project needs to stick to its current coding convention, just for sanity of reading code in one file.  Indentation, '{ }' placement, etc. Also, 'translate' is mispelled

TiemoJung:
Hi,

i did address your points. 

I was for some reason under the impression that including ../Include/BaseTypes.h would break something, So that's why i moved the enum over, i reverted it and added the required include.

All the other whitespace and indentation issues are due to automatic application of our internal coding rules.

sergeyreznik:
Thank you for implementing this. Can't wait this pull request to be merged and chnage my "workarounds" in code.
For the sake of the consistency you should use "builtIn" everywhere. Now it is spelled as "buildIn" in several places.

hlsl/hlslParseHelper.cpp
```diff
@@ -4688,7 +4605,7 @@ void HlslParseContext::setLayoutQualifier(const TSourceLoc& loc, TQualifier& qua
                 qualifier.layoutXfbOffset = value;
             return;
         } else if (id == "xfb_stride") {
-            // "The resulting stride (implicit or explicit), when divided by 4, must be less than or equal to the 
+            // "The resulting stride (implicit or explicit), when divided by 4, must be less than or equal to the
```
johnkslang: Would like to see no unrelated changes for something like this.  I'll make a pass that is clean up only.  But be sure you won't ever do this to in test input. (Tools that go make changes in other places behind your back are a bit dangerous... is there a mode to restrict it to the lines of code you are actually modifying?)

hlsl/hlslParseHelper.cpp
```diff
@@ -65,7 +65,8 @@ HlslParseContext::HlslParseContext(TSymbolTable& symbolTable, TIntermediate& int
     nextInLocation(0), nextOutLocation(0),
     sourceEntryPointName(sourceEntryPointName),
     builtInIoIndex(nullptr),
-    builtInIoBase(nullptr)
+    builtInIoBase(nullptr),
+    semanticResolver(semantic_resolver)
```
johnkslang: Note, for any that don't know: if you use `semanticResolver` both as an argument and member name, you can write this as `semanticResolver(semanticResolver)`.

glslang/Public/ShaderLang.h
```diff
@@ -284,6 +285,145 @@ bool InitializeProcess();
...
+            { "SV_GROUPID", EbvWorkGroupId },
+            { "SV_DOMAINLOCATION", EbvTessCoord },
+            { "SV_DEPTH", EbvFragDepth },
+            { "SV_COVERAGE", EbvSampleMask },
+            { "SV_DEPTHGREATEREQUAL", EbvFragDepthGreater },
+            { "SV_DEPTHLESSEQUAL", EbvFragDepthLesser },
+            //{ "SV_STENCILREF", not implemented },
+            //{ "SV_GROUPINDEX", not implemented },
+        };
+        for (const auto& entry : allOtherBuildIns)
```
johnkslang: The colon syntax (range-based) is a feature that not all current users can compile. Can you replace with traditional semicolons? `auto` is okay.

hlsl/hlslParseHelper.cpp
johnkslang: White space is now handled in ecba76fe73b68bd62d2037f9d5d2ff4a9afd7dc0.

TiemoJung:
Hi,

@johnkslang Regarding your concerns about making too many HLSL specific additions.
If preferred, there is an alternative way to handle HLSL semantics. By adding the semantics to the type qualifier and then extend the TIoMapResolver interface to handle in/out variable too. This also has the advantage that are the full in/out type info is available and the user can better decide what to do and there is only one HLSL specific addition to the interface, the semantic name.
In fact the first approach to handle HLSL semantics was to do it in this way.

hlsl/hlslParseHelper.cpp
TiemoJung: Its only changing line endings and removes trailing whiespaces on save, i have turned it off for now, so i hope that reoccurring problem is for future pull request resolved.

TiemoJung:
Added pull request #669 with the alternative approach.

johnkslang:
#669 is in now, closing this one, which greatly overlaps both #669 and some other changes made independently of these, but are quite similar.  We should increment on master now.

PR #661, opened 2017-01-05
PP: #include: simplify the different paths out of #include.

johnkslang:
- some paths didn't release 'res'
- token is always '\n' after proper acceptance of the directive itself,
  so no need to test it, change it to '\n', etc.
- assuming setCurrentColumn(0) is not needed unless there are header tokens,
  but not clear why it is ever needed

Note: much of the simplified code read as if the included header tokens had
actually been processed, versus queued up for processing; maybe that explains
some things.

### DIFF
--- a/glslang/MachineIndependent/preprocessor/Pp.cpp
+++ b/glslang/MachineIndependent/preprocessor/Pp.cpp
@@ -593,37 +593,42 @@ int TPpContext::CPPinclude(TPpToken* ppToken)
 
     // Make a copy of the name because it will be overwritten by the next token scan.
     const std::string filename = ppToken->name;
+
+    // See if the directive was well formed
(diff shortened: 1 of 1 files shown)


dneto0:
The code looks much nicer now.  I think the setCurrentColumn(0) is needed so that you get proper columns in error messages if there is a problem with the first line of the included text.  Otherwise the reported column is too far to the right, offset by the number of characters until you got to the (end of?) header name token.

I'm currently testing our downstream glslc tests.

dneto0:
Ok.  This change looks good to me.

(I had to fix some things in downstream Shaderc to adapt to your earlier change: 
*  error message change
*  we had cases where there were #include directives without the terminating newline.  Those test cases have now been fixed.
)

LGTM

PR #663, opened 2017-01-06
Includer interface change to support full include semantics (requires downstream changes)

johnkslang:
This could require downstream changes, so leaving open for review for a few days.

The first commit is just nomenclature consolidation to one model.

The second commit is an interface change, to simplify alignment with C++ semantics for <> and "".

Any previous subclass implementation would only be for "", and likely didn't check system locations, which would just mean changing

```
    include(...)  -> includeLocal(...)
```

(and possibly leaving the default implementations in the base class for `includeSystem` and `releaseInclude`). See comments about `includeLocal` being an additional search over `includeSystem`, not a superset search.

This also removed `ForbidIncluder` (the base class works for that), as
 
- the error message in `ForbidIncluder` was redundant: error results were already returned to the caller, which can then give the error it wants to in the proper channel for its use
- given that there is a trivial default implementation that a subclass can override any subset of, moved to that instead of the ABC (I still like abstract base classes in general though, it was this particular implementation that was no longer wanted)
- trying to get less implementation out of the interface file anyway


johnkslang:
@dneto0 and others that consume the Includer interface: Any issues adapting to this change? And, of course, does it look correct?  Thanks.

glslang/MachineIndependent/preprocessor/Pp.cpp
```diff
@@ -576,12 +576,12 @@ int TPpContext::CPPifdef(int defined, TPpToken* ppToken)
 int TPpContext::CPPinclude(TPpToken* ppToken)
 {
     const TSourceLoc directiveLoc = ppToken->loc;
-    TShader::Includer::IncludeType includeType = TShader::Includer::EIncludeRelative;
+    bool addLocalSearch = true; // to additionally include the extra "" paths
```
dneto0: Naming:
Depending on point of view, "adding" a step usually means the new step occurs after other steps already defined.
How about enableLocalSearchFirst ?

glslang/Public/ShaderLang.h
```diff
@@ -331,72 +331,68 @@ class TShader {
...
-        // Resolves an inclusion request by name, type, current source name,
+        // For both include methods below:
+        //
+        // Resolves an inclusion request by name, current source name,
         // and include depth.
         // On success, returns an IncludeResult containing the resolved name
         // and content of the include.  On failure, returns an IncludeResult
-        // with an empty string for the file_name and error details in the
-        // file_data field.  The Includer retains ownership of the contents
+        // with an empty string for the headerName and error details in the
```
dneto0: Alternate failure mode is it returns nullptr, as the defaults do.

glslang/Public/ShaderLang.h
```diff
@@ -331,72 +331,68 @@ class TShader {
...
+        // "system" paths, because on returning a failure, the parser will
+        // call includeSystem() to look in the "system" locations.
+        virtual IncludeResult* includeLocal(const char* headerName,
+                                            const char* includerName,
+                                            size_t inclusionDepth) { return nullptr; }
+
         // Signals that the parser will no longer use the contents of the
         // specified IncludeResult.
-        virtual void releaseInclude(IncludeResult* result) = 0;
+        virtual void releaseInclude(IncludeResult* result) { }
```
dneto0: This is almost certainly a bad default implementation when result is non-null.  It probably entails a leak.

I'd probably make this virtutal = 0, and re-instate the ForbidIncluder and have it just override this method with an empty body.

johnkslang: Oops, that was a `delete result`, I missed it on check in.  Still, the point is valid, that in the abstract it's odd to mix and match defaults of generators and releasers.

glslang/MachineIndependent/preprocessor/Pp.cpp
johnkslang: Sure; it's just a function-local variable.

glslang/Public/ShaderLang.h
johnkslang: Right, I thought that was already exercised, because consumers will already checking against nullptr, but will update the comment.

PR #664, opened 2017-01-07
add other builtins to interstage IO

steve-lunarg:
WIP: add other builtins to interstage IO

This fixes PR #632 so that:

(a) The 4 PerVertex builtins are added to an interface block for all stages except fragment.

(b) Other builtin qualified variables are added as "loose" linkage members.

(c) Arrayness from the PerVertex builtins is moved to the PerVertex block.

(d) Sometimes, two PerVertex blocks are created, one for in, one for out (e.g, for some GS that both reads and writes a Position)


hlsl/hlslParseHelper.cpp
```diff
@@ -1446,8 +1449,75 @@ void HlslParseContext::addInterstageIoToLinkage()
...
+                }
+                member.type->clearArraySizes();
+            }
+
+            ioBlockTypes[blockId]->push_back(member);
+        }
+    }
+
+    // If there were perVertex items, add the block to the linkage.  Handle in and out separately.
+    for (int blockId=0; blockId<=1; ++blockId) {
```
johnkslang: Use spaces around ops, like other loops/etc. do.

glslang/Include/Types.h
```diff
@@ -1333,6 +1342,53 @@ class TType {
         }
     }
 
+    virtual bool isLooseBuiltIn(EShLanguage language) const
+    {
+        // Any stage
```
johnkslang: Might be simpler to start:

```
if (language == EShLangFragment || language == EShLangCompute)
    return true;
```

steve-lunarg: Agreed, that's better.  Will fix.

steve-lunarg:
Repushed with review changes:

1. Add spaces around for() operators
2. Changed name isLooseBuiltIn -> isLooseAndBuiltIn
3. Rephrased isLooseAndBuiltIn in terms of isPerVertexAndBuiltIn, which is much shorter and more robust against new ones appearing.

Issue #666, opened 2017-01-08
Add ability to conditionally disable HLSL code compilation


slime73:
The code in glslang which provides HLSL support adds a decent amount to the compiled binary size of glslang – it would be handy to be able to prevent HLSL support from being compiled in, for cases where binary size is a factor and HLSL input support is unneeded.

sergeyreznik:
@johnkslang Suddenly my code stopped working :)
**ENABLE_HLSL** is not quite proper define for this feature request. By default it should work, and must be conditionally disabled with define like **GLSLANG_DISABLE_HLSL_SUPPORT**

sergeyreznik:
@johnkslang I'm using glslang not as a separate library, but compiling it from sources inside my project (I'm aware this is not quite proper way, but I don't have build system now to use it as a separate project). And ENABLE_HLSL is not defined anywhere for me. 
Also there is a crash inside **InitializeSymbolTables** (ShaderLang.cpp, line 297) when ENABLE_HLSL is not defined. Caused by `unique_ptr<TBuiltInParseables> builtInParseables` to be nullptr and then unconditional call to
`builtInParseables->initialize(version, profile, spvVersion);`

johnkslang:
Is it possible to externally define `ENABLE_HLSL`, or add to your code?

Apologies for driving it that direction instead of disable, but there are other things that must be enabled, and it seemed to better to request what was needed rather than the other way. Can reverse that decision if it's really not the right way to do it.

slime73:
> ENABLE_HLSL is not quite proper define for this feature request.

There was discussion on the PR about this (not sure if you saw or not): https://github.com/KhronosGroup/glslang/pull/668

> Also there is a crash inside InitializeSymbolTables (ShaderLang.cpp, line 297) when ENABLE_HLSL is not defined.

This crash will also happen if the `source` parameter is set to EShSourceNone, although I don't believe a codepath does that currently. So I think the crash will only happen if you compile glslang without HLSL support and then try to use it with HLSL.

johnkslang:
Commit 65d538c832670e1a0a0da469ceb345c3e9622ada is more robust against the `source` parameter.

However, trying with HLSL disabled, something still tries to bring in the non-built hlsl.lib.  Looking at that.

johnkslang:
Commit dc1a8196cf4e717af8b1e64cf82ca3d13e4dd2d8 makes the normal build work with `ENABLE_HLSL` set to off.

[...]

PR #668, opened 2017-01-09
Add a CMake option to disable compilation of HLSL input support.

slime73:
This adds the ability to build glslang without support for HLSL input, which reduces the size of the binary.

Resolves isssue #666

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ option(ENABLE_GLSLANG_BINARIES "Builds glslangValidator and spirv-remap" ON)
 
 option(ENABLE_NV_EXTENSIONS "Enables support of Nvidia-specific extensions" ON)
 
+option(DISABLE_HLSL "Disables HLSL input support" OFF)
+
 enable_testing()
 
 set(CMAKE_INSTALL_PREFIX "install" CACHE STRING "prefix")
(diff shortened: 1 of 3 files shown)


johnkslang:
This looks reasonable.  One question, and one comment:

Question:  Does it all need to be expressed in the negative, or can it be "enable-HLSL, on by default"?

Comment: Glslang master and testing will generally have these switches (HLSL, AMD, NVIDIA, etc.) set to enabled ON, and not be testing combinations.  So, there is some risk this will occasionally break for the disabled settting.

slime73:
> Does it all need to be expressed in the negative, or can it be "enable-HLSL, on by default"?

It could be enable / on-by-default - I chose disable/off-by-default because in my own project I have the glslang source code in-tree with my own source code and I don't use glslang's own CMake build files, so my thought was that if anyone else does that, with it enable/on-in-cmake-by-default they'd have to explicitly set the ENABLE_HLSL preprocessor define to get all of glslang's functionality, whereas they wouldn't need any extra preprocessor definition set for HLSL to work if the preprocessor define disables HLSL.

That's not a strong reason to keep it the way I've wrote it though, I'll change it ASAP if you think I should.

> Glslang master and testing will generally have these switches (HLSL, AMD, NVIDIA, etc.) set to enabled ON, and not be testing combinations. So, there is some risk this will occasionally break for the disabled settting.

I have not really looked into how testing is done yet - maybe this warrants some additional build steps (in a separate issue / PR) to at least compile glslang with all possible build flag combinations? That said, as a user of the library I can completely understand "use at your own risk" build settings that are off by default.

johnkslang:
> they'd have to explicitly set the ENABLE_HLSL preprocessor define to get all of glslang's functionality

Understood; that's a good reason. However, with the other enables being the opposite way, and it being more rational to have them all going the same direction, it would be better to reverse this one.  Then, a user always explicitly adds into their local environment the list of functionality they want to include.

PR #669, opened 2017-01-09
HLSL/TIoMapResolver: Extend to handle shader in/out variables and semantic capture

TiemoJung:
This change adds two things.

- Adds capture of HLSL semantics into TQualifier
- Extends TIoMapResolver to remap shader in/out variables

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -392,6 +392,7 @@ public:
 
     void clear()
     {
+        semanticName = nullptr;
         precision = EpqNone;
         invariant = false;
         noContraction = false;
@@ -401,6 +402,7 @@ public:
     // drop qualifiers that don't belong in a temporary variable
     void makeTemporary()
     {
+        semanticName = nullptr;
         storage = EvqTemporary;
         builtIn = EbvNone;
         clearInterstage();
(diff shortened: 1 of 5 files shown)


johnkslang:
Seems reasonable, but would you mind rebasing it against master so I can pull it?  Thanks.

TiemoJung:
@johnkslang Sorry for the delay, i was busy with other things. Did a rebase it as you requested.

zeux:
This change seems to bitrot faster than it is merged. Without this a large class of shader codebases can't be compiled to SPIRV with predictable I/O locations - imagine glslang didn't support location layout on vertex inputs...

It would be great to get this to master. Please let me know if I can do anything to help - it seems like it just needs rebasing. If the rebase is not complicated then anybody with write access can do it ( @johnkslang ?)

johnkslang:
I will have a look at finishing this off, or what the blockers were (there was some peope/workflow blocker, will verify current state later today).

johnkslang:
I fixed this up, including some parts a merge wouldn't have noticed.

My concern is that this adds a mostly unused address to every use of TQualifier, which is used a lot, everywhere. While at the same time, the information is copied to a set in the TIntermediate, while the originals still live on. It feels like something is lurking here where we just need a map in the TIntermediate, and no pointer in the TQualifier.

PR #674, opened 2017-01-11
WIP: Combine pure samplers and textures option

TiemoJung:
This change adds the ability to combine pure samplers and textures into sampled textures.

Adds a option to TShader setMergeTextureSamplerPairs to enable this feature (off by default).

Limitations:
- Can not combine samplers and textures of functions parameters
- Can not combine arrays of samplers and textures

Todo:
- Add more tests
- Check out possibility to handle arrayed samplers/textures
- Validate that both sampler and texture do bind to the same binding/set (or none)
- Handle cases correctly when samplers and/or textures are used separate and not in sampled image construction
- Move code to its own file?


glslang/Public/ShaderLang.h
```diff
@@ -313,6 +313,18 @@ class TShader {
     void setAutoMapBindings(bool map);
     void setFlattenUniformArrays(bool flatten);
     void setNoStorageFormat(bool useUnknownFormat);
+    // If set to true, the parser/compiler trys to merge
+    // sperate sampler and texture combinations into one
```
johnkslang: separate

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2622,4 +2623,274 @@ void TIntermAggregate::addToPragmaTable(const TPragmaTable& pTable)
...
+    return true;
+  }
+
+private:
+  TIntermSymbolVector&    symbols;
+};
+
+class ConstructorGatherTraverser
+  : public TIntermTraverser
+{
```
johnkslang: Will need to follow the local coding convention.  Four spaces, {} placement, etc.

TiemoJung:
Update:

- there are now 2 modes
-- remove sampler and upgrade texture to sampled texture. Works for all cases and update functions and its parameters correctly too
-- merge only mode as described in inita pull request description

hrydgard:
Anything happening with this one? It needs a rebase, anyway.

TiemoJung:
Sorry for the long absence, i created #856 as a replacement for this.
Starting from scratch with only one transformation mode was simpler than rebase and clean this up.
The new pull request has only the simple 'upgrade and drop' mode, its simple and covers probably most cases for anyone that needs to combine samplers and textures.

PR #675, opened 2017-01-16
Implement new revision of extension GL_AMD_shader_ballot

amdrexu:
- Add support for invocation functions with "InclusiveScan" and
  "ExclusiveScan" modes.
- Add support for invocation functions taking int64/uint64/doube/float16
  as inout data types.


amdrexu:
@johnkslang It is very strange that the new test "spv.shaderBallotAMD.comp" I add fails in certain configurations. But I could successfully run it locally. Maybe you could help me to check it.

johnkslang:
Two different platforms are saying the shader ballot test failed, so something real is going on. Is the test result correct, with correct line endings?

One suspicious thing I saw in the output was `���EOpImageAtomicExchange’` which implies some bad memory or initialization?

amdrexu:
I  find the root cause. The difference between the two files generated by spv.shaderBallotAMD.comp is caused by the ordering of OpExtension XXX. In original design, I define this `std::set<const char*> extensions` in `spvBuilder`. The ordering of the set will be based on pointer value instead of the contained string. This is why I could successfully run the test locally while it fails on your pre-checkin farm. Pointer value is very "unreliable" and depends on system address allocation. So this time, I change the design to this: `std::set<std::string> extensions`. It is on the right track now.

PR #678, opened 2017-01-18
support SPV_NV_viewport_array2 and SPV_NV_stereo_view_rendering

chaoc:
Two extensions have dependency. therefore merging together.


chaoc:
@johnkslang It seems that I am having a fail on spv.stereoViewRendering.vert.out and spv.stereoViewRendering.tesc.out for clang tests due to the order of the two extension:

-                              Extension  \"SPV_NV_viewport_array2\"
-                              Extension  \"SPV_NV_stereo_view_rendering\"
+                              Extension  \"SPV_NV_stereo_view_rendering\"
+                              Extension  \"SPV_NV_viewport_array2\"

But it is passing for me locally and if I reverse the order in the out file I will get fail at gcc tests: 

-                              Extension  \"SPV_NV_stereo_view_rendering\"
-                              Extension  \"SPV_NV_viewport_array2\"
+                              Extension  \"SPV_NV_viewport_array2\"
+                              Extension  \"SPV_NV_stereo_view_rendering\"


I saw other pull request seem to have similar test issues. So could you help take a look?

johnkslang:
These should be in the order std::set orders strings by default, which should be less than, which I think means they should come out in alphabetical order.

Is it possible one of the C++ compilers is not putting its members in the order required by the C++ specification?

johnkslang:
@amdrexu figured this out `std::set<const char*> extensions;` needs to be `std::set<const char> extensions;` to get the sort to work portably (to sort names instead of addresses).

amdrexu:
Please sync up latest codes. I think the problem is gone if you update your test output file and resubmit a merge request.

chaoc:
@amdrexu Thanks for solving that. It worked for me.

johnkslang:
@chaoc General comment: After you think this PR is ready, remove your "[WIP]".

SPIRV/GlslangToSpv.cpp
```diff
@@ -2275,6 +2311,24 @@ void TGlslangToSpvTraverser::decorateStructType(const glslang::TType& type,
             spv::BuiltIn builtIn = TranslateBuiltInDecoration(glslangMember.getQualifier().builtIn, true);
             if (builtIn != spv::BuiltInMax)
                 addMemberDecoration(spvType, member, spv::DecorationBuiltIn, (int)builtIn);
+
+#ifdef NV_EXTENSIONS
+            if (builtIn == spv::BuiltInLayer) {
+                // SPV_NV_viewport_array2 extension
+                if (glslangMember.getQualifier().layoutViewportRelative)
```
johnkslang: Use the local coding convention 'if (....) {'

chaoc:
@johnkslang Thanks for pointing that out. All work is now done and change request has been made. WIP in title removed. Please feel free to merge in.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3417,8 +3424,16 @@ void TParseContext::redeclareBuiltinBlock(const TSourceLoc& loc, TTypeList& newT
...
+#ifdef NV_EXTENSIONS
+            if (member->type->getFieldName() == "gl_Layer") {
+                if (!newType.getQualifier().layoutViewportRelative && newType.getQualifier().layoutSecondaryViewportRelativeOffset == -2048)
+                    error(loc, "redeclaration only allowed for viewport_relative or secondary_view_offset layout", "redeclaration", member->type->getFieldName().c_str());
+                oldType.getQualifier().layoutViewportRelative = newType.getQualifier().layoutViewportRelative;
+                oldType.getQualifier().layoutSecondaryViewportRelativeOffset = newType.getQualifier().layoutSecondaryViewportRelativeOffset;
+            }
             if (oldType.isImplicitlySizedArray() && newType.isExplicitlySizedArray())
                 oldType.changeOuterArraySize(newType.getOuterArraySize());
+#endif
```
johnkslang: This #endif made long-existing statements only exist for NV_EXTENSIONS.

chaoc:
Sorry about that. I see someone got to the fix already. Thanks for the help.

PR #683, opened 2017-01-20
Fix compilation with MSVC 2010

rdb:
As the title states.

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -1079,11 +1079,10 @@ void HlslParseContext::flatten(const TSourceLoc& loc, const TVariable& variable)
 {
     const TType& type = variable.getType();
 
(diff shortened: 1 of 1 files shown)


johnkslang:
@steve-lunarg does this break your intent?

steve-lunarg:
No.  It would be slightly less efficient, but not enough to worry about.  Seems OK.

Only if it's not to use emplace() any more, I'd fix the comment just above it which still says this in the revised version:

`// emplace gives back a pair whose .first is an iterator to the item...`

johnkslang:
The other aspect is using `[ ]` instead of `find` changes the tree even if the key is missing. Is this search always going to find it, or is it sometimes missing?

johnkslang:
Nevermind, it looks like it has to find it. @rdb  can you delete the comment and resubmit?

rdb:
Amended with requested change.  Thanks!

PR #684, opened 2017-01-20
WIP: vertex shader auto y negate for vulkan

TiemoJung:
This change adds the option to automatically negate y in vertex shaders.
It also has a mode where this can be controlled via a spec constant.

TODO:
- add tests
- vec4 constructor from scalar may be incorrect (need to clone nodes?)
- set and allow to configure default value for spec constant


johnkslang:
If I recall correctly, this was no longer needed now that the viewport can be set up with y inverted.

TiemoJung:
Only if the implementation supports VK_AMD_negative_viewport_height, then you can use a negative value for the viewport height to turn off this unfortunate feature in vulkan.

hesiod:
Actually, Vulkan 1.0.39 now includes the `VK_KHR_maintenance1` extension, which in turn incorporates the negative height feature from `VK_AMD_negative_viewport_height`, so this shouldn't be necessary anymore.

TiemoJung:
Yes we did decide that we now require it and this becomes obsolete.

PR #685, opened 2017-01-20
HLSL: Wrap entry-point; needed to write 'in' args, and support 'inout' args

johnkslang:
This needs some render testing, but is destined to be part of master.

This also leads to a variety of other simplifications.
 - IO are global symbols, so only need one list of linkage nodes (deferred)
 - no longer need parse-context-wide 'inEntryPoint' state, entry-point is localized
 - several parts of splitting/flattening are now localized


johnkslang:
This has tested well, will merge into master, and move on to other features.

PR #690, opened 2017-01-21
CMake: add target_link_libraries internally

ligfx:
Makes it easier to include glslang in a larger CMake project: instead of having to call `target_link_libraries(glslang OSDependent OGLCompiler HLSL)`, for example, you only need to call `target_link_libraries(glslang)` and it will pull in the helpers it needs.

This is also better in terms of cleaning up the "public interface", of sorts, for building glslang: end-users probably shouldn't need to know or be explicitly dependent on internal targets.


johnkslang:
Looks good. However, will this break anyone doing it the old way of listing more libraries?

ligfx:
It shouldn't, since those targets still exist. Either the compiler or CMake will notice that you're linking to the same library twice and deal with it appropriately.

PR #694, opened 2017-01-24
Use correct OriginUpperLeft when linking more than one shader.

mtavenrath:


johnkslang:
Good catch. It looks like this path (linking multiple compilation units into a single stage for Vulkan) is not fully enabled yet. This is one of the holes. The other obvious hole I see is the `spvVersion` is not getting set, and should happen at the same time (it's possible that's not yet getting used though).

Ideally, this also comes with a link test.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -1707,6 +1707,13 @@ bool TProgram::linkStage(EShLanguage stage, EShMessages messages)
         intermediate[stage] = new TIntermediate(stage,
                                                 firstIntermediate->getVersion(),
                                                 firstIntermediate->getProfile());
+
+        // The new TIntermediate must use the same origin as the original TIntermediates.
+        // Otherwise linking will fail due to different coordinate systems.
+        if (firstIntermediate->getOriginUpperLeft()) {
+            intermediate[stage]->setOriginUpperLeft();
```
johnkslang: Would you be willing to update this to also include this?

```
intermediate[stage]->setSpv(firstIntermediate->getSpv());
```

We maybe want a better mechanism for doing this, but at least that would be a complete fix.

(Otherwise, if you prefer, I can do a commit that does both these and adds a test too.)

mtavenrath: I'll add the change and update the PR.

mtavenrath:
@johnkslang, I've added the requested change and wrote a test for originUpperLeft. I haven't found a way to test the Spv propagation which wouldn't also fail for the originUpperLeft case.

PR #702, opened 2017-01-26
SpvBuilder: add const specifier to vector reference arguments

resetnow:
This allows calling functions with temporary objects like this:

```C++
b.makeCompositeConstant(
      vec4_float_type_,
      std::vector<Id>({b.makeFloatConstant(1.f), b.makeFloatConstant(1.f),
                       b.makeFloatConstant(1.f), b.makeFloatConstant(1.f)}));
```


johnkslang:
Adding the consts looks fine (will accept when tests pass) but keep in mind that the `{...}` syntax is not accepted by all the compilers in common use for this project.

PR #704, opened 2017-01-27
WIP: HS/DS

steve-lunarg:
In progress: do not merge.

This is a limited implementation of tessellation shaders.  In particular, the following are not functional:

* patchconstantfunctions accepting per-control-point input values, such as const OutputPatch <hs_out_t, 3> cpv

* patchconstantfunctions used with shaders whose entry point has multiple return points.

* patchconstantfunctions whose signature requires an aggregate input type such as a structure containing builtin variables.  Code to synthesize such calls is not yet present.

These restrictions will be relaxed as soon as possible, pending other incoming PRs such as for function wrapping.  Simple cases can compile now: see for example Test/hlsl.hull.1.tesc - e.g, writing to inner and outer tessellation factors.

PCF invocation is synthesized as an entry point epilogue protected behind a barrier and a test on invocation ID == 0.  If there is an existing invocation ID variable it will be used, otherwise one is added to the linkage.  The PCF and the shader EP interfaces are unioned and builtins appearing in the PCF but not the EP are also added to the linkage and used to synthesize the PCF invocation. Parameter matching is by builtin variable type.  Any user variables in the PCF signature will result in an error.  Overloaded PCF will also result in an error.

[domain()], [partitioning()], [outputtopology()], [outputcontrolpoints()], and [patchconstantfunction()]
attributes to the shader entry point are in place, with the exception of the Pow2 partitioning mode.


steve-lunarg:
Closing to replace with version built on wrap-entry-point code.

PR #712, opened 2017-02-01
HLSL: Type sanitization: create non-IO types for var decl and fn para…

steve-lunarg:
…m/returns

This introduces parallel types for IO-type containing aggregates used as non-entry point function parameters or return types, or declared as variables. Further uses of the same original type will share the same sanitized deep structure.

This is intended to be used with the wrap-entry-point branch.


hlsl/hlslParseHelper.h
```diff
@@ -296,9 +297,10 @@ class HlslParseContext : public TParseContextBase {
     TVector<int> flattenLevel;  // nested postfix operator level for flattening
     TVector<int> flattenOffset; // cumulative offset for flattening
 
-    // Sanitized type map.  During declarations we use the sanitized form of the type
-    // if it exists.
-    TMap<const TTypeList*, TType*> sanitizedTypeMap;
+    // Sanitized type map.  If the same type is sanitized again, we want to reuse it.
+    // We cannot index by the TType, since the same memory can be reused, e.g, from a
+    // stack variable, so the index is on the struct's typelist.
+    TMap<const TTypeList*, TTypeList*> nonIoTypeMap;
```
johnkslang: Would TUnorderedMap be better here?

BTW, on the comment: Generally TType itself (shallow, not what it points to) is copied a lot and is not supposed to mean the type is different. Not sure about "memory can be reused", just sure that things can be the same type even though they have different TType*, but this is less true of TTypeList*.

hlsl/hlslParseHelper.cpp
```diff
@@ -5457,19 +5457,27 @@ TIntermNode* HlslParseContext::declareVariable(const TSourceLoc& loc, TString& i
...
+    case EvqTemporary:
+        nonIoType = makeNonIoType(nonIoType);
+        // nonIoType = type.clone();
+    default:
+        break;
+    }
 
     // Declare the variable
-    if (type.isArray()) {
+    if (nonIoType->isArray()) {
```
johnkslang: Actually, `nonIoType` could still be an IO type, right?  Maybe just make input a different name, and keep all these as `type`.

hlsl/hlslParseHelper.cpp
```diff
@@ -5457,19 +5457,27 @@ TIntermNode* HlslParseContext::declareVariable(const TSourceLoc& loc, TString& i
...
-    // interstage IO, we want to make it a temporary.
-    TType* sanitizedType = sanitizeType(&type);
+    TType* nonIoType = &type;
+
+    // make non-IO version of type
+    switch (type.getQualifier().storage) {
+    case EvqGlobal:
+    case EvqTemporary:
+        nonIoType = makeNonIoType(nonIoType);
+        // nonIoType = type.clone();
```
johnkslang: Delete.

johnkslang:
Looks good modulo a few minor points, if you don't mind addressing those.

hlsl/hlslParseHelper.cpp
steve-lunarg: Yeah, and in fact it doesn't even have to be a separate variable at all, since it's modified in place - there's no point.  Will fix+repush.

PR #713, opened 2017-02-02
HLSL: fix copies between arrays of structs of builtins, and arrayed b…

steve-lunarg:
…uiltins.

Structs are split to remove builtin members to create valid SPIR-V.  In this process, an outer structure array dimension may be propegated onto the now-removed builtin variables.  For example, a mystruct[3].position ->
position[3].  The copy between the split and unsplit forms would handle this in some cases, but not if the array dimension was at different levels of aggregate.

It now does this, but may not handle arbitrary composite types.  Unclear if that has any semantic meaning for builtins anyway.

### DIFF
--- a/Test/baseResults/hlsl.struct.split.assign.frag.out
+++ b/Test/baseResults/hlsl.struct.split.assign.frag.out
@@ -37,6 +37,18 @@ gl_FragCoord origin is upper left
 0:7                0 (const int)
 0:7            Constant:
 0:7              0 (const int)
+0:7        move second child to first child (temp 4-component vector of float)
+0:7          pos: direct index for structure (temp 4-component vector of float)
+0:7            direct index (temp structure{temp float f, temp 4-component vector of float pos})
+0:?               'input' (temp 3-element array of structure{temp float f, temp 4-component vector of float pos})
+0:7              Constant:
+0:7                0 (const int)
+0:7            Constant:
+0:7              1 (const int)
+0:7          direct index (in 4-component vector of float FragCoord)
+0:?             'input_pos' (in 3-element array of 4-component vector of float FragCoord)
+0:7            Constant:
+0:7              0 (const int)
 0:7        move second child to first child (temp float)
 0:7          f: direct index for structure (temp float)
 0:7            direct index (temp structure{temp float f, temp 4-component vector of float pos})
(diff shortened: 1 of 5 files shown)


Test/baseResults/hlsl.struct.split.assign.frag.out
```diff
@@ -184,14 +256,14 @@ gl_FragCoord origin is upper left
...
-                              Name 62  "input_pos"
+                              Name 48  "input_pos"
+                              Name 67  "@entryPointOutput"
+                              Name 68  "param"
+                              Name 70  "param"
                               Decorate 32(i) Location 0
                               Decorate 39(input) Location 1
-                              Decorate 54(@entryPointOutput) Location 0
-                              Decorate 62(input_pos) BuiltIn FragCoord
+                              Decorate 48(input_pos) BuiltIn FragCoord
```
johnkslang: Is an array of FragCoord supposed to mean something? I don't think it is valid SPIR-V, and wouldn't be surprised if it is rejected.

I see it is true to the test source though, and may be an interesting test.

However, if this is a shader that is supposed to work, I wonder if it should become a scalar.

steve-lunarg: That one is an invented test to test some splitting functionality, and doesn't necessarily make semantic sense.

Test/baseResults/hlsl.struct.split.nested.geom.out
```diff
@@ -267,37 +341,53 @@ output primitive = triangle_strip
...
+              66:    7(fvec4) Load 65
+              67:     33(ptr) AccessChain 42(tin) 64 27
+                              Store 67 66
+              68:     54(ptr) AccessChain 53(tin) 64 27
+              69:    8(fvec2) Load 68
+              70:     39(ptr) AccessChain 42(tin) 64 35
+                              Store 70 69
+              73:          12 Load 42(tin)
+                              Store 72(param) 73
+              75:           2 FunctionCall 24(@main(struct-PS_IN-vf4-vf21[3];struct-GS_OUT-struct-PS_IN-vf4-vf21-struct-STRUCT_WITH_NO_BUILTIN_INTERSTAGE_IO-f1[2]-i111;) 72(param) 74(param)
```
johnkslang: The output parameter, and copy if it to the IO output seems missing, though that is probably a separate problem. I'll look at that after accepting this.

PR #720, opened 2017-02-06
Added --vn option to generate a C header file containing a variable a…

flaviobortot:
…ssigned to the hex representation of the shader. This is a standard feature on Microsoft's HLSL compiler and it allows developers to include pre-compiled shaders directly into the code. This option enables "Hex output", so it is NOT required to specify -x as well. The output file name is preserved, so no ".h" extension is added. If you want the output file to have ".h" extension then you have to specify it on the output file name. The generated header file uses the "#pragma once" pragma to avoid multiple inclusions.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -5254,11 +5254,15 @@ void OutputSpvBin(const std::vector<unsigned int>& spirv, const char* baseName)
 }
 
 // Write SPIR-V out to a text file with 32-bit hexadecimal words
-void OutputSpvHex(const std::vector<unsigned int>& spirv, const char* baseName)
+void OutputSpvHex(const std::vector<unsigned int>& spirv, const char* baseName, const char* varName)
 {
     std::ofstream out;
     out.open(baseName, std::ios::binary | std::ios::out);
     out << "\t// " GLSLANG_REVISION " " GLSLANG_DATE << std::endl;
+	if (varName != nullptr) {
+		out << "\t #pragma once" << std::endl;
+		out << "const uint32_t " << varName << "[] = {" << std::endl;
+	}
     const int WORDS_PER_LINE = 8;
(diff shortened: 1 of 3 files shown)


SPIRV/GlslangToSpv.cpp
```diff
@@ -5254,11 +5254,15 @@ void OutputSpvBin(const std::vector<unsigned int>& spirv, const char* baseName)
 }
 
 // Write SPIR-V out to a text file with 32-bit hexadecimal words
-void OutputSpvHex(const std::vector<unsigned int>& spirv, const char* baseName)
+void OutputSpvHex(const std::vector<unsigned int>& spirv, const char* baseName, const char* varName)
 {
     std::ofstream out;
     out.open(baseName, std::ios::binary | std::ios::out);
     out << "\t// " GLSLANG_REVISION " " GLSLANG_DATE << std::endl;
+	if (varName != nullptr) {
```
johnkslang: No tabs please; use four spaces per indent.

StandAlone/StandAlone.cpp
```diff
@@ -302,7 +303,20 @@ void ProcessArguments(int argc, char* argv[])
                     } else if (lowerword == "no-storage-format" || // synonyms
                                lowerword == "nsf") {
                         Options |= EOptionNoStorageFormat;
-                    } else if (lowerword == "source-entrypoint" || // synonyms
+					}
+					else if (lowerword == "variable-name" || // synonyms
```
johnkslang: use `} else` on the same line.

StandAlone/StandAlone.cpp
```diff
@@ -302,7 +303,20 @@ void ProcessArguments(int argc, char* argv[])
...
+						lowerword == "vn") {
+						Options |= EOptionOutputHexadecimal;
+						variableName = argv[1];
+						if (argc > 0) {
+							argc--;
+							argv++;
+						}
+						else
```
johnkslang: `} else` and spaces, not tabs.

PR #722, opened 2017-02-09
Add basic HS/DS implementation.

steve-lunarg:
This obsoletes WIP PR #704, which was built on the pre entry point wrapping master.  New version here uses entry point wrapping.

This is a limited implementation of tessellation shaders.  In particular, the following are not functional, and will be added as separate stages to reduce the size of each PR.

* patchconstantfunctions accepting per-control-point input values, such as   const OutputPatch <hs_out_t, 3> cpv are not implemented.

* patchconstantfunctions whose signature requires an aggregate input type such as   a structure containing builtin variables.  Code to synthesize such calls is not yet present.

* patchconstantfunction return values are handled, but output params are not yet copied to shader outputs.

These restrictions will be relaxed as soon as possible.  Simple cases can compile now: see for example Test/hulsl.hull.1.tesc - e.g, writing to inner and outer tessellation factors.

PCF invocation is synthesized as an entry point epilogue protected behind a barrier and a test on invocation ID == 0.  If there is an existing invocation ID variable it will be used, otherwise one is added to the linkage.  The PCF and the shader EP interfaces are unioned and builtins appearing in the PCF but not the EP are also added to the linkage and synthesized as shader inputs. Parameter matching to (eventually arbitrary) PCF signatures is by builtin variable type.  Any user variables in the PCF signature will result in an error.  Overloaded PCF functions will also result in an error.

[domain()], [partitioning()], [outputtopology()], [outputcontrolpoints()], and [patchconstantfunction()] attributes to the shader entry point are in place, with the exception of the Pow2 partitioning mode.

_Note to reviewers_: I'm not very happy with the TQualifier tracking the original declaration builtin type, but not sure what to do about it, either.  That functionally is not replaced by the HlslParseContext::ioTypeMap, which only enters the picture for structures.  The TType can't be indexed by pointer, and I don't know of any other handle to build an analogous map in the HlslParseContext.


johnkslang:
> TQualifier tracking the original declaration builtin type, but not sure what to do about it

An alternative might be to have a shadow list of original shallow types of the parameters in the TFunction, given the issue here is that we make parameters non-entry-point like before finding out later it really was an entry point.  That localizes the shadowing to the functional area with the problem.

hlsl/hlslParseHelper.h
```diff
@@ -242,6 +243,10 @@ class HlslParseContext : public TParseContextBase {
     void correctUniform(TQualifier& qualifier);
     void clearUniformInputOutput(TQualifier& qualifier);
 
+    // Pass through to base class after remembering builtin mappings.
+    using TParseContextBase::trackLinkage;
+    void trackLinkage(TVariable& variable);
```
johnkslang: Should this say `override`?

glslang/Include/ConstantUnion.h
```diff
@@ -136,6 +143,22 @@ class TConstUnion {
         return false;
     }
 
+    bool operator==(const TString& s) const
+    {
+        if (s == *sConst)
```
johnkslang: `return s == *sConst;` ?  I don't care much, just wondering.

glslang/Include/ConstantUnion.h
```diff
@@ -136,6 +143,22 @@ class TConstUnion {
...
+            return true;
+
+        return false;
+    }
+
+    bool operator==(const char* s) const
```
johnkslang: Does it need a `operator!=` ?

glslang/Include/ConstantUnion.h
```diff
@@ -243,6 +276,12 @@ class TConstUnion {
                 return true;
 
             return false;
+        case EbtString:
```
johnkslang: I see that over here, none are using the more concise `return expr;` syntax. Up to you about the other place.

glslang/Include/Types.h
```diff
@@ -395,6 +395,7 @@ class TQualifier {
         precision = EpqNone;
         invariant = false;
         noContraction = false;
+        builtInOrig  = EbvNone;
```
johnkslang: See comment in the conversation.

hlsl/hlslParseHelper.h
steve-lunarg: I can add override if I make it accept a TSymbol instead of a TVariable.  I'll do that; seems better.

glslang/Include/ConstantUnion.h
steve-lunarg: Yeah, I thought about that at the time but decided to make it match the others.  IMHO we should either change all of them to be that way, or leave this one matching the others.  I can change all of them easily enough; it's just not directly related, so maybe in another PR?

glslang/Include/ConstantUnion.h
steve-lunarg: Yep, should.  (Nothing calls it, but it should be there).  I'll add that.

steve-lunarg:
Repushed with these changes:

* Removed unneeded string methods in TConstUnion.  There are only getSConst/setSConst now: no more comparisons etc.

* Remove tracking of original builtin type from TQualifier.  The TParameter and TFunction
classes now do this.  External callers don't worry about it: it's handled automatically
upon addParameter.

* HlslParseContext::trackLinkage now accepts a TSymbol, and uses the override keyword.

* HlslParseContext::addPatchConstantInvocation now tracks storage qualifier along with
  builtin type, so if the same builtin appears on an in and an out param, the right one will
  be selected when synthesizing the function call.  (Any individual parameter has only
  one storage qualifier).

PR #724, opened 2017-02-14
Implement NVX_multiview_per_view_attributes

chaoc:


Test/baseResults/310.tese.out
```diff
@@ -78,8 +78,8 @@ ERROR: node is still EOpNull!
 0:36        move second child to first child (temp highp 4-component vector of float)
 0:36          'p' (temp highp 4-component vector of float)
 0:36          gl_Position: direct index for structure (in highp 4-component vector of float Position)
-0:36            direct index (temp block{in highp 4-component vector of float Position gl_Position, in highp float PointSize gl_PointSize, in highp 4-component vector of float gl_SecondaryPositionNV})
-0:36              'gl_in' (in 32-element array of block{in highp 4-component vector of float Position gl_Position, in highp float PointSize gl_PointSize, in highp 4-component vector of float gl_SecondaryPositionNV})
+0:36            direct index (temp block{in highp 4-component vector of float Position gl_Position, in highp float PointSize gl_PointSize, in highp 4-component vector of float gl_SecondaryPositionNV, in implicitly-sized array of highp 4-component vector of float gl_PositionPerViewNV})
```
johnkslang: I might not be seeing something correctly here:  Why do tests not using this extension (or any NV extension?) pick up NV-specific decorations?

johnkslang:
Seems good, except that I don't understand unrelated tests (310.tes*) changing their results.

Test/baseResults/310.tese.out
johnkslang: I think I realize now: This (and previous) extension unconditionally extends the per-vertex data structure. I think that is okay, as long as it requires redeclaration to use, and the unused members stay hidden on translation to other layers.

PR #727, opened 2017-02-17
Implement GL_EXT_shader_image_load_formatted

jekstrand:


glslang/MachineIndependent/Versions.cpp
```diff
@@ -275,6 +276,7 @@ void TParseVersions::getPreamble(std::string& preamble)
             "#define GL_OES_texture_buffer 1\n"
             "#define GL_OES_texture_cube_map_array 1\n"
             "#define GL_EXT_shader_non_constant_global_initializers 1\n"
+            "#define GL_EXT_shader_image_load_formatted 1\n"
```
johnkslang: Is this really written against ES too, or just desktop?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4499,7 +4499,8 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
                 }
             }
         }
-    } else if (type.isImage() && ! qualifier.writeonly)
+    } else if (type.isImage() && ! qualifier.writeonly &&
+               !extensionTurnedOn(E_GL_EXT_shader_image_load_formatted))
```
johnkslang: You can get more informative error messages using the common pattern (see other parts of this file):

```
requireProfile(...)
profileRequires(...)
```

Typically, the `extensionTurnedOn` is for taking a different control path (other than error messaging).

This also gives a chance to, say, support it for desktop and not ES, if that's the situation.

johnkslang:
Nice. It looks like the Travis CI error is not caused by your change. I noted a couple places this extension can be refined, especially if it is profile specific, (or the extension gets pulled into core of some profiles in the future).

glslang/MachineIndependent/Versions.cpp
jekstrand: At the moment, just desktop.  However, it would be fairly trivial if someone wanted to update the spec to make it also apply against ES.  Would you like me to disable it on ES for now?  I don't really like glslang having more desktop-only features.

jekstrand:
I've pushed a new version which uses `requiresProfile` and `profileRequires` and doesn't expose it in ES.

johnkslang:
This should probably have changed existing test results, to say the extension is required instead of just giving an error.  Hopefully that's also the reason the bots are failing.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4499,8 +4499,11 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
                 }
             }
         }
-    } else if (type.isImage() && ! qualifier.writeonly)
-        error(loc, "image variables not declared 'writeonly' must have a format layout qualifier", "", "");
+    } else if (type.isImage() && ! qualifier.writeonly) {
+        const char *explanation = "image variables not declared 'writeonly' must have a format layout qualifier";
```
johnkslang: This should describe the feature now:  "image variables declared 'writeonly' without a format layout qualifier"

The improved error message should also show up in existing negative tests for this feature, changing at least one other test file result.

jekstrand: Yup, I've updated the tests now.  It took me a little while to remember how to do it.

PR #730, opened 2017-02-20
Fix #726: Implement GL_EXT_texture_array

### LINKED ISSUE #726: GLSL: Add support for GL_EXT_texture_array
My GLSL 1.20 shaders make use of GL_EXT_texture_array (and require the extension), but glslang doesn't support/export that extension so validation of those shaders fails.

It would be useful to me if glslang supported that extension, for pre-1.30 GLSL shaders.

slime73:
Implementing this has been my only real exposure to this area of the codebase, so please let me know if it can be improved. :)


Test/120.frag
```diff
@@ -243,4 +243,40 @@ void voidTernary()
 	b ? foo121111() : foo12111();
 	b ? foo121111() : 4;  // ERROR
 	b ? 3 : foo12111();   // ERROR
-}
\ No newline at end of file
+}
+
```
johnkslang: Can you add a few calls to the extension built-in functions before the `#extension` so that we see error messages that these are not excepted unless the extension is enabled?

johnkslang:
It looks pretty good to me.  The main thing I'd like to see is some negative testing. That is, that calling these added functions does not work unless using the extension.

Test/120.frag
slime73: Hi. Apologies for the extended delay.

If I'm understanding things correctly, doing negative testing in the same file as the positive tests won't be a good idea because `sampler2DArray` and friends are not reserved keywords in GLSL 1.20, so glslang causes a shader syntax error when parsing them without the extension. The syntax error is correctly written to the test output file, but the positive tests below the syntax error will fail to execute.

Should the negative test be put into a separate (perhaps new) test shader file?

glslang/MachineIndependent/Scan.cpp
```diff
@@ -1117,6 +1114,37 @@ int TScanContext::tokenizeIdentifier()
...
+            return identifierOrType();
+        }
+        return keyword;
+
+    case SAMPLER2DARRAY:
+    case SAMPLER2DARRAYSHADOW:
+        afterType = true;
+        if (parseContext.symbolTable.atBuiltInLevel())
+            return keyword;
+        else if (parseContext.version < 130 && (parseContext.profile == EEsProfile || !parseContext.extensionTurnedOn(E_GL_EXT_texture_array))) {
```
johnkslang: This used to call nonreservedKeyword(), with a couple important points:

- comparisons to version were not made without first saying whether is was for ES or non-ES
- an ES 300 was involved, but got dropped

Is it just picking up that ES 100 happens to be less than desktop 130?  It takes quite a bit of time to verify this logic.

Is it possible instead to write these by first checking if the extension allows it, and if not, keeping existing calls to `nonreservedKeyword()` and `es30ReservedFromGLSL()`?

glslang/MachineIndependent/Scan.cpp
```diff
@@ -1130,15 +1158,6 @@ int TScanContext::tokenizeIdentifier()
         }
         return keyword;
 
-    case SAMPLER1DARRAY:
-        afterType = true;
-        if (parseContext.profile == EEsProfile && parseContext.version == 300)
```
johnkslang: It looks like this should have been a >=300 in the code anyway, which you fixed, so that is good to preserve.

arcady-lunarg:
Closing this PR as there hasn't been any activity in over 5 years.

PR #734, opened 2017-02-22
Add reflection queries for thread local size.  Add ssbo auto-binding.

steve-lunarg:
New command line option --shift-ssbo-binding.  Mirrors --shift-ubo-binding, etc.

New reflection query getLocalSize(int dim) queries CS thread local size.


danginsburg:
Tested with my compute shader test.  Looks good.

PR #736, opened 2017-02-24
HLSL: add structuredbuffer pass by reference in fn params

steve-lunarg:
This PR adds the ability to pass structuredbuffer types by reference as function parameters.

It also changes the representation of structuredbuffers from anonymous blocks with named members, to named blocks with pseudonymous members. That should not be an externally visible change.

Local copies of struct buffer references are not currently handled, as in:
   StructuredBuffer&lt;uint4&gt; sbuf_copy = sbuf;


steve-lunarg:
Repushed, still WIP, with proper type sharing between globally scoped variables and fn param types to make the validator happy.

PR #737, opened 2017-02-25
HLSL - Add half to float aliasing for literals when actual half16 support is disabled

gwihlidal:
Add half to float aliasing for literals when actual half16 support is disabled. It would be nice to also add type name aliasing from half to float in this case, but it's easy enough to add "#define half float" as a shader preamble, and this change then allows 'h' and 'H' literals to compile.

Previously the errors would be thrown:
```
D:\ExampleData\Source\Systems\Shader\Common.hlsl:861: error: ')' : Expected
D:\ExampleData\Source\Systems\Shader\Common.hlsl:861: error: 'function call arguments' : Expected
D:\ExampleData\Source\Systems\Shader\Common.hlsl:861: error: ';' : Expected
D:\ExampleData\Source\Systems\Shader\Common.hlsl:861: error: 'h' : unknown variable
D:\ExampleData\Source\Systems\Shader\Common.hlsl:861: error: ';' : Expected
```

This case now compiles:
```
#define half float

half fbSelect(bool cnd, half src0, half src1)
{
	return cnd ? src0 : src1;
}

half overlayBlend(half value1, half value2)
{
	return fbSelect(value1 < 0.5h, 2.0h*value1*value2, 1.0h - 2.0h*(1.0h-value1)*(1.0h-value2));
}
```

Thanks!
Graham


johnkslang:
I don't think we can do that, wouldn't incorrectly change semantics of GLSL validation? Maybe it needs to be tied to 

```
parseContext.intermediate.getSource() == EShSourceHlsl
```

Would also want to check the other places looking for `F` and see why the also shouldn't (or should) be checking for `H`.

johnkslang:
FYI, I'm doing a full implementation/test of this now.

PR #744, opened 2017-03-01
Sanitize ShaderLang.h

realitix:
`ShaderLang.h` contains declaration for `ShLink` function which is never
defined.

See this grep output:

```
$ grep -Hrn 'ShLink'
glslang/Public/ShaderLang.h:206:SH_IMPORT_EXPORT int ShLink(
glslang/Public/ShaderLang.h:214:SH_IMPORT_EXPORT int ShLinkExt(
glslang/MachineIndependent/ShaderLang.cpp:1268:int ShLinkExt(
```

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -203,14 +203,6 @@ SH_IMPORT_EXPORT int ShCompile(
     EShMessages messages = EShMsgDefault // warnings and errors
     );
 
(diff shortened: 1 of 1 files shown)


johnkslang:
Yes all the linker stuff should be dead at this point. It is historical (from use of this compiler to drive back-end compilation as well). It's conceivable someone still uses it, but I'd actually like to remove all the ShLink stuff.  Will start with this.

PR #745, opened 2017-03-01
Add comments in ShaderLang.h

realitix:
`ShCompile` function is not so easy to use and need more explanations.

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -190,13 +190,16 @@ SH_IMPORT_EXPORT void ShDestruct(ShHandle);
 // The info-log should be written by ShCompile into
 // ShHandle, so it can answer future queries.
 //
+// You can put all your shader content inside shaderStrings[0] and
+// set numStrings to 1 or you can split it into several strings.
+//
 SH_IMPORT_EXPORT int ShCompile(
-    const ShHandle,
-    const char* const shaderStrings[],
-    const int numStrings,
-    const int* lengths,
-    const EShOptimizationLevel,
-    const TBuiltInResource *resources,
+    const ShHandle,                      // compiler object
+    const char* const shaderStrings[],   // shader splitted in numStrings part
+    const int numStrings,                // length of shaderStrings
+    const int* lengths,                  // list of lengths (corresponding to each shaderStrings), can be NULL to use \0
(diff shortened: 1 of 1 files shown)


johnkslang:
This path is deprecated. Is this a legacy use?  Hopefully, any new use would be through TShader.parse(), and hopefully that interface makes more sense.  See comments in the middle of ShaderLang.h.

glslang/Public/ShaderLang.h
```diff
@@ -190,13 +190,16 @@ SH_IMPORT_EXPORT void ShDestruct(ShHandle);
...
+//
 SH_IMPORT_EXPORT int ShCompile(
-    const ShHandle,
-    const char* const shaderStrings[],
-    const int numStrings,
-    const int* lengths,
-    const EShOptimizationLevel,
-    const TBuiltInResource *resources,
+    const ShHandle,                      // compiler object
+    const char* const shaderStrings[],   // shader splitted in numStrings part
```
johnkslang: -> shader split into 'numString's parts

glslang/Public/ShaderLang.h
```diff
@@ -190,13 +190,16 @@ SH_IMPORT_EXPORT void ShDestruct(ShHandle);
...
+    const int numStrings,                // length of shaderStrings
```
johnkslang: -> number of shader strings

(The variable name was more accurate.)

glslang/Public/ShaderLang.h
```diff
@@ -190,13 +190,16 @@ SH_IMPORT_EXPORT void ShDestruct(ShHandle);
...
+    const int* lengths,                  // list of lengths (corresponding to each shaderStrings), can be NULL to use \0
```
johnkslang: NULL means a normal 0-terminated C string was passed in

realitix:
@johnkslang I am doing a Python wrapper with Cffi so I need C wrapper. But after 2 days, I have realized that this C API is not usable. The C header can't compile, I don't manage to compile a single glsl shader to Spir-V... So I now use libshaderc which provide good C headers.
Maybe this C header must be removed and the documentation updated to redirect to shaderc for C header.
What do you think about that?

johnkslang:
@realitix I created issue #751 for this, thanks.

PR #761, opened 2017-03-09
Compile fixes for VS2010

baldurk:
* Removed range-based for
* Added explicit return type to non-trivial lambda
* Added explicit scope for `tInterstageIoData`

The last one I presume is just a compiler scoping quirk - it's a struct member of `HlslParseContext` so the lambda doesn't seem to have default access to it.

Tests all look OK, although it's a pretty trivial set of changes so I'd hope it's safe.

FWIW I'm probably dropping VS2010 soon for my project, because most other dependencies are sadly increasingly unsupported on that version. Take that as you wish for your own compiler support decisions :smile:.

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -1892,7 +1892,7 @@ void HlslParseContext::remapEntryPointIO(TFunction& function, TVariable*& return
 {
     // Do the actual work to make a type be a shader input or output variable,
     // and clear the original to be non-IO (for use as a normal function parameter/return).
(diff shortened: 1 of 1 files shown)


johnkslang:
If you dropped VS2010, what would you step up to?

baldurk:
I'm planning to go all the way up to VS2015. In theory I could maybe get away with moving to say VS2013 currently, but there's not much point in doing a half-way step like that as it just means I need to repeat the process that much sooner if Qt drops 2013 in a year or whatever.

I'm not going up to VS2017 since it's just come out and not everyone will have it, but I think it's safe to assume by now that hobbyists and professionals will have 2015, and I don't need bleeding edge standards support from 2017.

PR #762, opened 2017-03-09
Added 'GL_core_profile' and 'GL_compatibility_profile' to macro expansion

LukasBanana:
As stated in the [OpenGL wiki](https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#Standard_macros), `GL_core_profile` is always defined as standard macro to "1", and `GL_compatibility_profile` is only defined as standard macro to "1", if the version profile was set to "compatibility".

**Note:**
All standard macros (i.e. `__VERSION__`, `__LINE__`, `__FILE__`, `GL_core_profile`, and `GL_compatibility_profile`) are still not identified in an '#ifdef' or '#if defined' directive!
I.e. the following pre-processor directives will not result in a compiler error, but it should be:
```
#ifdef __VERSION__
#error foo bar
#endif
```

### DIFF
--- a/Test/150.vert
+++ b/Test/150.vert
@@ -1,5 +1,9 @@
 #version 150 core
 
+#ifndef GL_core_profile
+#	error standard macro GL_core_profile not defined
+#endif
+
 in vec4 iv4;
 
 uniform float ps;
--- a/Test/baseResults/150.vert.out
+++ b/Test/baseResults/150.vert.out
@@ -1,44 +1,44 @@
 150.vert
-ERROR: 0:22: 'a' : cannot redeclare a user-block member array 
+ERROR: 0:26: 'a' : cannot redeclare a user-block member array 
 ERROR: 0:3001: '#error' : line of this error should be 3001  
 ERROR: 2 compilation errors.  No code generated.
 
 
 Shader version: 150
 ERROR: node is still EOpNull!
-0:9  Function Definition: main( (global void)
-0:9    Function Parameters: 
-0:11    Sequence
-0:11      move second child to first child (temp 4-component vector of float)
-0:11        gl_Position: direct index for structure (invariant gl_Position 4-component vector of float Position)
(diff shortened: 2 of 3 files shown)


johnkslang:
Why wouldn't these be done in Versions.cpp, in `getPreamble()`? They are not special tokens to the preprocessor.

LukasBanana:
Reverted previous changes and now added 'GL_core_profile' and 'GL_compatibility_profile' macro definitions to preamble.

johnkslang:
That looks much closer.  These were not defined until `#version 150`, so should be protected by a test for that:

```
if (version >= 150)
    ...
```

Also, can you add a test that consumes them that shows it worked?  Alternatively, I don't mind doing this, if it's not worth figuring out the test system for this simple change.  I'd probably add this to `150.vert`.

LukasBanana:
It looks like `spvVersion.openGl` is always 0 at the moment `TParseVersions::getPreamble` is called, so the check `if (spvVersion.openGl >= 150)` does not work.

johnkslang:
It should not be a `spvVersion` check. Just a `version` check. It is orthogonal to SPIR-V support. `version` holds the `#version` number in it.

LukasBanana:
Should work now after lots of changes, but that's a good way to get used to the framework :)

johnkslang:
The code is looking good.  I assume the tests will fail, because there is no test result update included in the PR.  Need to run the google-test suite with `--update-mode`, verify the resulting test-result changes, and include them.  There is a bit more detail about that in the readme.

LukasBanana:
Tests have all passed now on my system.

johnkslang:
Final request: make this a single commit, instead of 7.  You can squash and force push.
(I tried the squash, but it wouldn't record you as the committer, so better for you to do it.)

LukasBanana:
Yeah, my first git squash \o/
In my own projects I didn't have to merge in many cases so far :)

johnkslang:
Looks good, thanks.

PR #763, opened 2017-03-10
HLSL: preserve empty structures after splitting

steve-lunarg:
WIP pending discussion.  Also, this may be just a subset of a recently discussed problem.

If a user structure contained only builtin types, those types were assigned to their respective builtins, but the resulting empty structure was being dropped (e.g, not appearing in the linkage).  That created an interface mismatch: e.g, if the output side of the interface dropped the structure, the input side was expecting an empty structure on input and didn't see one.

Note: this would expose consuming code to more empty structures, but correctly so.


steve-lunarg:
Removing WIP: this was tested and fixes a validator complaint about interface mismatch.

PR #766, opened 2017-03-11
Memory leaks fixed in TShader and TProgram

sergeyreznik:
There were several places, where TPoolAllocator being unconditionally created inside TShader and TProgram. I've added check for existing allocator and also I'm preserving previous one (and restoring it in TShader/TProgram dtor)


johnkslang:
I might not be fully understanding this yet.  At first glance, it doesn't seem like it would work, or would only work for a particular pattern of creating/destroying programs and shaders. So, I haven't been sure what to suggest yet, so I'll just ask.

Is this intended to be insensitive to order of creation of a bunch of programs and shaders, and say, keep working correctly after half have been deleted, but the other half continue to get used?  Or, is this implying there is some rule for the pattern of create/use/destroy in order to work correctly?

Thanks.

sergeyreznik:
@johnkslang 
Here is my flow (HLSL -> SPIR): I'm creating shader and call "preprocess" method to get "clean" source, after that I'm validating preprocessed HLSL with DX compiler, and only then calling "parse" method for shader. So, looks like sequential calls to preprocess and the parse causes memory leaks. 

I'm aware this is not best way from performance point of view, but this is debug flow for now.

jantlo:
Hi, maybe you want to take a look to my pull request #826. I think it addresses this issue as well. Part of then changes is to clearly define who has the ownership of the pool in any moment.

johnkslang:
These leaks have been fixed by recent commits that made shaders, compilers, linkers, and programs all own their own pool creation and deletion. There is no longer an "original" per-thread pool to track and free.

PR #767, opened 2017-03-12
Fix building on MinGW-w64

DragoonX6:
Some HLSL to Spirv tests fail.

I don't have sufficient knowledge about Vulkan, HLSL or Spirv to fix this, so I don't feel comfortable making changes to the code in order to try and make the test pass.

I've built it with both MinGW-w64 Clang and MSVC to make sure nothing broke, even if very unlikely from such a small change.

This is the output from the failing tests:
```
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_attribute_expression_comp, where GetParam() = 16-byte object <7F-75 62-00 00-00 00-00 56-75 62-00 00-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_forLoop_frag, where GetParam() = 16-byte object <16-77 62-00 00-00 00-00 0B-75 62-00 00-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_struct_split_array_geom, where GetParam() = 16-byte object <6D-83 62-00 00-00 00-00 56-75 62-00 00-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_struct_split_trivial_geom, where GetParam() = 16-byte object <E2-83 62-00 00-00 00-00 56-75 62-00 00-00 00-00>
```


johnkslang:
I don't understand yet. Does the PR fix 4 failing tests from MinGW-w64, or do you think the change is needed, but it causes the failures?

I think @dneto0 does some MinGW work, do you have input on this?

DragoonX6:
The change is needed because it would otherwise add flags that gcc and clang do not understand and thus error out on.
So in short, the PR fixes the build process, which is needed to get the vulkan SDK compiled with MinGW-w64.
I have [opened an issue there](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/1544) about having made changes that get it to build, any problems strictly related to vulkan I don't know how to appropriately deal with.

PR #774, opened 2017-03-15
HLSL: support per control point patch const fn invocation

steve-lunarg:
This PR emulates per control point inputs to patch constant functions. Without either an extension to look across SIMD lanes or a dedicated stage, the emulation must use separate invocations of the wrapped
entry point to obtain the per control point values.  This is provided since shaders are wanting this functionality now, but such an extension is not yet available.

Entry point arguments qualified as an invocation ID are replaced by the current control point number when calling the wrapped entry point.  There is no particular optimization for the case of the entry point not having
such an input but the PCF still accepting ctrl pt frequency data.  It'll work, but anyway makes not so much sense semantically.

The wrapped entry point must return the per control point data by value. At this time it is not supported as an output parameter.


steve-lunarg:
Repush (still WIP) with linkage fixes:

- SPIR-V requires fixed array sizes for tessellation factors: 4 (outer) and 2 (inner).  HLSL syntax allows sizes such as 3, or scalars.  The externally facing variable now abides the SPIR-V requirements, and the internal form is copied to a possible subset of it in the wrapped entry point.

- Location assignment now strips off the array dimension from arrayed IO variables so locations match between stages HS/DS stages.

- The patch qualifier is set on per-patch data.

- Tessellation coordinate (SV_DomainLocation) is no longer being silently ignored.

- Patch constant function output is now added to the linkage (it was being ignored before).

**There is currently a limitation: the patch constant function output must be given as the LAST parameter to the domain shader entry point if it is read.  Anything else will not work.**

steve-lunarg:
Small re-push for matching 'patch' qualifiers on patch frequency user data output from the HS's PCF and input by DS inputs.

steve-lunarg:
Re-push adding HS return value translation to an array of same type, for matching SPIR-V rules.  It's indexed by invocation ID, either from an existing entry point input if one exists, or an introduced variable otherwise.

steve-lunarg:
Repush with fix to pass topology data through to domain shader SPIR-V.  There was a hardcoded assumption it would be in the hull shader, but HLSL allows it (possibly requires it) to be in the hull shader, and SPIR-V allows it to be in either stage.

steve-lunarg:
Removed WIP.  Repushed after rebase and minor code cleanup.

**NOTE: Known limitations remaining:**

- Patch constant function's output, if read by the domain shader, must be the last entry point parameter.  Failure to do this may result in a link validation error.

- PatchInput<> types cannot be passed to a patch constant function.

- Tessellation coordinate (SV_DomainLocation) qualified builtin is always a vec3, which is wrong for some topology types (e.g, should be a vec2).

These will be addressed in subsequent PRs.

Note also that hardware compilers are still addressing some issues in this space.

hlsl/hlslParseHelper.cpp
```diff
@@ -1381,6 +1383,34 @@ void HlslParseContext::trackLinkage(TSymbol& symbol)
 }
 
 
+// Some types require fixed array sizes in SPIR-V, but can be scalars or
+// arrays of sizes SPIR-V doesn't allow.  For example, tessellation factors.
+// This creates the right size.  A conversion is performed when the internal
+// type is copied to or from the external
```
steve-lunarg: Yeah; will fix and repush.

hlsl/hlslParseHelper.cpp
```diff
@@ -1381,6 +1383,34 @@ void HlslParseContext::trackLinkage(TSymbol& symbol)
...
+        return;
+    }
+
+    if (type.isArray()) {
+        // Already an array.  Fix the size.
+        type.changeOuterArraySize(requiredSize);
+    } else {
+        // it wasn't an array, but needs to be.
+        TArraySizes arraySizes;
+        arraySizes.addInnerSize(requiredSize);
```
steve-lunarg: I think this should be OK, because the type being changed is not arbitrary.  It can only be tessellation factors.  The two cases are:

- It's a scalar, and needs to become an array
- It's an array of the wrong size, and the size must be fixed (but no more dimensions added).

SPIRV/GlslangToSpv.cpp
```diff
@@ -869,14 +869,20 @@ TGlslangToSpvTraverser::TGlslangToSpvTraverser(const glslang::TIntermediate* gls
         builder.addCapability(spv::CapabilityShader);
         break;
 
+    case EShLangTessEvaluation:
```
steve-lunarg: Yes, GLSL gets the same results since it never had the topology data present in the other stage.  This merely lets it be in either stage (as does SPIR-V), and it's up to the translator (GLSL, HLSL, etc) which stage it goes in, or if it goes in both as allowed by the spec as long as there's no inconsistency.

Test/baseResults/hlsl.basic.geom.out
```diff
@@ -173,7 +173,7 @@ output primitive = line_strip
                               Name 53  "param"
                               Name 55  "param"
                               Decorate 45(VertexID) Location 0
-                              Decorate 48(test) Location 3
+                              Decorate 48(test) Location 1
```
steve-lunarg: Yes, exactly.  We now remove the arrayness dimension when assigning locations on things for which isArrayedIo() returns true.

steve-lunarg:
Repush with fix for truncated comment.

PR #779, opened 2017-03-17
HLSL: Buffer types should be unsampled.

steve-lunarg:
RWBuffer objects were generating (properly) unsampled resoures, while Buffer objects were marked as sampled when they should not be.


steve-lunarg:
Repush with fix for dropped readonly qualifier on buffer types.

steve-lunarg:
Removed WIP after testing in duplicating environment.

PR #801, opened 2017-03-28
HLSL: Fix an issue of frexp().

amdrexu:
The "exp" parameter is floating-point type in HLSL intrinsic while it is integer type in GLSL built-in function.


amdrexu:
The output test result file is somewhat messy. The major change is a OpConvertSToF is inserted to correct the result from extended instruction FrexpStruct.

SPIRV/SpvBuilder.h
```diff
@@ -134,6 +134,9 @@ class Builder {
     bool isSampledImage(Id resultId) const { return isSampledImageType(getTypeId(resultId)); }
 
     bool isBoolType(Id typeId)         const { return groupedTypes[OpTypeBool].size() > 0 && typeId == groupedTypes[OpTypeBool].back()->getResultId(); }
+    bool isIntType(Id typeId)          const { return getTypeClass(typeId) == OpTypeInt && module.getInstruction(typeId)->getImmediateOperand(1); }
```
johnkslang: Because `getImmediateOperand` is not a bool, I'd rather see these as `== 0` and `!= 0`.

johnkslang:
The code looks good (with one small request for a minor tweak).

For testing, I think it looks okay, but it is difficult. There is a way to be more sure, with two commits instead. Each of this can be one commit:

1. Remove `frexp` (whatever is affected by this change) from the existing test, and put it into a new test file. This commit would show differences in tests only, which we'd have confidence in, because of no code changes.
2. Apply the code change.  We'll see no test differences in the original files, and more clearly see the fix in the new test file.

amdrexu:
Thanks for your suggestions. I have split the change to two ones. Please have a look.

PR #806, opened 2017-03-29
Don't emit NV-specific interface members if NV extensions are disabled.

amdrexu:
Add some missing handling.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2272,10 +2272,19 @@ bool TGlslangToSpvTraverser::filterMember(const glslang::TType& member)
 {
     auto& extensions = glslangIntermediate->getRequestedExtensions();
 
+    if (member.getFieldName() == "gl_ViewportMask" &&
+        extensions.find("GL_NV_viewport_array2") == extensions.end())
+        return true;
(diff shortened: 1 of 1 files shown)


johnkslang:
I accepted this, assuming it helps you, but I believed these symbols are a different category than the ones in the blocks, and have available a better solution than this.  Perhaps not, but without having tests present to show the delta it is harder to see what is getting fixed.

We should revisit this with tests, and possible better solutions.  Ideally, the AST never has these to begin with if the extension was not enabled.  (Much harder to do for built-in blocks, but possible.)  For now, we are accepting that expedience is more important than getting the ideal solution.

amdrexu:
Yes, you are right. I admit we should have a formal solution to get rid of those built-in declarations from AST if they should not appear. The problematic shader is very simple (a passthrough tessellation control shader).

```
#version 450 core

layout(vertices = 3) out;

void main()
{
    gl_TessLevelOuter[0] = 63.0;
    gl_TessLevelOuter[1] = 63.0;
    gl_TessLevelOuter[2] = 63.0;
    gl_TessLevelOuter[3] = 63.0;
    gl_TessLevelInner[0] = 63.0;
    gl_TessLevelInner[1] = 63.0;

    **gl_out[gl_InvocationID].gl_Position = gl_in[gl_InvocationID].gl_Position;**
}
```

PR #808, opened 2017-03-30
HLSL: fix GS to work with EP wrapping

steve-lunarg:
The prior implementation of GS did not work with the new EP wrapping architecture, since the vertex emit happened to an internal sanitized type. This fixes that: the Append() method now uses the actual GS output rather than the internal sanitized temporary type, and writes to that.


steve-lunarg:
Updated to use pseudo-builtin variable type, removed WIP.

This may have a (trivial) merge conflict with #812, which I'll fix on that branch if needed.

steve-lunarg:
Remerge, repush, no changes.

PR #812, opened 2017-04-01
HLSL: Support InputPatch variables in patch constant functions

steve-lunarg:
Previously, patch constant functions only accepted OutputPatch.  This adds InputPatch support, via a pseudo-builtin variable type, so that the patch can be tracked clear through from the qualifier.


steve-lunarg:
Repush after resolving minor merge conflict.

PR #813, opened 2017-04-02
Cpp11 feat

jantlo:
Willing to contribute to this project, first pull request on it. This pull request basically contains changes affecting StandAlone application (glslangvalidator). Basically using/adding more C++11 features. The pull request should contain 4 patches:
- The first one is just applying the automatic coding style in order to reduce the number of changes for the next reviews.
 - The second one is replacing the global mutex in WorkList by the one from the std library and store it as a member variable (just in case several instances are used). This change is fixing an issue in remove function where the mutex were not released in case of being the wok list empty.
- The third patch enables concurrency processing in non-windows platforms and uses std::threads instead the OS dependent thread code.
- And the forth one just refactor some code in order to convert the workList global variable in a local one. Since this variable has the ownership of all the WorkItems, it has been replaced by an std::vector<std::unique_ptr< glslang::WorkItem >>. This should make the code easier to maintain since the memory management is handle automatically.

For testing these changes basically I ran the glslangvalidator application (only on a GNU/Linux machine) with a bunch of shaders from the Test folder passing the -t option in order to enable multithreading, with and without my changes, and no regression has been found. 

Please fell free to apply this pull request and any comment/feedback will be really welcome. 

Thanks,
    Juan

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -154,13 +154,6 @@ void ProcessConfigFile()
         delete[] config;
 }
 
-// thread-safe list of shaders to asynchronously grab and compile
-glslang::TWorklist Worklist;
-
-// array of unique places to leave the shader names and infologs for the asynchronous compiles
-glslang::TWorkItem **Work = 0;
-int NumWorkItems = 0;
-
 int Options = 0;
 const char *ExecutableName = nullptr;
 const char *binaryFileName = nullptr;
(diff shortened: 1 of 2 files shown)


johnkslang:
This includes a large amount of coding-convention change that we would not want to accept.

Can you make a new submission, containing a single commit, that shows the desired change?

jantlo:
I can remove the first commit where the automatic coding style is applied. This will left the other 3 commits in order to make the changes a much atomic as possible. Is this fine or do you consider is better to squash all of them in just one commit?

PR #816, opened 2017-04-03
Cpp11 feat

jantlo:
This pull request basically contains changes affecting StandAlone application (glslangvalidator). Basically using/adding more C++11 features. The pull request should contain 3 patches:

-The first one is replacing the global mutex in WorkList by the one from the std library and store it as a member variable (just in case several instances are used). This change is fixing an issue in remove function where the mutex were not released in case of being the wok list empty.

-The second patch enables concurrency processing in non-windows platforms and uses std::threads instead the OS dependent thread code.

-And the third one just refactor some code in order to convert the workList global variable in a local one. Since this variable has the ownership of all the WorkItems, it has been replaced by an std::vector<std::unique_ptr< glslang::WorkItem >>. This should make the code easier to maintain since the memory management is handle automatically.

For testing these changes basically I ran the glslangvalidator application (only on a GNU/Linux machine) with a bunch of shaders from the Test folder passing the -t option in order to enable multithreading, with and without my changes, and no regression has been found.

Please fell free to apply this pull request and any comment/feedback will be really welcome.

Thanks,
Juan

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -51,6 +51,7 @@
 #include <cctype>
 #include <cmath>
 #include <array>
+#include <thread>
 
 #include "../glslang/OSDependent/osinclude.h"
 
@@ -420,9 +421,7 @@ void ProcessArguments(int argc, char* argv[])
                 Options |= EOptionSuppressInfolog;
                 break;
             case 't':
-                #ifdef _WIN32
-                    Options |= EOptionMultiThreaded;
-                #endif
+                Options |= EOptionMultiThreaded;
                 break;
             case 'v':
                 Options |= EOptionDumpVersions;
@@ -489,13 +488,13 @@ void SetMessageOptions(EShMessages& messages)
 //
 // Return 0 for failure, 1 for success.
 //
-unsigned int CompileShaders(void*)
+void CompileShaders()
 {
     glslang::TWorkItem* workItem;
     while (Worklist.remove(workItem)) {
         ShHandle compiler = ShConstructCompiler(FindLanguage(workItem->name), Options);
         if (compiler == 0)
-            return 0;
+            return;
 
         CompileFile(workItem->name.c_str(), compiler);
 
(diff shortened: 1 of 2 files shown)


jantlo:
I removed the first review that was applying the coding style to the modified files.

johnkslang:
One commit per functional change is okay, so would one commit total for this.  Couple of other notes:

1. Not all platforms were supporting all C++11 features, so we haven't been whole-sale using as many as we can. I see novel ones in your commits. Do you know what version of visual studio (on Windows) they work with?

2. There are still coding convention modifications in the commits.

jantlo:
Only std::thread, std::mutex and std::unique_ptr are used on this request. They should work fine with at least Visual Studio 2013 and according to the CI it can be compiled with no errors.  Do you know what compiler must be officially supported?

Sorry, regarding the coding conventions, could you point to some of them? I am using clang-format so it must apply automatically the coding style. So I guess it is something related with the variable naming.

StandAlone/Worklist.h
```diff
@@ -56,26 +57,21 @@ namespace glslang {
...
-            GetGlobalLock();
-
+            std::lock_guard<std::mutex> guard(mutex);
             worklist.push_back(item);
-
-            ReleaseGlobalLock();
         }
 
-        bool remove(TWorkItem*& item)
+        bool remove(TWorkItem *&item)
```
johnkslang: Coding style is to keep `*` and `&` with the type, not the variable.

StandAlone/StandAlone.cpp
```diff
@@ -440,8 +429,7 @@ void ProcessArguments(int argc, char* argv[])
         } else {
             std::string name(argv[0]);
             if (! SetConfigFile(name)) {
-                Work[argc] = new glslang::TWorkItem(name);
-                Worklist.add(Work[argc]);
+                workItems.emplace_back(new glslang::TWorkItem(name));
```
johnkslang: We've had past experience of emplace not being supported on all platforms.

Issue #786, opened 2017-03-20
HLSL - isfinite() not supported for shader other than OpenCL


redorav:
Hi,
I have this minimal shader:

```
cbuffer Example
{
	float4 input;
};

float main() : SV_Target0
{
    // I need a possible source of nan/infinity or fxc will ignore the isfinite test
    float testValue = input.x / input.y;
    if(isfinite(testValue))
        return input.z;
    return input.w;
}
```

This doesn't work because (I'm told from SPIRV-Cross) isfinite isn't defined for shaders other than Kernel. Is there a chance we could translate this to !isnan(value) && !isinf(value) for other shaders?

steve-lunarg:
Ah, so it's not.  #817.  It's decomposed unconditionally, though strictly speaking it could pass through for CS.

PR #817, opened 2017-04-04
HLSL: Decompose OpIsFinite to avoid capability restrictions

steve-lunarg:
OpIsFinite is not available everywhere: decompose isfinite to !isinf && !isnan.


johnkslang:
Helpful tip: If a merge request fixes an issue, say #786, and the commit message says "Fixes #786", then GitHub automatically connects the two and closes the issue.

(Generally, even if unsure of closing an issue, it's nice for the commit message to mention an issue number, like "Addresses #786".)

PR #818, opened 2017-04-04
Refactor the StandAlone.cpp and WorkList.h files in order to use more C++ feature and remove dependencies with OS specific code.

jantlo:
Specific changes:

- Making WorkList class to have its own mutex instead of the OS specific
global one. The new mutex is the one from std library.

- Using the C++11 std library to handle threads in StandAlone
application
and enabling concurrent processing on non-windows platforms.

- converting the global variable Worklist into local variable workList.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -51,6 +51,7 @@
 #include <cctype>
 #include <cmath>
 #include <array>
+#include <thread>
 
 #include "../glslang/OSDependent/osinclude.h"
 
@@ -150,13 +151,6 @@ void ProcessConfigFile()
         delete[] config;
 }
 
-// thread-safe list of shaders to asynchronously grab and compile
-glslang::TWorklist Worklist;
-
-// array of unique places to leave the shader names and infologs for the asynchronous compiles
-glslang::TWorkItem** Work = 0;
-int NumWorkItems = 0;
-
 int Options = 0;
 const char* ExecutableName = nullptr;
 const char* binaryFileName = nullptr;
@@ -253,7 +247,7 @@ void ProcessBindingBase(int& argc, char**& argv, std::array<unsigned int, EShLan
 //
 // Does not return (it exits) if command-line is fatally flawed.
 //
-void ProcessArguments(int argc, char* argv[])
+void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItems, int argc, char* argv[])
 {
     baseSamplerBinding.fill(0);
     baseTextureBinding.fill(0);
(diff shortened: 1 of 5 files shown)


jantlo:
All the changes have been squashed in just one commit. The coding style issues have been fixed.

I strongly recommend to adjust the .clang-format file in order to match the coding style otherwise is really easy to commit mistakes.

Thanks,
    Juan

StandAlone/StandAlone.cpp
```diff
@@ -489,13 +478,13 @@ void SetMessageOptions(EShMessages& messages)
 //
 // Return 0 for failure, 1 for success.
 //
-unsigned int CompileShaders(void*)
+void CompileShaders(glslang::TWorklist& worklist)
```
johnkslang: `void` is contradictory to the comment above.

StandAlone/StandAlone.cpp
```diff
@@ -784,47 +779,43 @@ int C_DECL main(int argc, char* argv[])
...
-        if (Options & EOptionMultiThreaded) {
-            const int NumThreads = 16;
-            void* threads[NumThreads];
-            for (int t = 0; t < NumThreads; ++t) {
-                threads[t] = glslang::OS_CreateThread(&CompileShaders);
-                if (! threads[t]) {
+        if (Options & EOptionMultiThreaded)
+        {
+            std::array<std::thread, 32> threads;
+            const unsigned int numThreads = std::min<unsigned int>(threads.size(), std::thread::hardware_concurrency());
```
johnkslang: Does this mean OS_CreateThread(), et. al., are now dead in ossource.cpp?

If this really works on all platforms, it would be nice to get rid of the os-dependent code.  If not, this removes the only thing the test has to exercise it.  So they should go hand-in-hand, both staying or both going.

johnkslang: What range might `numThreads` take on?  I would not want it to, say, go down to a small value, just because of hardware.  This is used for testing multi-threadedness of the library, and preemption is useful for that test, even on hardware supporting only one thread.

Basically, I don't want this to suddenly change testing from 16 thread threads to a small number of threads.

jantlo: Yes right, my idea was to submit this and the prepare other request with this change.

jantlo: So you prefer a fix value, I thought the idea was to speed up the shader compilation but if is not the case then we get back the original behavior. Weird that is used for testing since is a public option that everybody can use. Should it be not exposed in the usage message?

johnkslang:
Okay, starting to make good sense.  Don't want to see separate commits between headers, the cpp files etc.  It is okay to do `push --force` to update the PR, to keep commits down to 1 or 2. Thanks.

PR #799, opened 2017-03-27
Assign a constant number for each descritor set.

xlpiao:
I've commited the patch to [https://github.com/xlpiao/glslang.git](url)
The commit number is d1bdde6.

The test results:

**Before the patch:**
```c
OpDecorate %17 DescriptorSet 0
OpDecorate %17 Binding 0
OpDecorate %21 DescriptorSet 0
OpDecorate %21 Binding 0
OpMemberDecorate %32 0 RowMajor
OpMemberDecorate %32 0 Offset 0
OpMemberDecorate %32 0 MatrixStride 16
OpMemberDecorate %32 1 Offset 64
OpDecorate %32 Block
OpDecorate %34 DescriptorSet 0
OpDecorate %34 Binding 2
OpDecorate %44 BuiltIn FragCoord
OpDecorate %50 Location 0
OpDecorate %56 Location 0
OpMemberDecorate %60 0 RowMajor
OpMemberDecorate %60 0 Offset 0
OpMemberDecorate %60 0 MatrixStride 16
OpDecorate %60 Block
OpDecorate %62 DescriptorSet 0
OpDecorate %62 Binding 0
OpMemberDecorate %63 0 RowMajor
OpMemberDecorate %63 0 Offset 0
OpMemberDecorate %63 0 MatrixStride 16
OpDecorate %63 Block
OpDecorate %65 DescriptorSet 0
OpDecorate %65 Binding 1
```
**After the patch:**
```c
OpDecorate %17 DescriptorSet 1
OpDecorate %17 Binding 0
OpDecorate %21 DescriptorSet 3
OpDecorate %21 Binding 0
OpMemberDecorate %32 0 RowMajor
OpMemberDecorate %32 0 Offset 0
OpMemberDecorate %32 0 MatrixStride 16
OpMemberDecorate %32 1 Offset 64
OpDecorate %32 Block
OpDecorate %34 DescriptorSet 0
OpDecorate %34 Binding 2
OpDecorate %44 BuiltIn FragCoord
OpDecorate %50 Location 0
OpDecorate %56 Location 0
OpMemberDecorate %60 0 RowMajor
OpMemberDecorate %60 0 Offset 0
OpMemberDecorate %60 0 MatrixStride 16
OpDecorate %60 Block
OpDecorate %62 DescriptorSet 0
OpDecorate %62 Binding 0
OpMemberDecorate %63 0 RowMajor
OpMemberDecorate %63 0 Offset 0
OpMemberDecorate %63 0 MatrixStride 16
[...]

PR #820, opened 2017-04-05
HLSL: Different descritptor set number for different resource type

xlpiao:
Since there's no concept about descriptor set in HLSL, we assign the descriptor set number in the order of declaration sequence.
This is the same issue with #799 , but slightly different patch.


xlpiao:
@johnkslang 
I don't know why the errors ccur on appveyor: 
```
fatal: unable to access 'https://github.com/KhronosGroup/glslang.git/': Unknown SSL protocol error in connection to github.com:443 
Command exited with code 128
```
The release version is successfully built in my PC, but not on that appveyor site.

johnkslang:
There are two issues with this:

1. It is not thread safe.  The `static` variables are effectively globals shared by multiple threads.

2. It seems quite arbitrary that the first texture, first sample, first buffer, etc. all go together in the first descriptor set, and then the next of each go into the next descriptor set, etc.

Is it possible to do what you want by instead overriding the IO remapper?  Can you see around line 450 in ShaderLang.h?

xlpiao:
@johnkslang 
Thanks for your reply. I'll start doing this one week later.

PR #826, opened 2017-04-07
Using C++11 features for replacing current OS specific code.

jantlo:
TLS code remove at all, replaced by the thread_local specifier. This also
impacts in the initialization process simplifying it a lot. The
functions Init/DetachProcess, Init/DetachThread have been removed. The
only function needed to be called for a proper initializations is
InitializeMemoryPools (FreeMemoryPools is optional during th shutdown).

TLS Support for MSVC 2013 is limited, only PODs are allowed to be used
and the specific extension should be used for it. This prevents us to do
further code simplifications.

The OGLCompilersDLL dynamic library is not needed anymore and has been
removed from the solution.

The way how is handle the PoolAllocators has been slightly modified. The
SetThreadPoolAllocator now receives a unique_ptr for the new
PoolAllocator and returns another unique_ptr for the old one. The goal
is to clearly define who is the ownership of those pool allocators in any
moment. Therefore the member variables pools in any object must be from
now a unique_ptr.

Overall this change removes 560 lines and adds 87.


jantlo:
After this changes the OSDependent dynamic library contains only one method: OS_DumpMemoryCounters which is only used for debug purposes in StandAlone.cpp file. I would like to move this function there and remove the OSDependent library.

johnkslang:
This would be nice. Do you know what performance impact it has? Is it the same underlying mechanism, just abstracted, or is it possibly different on some platforms?

jantlo:
I wouldn't expect any impact in the performance. The C++11 version behaviour should be very well defined, regarding the pre-C++11 version extensions it is hard to say.

xorgy:
Seems as though you could ditch `OSDependent/Unix` altogether, since it is completely inert.

johnkslang:
Closing this one, as the key feature was covered by #2348, a more up-to-date request, triggered by dropping VS 2013.

PR #827, opened 2017-04-07
Implement the extension SPV_KHR_16bit_storage

amdrexu:


SPIRV/GlslangToSpv.cpp
```diff
@@ -2075,6 +2075,24 @@ spv::Id TGlslangToSpvTraverser::createSpvVariable(const glslang::TIntermSymbol*
     spv::StorageClass storageClass = TranslateStorageClass(node->getType());
     spv::Id spvType = convertGlslangToSpvType(node->getType());
 
+#ifdef AMD_EXTENSIONS
```
johnkslang: Isn't the extension this implements Khronos level, not AMD specific?

That is, shouldn't the #if_def go away?

amdrexu: There are two ways to handle this extension and it depends on your decision:
(1) Insert the implementation of this extension to SPIR-V builder.
This means we could do some processing in makePointer() or createVariable() and do not care about what the original glslang types. Because we are lack of underlying utilities to analyze the input SPIR-V type thoroughly like TType::constainsBasicType(), we need add codes to determine this along with the "Buffer" decoration. In this sense, this extension is AMD-independent.

(2) Insert the implementation of this extension to the uses of  SPIR-V builder.
This way is showed in my codes, we add the capability when translating glslang::TType to SPIR-V type. We have a lot of knowledge to do this because TType in glslang is well built. It includes all information we need. But doing this means glslang must support EbtFloat16 type (or EbtInt16/EbtUint16 type). Those types are currently missing in standard GLSL grammar (only in AMD extensions). So I have to add the #ifdef/#endif delimiter.

If you do not satisfy with option(2), please let me know and I will switch to option(1).

johnkslang:
Thank you for the explanation. I agree with your change, will accept.

PR #828, opened 2017-04-08
Modernise CMake

SoapGentoo:
* Use `GNUInstallDirs` in order to respect GNU conventions.
  This is especially important for multi-arch/multi-lib setups.
* Specify position independent mode building properly, without
  using the historic hack of adding `-fPIC` as a definition.
  This makes the build system more portable.
* Only detect C++ (and not C) to slightly speed up configuring.
* Specify C++11 mode using modern CMake idioms.
* Fix some whitespace issues.


CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.1)
```
turol: Is this absolutely necessary? It makes glslang unbuildable on current Debian stable (Jessie) since it only has CMake 3.0. Since SteamOS is based on Debian it might also have this issue but I'm not sure if they've upgraded CMake. Does someone know?

Next version of Debian (Stretch) is currently frozen and will hopefully release this year. It has CMake 3.7 so this problem will go away then.

SoapGentoo: 3.1 is required for the `set(CMAKE_CXX_STANDARD 11)` stuff. I know this is kinda painful, but the current solution is less portable in my opinion. Alternatively, you could dispatch on the version, and only add `-std=c++11` for < 3.1 and use the correct stuff on 3.1.

SoapGentoo: @turol added the <3.1 case discriminating conditional

SoapGentoo:
@johnkslang any chance of getting this merged?

johnkslang:
Moving to 3.0 will make some downstream uses need to adapt (they will initially break), so it is important to know what the motivating problem is.  I see many independent changes, many of which don't seem to really require 3.0, so I wonder how much of the real problem can be solved without changing the base requirement for CMake.

SoapGentoo:
@johnkslang 
I've rechecked all minimum requirements for the features:

1. `GNUInstallDirs` - >= **2.8.5**
2. `add_compile_options` - >= **2.8.12**
3. `PROPERTY POSITION_INDEPENDENT_CODE ON` - >= **2.8.9**

so if I increase the minimum requirement from **2.8.11** to **2.8.12**, would that be ok?

johnkslang:
Yes, I believe going just to 2.8.12 will not create any downstream problems, thanks.

SoapGentoo:
@johnkslang done, please have another look

johnkslang:
This fails on my Windows system:

6>  CMake Error at External/googletest/googlemock/cmake_install.cmake:34 (file):
6>    file cannot create directory: C:/Program Files (x86)/glslang/lib.  Maybe
6>    need administrative privileges.
6>  Call Stack (most recent call first):
6>    External/googletest/cmake_install.cmake:32 (include)
6>    External/cmake_install.cmake:32 (include)
6>    cmake_install.cmake:32 (include)

SoapGentoo:
@johnkslang given that `GNUInstallDirs` just specifies some defaults and cache entries, I cant see how this causes such an error. On the other hand, trying to install the package as a normal user to `/usr` also gives the same error due to missing permissions. I don't have a windows machine to try, but have you checked whether this isn't some issue relating to UAC or such?

Issue #794, opened 2017-03-23
HLSL: Explicit descriptor offsets for b#, t#, s#, u#


punkUser:
As with many other folks, when compiling HLSL shaders we need a way to map the four standard HLSL binding namespaces (cbuffers: b#, SRVs: t#, samplers: s# and UAVs: u#) to unique descriptor indices in SPIRV/Vulkan. Since our shaders have a mix of explicit and implicit mappings, the best way to guarantee they end up unique is to just assign ranges in the resulting descriptor set for each of the binding spaces.

There's some support for this with the IoRemapper for this in the "set shift [] binding", but the naming of these implies it's more in terms of the Vulkan types rather than the HLSL namespaces. While in a lot of cases there's a 1:1 mapping between these, there are certainly cases today where this is not the case and it's probably not something we want to always assume will be perfect 1:1 in the future.

Thus what we'd really like to see in the HLSL path is:

1) Auto map bindings done in the HLSL namespaces before translation, similar to what FXC would do. Doesn't have to match perfectly, but should be done in those namespaces to ensure uniqueness.

2) Application provides 4 offsets - one each for b#, t#, s# and u# - named as per the HLSL conversions. ex. Cbuffer offset, SRV offset, UAV offset, sampler offset. When mapping the now fully resolved bindings from HLSL one simply adds the namespace offset to the #.

Today we're implementing something similar by modifying the IoRemapper but it would be great to have this directly done as described above for future compatibility and to ensure that the semantics we're looking for remain.

Thanks!

danginsburg:
To add some more detail, because I had a conversation with @punkUser and at first didn't understand the ask here.  So related to #795 I believe what he is saying is that in HLSL both Buffer<> and regular textures share a namespace t#.  That is, both textures and texture buffers are specified in HLSL with t#.  In the I/O remapper, these are treated as different binding spaces (Buffer<> currently as Ssbo although that's possibly wrong, depends on how #795 is resolved) and textures as textures.  

He wants to be able to preserve the t# even though in Vulkan these are going to correspond to different resource types.  So whereas now the I/O remapper would treat the Buffer<> and textures as separate namespaces with different starting offsets, he wants to use the same namespace.

I think maybe if the I/O remapper was able to specify shift bindings in an alternate way using HLSL namespaces that might give him what he wants.

johnkslang:
> He wants to be able to preserve the t# even though in Vulkan these are going to correspond to different resource types.

[...]

PR #840, opened 2017-04-18
HLSL: hlsl register class iomapping

steve-lunarg:
Adds --hlsl-iomap option to perform IO mapping in HLSL register space.

--shift-cbuffer-binding is now a synonym for --shift-ubo-binding.

The idea way to do this seems to be passing in a dedicated IO resolver, but that would require more intrusive restructuring, so maybe best for its own PR.

The TDefaultHlslIoResolver class and the former TDefaultIoResolver class share quite a bit of mechanism in a common base class.

~~TODO: tbuffers are landing in the wrong register class, which needs some investigation.  They're either wrong upstream, or the detection in the resolver is wrong.~~  DONE

Addresses #794.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -87,6 +87,7 @@ enum TOptions {
     EOptionNoStorageFormat      = (1 << 21),
     EOptionKeepUncalled         = (1 << 22),
     EOptionHlslOffsets          = (1 << 23),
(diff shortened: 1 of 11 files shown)


steve-lunarg:
Repush to resolve minor merge conflict, fix tbuffer register class problem, and add some hopefully useful comments above the TDefaultHlslIoResolver class.

PR #842, opened 2017-04-20
HLSL: cast non-int types to uint on Load/Store indexes

steve-lunarg:
Some texture and SB operations can accept non-integer indexes, which should be auto-cast to integers before use if they are not already.  This adds makeIntegerIndex() for the purpose.  Integer types are passed through unchanged.

(This was done before for operator[], but needs to apply to some other ops too which bypass the normal argument conversion process, hence its extraction into common function)


steve-lunarg:
I don't know what happened to the other auto-build.  Only one ran at all, though the "All checks have passed" status seems unconcerned.

PR #843, opened 2017-04-20
Fix logic of do-while hlsl parsing.

gwihlidal:


gwihlidal:
Tests fail, but likely just a difference in AST

johnkslang:
Sorry, I didn't see this while fixing the issue I saw. Anyway, it's all fixed now.

Do note though, that all functional changes need to be accompanied by testing changes that show them in action and serve as regression tests.

PR #852, opened 2017-04-24
track declared builtin type

steve-lunarg:
Marking as WIP since it might deserve discussion or at least explicit consideration, as it touches the TType member data.

During type sanitization, the TQualifier's TBuiltInVariable type is lost.  However, sometimes it's needed downstream.  There were already two methods in use to track it through to places it was needed: one in the TParameter, and another in a map in the HlslParseContext used for structured buffers.

The latter was going to be insufficient when SB types with counters are passed to user functions, and it's proving awkward to track the data to where it's needed. This PR holds a proposal: track the original declared builtin type in the TType, so it's trivially available anywhere needed.

This lets the other two mechanisms be removed (as they are in this PR).  There's also a side benefit of not losing certain types of information before the reflection interface.

This PR contains only that proposal, so it changes no test results.  If it's acceptable, I'll build on it for the last piece of SB counter functionality.

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -400,6 +400,7 @@ public:
         invariant = false;
         noContraction = false;
         makeTemporary();
+        declaredBuiltIn = EbvNone;
     }
 
     // drop qualifiers that don't belong in a temporary variable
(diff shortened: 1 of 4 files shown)


steve-lunarg:
Merge with latest code, repush, remove WIP.

PR #856, opened 2017-04-25
Pure Texture to Sampled Texture Transform

TiemoJung:
Adds a transformation step at the post processing step.
Two modes are available:
1) keep
- Keeps samplers, textures and sampled textures as is
2) transform pure texture into sampled texture and remove pure samplers
- removes all pure samplers
- transforms all pure textures into its sampled counter part

replacement for #674


hrydgard:
Sorry for the bump, but any plans to merge this? Would be useful.

Needs a rebase by the way.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2854,4 +2862,41 @@ bool TIntermediate::specConstantPropagates(const TIntermTyped& node1, const TInt
            (node2.getType().getQualifier().isSpecConstant() && node1.getType().getQualifier().isConstant());
 }
 
+struct TextureUpgradeAndSamplerRemovalTransform : public TIntermTraverser {
+    bool visitAggregate(TVisit, TIntermAggregate* ag) override {
+        using namespace std;
+        TIntermSequence& seq = ag->getSequence();
+        // remove pure sampler variables
+        TIntermSequence::iterator newEnd = remove_if(seq.begin(), seq.end(), [=](TIntermNode* node) {
```
johnkslang: What is it you are trying to capture by value here?  Would `[&]` or even `[ ]` be better?

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2854,4 +2862,41 @@ bool TIntermediate::specConstantPropagates(const TIntermTyped& node1, const TInt
...
+            TIntermSymbol* symbol = node->getAsSymbolNode();
+            if (!symbol)
+                return false;
+
+            return (symbol->getBasicType() == EbtSampler && symbol->getType().getSampler().isPureSampler());
+        });
+        seq.erase(newEnd, seq.end());
+        // replace constructors with sampler/textures
+        // update textures into sampled textures
+        for_each(seq.begin(), seq.end(), [=](TIntermNode*& node) {
```
johnkslang: Same question; unclear why to request capturing everything by value.

johnkslang:
If this is useful, it seems reasonable. Can merge in when it is ready (and question answered).

TiemoJung:
@johnkslang The two lambda captures where not needed, and i removed it. Also rebased it.

PR #860, opened 2017-04-27
HLSL: add ability to pass struct buffers with counters to fns

steve-lunarg:
HLSL: add ability to pass struct buffers with counters to fns

This modifies function parameter passing to pass the counter buffer associated with a struct buffer to a function as a hidden parameter.  Similarly function declarations will have hidden parameters added to accept the associated counter buffers.

There is a limitation: if a SB type may or may not have an associated counter, passing it as a function parameter will assume that it does, and the counter will appear in the linkage whether or not there is a counter method used on the object.


steve-lunarg:
Repush with fix for access off end of array.

Still WIP.

steve-lunarg:
Repush after code cleanup and mild refactoring.

steve-lunarg:
Minor code cleanup, removed WIP.

PR #867, opened 2017-04-28
HLSL: Allow empty struct initializers

### LINKED ISSUE #862: HLSL - can't initialize an empty struct
The following code fails to compile with glslangValidator.exe -V "VertexShader.hlsl" -e ShadeVertex -S vert -D :

---8<---

struct Test
{

};

static const Test Test_Empty;
float4 ShadeVertex(in uint vertexIndex : VERTEXID) : VS_OUT_POSITION
{ 
    return float4(0,0,0,0);
}

---8<---

If you change the Test struct:

---8<---

struct Test
{
int foo;
};

---8<---

it works.

steve-lunarg:
Review request: is this a problem for GLSL in any way?  All the GLSL tests passed, but I'm uncertain if there's an undesired interaction.

Fixes #862


johnkslang:
The GLSL grammar can't accept an empty structure.

PR #876, opened 2017-05-02
CMake: use target_include_directories to expose includes

ligfx:
This makes glslang simpler to embed in larger projects.


scribam:
@ligfx Would you mind rebasing your PR? Would it be still useful or you found an other way to integrate glslang in your project?

Tilka:
Ping for rebase.

ligfx:
Rebased

johnkslang:
It seems like deleting the `Public` directory could break a lot of things. Is that actually a required part of embedding in other projects?

ligfx:
The commit message claimed it was to match the installed location, but glslang does install that header in a `Public` directory (like `/usr/local/include/glslang/Public/ShaderLang.h`). Not sure what I was thinking back then.

Anyways, that's removed now: all this PR does is add the root directory to `target_include_directories` for both the glslang and SPIRV targets.

PR #878, opened 2017-05-03
HLSL: implement [unroll] and [loop] attributes

### LINKED ISSUE #836: HLSL: emit OpLoopMerge Unroll for HLSL [unroll]
I believe that currently [unroll] in the HLSL does not result in OpLoopMerge Unroll in the SPIR-V.  Please add support for this, it is a feature relied upon by a lot of HLSL for performance.

steve-lunarg:
This adds infrastructure suitable for any front end to create SPIR-V loop control flags.  The only current front end doing so is HLSL.

[unroll] turns into spv::LoopControlUnrollMask
[loop] turns into spv::LoopControlDontUnrollMask
no attribute means spv::LoopControlMaskNone

Fixes #836


SPIRV/GlslangToSpv.cpp
```diff
@@ -767,6 +768,16 @@ spv::ImageFormat TGlslangToSpvTraverser::TranslateImageFormat(const glslang::TTy
...
 }
 
+spv::LoopControlMask TGlslangToSpvTraverser::TranslateLoopControl(glslang::TLoopControl loopControl) const
+{
+    switch (loopControl) {
+    case glslang::ELoopControlNone:       return spv::LoopControlMaskNone;
+    case glslang::ELoopControlUnroll:     return spv::LoopControlUnrollMask;
+    case glslang::ELoopControlDontUnroll: return spv::LoopControlDontUnrollMask;
+    default:                              return spv::LoopControlMaskNone;
+    }
```
johnkslang: Don't we want something that takes a number too, to translate to expose `DependencyLength`?

hlsl/hlslParseHelper.cpp
```diff
@@ -7568,6 +7568,7 @@ bool HlslParseContext::handleOutputGeometry(const TSourceLoc& loc, const TLayout
     return true;
 }
 
+
```
steve-lunarg: Accidental blank line; will remove in next iteration.

SPIRV/GlslangToSpv.cpp
steve-lunarg: There is an optional integer count to the unroll attribute that can be passed through.  I haven't yet found info about `DependencyLength` - do you have a reference to it?  I searched for that (and a few similar spellings) in the SPIR-V spec and in glslang and haven't found the magic phrase yet.

johnkslang: Look at the 1.1 SPIR-V specification. There is also `DependencyInfinite`.  Maybe can't be used immediately, but at some point, there will be a new version, and we'll have `spvVersion` saying what code can be generated.

I guess since you said "creating infrastructure for any front end" I was just reviewing from a broad perspective.  Since we'd have to include a different header, I'd understand deferring this for now.

johnkslang: (Could leave a commented-out placeholder for the extra enums.)

steve-lunarg: Ah, I see... I was looking at 1.0 still.  Will update to add placeholders.

steve-lunarg:
Repush after adding placeholder comments for other loop flags, and minor code cleanup to move some non-grammar code into the parse context.

Removed WIP.

PR #899, opened 2017-05-19
Remove empty cpp files

antiagainst:
This solves ranlib warnings on MacOS.


johnkslang:
There were a few consumers that break every time the set of files changes, so I have been conservative in only changing the file list when necessary.  Maybe will combine with this with something else, or just do the CMakeLists.txt change.

antiagainst:
@johnkslang: I think you mean combining this with more "breaking" changes? I'm fine with that. But right now I don't have more in my hand.

johnkslang:
The includer change effected the needed files, so pulling this in as well.

PR #907, opened 2017-05-29
HLSL: Add an Includer to handle #include for local HLSL paths.

johnkslang:


johnkslang:
FYI, this #include functionality is working correctly. There is a google-test problem with having either no paths (don't know where to include from) or absolute paths (not portable across systems for results).

johnkslang:
#include testing moved to runtests, which does not have working-directory variance.

PR #912, opened 2017-06-01
Implement the extension GL_AMD_texture_gather_bias_lod

amdrexu:
The OpenGL spec is: https://www.khronos.org/registry/OpenGL/extensions/AMD/AMD_texture_gather_bias_lod.txt

The SPIR-V spec is: https://www.khronos.org/registry/spir-v/extensions/AMD/SPV_AMD_texture_gather_bias_lod.html

This extension adds some new built-in functions to allow texture gather to support `bias` and `lod`.


oscarbg:
Are public drivers exposing it in both Vulkan and OpenGL? also is a Vega GPU feature or supported on Polaris cards?

SPIRV/SpvBuilder.cpp
```diff
@@ -56,6 +56,8 @@
 
 namespace spv {
 
+#include "GLSL.ext.AMD.h"
```
johnkslang: Why is this needed in this file?  Can it go away?

SPIRV/GlslangToSpv.cpp
```diff
@@ -3308,21 +3342,20 @@ spv::Id TGlslangToSpvTraverser::createImageTextureFunctionCall(glslang::TIntermO
...
     if (cracked.gather && ! sampler.shadow) {
         // default component is 0, if missing, otherwise an argument
         if (2 + extraArgs < (int)arguments.size()) {
             params.component = arguments[2 + extraArgs];
             ++extraArgs;
-        } else {
+        } else
             params.component = builder.makeIntConstant(0);
-        }
+    }
```
johnkslang: The order of this code is reversed now.  Before, if there was both bias and gather, were arguments taken from the wrong place?  Or the combination never existed until now?

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5366,6 +5478,20 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
+        // E_GL_AMD_texture_gather_bias_lod
+        if (profile != EEsProfile) {
+            symbolTable.setFunctionExtensions("textureGatherLodAMD",                1, &E_GL_AMD_texture_gather_bias_lod);
+            symbolTable.setFunctionExtensions("textureGatherLodOffsetAMD",          1, &E_GL_AMD_texture_gather_bias_lod);
+            symbolTable.setFunctionExtensions("textureGatherLodOffsetsAMD",         1, &E_GL_AMD_texture_gather_bias_lod);
+
+            const char* const sparse_gather_lod[] = { E_GL_ARB_sparse_texture2, E_GL_AMD_texture_gather_bias_lod };
+            const int Num_sparse_gather_lod = sizeof(sparse_gather_lod) / sizeof(sparse_gather_lod[0]);
+
+            symbolTable.setFunctionExtensions("sparseTextureGatherLodAMD",        Num_sparse_gather_lod, sparse_gather_lod);
```
johnkslang: Does the ARB extension really turn on AMD-specific features?  This extension list enables the feature *if any* of the extensions are present, not if all are present.  

I suspect you want to just tie this to `E_GL_AMD_texture_gather_bias_lod`. Then, in `TParseContext::builtInOpCheck` you can do additional error checking for extension interactions.

amdrexu:
@oscarbg Currently, no public driver has exposed  it. In the near future, Vulkan driver will support it soon. For OpenGL, the development is still in progress. This extension is fully supported in Vega and is partially supported in Polaris (some functions are unavailable).

SPIRV/SpvBuilder.cpp
amdrexu: This is not necessary. I will remove it.

SPIRV/GlslangToSpv.cpp
amdrexu: The order is adjusted. Because originally, `comp `(for texture gather) and `bias` do not coexist (no such function prototype). Thus, the codes checking `bias` could be prior to the codes checking `comp` or be after it. The order is not important. But now, new function prototypes assume `bias` is the last parameter. Hence, the order is adjusted.

glslang/MachineIndependent/Initialize.cpp
amdrexu: Thank you. You are right. I thought the extensions listed here are AND relationship but actually they are OR. I will correct it.

amdrexu:
Thank you for feedbacks. I modify the codes.

oscarbg:
@amdrexu thanks.. still waiting/wanting to see "new programmable geometry pipeline" Vega Vulkan/OpenGL extension.. hope more info by Siggraph time and supported on initial Radeon Vega drivers..

PR #916, opened 2017-06-03
Fix a few memory leaks

polarina:


Kwizatz:
Hello,
Is there a use for the cached pointer other than to delete it later?
It may make more sense to delete it whenever it is replaced and if deleted for good, set the pointer to nullptr, if there is indeed a double free "delete nullptr" should be safe.

polarina:
@Kwizatz There are a few places where the pool allocator is temporarily replaced and restored at a later time (a pair call to `GetThreadPoolAllocator` and `SetThreadPoolAllocator`). For example, `SetupBuiltinSymbolTable` in `glslang/MachineIndependent/ShaderLang.cpp` does this.

Kwizatz:
@polarina Ok just checking, I am not a glslang developer, but thanks for your patch, I cherry picked it and it solves my leaks.

glslang/Include/Common.h
```diff
@@ -235,7 +235,11 @@ struct TSourceLoc {
     int column;
 };
 
-typedef TMap<TString, TString> TPragmaTable;
+class TPragmaTable : TMap<TString, TString> {
+public:
+	POOL_ALLOCATOR_NEW_DELETE(GetThreadPoolAllocator())
```
johnkslang: `TMap` is templatized with a `pool_allocaton<>`.  This just effects the `new` of the table itself, not what happens inside. That has a delete in `~TIntermAggregate()` so, I suppose the problem is without explicit destruction of all nodes, this destructor might not get called.

glslang/MachineIndependent/PoolAlloc.cpp
```diff
@@ -53,6 +53,7 @@ void InitializeMemoryPools()
     TThreadMemoryPools* threadData = new TThreadMemoryPools();
 
     threadData->threadPoolAllocator = threadPoolAllocator;
+    threadData->originalThreadPoolAllocator = threadPoolAllocator;
```
johnkslang: What is the code path that triggers the need for this?  Thanks.

polarina: @johnkslang The methods `TShader::parse`, `TShader::preprocess`, and `TProgram::link` create a new allocator and call `SetThreadPoolAllocator` with it. The allocators they create get deleted in the destructors for `TShader` and `TProgram`. In that process we have lost the reference to the allocator that was in effect prior to calls to any of the three above-mentioned methods, leading to a memory leak.

glslang/Include/Common.h
polarina: @johnkslang `TIntermAggregate` inherits from `TIntermOperator`, which inherits `TIntermTyped`, which inherits `TIntermNode`, which uses `POOL_ALLOCATOR_NEW_DELETE`.

Because how that macro implements the `new` and `delete` operations, the destructors for any class that uses it will not get called. The destructor in `TIntermAggregate` is dead code.

`TIntermAggregate::addToPragmaTable`, without this patch, will allocate `TPragmaTable` from the standard heap, not from the pool allocator. Because the destructor for `TIntermAggregate` never gets called, this results in a leak.

johnkslang: Yes, that's all what I meant by my last phrase, thanks.

glslang/MachineIndependent/PoolAlloc.cpp
johnkslang: So, I think this interacts with the `parse()` and `preprocess` collision, where it currently depends on use model how many pools there are and where they went.  It could be that explicitly tracking all these uses solves everything.  (Maybe there is a more automatic way.)

PR #917, opened 2017-06-04
Replace #422: Remove the redundant location setting in AST->SPIR-V.

johnkslang:
This was redundant in two ways:
1) it replicated algorithms owned in the front end, and
2) it sometimes left location information on both a block and its members.


johnkslang:
This does the job that #422 was trying to: don't have the location both on the block and on all the members.

The question is whether removing the redundant location will break anything. It should not, the specifications don't expect it, and indeed a driver would more likely be confused by the redundant information, but maybe a driver depends on it.

Will leave here for a week or so to discuss.

PR #929, opened 2017-06-09
Implement extension GL_AMD_gpu_shader_int16

amdrexu:
- Add int16 types (int16_t, uint16_t, i16vec, u16vec).
- Add int16 support to GLSL operators.
- Add int16 type conversions (to int16, from int16).
- Add int16 built-in functions.

The GLSL spec is here: https://khronos.org/registry/OpenGL/extensions/AMD/AMD_gpu_shader_int16.txt.
Not introduce new semantics elements in SPIR-V.


glslang/Include/Types.h
```diff
@@ -190,8 +190,6 @@ struct TSampler {   // misnomer now; includes images, textures without sampler,
         case EbtFloat:               break;
         case EbtInt:  s.append("i"); break;
         case EbtUint: s.append("u"); break;
-        case EbtInt64:  s.append("i64"); break;
```
johnkslang: Good catch.

glslang/Include/intermediate.h
```diff
@@ -141,6 +141,42 @@ enum TOperator {
     EOpConvFloat16ToDouble,
     EOpConvFloat16ToInt64,
     EOpConvFloat16ToUint64,
+
+    EOpConvBoolToInt16,
```
johnkslang: It would be a big change, to be made independently, but we should consider moving to a "conversion" op that has two subtypes for from -> to.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2327,8 +2524,14 @@ bool TIntermediate::promoteBinary(TIntermBinary& node)
 
         // Check for integer-only operands.
         if ((left->getBasicType() != EbtInt &&  left->getBasicType() != EbtUint &&
-             left->getBasicType() != EbtInt64 &&  left->getBasicType() != EbtUint64) ||
+#ifdef AMD_EXTENSIONS
+             left->getBasicType() != EbtInt16 && left->getBasicType() != EbtUint16 &&
```
johnkslang: We probably need to add a helper function somewhere (next time) for this, like the already existing `isFloatingDomain()`.

johnkslang:
Why are there float16 changes?  Is this just missed previously and combined here?

Note: I plan to do the merge locally, and regenerate the grammar files manually, as I suspect this will overwrite recent grammar changes.

glslang/Include/intermediate.h
amdrexu: Yes, I agree. If int8 is added, the combinations will increase rapidly.

glslang/MachineIndependent/Intermediate.cpp
amdrexu: Yes, since new integer types are added, it is convenient to call a helper function to check this.

amdrexu:
I use bison 2.4. Perhaps, it is old and has to be upgraded. That might be be root cause why you see the grammar file is not as expected. Also, I think the guidance page of glslang should provide new bison link (it is 3.0+ now).

johnkslang:
My bison comment was not about the version. I'd thought we had overlapping modifications to glslang.y, and I didn't trust git to correctly merge the generated files.

PR #931, opened 2017-06-09
HLSL: fix several issues in mat construction from scalars

### LINKED ISSUE #923: HLSL: Assigning scalar to matrix allowed by fxc, but not glslangValidator
fxc allows assignment of a matrix to a scalar, but this is not supported glslangValidator.  See shader for examples:

```
float4 main() : SV_TARGET
{
    /// Case 1
    float4x4 mat1 = 0.0;                                                 // Works in fxc. glslangValidator gives ERROR: mini_test_shader.cg:8: '=' :  cannot convert from ' const float' to ' temp 4X4 matrix of float'
    float4x4 mat2 = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};    // Works in fxc and glslangValidator
    float4x4 mat3 = (float4x4)0.0;                                       // Works in fxc and glslangValidator

    /// Case 2
    float4x4 mat4;
    mat4 = 0.0;                                                         // Works in fxc and glslangValidator gives ERROR: mini_test_shader.cg:14: '' : could not create assignment
    mat4 = float4x4(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);    // Works in fxc and glslangValidator
    mat4 = (float4x4)0.0;                                               // Works in fxc and glslangValidator
   
    float4 f4 = 0.0;                                                    // Works in fxc and glslangValidator

    return f4;
}
```

LoopDawg:
This fixes:

1. A compilation error when assigning scalars to matrices

2. A semantic error in matrix construction from scalars.  This was initializing the diagonal, where HLSL semantics require the scalar be replicated to every matrix element.

3. Functions accepting mats can be called with scalars, which will be shape-converted to the matrix type.  This was previously failing to match the function signature.

**NOTE: this does not yet handle complex scalars (a function call, say) used to construct matricies.  That'll be added when the node replicator service is available.  For now, there's an assert.**

There's one new test (hlsl.scalar2matrix.frag).  An existing test lsl.type.half.frag changes, because of (2) above, and a negative test error message changes due to (3) above.

Fixes #923.


johnkslang:
We have some other cases where we need to properly replicate complex values, but are not yet, and some cases where it is already done ad hoc.

I'd like to formalize this in the glslang AST builder, and then have all the candidate users start using it.

In the meantime, I think this should go in, for two reasons:
1. better to have this incremental functionality, than none, especially since it would cover most the cases
2. better to see three different commits anyway: a) this commit, b) the "replicator" shows up, used by one place, and c) all the other places switch over to using it

I'd ask to get the WIP down to just needing the "replicator", and then removing "WIP".

LoopDawg:
OK, cleaned up, added assert and TODO comment about replicator, removed WIP.  Will revisit when replicator service is available.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -1043,6 +1043,32 @@ TIntermTyped* TIntermediate::addShapeConversion(const TType& type, TIntermTyped*
...
+            const int matSize = type.getMatrixRows() * type.getMatrixCols();
+            TIntermAggregate* rhsAggregate = new TIntermAggregate();
+
+            const bool isSimple = (node->getAsSymbolNode() != nullptr) || (node->getAsConstantUnion() != nullptr);
+
+            if (!isSimple) {
+                assert(0); // TODO: use node replicator service when available.
+            }
+            
+            for (int x=0; x<matSize; ++x)
```
johnkslang: FYI, convention is still to use spaces around operators.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -1043,6 +1043,32 @@ TIntermTyped* TIntermediate::addShapeConversion(const TType& type, TIntermTyped*
...
+            // Note that if the node is complex (e.g, a function call), we don't want to duplicate it here
+            // repeatedly, so we copy it to a temp, then use the temp.
+            const int matSize = type.getMatrixRows() * type.getMatrixCols();
+            TIntermAggregate* rhsAggregate = new TIntermAggregate();
+
+            const bool isSimple = (node->getAsSymbolNode() != nullptr) || (node->getAsConstantUnion() != nullptr);
+
+            if (!isSimple) {
+                assert(0); // TODO: use node replicator service when available.
+            }
```
johnkslang: Note that even with this "simple" test, a DAG is still being made.  We already make DAGs in other cases and it seems okay. I think the spirit here seems to be to avoid side effects, but the future direction of having a replicator will abstract that.

LoopDawg: Agreed.  I figured as soon as we switch over to the replicator (for all cases, simple or not), that'll address it properly.

PR #939, opened 2017-06-17
Test: Add a function to the HLSL buffer test.

Armada651:
Currently the HLSL parser fails to correctly interpret the function parameters when the buffers don't end with semicolons.

### DIFF
--- a/Test/hlsl.buffer.frag
+++ b/Test/hlsl.buffer.frag
@@ -26,7 +26,11 @@ tbuffer tbufName : register(b8) {
                  float3x4 m4;
 }  // no semicolon is okay
 
+float func0(float c) {
+    return c;
+}
+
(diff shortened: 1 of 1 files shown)


johnkslang:
This is not a valid pull request, as it changes tests results, without including the result changes.

However, it points out a valid issue. Will treat it as an issue instead of a PR.

Glslang was allowing for non-anonymous cbuffer/tbuffer, and `float` is a valid identifier (HLSL is wacky: you can say `float float` to declare a variable whose name is float). Yes, since cbuffer/tbuffer don't have semicolons, HLSL has precluded the non-anonymous cases, so the `float` shouldn't be considered a potential instance name of the buffer.

So glslang should remove this possibility, and just always end cbuffer/tbuffer with the closing `}`.

PR #943, opened 2017-06-20
hlsl: "in out" is also an inout qualifier.

xxxbxxx:


xxxbxxx:
caveat: I have no real knowledge of neither hlsl nor glslang...

Issue #946, opened 2017-06-21
ClipDistance decoration applied to scalar from HLSL shader


baldurk:
This short HLSL vertex shader

```hlsl
// glslangValidator -e main -S vert -D -V test.hlsl -o test.spv

void main(out float4 pos : SV_Position, out float clip : SV_ClipDistance)
{
	pos = 1.0f.xxxx;
	clip = 0.5f;
}
```

Produces SPIR-V that applies the ClipDistance to a scalar float output variable:

```asm
      %float = OpTypeFloat 32
%_ptr_Output_float = OpTypePointer Output %float
     %clip_1 = OpVariable %_ptr_Output_float Output
```

But the SPIR-V spec says:

> **ClipDistance**
> _Array_ of clip distances. See Vulkan or OpenGL API specifications for more detail.

And Vulkan's spec says even more explicitly:

> Any variable decorated with `ClipDistance` **must** be declared as an array of 32-bit floating-point values.

So I think any/all ClipDistance outputs from HLSL should this be turned into an implicit array output?

LoopDawg:
There is some code doing a similar(ish) thing for tessellation level builtins (HlslParseContext::fixBuiltInIoType), with some TODO comments about handling clip distance and cull distance.

If I'm reading the HLSL documentation correctly, it seems multiple clip distances can be packed into components of registers, per: _"Multiple clip planes can be implemented simultaneously, by declaring multiple component(s) of one or more vertex elements as the SV_ClipDistance."_  I guess it means something like this?

```
void main(out float4 pos : SV_Position, out float2 clip[2] : SV_ClipDistance)
{
    pos = 1.0f.xxxx;
    clip[0].x = 0.5f;
    clip[0].y = 0.6f;

    clip[1].x = 0.7f;
    clip[1].y = 0.8f;
}
```

Does that match your understanding around multiple clip / cull distance values?  There is some plumbing missing to unpack that into the array of floats expected by SPIR-V.  I could try to add it, although I'm not entirely clear about the rules.  E.g, do the values 0.5 .. 0.8 above become sequential in the clip distance array?  Is the above code a synonym for an array of 4 float (not float2) set with the same values?

baldurk:
I don't honestly know myself, I'd need to leave that for someone with an understanding of how multiple clip distances are declared in HLSL. This wasn't my shader where I encountered it - I was doing some SPIR-V patching/modification and ended up running into an issue where the source SPIR-V seemed to come from something like this.

[...]

PR #947, opened 2017-06-22
HLSL: handle clip/cull distance array semantic matching

LoopDawg:
In HLSL, there are three (TODO: sanity verification from HLSL expert) dimensions of clip and cull distance values:

  * The semantic's value N, ala SV_ClipDistanceN.
  * The array demension, if the value is an array.
  * The vector element, if the value is a vector or array of vectors.

In SPIR-V, clip and cull distance are arrays of scalar floats, always.

This PR currently ignores the semantic N axis entirely, which is probably incorrect, and handles the other two axes by sequentially copying each vector element of each array member into sequential floats in the output array.  This will handle some subset of the possible situations.  It will not handle N &gt; 0 in the semantic.  That will happen in a dedicated PR.

The relative priority of the vector and array dimensions are unspecified in documentation.  This is an educated guess (this order makes some sense).  It can be altered if need be.  In any case, the defect cases do not depend on that order.

Fixes: #946


xxxbxxx:
I also did a patch to solve part of this problem (the vector float4 clipdist).
But it's more of a hack, since I don't quite grasp the general design of glslang:
https://github.com/xxxbxxx/glslang/commit/ab4ee35c455a41efc22b13c1d92aa434f4a67cc8

I'll give a spin to your patch and see how it behaves on my test.

LoopDawg:
> I'll give a spin to your patch and see how it behaves on my test.

OK, thanks @xxxbxxx - appreciated.  Note that at the moment it may not support having these cases inside mixed interface structures of builtins and user variables.  I'll test that tomorrow and fix it if it's broken.

xxxbxxx:
yeah, looks like it doesn't copies all four values. 
Looks like it doesn't get to call assignClipCullDistance() in the "// Output copies" loop.

LoopDawg:
OK thanks - that's for a case with the clipdistance in a structure?  I'll get that working and amend the PR.

xxxbxxx:
> that's for a case with the clipdistance in a structure?
indeed.
https://github.com/xxxbxxx/glslang/blob/master/Test/hlsl.clipdistance.vert

xxxbxxx:
> @xxxbxxx - I believe this should handle your case now. I added two test cases of similar form.

it does, thanks.

PR #965, opened 2017-07-01
Implement SPV_KHR_post_depth_coverage

chaoc:
Added support for both extension GL_ARB_post_depth_coverage and GL_EXT_post_depth_coverage.


glslang/MachineIndependent/localintermediate.h
```diff
@@ -165,7 +165,7 @@ class TIntermediate {
         invocations(TQualifier::layoutNotSet), vertices(TQualifier::layoutNotSet), inputPrimitive(ElgNone), outputPrimitive(ElgNone),
         pixelCenterInteger(false), originUpperLeft(false),
         vertexSpacing(EvsNone), vertexOrder(EvoNone), pointMode(false), earlyFragmentTests(false), depthLayout(EldNone), depthReplacing(false), blendEquations(0),
-        xfbMode(false), multiStream(false),
+        xfbMode(false), multiStream(false), postDepthCoverage(false),
```
johnkslang: These initializers needs to go in order of declaration.  `postDepthCoverage` was declared right after `earlyFragmentTests`.

SPIRV/GlslangToSpv.cpp
```diff
@@ -980,6 +980,11 @@ TGlslangToSpvTraverser::TGlslangToSpvTraverser(const glslang::TIntermediate* gls
         if (glslangIntermediate->getEarlyFragmentTests())
             builder.addExecutionMode(shaderEntry, spv::ExecutionModeEarlyFragmentTests);
 
+        if (glslangIntermediate->getPostDepthCoverage()) {
+            builder.addExecutionMode(shaderEntry, spv::ExecutionModePostDepthCoverage);
```
johnkslang: Should the `SampleMaskPostDepthCoverage` capability be emitted somewhere?

SPIRV/doc.cpp
```diff
@@ -843,6 +844,7 @@ const char* CapabilityString(int info)
     case 5009: return "ImageGatherBiasLodAMD";
 #endif
 
+    case 4447: return "SampleMaskPostDepthCoverage";
```
johnkslang: I don't see either this enum value or string used anywhere, is something missing?

(I also wonder if bringing is these changes (and the official headers) from SPIRV-headers is a good idea.)

chaoc:
All fixed accordingly. Thanks for pointing out.

PR #969, opened 2017-07-04
Add option to skip installation

d3x0r:
add option SKIP_GLSLANG_INSTALL which can be set to skip install() rules.
SKIP_GLSLANG_INSTALL turns into ENABLE_GLSLANG_INSTALL internally which is what is actually used to control installation.

When used as a static library as part of another project, this does not need to be installed, and just clutters application target.


johnkslang:
I didn't notice at first there is one commit per file, versus one commit per conceptual change. Next time, all these commits should be squashed into a single commit.

d3x0r:
sure; but I was editing it on github which only allows me to edit one file at a time.

PR #971, opened 2017-07-04
HLSL: Translate directive [flatten] and [branch] to SPV control mask.

amdrexu:


SPIRV/SpvBuilder.cpp
```diff
@@ -2064,7 +2065,7 @@ void Builder::If::makeEndIf()
 }
 
 // Comments in header
-void Builder::makeSwitch(Id selector, int numSegments, const std::vector<int>& caseValues,
+void Builder::makeSwitch(Id selector, unsigned int control, int numSegments, const std::vector<int>& caseValues,
```
johnkslang: Can we use the type `SelectionControlMask` from the declaration in spirv.hpp?

(Also, any other places, a recurring theme.)

SPIRV/SpvBuilder.h
```diff
@@ -385,7 +385,7 @@ class Builder {
     // Helper to use for building nested control flow with if-then-else.
     class If {
     public:
-        If(Id condition, Builder& builder);
+        If(Id condition, unsigned int ctrl, Builder& builder);
```
johnkslang: `SelectionControlMask`

SPIRV/SpvBuilder.h
```diff
@@ -397,6 +397,7 @@ class Builder {
 
         Builder& builder;
         Id condition;
+        unsigned int control;
```
johnkslang: `SelectionControlMask`

SPIRV/GlslangToSpv.cpp
```diff
@@ -1936,7 +1947,7 @@ bool TGlslangToSpvTraverser::visitSelection(glslang::TVisit /* visit */, glslang
     }
 
     // Instead, emit control flow...
-
+
```
johnkslang: Nicer to undo this white-space change, given that the type change will likely need to be made anyway.

amdrexu: Sure, I will undo it.

SPIRV/SpvBuilder.cpp
```diff
@@ -2009,9 +2009,10 @@ Id Builder::createMatrixConstructor(Decoration precision, const std::vector<Id>&
 }
 
 // Comments in header
-Builder::If::If(Id cond, Builder& gb) :
+Builder::If::If(Id cond, unsigned int ctrl, Builder& gb) :
```
amdrexu: The reason is that this is acutally an integer mask. It might contains multiple emuerants of `SelectionControlMask`. So we use unsigned integer here. Codes might be something like this in the future:
 
`Builder::If ifBuilder(cond, SelectionControlXxxMask | SelectionControlYyyMask, gb)`

If the type `SelectionControlMask` is specified here, we have to do `static_cast`. In the existing codes, loop control is defined similarly. See this:

`Builder::createLoopMerge(Block* mergeBlock, Block* continueBlock, unsigned int control)`

PR #978, opened 2017-07-11
HLSL: support global const initializers from non-constant rvalues

LoopDawg:
WIP: HLSL: support global const initializers from non-constant rvalues

Semantic test left over from other source languages is removed, since this is permitted by HLSL. Also, to support the functionality, a targeted test is performed for this case and it is turned into a EvqGlobal qualifier to create an AST initialization segment when needed.

Constness is now propagated up aggregate chains during initializer construction.  This handles hierarchical cases such as the distinction between:

    static const float2 a[2] = { { 1, 2 }, { 3, 4} };

vs

    static const float2 a[2] = { { 1, 2 }, { cbuffer_member, 4} };

The first of which can use a first class constant initalization, and the second cannot.


hlsl/hlslParseHelper.cpp
```diff
@@ -7149,6 +7149,26 @@ TSymbol* HlslParseContext::lookupUserType(const TString& typeName, TType& type)
         return nullptr;
 }
 
+// Returns true if all nodes in potential aggregate collection, recursively, pass the predicate.
+bool HlslParseContext::recursiveNodeTest(const TIntermTyped* node, std::function<bool(const TIntermTyped*)> pred)
+{
+    // must be typed and non-null.
+    if (node == nullptr || node->getAsTyped() == nullptr)
```
johnkslang: Is there a subtlety here between being a "typed node" versus "having a type"?

If the node has a type, and since it has been built bottom up, the type should reflect its subtree's type. (I somewhat expected this test to go the other way: if it was not typed, then need to traverse subtree to find out, rather than kicking out early.)

Note that all aggregates are typed.

hlsl/hlslParseHelper.cpp
```diff
@@ -7149,6 +7149,26 @@ TSymbol* HlslParseContext::lookupUserType(const TString& typeName, TType& type)
...
+    const TIntermAggregate* aggNode = node->getAsAggregate();
+
+    // recurse into aggregates
+    if (aggNode != nullptr) {
+        return std::all_of(aggNode->getSequence().begin(),
+                           aggNode->getSequence().end(), 
+                           [pred](const TIntermNode* n) { return recursiveNodeTest(n->getAsTyped(), pred); });
+    }
+
+    return pred(node);
```
johnkslang: Ah, so this is working for non-aggregates too. Maybe the top comment (before the function) is wrong about aggregates?

hlsl/hlslParseHelper.cpp
```diff
@@ -7165,6 +7185,21 @@ TIntermNode* HlslParseContext::declareVariable(const TSourceLoc& loc, const TStr
     if (voidErrorCheck(loc, identifier, type.getBasicType()))
         return nullptr;
 
+    // Global consts with initializers that are non-const act like EvqGlobal in HLSL.
+    // This test is recursive, for aggregate initialization, ala:
+    // static const float foo[3] = { 1, 2, 3 };
+    // That gets to stay constant.  However, if it's an aggregate containing a non-constant,
+    // such as { 1, { myvar, 2 }, 3 }, it still becomes EvqGlobal.
+
+    const bool entirelyConstInitializer =
```
johnkslang: Typically, information is aggregated bottom up, in the order the tree is assembled.  Even for aggregate nodes representing constructors and built-in calls, the aggregate should already know the constness of the subtree.  We should not need to revisit the whole subtree to reproduce the information that was just figured in assembling the subtree.

Is it just a special case for `{...}` initializers that the type is missing or not accurate?

hlsl/hlslParseHelper.cpp
```diff
@@ -7165,6 +7185,21 @@ TIntermNode* HlslParseContext::declareVariable(const TSourceLoc& loc, const TStr
...
+    // That gets to stay constant.  However, if it's an aggregate containing a non-constant,
+    // such as { 1, { myvar, 2 }, 3 }, it still becomes EvqGlobal.
+
+    const bool entirelyConstInitializer = 
+        recursiveNodeTest(initializer, [](const TIntermTyped* node) { return node->getQualifier().storage == EvqConst; });
+
+    if (type.getQualifier().storage == EvqConst && symbolTable.atGlobalLevel() && 
+        initializer != nullptr && !entirelyConstInitializer) {
+        // Force to global
+        type.getQualifier().storage = EvqGlobal;
```
johnkslang: Note that glslang has both types of `const`:

- `EvqConst,         // User-defined constant values, will be semantically constant and constant folded`
- `EvqConstReadOnly, // input; also other read-only types having neither a constant value nor constant-value semantics`

Is there an issue using the latter at global scope?  It might help with error checking, which admittedly is not top priority?

This could be a new concept, missing a way of representing it.

johnkslang:
How troublesome is it to fix any remaining corner cases of not knowing constness at each node, of that node's subtree, from the O(1) tests already being done while building the subtree, bottom up?

My guess would be that this is almost always already known, but perhaps not for `{...}` sequences, yes?  Note that `{...}` sequences are typed nodes, so have a place to store the type.

At minimum, the recursive code should skip out early when seeing a valid type, and using that type, rather the recomputing what was already figured out.

hlsl/hlslParseHelper.cpp
LoopDawg: The comment says "potential aggregate", (typo: should have been _potentially_), but the intent was that it can be a typed node, or a (possibly nested) aggregate.  I'll spruce the comment up a bit.

LoopDawg:
> How troublesome is it to fix any remaining corner cases of not knowing constness at each node, of that node's subtree, from the O(1) tests already being done while building the subtree, bottom up?

> My guess would be that this is almost always already known, but perhaps not for {...} sequences, yes? 

Yes, I think that's correct.  Most cases know, but {...} sequences do not.  I'll investigate tracking the constness during the building of an aggregate tree.  If that's easy (maybe so), then this test can be O(1).  (It is already except for aggregate initializers, but could be in all cases).

hlsl/hlslParseHelper.cpp
LoopDawg: That test was a safety mechanism from an older version of the code, but isn't needed in this one.  Left over - I'll remove (or maybe the whole function can simplify down to a single test, if aggregate collections track constness, per other comments below).

LoopDawg:
Pushed update to track EvqConst-ness bottom up during aggregate {...} initializer construction, to avoid recursive query.  Node is EvqConst if all members are, else is EvqTemporary.  External functionality is unchanged.  Removed the now-unneeded query.

Also, added tests for both {...} initializer cases to hlsl.global-const-init.frag

Only constness is propagated, since for example an aggregate of EvqPositions is not an EvqPosition.  If needed, others could be treated like EvqConst.

johnkslang:
Looks good. Will revisit when WIP is removed.

LoopDawg:
Removed WIP.

PR #986, opened 2017-07-13
SPV: Do not generate ImageMSArray capability for sampled images

itoral:
According to the specification description for the capability:

"ImageMSArray: An MS operand in OpTypeImage indicates multisampled,
 used without a sampler."

So the capability must not be included when the only images
used by the shader are sampled. The description of the Sampled
member of OpTypeImage is:

"Sampled indicates whether or not this image will be accessed
 in combination with a sampler, and must be one of the
 following values:
 0 indicates this is only known at run time, not at compile time
 1 indicates will be used with sampler
 2 indicates will be used without a sampler (a storage image)"

So we do not want to generate this capability when the image
Sampled value is 1.

There are a few tests that seemed to expect the capability to be
generated when only sampled images are used. The patch updates these
tests accordingly.

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -466,7 +466,7 @@ Id Builder::makeImageType(Id sampledType, Dim dim, bool depth, bool arrayed, boo
     }
 
     if (ms) {
-        if (arrayed)
+        if (arrayed && ! (sampled == 1))
             addCapability(CapabilityImageMSArray);
         if (! sampled)
             addCapability(CapabilityStorageImageMultisample);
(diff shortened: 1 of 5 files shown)


itoral:
BTW, while fixing this I also noticed that various other checks on the 'sampled' parameter in Builder::makeImageType() seem to do things like:

if (sampled)
...
else
...

Instead of checking whether the value of sampled is exactly '1'. Because sampled can have a value of '2' that means that the image is not sampled I guess these checks are probably not correct and should be fixed too.

johnkslang:
Thanks for catching this: These are good points, and it appears the code was written to an earlier semantic of *Sampled* being a Boolean, before the introduction of an "unknown" state. In practice, the latter is only used for OpenCL, so the only values glslang needs to consider are 1 and 2, though, historically, this was Boolean.

I'll make a comprehensive fix.  (Also, it reads a bit odd, I think to say `! (sampled == 1)` instead of `sampled != 1`.)

PR #991, opened 2017-07-14
HLSL: Fix crash with --resource-set-binding [n] (global form, not per-register form)

LoopDawg:
The ``--resource-set-binding`` command line option has a mode which allows per-register assignments of bindings and descriptor sets on the command line, and another accepting a single descriptor set value to assign to all variables.

The former worked, but the latter would crash when assigning the values. This fixes it, and makes the former case a bit more robust against premature termination of the pre-register values, which must come in ``(regname set binding)`` triples.

Example: ``--resource-set-binding 4``

This also allows the form "--resource-set-binding stage setnum", which was  mentioned in the usage message, but did not parse.

The operation of the per-register form of this option is unchanged.  The global form will also handle variables without register() specifiers ~~but only for HLSL (that's unchanged; --resource-set-binding was HLSL-specific before this PR too)~~, for any language that uses the iomap resolver.

There's a new test for this: hlsl.explicitDescriptorSet.frag, used by two runtests.


LoopDawg:
Repush with changes to allow ``--resource-set-binding N`` form to work with variables lacking register() specifiers.

LoopDawg:
Repush: move some things into the iomap resolver.  This lets the global form play nicely with --auto-map-bindings, and also should work with any language that uses the iomapper functionality.  (The register specific form is still HLSL specific, since the register concept is).

StandAlone/StandAlone.cpp
```diff
@@ -289,31 +289,29 @@ void ProcessResourceSetBindingBase(int& argc, char**& argv, std::array<std::vect
...
-                base[lang].push_back(argv[1]);
-                base[lang].push_back(argv[2]);
-                base[lang].push_back(argv[3]);
-                argc -= 3;
-                argv += 3;
-            }
-            else {
-                break;
-            }
+        argc--;
```
johnkslang: See bumpArg() lambda.

StandAlone/StandAlone.cpp
```diff
@@ -289,31 +289,29 @@ void ProcessResourceSetBindingBase(int& argc, char**& argv, std::array<std::vect
...
+        argv++;
+
+        while (argc > 1 && argv[1] != nullptr && argv[1][0] != '-') {
+            base[lang].push_back(argv[1]);
+
+            argc--;
```
johnkslang: See bumpArg() lambda.

hlsl/hlslParseHelper.cpp
```diff
@@ -5370,19 +5370,25 @@ void HlslParseContext::handleRegister(const TSourceLoc& loc, TQualifier& qualifi
...
         qualifier.layoutBinding = regNumber + subComponent;
-        for (auto it = resourceInfo.cbegin(); it != resourceInfo.cend(); it = it + 3) {
-            if (strcmp(desc.c_str(), it[0].c_str()) == 0) {
-                qualifier.layoutSet = atoi(it[1].c_str());
-                qualifier.layoutBinding = atoi(it[2].c_str()) + subComponent;
-                break;
+
+        // This handles per-register layout sets numbers.  For the global mode which sets
+        // every symbol to the same value, see setLinkageLayoutSets().
+        if ((resourceInfo.size() % 3) == 0) {
```
johnkslang: Do we need to tell the user if there weren't a multiple of 3?  Or did that happen earlier with a usage statement?

StandAlone/StandAlone.cpp
johnkslang: Sorry, don't worry about it, different function.

StandAlone/StandAlone.cpp
johnkslang: Sorry, don't worry about it, different function.

hlsl/hlslParseHelper.cpp
LoopDawg: > Or did that happen earlier with a usage statement?

Correct - there's some code like so:

        // Must have one arg, or a multiple of three (for [regname set binding] triples)
        if (base[lang].size() != 1 && (base[lang].size() % 3) != 0)
             usage();

The multiple-of-3 case is handled at register assignment time (since they apply per HLSL register) and the global case in the io resolver.

It's surely possible to get bad data past it still - e.g, non-numbers, but the input checking should be improved over the previous version.

PR #997, opened 2017-07-20
HLSL: handle multiple clip/cull semantic IDs

### LINKED ISSUE #951: HLSL: support SV_(Clip/Cull)DistanceN for N>0
PR #947 has support for handling some varieties of HLSL clip and cull distance syntax.  HLSL allows values in: vectors, arrays, and semantic IDs, whereas SPIR-V requires a 1D array of scalar floats.

The following appears to be a valid construct which mixes the vector dimension and the semantic ID dimension:

```
struct PS_INPUT 
{ 
    float3 clip0 : SV_ClipDistance0;
    float3 clip1 : SV_ClipDistance1;
}; 
```

I think the separate clip0 and clip1 structure member values should be merged into the single SPIR-V clipdistance output array (because where else would they go?)  The intended semantics are a little unclear though.  Does anyone know this?  I guess that the above (once you write to all the components) should produce the 6 consecutive clip distance values in array elements [0..5], but I can imagine other possibilities such as 8 values with indexes 3 and 7 being implicitly 0, and indexes 0,1,2,4,5,6 are the values the shader wrote.

LoopDawg:
HLSL allows several clip or cull variables to be declared, differentiated by semantic ID.  There are packing rules involved: e.g, a float3 and a float1 can be packed into a single array[4], while for a float3 and another float3, the second one will skip the third array entry to avoid straddling.

This is implements that ability.  Because there can be multiple variables involved, and the final output array will often be a different type altogether (to fuse the inputs appropriately), a new variable is synthesized, unlike the prior clip/cull support which used the declared variable with a modified type.  The new variable name is taken from one of the declared ones, so the old tests are unchanged.

Several new tests are added to test various packing scenarios.  They have to be separate test files, unfortunately, so there are a number of them.

Only two semantic IDs are supported: 0, and 1, per HLSL rules.  This is encapsulated in

     static const int maxClipCullRegs = 2;

and the algorithm (probably :) ) generalizes to larger values, although there are a few questions around how HLSL would pack (e.g, would 4 scalars be packed into a single HLSL float4 out reg?  This algorithm assumes so).

Fixes #951.


LoopDawg:
Repush with existing test fix to avoid invalid semantic IDs, and some new tests to exercise various packing scenarios with struct members and output parameters.

hlsl/hlslParseHelper.cpp
```diff
@@ -2353,23 +2358,99 @@ void HlslParseContext::handleFunctionArgument(TFunction* function,
...
+    }
+
+    // This is the offset in the destination array of a given semantic's data
+    std::array<int, maxClipCullRegs> semanticOffset;
+
+    // Calculate offset of variable of semantic N in destination array
+    int arrayLoc = 0;
+    int vecItems = 0;
+
+    for (int x=0; x<maxClipCullRegs; ++x) {
```
johnkslang: Convention is to use spaces around operators.

hlsl/hlslParseHelper.cpp
```diff
@@ -5314,13 +5395,29 @@ void HlslParseContext::handleSemantic(TSourceLoc loc, TQualifier& qualifier, TBu
         return (unsigned int)atoi(semantic.c_str() + pos + 1);
     };
 
+    // Parse and range check clip/cull distance register ID
+    const auto parseClipCullReg = [getSemanticNumber, loc, this](const TString& semantic) -> unsigned int {
+        const unsigned reg = getSemanticNumber(semantic);
+        if (reg >= maxClipCullRegs) {
```
johnkslang: It seems maxClipCullRegs is 2, so would only need 3 bits of `builtInIndex`. Unless I'm off by a factor of 4 or so between the count of scalars and count of registers.

However, can't we have all such semantic values use layoutLocation instead of adding a new field?  Possibly some code for handling these could be unified, except for the range checking. I think having all these use layoutLocation is rational, not an artificial reuse of a field.

hlsl/hlslParseHelper.cpp
```diff
@@ -5314,13 +5395,29 @@ void HlslParseContext::handleSemantic(TSourceLoc loc, TQualifier& qualifier, TBu
...
+            error(loc, "invalid clip register", semantic.c_str(), "");
```
johnkslang: Message, if we really need it, doesn't  work for cull distance.

I think better to just use a single lambda (start with the existing one), but expand it to accept a limit and an error-message label as parameters, rather than specializing multiple lambdas just for the error message.

hlsl/hlslParseHelper.cpp
LoopDawg: Will fix.

hlsl/hlslParseHelper.cpp
LoopDawg: Agreed.  Will fix.

hlsl/hlslParseHelper.cpp
LoopDawg: That was actually my first approach, and it might be made to work, but without other cleanup it ends up with layoutLocations showing up in the linkage for every use of clip or cull distance, like so:

0:?     'clip' (layout( **location=0**) out 1-element array of float ClipDistance)

It's a bunch of test diffs just for that.  The 0 is not really a linkage location in the above example, it's from the semantic ID.  I might be able to scrub the layoutLocations back off late in the process though, if you want me to push on this approach a little more to save the bits.

(And yeah, it really needs just 1 bit for the builtInIndex, which can be just 0 or 1.  I was thinking about future use with other semantics having number spaces, but those could expand it at the time if they needed to.  This could be 1 bit.)

johnkslang: I see what you mean, and am almost on the fence.

Isn't there a clear point of consumption though? Where it the "location" is processed/refined to become something else, at which point it "moves" from `layoutLocation` to the new way, and `layoutLocation` can be cleared?

LoopDawg: I think this will work out.  There are two places to fix the layoutLocation, one for input side, one for output side.  Will repush a new version removing the builtInIndex and using layoutLocation.

LoopDawg:
Repush with changes as follows:

1. Add spaces around operators.

2. Expanding existing semantic parsing lambda to accept a limit and error message.

3. Use layoutLocation instead of adding a new index field.  This requires a few other changes, but nothing major.  There are two places which reset it.

Aside from this entire change to get the semantic ID hooked up for clip and cull outputs, I'm a little worried about whether there could be a stage mismatch if the input side of a stage linkage tries to read it.  (This PR doesn't make that any *worse* - it's its own issue).  I need to experiment with that.  E.g, glslang changes the clip or cull outputs to a single array to match SPIR-V semantics.  However, if a linked downstream stage would try to read them with the original HLSL types, then maybe there would need to be an opposite input side transform.

PR #1008, opened 2017-07-31
WIP: HLSL: add methods to track user structures in texture return type.

LoopDawg:
**WIP: Incomplete and not yet completely functional**.  This will turn into support for user structure returns from textures.

Some languages allow a restricted set of user structure types returned from texture sampling operations.  The total vector size of all components may not exceed 4, and the basic types of all members must be identical.

This adds underpinnings for that ability.  Because storing a whole TType or even a TTypeList in the TSampler would be expensive, the structure definition is packed into a small bitfield, which can represent:

1. a scalar or vector type as before, or,
2. user structure types abiding the necessary restrictions.

The details of the storage are protected.  There are abstraction methods accepting and returning a TType (such as may have been parsed from a grammar).  The abstraction methods turn the type into the internal representation, or vice versa, so it is never stored directly.

Currently this is hooked up enough to do what it used to do for vectors and scalars. Places in the code which used to directly set or get the vector size now use the new abstraction.  The new abstraction cannot yet handle structures, but that's a matter of fleshing it out, and will be added soon.

NOTE: I believe it's possible to use even fewer bits than are currently being used (currently 12 bits), because not all combinations of vector sizes are valid.  E.g., you cannot have 4 vec4s, and the presence of a member 0 restricts the possible vector size of other members, so 12 bits can hold a lot more entropy than is actually required.  Maybe it really needs something like 5 bits.  That's a detail hidden behind the abstraction.  I'll improve that once full functionality is in place.

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -70,6 +70,8 @@ enum TSamplerDim {
     EsdNumDims
 };
 
+class TType;
+
 struct TSampler {   // misnomer now; includes images, textures without sampler, and textures with sampler
     TBasicType type : 8;  // type returned by sampler
     TSamplerDim dim : 8;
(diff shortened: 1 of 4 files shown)


LoopDawg:
Closing as obsoleted by #1017.

PR #1014, opened 2017-08-11
SPV: 1st pass implementation of SPV_EXT_shader_viewport_index_layer.

johnkslang:
I'm unclear on when to require which extension, given there is a choice, and the more general choice (EXT) is available after the specific choice (NV).

Feel free to update, etc.


amdrexu:
Is it OK for this change to move on? I think it is simple without any demur.

pdaniell-nv:
Fine with me.

PR #1017, opened 2017-08-15
HLSL: add methods to handle user structures in texture template type.

LoopDawg:
Some languages allow a restricted set of user structure types returned from texture sampling operations.  Restrictions include the total vector size of all components may not exceed 4, and the basic types of all members must be identical.

This adds underpinnings for that ability.  Because storing a whole TType or even a simple TTypeList in the TSampler would be expensive, the structure definition is held in a table outside the TType.  The TSampler contains a small bitfield index, currently 4 bits to support up to 15 separate texture template structure types, but that can be adjusted up or down.  Vector returns are handled as before.

There are abstraction methods accepting and returning a TType (such as may have been parsed from a grammar).  The new methods will accept a texture template type and set the sampler return type to the structure if possible, checking a range of error conditions such as whether the total structure vector components exceed 4, or whether their basic types differ, or whether the struct contains non-vector-or-scalar members.  Another query returns the appropriate TType for the sampler.

High level summary of design:

In the TSampler, this holds an index into the texture structure return type table:

    unsigned int structReturnIndex : structReturnIndexBits;

These are the methods to set or get the return type from the TSampler.  They work for vector or structure returns, and potentially could be expanded to handle other things (small arrays?) if ever needed.

    bool setTextureReturnType(TSampler& sampler, const TType& retType, const TSourceLoc& loc);
    void getTextureReturnType(const TSampler& sampler, const TType& retType, const TSourceLoc& loc) const;

The ``convertReturn`` lambda in ``HlslParseContext::decomposeSampleMethods`` is greatly expanded to know how to copy a vec4 sample return to whatever the structure type should be.  This is a little awkward since it involves introducing a comma expression to return the proper aggregate value after a set of memberwise copies.

### DIFF
--- a/Test/baseResults/hlsl.texture.struct.frag.out
+++ b/Test/baseResults/hlsl.texture.struct.frag.out
@@ -1,0 +1,1184 @@
+hlsl.texture.struct.frag
+Shader version: 500
+gl_FragCoord origin is upper left
+0:? Sequence
+0:38  Function Definition: fn1(t2-tx-struct0-1; ( temp structure{ temp float c0,  temp 2-component vector of float c1,  temp float c2})
+0:38    Function Parameters: 
+0:38      't1' ( in texture2D)
+0:?     Sequence
+0:38      Branch: Return with expression
+0:38        Sequence
+0:38          move second child to first child ( temp 4-component vector of float)
+0:38            '@sampleResultShadow' ( temp 4-component vector of float)
+0:?             texture ( temp 4-component vector of float)
+0:38              Construct combined texture-sampler ( temp sampler2D)
+0:38                't1' ( in texture2D)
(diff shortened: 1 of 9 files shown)


LoopDawg:
Current form calls sample operation too many times while unpacking results into user struct.  Probably nominally gets the right answer, but inefficiently.  Will push a fix.

LoopDawg:
Repush with changes to copy texture sample result into temporary variable, to avoid issuing redundant sample operations.  The temp is now indirected when copying into the user struct type from the vec4 sample result.

LoopDawg:
Repush with added test for textures templatized on the same core struct type, to ensure e.g, fn signature matching works as expected.

LoopDawg:
Minor nonfunctional code and comment updates, removed WIP.

PR #1028, opened 2017-08-23
HLSL: handle clip and cull distance input builtin type conversion

### LINKED ISSUE #1026: HLSL: clip and cull distance need input support
PR #947 and #997 added output variable type conversion for clip and cull distance builtins, e.g, between potentially multiple source level vectors, arrays, or scalars, and the SPIR-V required single array.  The opposite transform is needed for inputs.

LoopDawg:
(Probably works or mostly works, but still WIP).

HLSL allows a range of types for clip and cull distances.  There are three dimensions, including arrayness, vectorness, and semantic ID. SPIR-V requires clip and cull distance be a single array of floats in all cases.

This code provides input side conversion between the SPIR-V form and the HLSL form.  (Output conversion was added in PR #947 and #997).

This PR extends HlslParseContext::assignClipCullDistance to cope with the input side conversion.  There are new .frag tests matching the output structure of the similarly numbered .vert.

Currently, non-{frag,vert} stages are not handled, and are explicitly rejected.

Fixes #1026.


LoopDawg:
Repush: explicitly reject non-{frag,vert} stages.  The conversion logic isn't working with the implicit array dimension.  That can be addressed in its own PR.  For now, those cases error out.

Also minor comment improvements.

LoopDawg:
Removed WIP.

johnkslang:
Question: Has this been working correctly on output, for IO-implicit array interfaces?

johnkslang:
Note also this area was one of the ones blocking the I/O flattening from working on all stages.

PR #1030, opened 2017-08-24
HLSL: include the semanticName in getCompleteString()

xxxbxxx:
to be able to check the value in tests.


johnkslang:
Note that any semantic-name that has real semantic impact on the AST has been translated to information the AST already emits. It would actually not be a good idea to start looking at the text string instead of the other information. So, beyond this failing all tests, I'd like to know the motivation for really needing this.

xxxbxxx:
The motivation is that the semantic value is usefull for introspection when compiling hlsl ported from directx 
and being able to see if proper values are computed when running "make test" is nice.

Issue #772, opened 2017-03-15
HLSL token pastes after expand instead of before? (Opposite of C++.)


redorav:
I have another bug, this time for the preprocessor. It seems like we can't use a defined token inside a define expression. So something like the following won't work.

```
#define DEFINE_TEXTURE(name) Texture2D name; SamplerState name##_ss;
#define SAMPLE_TEXTURE(name, uv) name.Sample(name##_ss, (uv).xy)

#define test_texture2 test_texture

DEFINE_TEXTURE(test_texture)

float4 test_px(float4 input : TEXCOORD0)
{
    float4 tex = SAMPLE_TEXTURE(test_texture2, input.xy);
    return tex;
}
```

johnkslang:
This may be a difference from C++, and hence GLSL.  In C++ you paste before you expand.

Both C++ and glslang create a token `test_texture2_ss` because those are the tokens you requested to be pasted.

johnkslang:
Fixed by https://github.com/KhronosGroup/glslang/pull/1031

PR #1031, opened 2017-08-24
HLSL: fix preprocessor concatenation behaviour.

xxxbxxx:


johnkslang:
This PR has a problem, beyond failing tests, of replicating another PR, rather than being independent of it.

It would be good to see the just the preprocessor fix as its own PR.

xxxbxxx:
updated.

johnkslang:
This issue came up in #772. Does this get the correct results for the test there?  Or, you prefer I verify?

glslang/MachineIndependent/preprocessor/Pp.cpp
```diff
@@ -1068,6 +1068,10 @@ int TPpContext::tMacroInput::scan(TPpToken* ppToken)
         pasting = true;
     }
 
+    // HLSL fxc.exe does expand macros before concatenation
```
johnkslang: I hope this is part of HLSL, no matter the front end, not an aspect of fxc.  Can we remove the "fxc.exe"?

xxxbxxx:
I added #772 as a test case and compared the output with fxc and it looks the same.

(I have no idea where the hlsl preprocessor behaviour is documented.)

johnkslang:
Looks good, thanks.

PR #1032, opened 2017-08-24
HLSL: fix array[1] of vec4 constant declaration.

xxxbxxx:
+ const arrays with only one element where interpreted as an immediate. 
for example, 
`float4 a1[1] = { float4(1,2,3,4) };  `
a1[0] was the x component of the float4, instead of the float4 itself.

+ And a some bound checking tests, instead of accessing uninitialized memory in case of a typo..

### DIFF
--- a/Test/baseResults/hlsl.array.frag.out
+++ b/Test/baseResults/hlsl.array.frag.out
@@ -2,72 +2,134 @@ hlsl.array.frag
 Shader version: 500
 gl_FragCoord origin is upper left
 0:? Sequence
-0:8  Function Definition: @PixelShaderFunction(i1;vf4[3]; ( temp 4-component vector of float)
-0:8    Function Parameters: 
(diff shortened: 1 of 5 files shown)


johnkslang:
For the first commit, there is already GLSL code to do a more thorough job of this, and correct the index so parsing continues. I didn't at first know HLSL semantics for this, but it seems to match closely enough.  I will promote that from the GLSL parse class to the shared base class.

For the second commit, I'm worried that breaks something. I will analyze deeper.

Thank you for the contributions.

johnkslang:
Note the new `const` lines in the test file get warnings from FXC about not being static.

johnkslang:
I cherry picked your second commit, and somewhat generalized it.  So, ae8af5d33e11513ac93113c254f2dcaee809c6f1 now closes this.

PR #1036, opened 2017-08-29
Fix warning about conversion from size_t to int

baldurk:
Not necessarily to be merged as is unless someone can confirm it's safe. I ran into this on MSVC2015 with warnings-as-errors, but I can't tell if it's safe to silence the warning like this - because indirectly it's pointing to code doing something that might underflow.

I can't tell from the surrounding code if `args` could ever be empty - in which case doing `args.size() - 1` as unsigned size_t would underflow I think. `setArgList` only gets called further down below, and from an uninformed position I can't tell if it's theoretically possible to end up with an empty struct getting to there.


johnkslang:
Okay, will do an equivalent that is more clearly safe, thanks.

PR #1037, opened 2017-08-29
HLSL: add geometry stage support for clip/cull distance

LoopDawg:
Changes:

(1) Allow clip/cull builtins as both input and output in the same shader stage.  Previously, not enough data was tracked to handle this.

(2) Handle the extra array dimension in GS inputs.  The synthesized external variable can now be created with the extra array dimension if needed, and the form conversion code is able to handle it as well.

For example, all of these GS input lines would result in the same synthesized external type:

    triangle in float4 clip[3] : SV_ClipDistance;

    triangle in float2 clip[3][2] : SV_ClipDistance;

    triangle in float2 clip0[3] : SV_ClipDistance0;  triangle in float2 clip1[3] : SV_ClipDistance1;

In the second case, the inner array dimension packs with the 2-vector of floats into an array[4], which there is an array[3] of due to the triangle geometry.  This PR also handles separated semantic IDs (e.g, SV_ClipDistance0/1) on input side, or output side, or both, such as in the 3rd example above, where the variables will be merged.

I was going to add HS/DS into this PR (all the mechanisms are likely the same), but the language seems quite constrained when using clip/cull in those stages.  Most attempts at testing clip/cull in those stages failed to compile, although Vulkan allows them as both in/out there, per "14.6. Built-In Variables".  Perhaps we can revisit later if HS/DS support is needed, maybe with input guided by real workloads, which I don't currently have an example of.


LoopDawg:
Repush: add lambda function to clean up some previously replicated code.

LoopDawg:
Repush to add more GS testing combinations (e.g, separated 0 and 1 semantics for both input and output side).

LoopDawg:
Removed WIP.

PR #1039, opened 2017-08-31
Fix issues of GL_ARB_viewport_layer_array

amdrexu:
The definitions of gl_Layer and gl_ViewportIndex should be independent of vendor implementation.


amdrexu:
Last submission experiences unknown build failure. Now, it is OK.

PR #1044, opened 2017-09-05
Add support for GL_OES_EGL_image_external_essl3

dsrbecky:


dsrbecky:
PTAL :-)

julee:
Hi, 
Why use glslangValidator to validate the shader with  GL_OES_EGL_image_external_essl3 works,
but use -G to generate spv file fails, saying:
'texture' : no matching overloaded function found

PR #1047, opened 2017-09-06
Add stdin option (fixes #598)

### LINKED ISSUE #598: read file from stdin
I have a library, wher the shader is partially written by the library user, and partially generated at compile time. The full shader itself exists only in text form, and never as a file. It would be nice, if I could run this validator on the generated shader source, without the requirement, to write it into a temp file. If glslangValidator could read a file from stdin, I could leave out the creation of the temporary file, and have (in my opinion) more pretty code.

svenstaro:
This adds `--stdin` which allows you to specify that a file should be provided on stdin. Since we can't parse the filename in that case, we'll have to force the user to specify `-S` as well.

I handled both the legacy and modern ways for stdin and tested both.

I think me providing a dummy WorkItem called "stdin" in case of stdin is not too bad of a hack. I tried to make the best usage of what was already there in order to keep the patch as small as possible.

This also fixes #598 so that's nice.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -94,6 +94,7 @@ enum TOptions {
     EOptionHlslIoMapping        = (1 << 24),
     EOptionAutoMapLocations     = (1 << 25),
     EOptionDebug                = (1 << 26),
+    EOptionStdin                = (1 << 27),
 };
 
 //
@@ -469,6 +470,9 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                         sourceEntryPointName = argv[1];
                         bumpArg();
                         break;
+                    } else if (lowerword == "stdin") {
+                        Options |= EOptionStdin;
+                        shaderStageName = argv[1];
                     } else if (lowerword == "suppress-warnings") {
                         Options |= EOptionSuppressWarnings;
                     } else if (lowerword == "target-env") {
@@ -606,6 +610,10 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
         }
     }
 
+    // Make sure that -S is always specified if --stdin is specified
+    if ((Options & EOptionStdin) && shaderStageName == nullptr)
+        Error("must provide -S when --stdin is given");
+
     // Make sure that -E is not specified alongside linking (which includes SPV generation)
     if ((Options & EOptionOutputPreprocessed) && (Options & EOptionLinkProgram))
         Error("can't use -E when linking is selected");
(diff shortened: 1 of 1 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -396,7 +397,10 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                     std::transform(lowerword.begin(), lowerword.end(), lowerword.begin(), ::tolower);
 
                     // handle --word style options
-                    if (lowerword == "auto-map-bindings" ||  // synonyms
+                    if (lowerword == "stdin") {
```
johnkslang: These are currently being done in alphabetical order. Should follow suit.

Also, is there a convention of just "--" alone meaning stdin?

StandAlone/StandAlone.cpp
```diff
@@ -1231,6 +1278,8 @@ void usage()
            "                                       uint32_t array named <name>\n"
            "                                       initialized with the shader binary code.\n"
            "  --vn <name>                          synonym for --variable-name <name>\n"
+           "  --stdin                              Read from stdin instead of from a file. You'll have to provide\n"
```
johnkslang: Need alphabetical order again.

johnkslang: Also, looks like it overflows the current 80 character limit the others are adhering to.

johnkslang:
This should appear as a single commit to glslang, which does not need a history of the mistakes made and corrected while developing a single increment change. (E.g., use `squash` and `push --force`.)

The precedent for testing bits is not strong, and evolving, so there is a mix, but I pointed out the preferred form a few times.

svenstaro:
I addressed all of your issues. Additionally, while both --stdin and - exist, I think in this case it's good to make it explicit and make people use --stdin because we also require them to provide -S in this case. I don't have a very strong opinion on this though so whatever you prefer.

StandAlone/StandAlone.cpp
```diff
@@ -606,6 +610,10 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
         }
     }
 
+    // Make sure that -S is always specified if --stdin is specified
+    if ((Options & EOptionStdin) && !(shaderStageName != nullptr))
```
johnkslang: Better to say `shaderStageName == nullptr` than `!(shaderStageName != nullptr)`, yes?

johnkslang:
My apologies for not submitting my previous review comments. (I'd thought the were all instant comments, but same came in under an unsubmitted review.)

StandAlone/StandAlone.cpp
svenstaro: Uh, you're right, of course!

StandAlone/StandAlone.cpp
```diff
@@ -973,10 +1008,15 @@ int C_DECL main(int argc, char* argv[])
             return ESuccess;
     }
 
-    if (workList.empty()) {
+    if (workList.empty() && !(Options & EOptionStdin)) {
```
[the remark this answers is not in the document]
svenstaro: Nope, that would change the logic. However, I'm going to make it `((Options & EOptionStdin) == 0)` instead.

svenstaro:
I addressed your further issues.

Issue #1046, opened 2017-09-05
Invalid SPIR-V from function call in HLSL shader


daveh-lunarg:
Seeing a validation error from a production HLSL shader while porting to a Vulkan renderer.  It's possibly related to issue #990, but does not require passing a struct. A minimal pixel shader that provokes the error is:
```HLSL
Texture2D      MyTexture : register(t0);
SamplerState   MySampler : register(s0);

//----------------------------------------------------------------------------------------
void TexFunc(
     in const float2 UVs, in const SamplerState s2D, in const Texture2D t2D,
     out float3 RGB
    )
{    
  RGB = t2D.Sample(s2D, UVs).rgb;
  //RGB = MyTexture.Sample(MySampler, UVs).rgb;
}

//-----------------------------------------------------------------------------------
void main() {
  float2 UVs = {0.5, 0.5};
  float3 final_RGB;

  TexFunc(   
      // Inputs
      UVs, MySampler, MyTexture,
      // Outputs
      final_RGB
    );
}
```
The SPIR-V output looks like:
```
repro.hlsl
// Module Version 10000
// Generated by (magic number): 80001
// Id's are bound by 42

                              Capability Shader
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main"
                              ExecutionMode 4 OriginUpperLeft
                              Source HLSL 500
                              Name 4  "main"
                              Name 19  "TexFunc(vf2;p1;t21;vf3;"
                              Name 15  "UVs"
                              Name 16  "s2D"
                              Name 17  "t2D"
                              Name 18  "RGB"
                              Name 21  "@main("
                              Name 31  "UVs"
                              Name 35  "MySampler"
                              Name 36  "MyTexture"
                              Name 37  "final_RGB"
                              Name 38  "param"
                              Decorate 35(MySampler) DescriptorSet 0
                              Decorate 35(MySampler) Binding 14
                              Decorate 36(MyTexture) DescriptorSet 0
                              Decorate 36(MyTexture) Binding 30
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 2
               8:             TypeSampler
               9:             TypePointer UniformConstant 8
              10:             TypeImage 6(float) 2D sampled format:Unknown
              11:             TypePointer UniformConstant 10
              12:             TypeVector 6(float) 3
[...]

PR #1048, opened 2017-09-06
Fix lvalue test in SPIR-V generation

LoopDawg:
There were several locations in TGlslangToSpvTraverser::handleUserFunctionCall testing for whether a fn argument should be in the lvalue or rvalue array.  They must get the same result for indexing sanity, but had slightly different logic.

Logic is now encapsulated in a lambda.

Review request: Verification that I picked the right one of the two logics, and these really should be getting the same answer.  (I think so, but am only loosely familiar with handleUFC).

Fixes: #1046


daveh-lunarg:
Thanks, Dawg! :)  @LoopDawg

PR #1051, opened 2017-09-10
HLSL: allow split type in InputPatch template type, & HS return types.

LoopDawg:
This PR adds:

(1) InputPatch parameters to patch constant functions use the correct internal (temporary) variable type.  This avoids validation errors if the input patch had a mixture of builtins and user qualified members.

(2) Return structures from hull shader entry points are implicitly arrayed, and indexed via the InvocationId.  That implicit indirection was encountering a limitation of handleAssign which couldn't deal with a non-symbol lvalue.  handleAssign is slightly expanded to cope with either a simple symbol, or an index into a symbol.

**LIMITATION:** There is currently a limitation to (1) above: if an InputPatch is used in a PCF, it must also have appeared in the main entry point's parameter list. That is not a limitation of the HLSL language.  Currently that situation is detected and an "implemented" error results.  The limitation can be addressed, but isn't in this PR.


LoopDawg:
Repush after rebase.
Removed WIP.

PR #1063, opened 2017-09-19
Remapper: make remapper robust against non-exiting error handlers

### LINKED ISSUE #1062: spirv-remap should tolerate non-exiting error handlers
(Mentioned in #1052, but this is a different issue at heart than the compiler warnings).

If a non-exiting error handler is used with the spirv-remap tool, and an error is detected, potentially bad things can happen downstream.  This situation should be handled more gracefully.  (The default error handler exits).

LoopDawg:
Remapper errors are generally fatal: there has been some unexpected situation while parsing the SPV binary, and there is no reasonable way to carry on.  The errorHandler() function is called in this case, which by default exits, but it is possible to submit a handler which does not.  In that case the remapper would carry on in a bad state.

This change ensures a graceful termination of the remap() function.

WIP because (1) need to add a testing case, and (2) is there any reason to avoid a try block?  It's old functionality, at least 2003 I think, if not earlier, but I didn't see any other uses of it in the code.  This seems like the robust way to handle this though, in preference to trying manually recover from every possible bailout path.

While a try {} catch {} construct would be the ideal and safe way to do this, that's off limits for certain environments, so this tries to do the same thing with explicit code, to catch all the bailout paths.

Fixes #1062.


johnkslang:
I thought we were not doing exception handling in glslang.  Does this require turning something extra on for some builds? Is that contained in just the remapper, and does not require any change to glslang regarding exception handling?

LoopDawg:
AFAIK it doesn't require anything else be turned on or linked in.  There are no compile option or cmake changes, and it passed all the builds.   (I'll add a test to verify it functionally, too, regardless of whether it ends up using try/catch or explicit recovery logic).   And yes, just in the remapper binary, so the glslang binary would be unaffected.

This could potentially be done without the try block, but it would need a fair bit more code and be more fragile against logic errors and subsequent introduction of more error cases.

johnkslang:
Okay, thanks.

I know there are some consumers that don't want to run in an exception environment, but maybe the remapper is primarily an offline tool and that's okay.

greg-lunarg:
Well, my recent efforts in legalizing glslang have been linking in the remapper to do module-level dead code cleanup. Is this going to break the exception-free requirements of glslang?

LoopDawg:
If there's an exception-free requirement coming from somewhere, and the remapper is linked into that environment, then I guess this should be redone to avoid the try block.  (Weird though: I don't think try/except requires any OS support for catching throws, unlike threads say, and it's been standard C++ a very long time.  But... if there is such an environment, then there is :) )

I'll look into doing it without the try.

johnkslang:
Right, it's not as OS thing, it is a compile-more-efficiently thing, hence their avoidance in glslang (due to request).

LoopDawg:
Should we add -fno-exceptions to the build then?  It will trigger a compilation error if a try block is used.  Also, if the option _isn't_ given, I believe the penalty is still paid, even if there is no try block present, so unclear if the benefit is being obtained.

Also, for interest, some measurements for remapping a 14KB module with --do-everything, and compiling an (unrelated) 488 line HLSL shader:

**no try blocks, -fno-exceptions**:
- Remapping: 465,216 max heap, 15KB max stack, 226448 byte program binary
- Compiling: 12,792,191 max heap, 24KB max stack, 2431576 byte program binary

**try blocks (including adding one to compiler), -fexceptions**
- Remapping: 465,216 max heap, 15KB max stack, 242744 byte program binary
- Compiling: 12,792,191 max heap, 24KB max stack, 2492680 byte program binary

So identical stack use, identical heap use, but the program binary sizes are 7% bigger (remapper) and 2.5% bigger (compiler).

johnkslang:
The size difference might matter in some environments, and those environments would be compiling with their flags set the right way to get it.

Agreed that since avoiding this is intentional, it would help to get an error message out if it is used.

LoopDawg:
Repushed: removed the try... catch construct, and replaced with explicit unwind paths to catch error bailouts.   I tried to get them all, but it's a manual process rather than guaranteed by the compiler, and not immediately obvious, so this is almost certainly imperfect.  (glslangValidator also has post-input-error troubles).

Also 1: added a few testing cases using bad SPV binaries, named as "remap.invalid-spirv-?.spv" in the Test tree.  I did not test every possible bailout path, since the bad binaries are made via hand-hex-editing good binaries.  (You can't spirv-as into a bad binary, since the assembler would refuse, and some binary errors have no textual form).

Also 2: a compiler option is added so the build will fail if a try... catch is encountered in the code.  Previously it would compile successfully, so you'd have to know not to do it.  That flag is only set for g++, but that should be enough, given that the auto-builds exercise that environment and will fail.

johnkslang:
Yes, it is a large imperfect task to try to emulate what the try-catch would have done. Was hoping there was an alternative (didn't intend that as the consequence of not using this mechanism).

E.g., (and maybe doesn't help current situation) the GLSL front end goes and queues up an EOF on shader the input on error, as one localized thing all errors go through, rather than using 500 exta if tests (just a guess) or using try catch.

Up to you if you see a third more streamlined choice, or you want to go with the current PR.

LoopDawg:
I think there's a grammar-typo (there's gotta be a word for that...) in the line starting with "E.g.", but I think I got the gist of it about a centralized thing.

So far I didn't manage to think of anything like that in the remapper, since the binary could be from arbitrary sources and be corrupted in arbitrary ways.  Once something goes wrong while reading it, the stack has to be unwound before anything bad happens - either manually, as here, or via throw.  Even if I changed some underlying reader functions to stop reading, they still have nothing sensible to say.  Also there may have been bad IDs or bogus offsets (say) already read which are then used for all manner of data structure addressing and so forth.

I guess there's also setjmp and friends, but that does not seem so wonderful for C++ - there are limitations about undefined behavior for stack objects with nontrivial destructors for example.

I'm inclined to go with this version for the remapper, unless there's a nice centralized thing, but I didn't think of one.  At least all the bailout tests are obviously bailout tests, because they look at the errorLatch member.

greg-lunarg:
> I think there's a grammar-typo (there's gotta be a word for that...)

Syntax error? :)

PR #1068, opened 2017-09-21
CMake file changes for SPIR-V legalization for HLSL

greg-lunarg:
Also added known-good mechanism to fetch latest validated spirv-tools.
Also added -Od and -Os to disable optimizer and optimize for size.

Fetching spirv-tools is optional for both glsl and hlsl. Legalization
of hlsl is done by default if spirv-opt is present at cmake time.
Optimization for glsl is currently done through explicit option.

Also includes code from JohnK to disable current workarounds dealing
with writes of opaque types.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,8 @@ option(ENABLE_NV_EXTENSIONS "Enables support of Nvidia-specific extensions" ON)
 
 option(ENABLE_HLSL "Enables HLSL input support" ON)
 
+option(ENABLE_OPT "Enables spirv-opt capability if present" ON)
+
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND WIN32)
     set(CMAKE_INSTALL_PREFIX "install" CACHE STRING "..." FORCE)
 endif()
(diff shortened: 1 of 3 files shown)


greg-lunarg:
Will resolve tomorrow.

johnkslang:
> Also includes code from JohnK to disable current workarounds dealing
with writes of opaque types.

That should probably show up as its own commit, like a merge of hlsl-assume-opt, not an independent copy of my commits.  You could have a commit history that starts (or ends?) with such a merge. Or, you could check in the infrastructure here, and I could soon follow up with a merge of of hlsl-assume-opt.

greg-lunarg:
At JohnK's request, I scaled this PR back to just the CMake changes. I will add the remaining content in subsequent PRs.

Cmake now looks for External/spirv-tools. If found, it links in SPIRV-Tools-opt and SPVRemapper, and adds -DENABLE_OPT to build.

johnkslang:
The CMake changes look good, on code reading. Want to test them out; is there an update to the readme.md that says how to correctly populate the external projects that the CMake changes are referring to?

greg-lunarg:
The known-good mechanism (and the README update) is in the next PR. In the meantime you can just clone the latest SPIRV-Tools into External.

PR #1070, opened 2017-09-22
Enable HLSL legalization

greg-lunarg:
Also added known-good mechanism to fetch latest validated spirv-tools.
Also added -Od and -Os to disable optimizer and optimize for size.

Fetching spirv-tools is optional for both glsl and hlsl. Legalization
of hlsl is done by default if spirv-opt is present at cmake time.
Optimization for glsl is currently done through the option -Os.

Legalization testing is currently only done on four existing shaders.
A separate baseLegalResults directory holds those results. All previous
testing is done with the optimizer disabled.


greg-lunarg:
@johnkslang Any idea why the Linux/g++ compiles are failing? Linux/clang seems to run fine. I have retried this three times and there doesn't seem to be any consistency on which source files the fails occur on. Have you seen this before?

johnkslang:
I've only ever seen failures to get the whole process started, or real/repeatable compilation errors.

Is the apparent randomness on which one fails just due to threading?

dneto0:
I wonder if GCC is running out of memory on the VM.  Perhaps there is too much parallelism at once.

Also, the compressing code code in SPIRV-Tools (source/comp) has one autogenerated source file that compilers try hard to analyze.   It may be that compiling that one translation unit causes peers in the parallel build to fail.  (Fratricide).

Glslang's use of SPIRV-Tools does not care for the compressing codec.  
Perhaps we should add an option to disable compiling that code.
Hey @atgoo what do you think?

dneto0:
I'm adding an option to disable building of the SPIRV-Tools compressing codec

antiagainst:
+1 to adding options to disable compressing.

antiagainst:
+1 to what @dneto0 said. We've experienced the error before on SPIRV-Tools; pretty much is the parallel build to blame. So I changed SPIRV-Tools to use `make` instead of `ninja` and using 4 concrrent jobs for `make`: https://github.com/KhronosGroup/SPIRV-Tools/pull/686.

antiagainst:
I've made similar changes to Travis configuration in #1071.

greg-lunarg:
Thanks everybody. I will pull in these changes as they become available.

dneto0:
@atgoo's fix is to disable the build of the compressing codec by default.  The Cmake option to *enable* it is SPIRV_BUILD_COMPRESSION.

I pushed that into SPIRV-Tools master.

dneto0:
Forgot to mention the SPIRV-Tools PR is https://github.com/KhronosGroup/SPIRV-Tools/pull/835

greg-lunarg:
Disabling compression build still results in travis build failure. Will try #1071 next. Will wait for that to be merged and try again.

johnkslang:
Looks like the problem went away.

dneto0:
Looks great.
But I didn't see where the preprocessor symbol ENABLE_OPT is defined. ??

greg-lunarg:
JohnK asked to split the commit into three pieces. The define of ENABLE_OPT  happens in CMakeLists.txt in the root directory and was put in in the previous PR. The next PR will disable the opaque workarounds JohnK initially put into glslang.

StandAlone/StandAlone.cpp
```diff
@@ -1201,6 +1217,8 @@ void usage()
            "  -H          print human readable form of SPIR-V; turns on -V\n"
            "  -I<dir>     add dir to the include search path; includer's directory\n"
            "              is searched first, followed by left-to-right order of -I\n"
+           "  -Od         disables optimization. May cause illegal SPIR-V for HLSL.\n"
```
dneto0: Bikeshed moment:   GCC, Clang and other compilers use -O0   (oh zero) for this purpose.  I would prefer to follow that prior pattern.

dneto0:
My two cents: I would rather see -O0 (oh zero) to disable optimizations, following the example of GCC, Clang and many other compilers.

johnkslang:
What is the default state though? On the one hand we wanted automatic legalization of HLSL when necessary (starting with always for simplicity, but the front-end will have a flag saying when it is necessary, and that's in the middle of the compile that becomes known). On the other hand, all tests got a -O added, which I wasn't quite expecting, but implies it must be requested to get it, and on the other other hand, there is this option to turn it off.

johnkslang:
FYI, the hlsl-assume-opt branch has

```
    void setNeedsLegalization() { needToLegalize = true; }
    bool needsLegalization() const { return needToLegalize; }
'''

In the 'TIntermediate' and it is currently set when opaques are written too. I thought we'd worked out with Greg that he could use this, but at first turn on the front-end wouldn't be quite right, so he would write as if it was always needed.

File: StandAlone/StandAlone.cpp (hunk 1 above)
Comment (greg-lunarg) [reply StandAlone/StandAlone.cpp]: I thought of that but here is why I decided against it...

I wanted something to truly disable the optimizer. One reason was because we wanted to keep a lot of the pre-optimization testing that is currently done.

-Od can cause "illegal" SPIR-V to be generated. My thinking was we would want to generate only "legal" SPIR-V for  -O0. So my thinking is that -O0 will be the minimum optimization needed to generate legal SPIR-V.

So my proposal is to keep -Od and add -O0 at some point in the future. Right now, given these definitions, the default is basically -O0.

**Comment by greg-lunarg** [comment]
Re default, here is what happens now:

If glslang is built with a spirv-tools in External, then the default is to run the optimizer for HLSL compilations when SPIR-V is generated.

If glslang is not built with spirv-tools, the default is not to run the optimizer for HLSL. A (perhaps too subtle) warning is generated by cmake stating that illegal SPIR-V may be generated for HLSL, but no such messages are generated when invoking glslangValidator.

This was in accordance with what I thought was JohnK's request that users (even HLSL users) would be able to continue to work with glslang as they had before adding the spirv-opt capability. Let me know if I misunderstood.

And yes, I also was following JohnK's request that for the moment *all* HLSL compiles would get legalized.

**Comment by greg-lunarg** [comment]
The apparent Travis failure on the multithread capability of glslang seems on the surface to be related to the previous failure we were seeing where multiple parallel compilations of g++ were causing failures. Have we seen this before?

**Comment by greg-lunarg** [comment]
FWIW, I do have one last commit I would like to add to this checkin. It skips building the spirv-tools tests so that they are not run when ctest is run in the glslang/build director. I am holding off pushing that commit until people weigh in on the current travis failure.

**Comment by johnkslang** [comment]
The multi-threaded run of glslang against all tests in the test directory crashed. May be related to any new tests you added? They would need to error out appropriately.  I have been thinking of narrowing the focus of that test, as it can sometimes pick up unwanted input.

However, the time before this, they passed. Was there a recent commit change that effected tests?

Adding your final commit will be a good retrigger, if there are no other theories about this.

**Comment by johnkslang** [comment]
@greg-lunarg 

> If glslang is built with a spirv-tools in External, then the default is to run the optimizer for HLSL compilations when SPIR-V is generated.

> If glslang is not built with spirv-tools, the default is not to run the optimizer for HLSL.

I just would not deduce these defaults because you added -O to so many tests. If the above is true, it seems most tests could be left as they were. Yes?

**Comment by johnkslang** [comment]
@greg-lunarg 
> The apparent Travis failure on the multithread capability of glslang seems on the surface to be related to the previous failure we were seeing where multiple parallel compilations of g++ were causing failures. Have we seen this before?

No. Glslang's multithreading does not normally fail. If it seg faults, then an individual test segfaults. Did you invoke any optimization stuff (or any new code) in parallel? I thought not, as I think the parallel testing is all GLSL-interpreted, but perhaps there is a new piece of code somewhere that does not work in parallel?

**Comment by antiagainst** [comment]
I haven't checked the error carefully. But we recently pushed a fix into SPIRV-Tools (https://github.com/KhronosGroup/SPIRV-Tools/pull/830) regarding static global variable initialization. Not sure whether its that issue; but @greg-lunarg you may want to update SPIRV-Tools to the latest to give it a try. An easy step anyway.

**Comment by dneto0** [comment]
> I haven't checked the error carefully. But we recently pushed a fix into SPIRV-Tools (KhronosGroup/SPIRV-Tools#830) regarding static global variable initialization.

Background:  When incorporated into Microsoft's dxc, there was trouble with non-trivial class objects with static duration.  Their constructors were being called at the beginning of the process, but then the tool was substituting its own memory allocator process-wide, and the deallocations at the end were going to the wrong allocator.    The fix was to hide those objects behind a function call, so that allocation occurred after the allocator substitution took place.

**Comment by greg-lunarg** [comment]
@johnkslang I only added -Od to all existing HLSL tests. I added an additional run without -Od of the four existing HLSL shaders that require legalization. I did not add -Od to any GLSL tests. I did not add any additional shaders to Test so that should not have caused differences in the multithread test. The multithread test does not invoke HLSL or SPIR-V compilation, so theoretically spirv-opt should not be (explicitly) invoked during this test, although perhaps the issues that @dneto0 and @antiagainst may be coming into play during linking/initialization.

@dneto0 @antiagainst I will advance known-good to include the fixes you mentioned and try again.

**Comment by greg-lunarg** [comment]
Yeah, I rebuilt the branch from the current master and force-pushed it. Its a long story.

**Comment by antiagainst** [comment]
Now all tests passes! It seems that my guess was right. ;-P

**Comment by johnkslang** [comment]
Maybe. They had all passed before, and I did a close/reopen, and they failed. I don't know if that means before it was non deterministic or if something else changed.

**Comment by johnkslang** [comment]
So, testing this without adding SPIRV-tools, I get the following problems:

Code should be changed to build without warning, like this:

```
-                    spvOptions.disableOptimizer = Options & EOptionOptimizeDisable;
-                    spvOptions.optimizeSize = Options & EOptionOptimizeSize;
+                    spvOptions.disableOptimizer = (Options & EOptionOptimizeDisable) != 0;
+                    spvOptions.optimizeSize = (Options & EOptionOptimizeSize) != 0;
```

Test failures:

```
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] ToSpirv/HlslLegalizeTest.FromFile/hlsl_aliasOpaque_frag, where GetParam() = 8-byte object <F8-99 35-01 10-9A 35-01>
[  FAILED  ] ToSpirv/HlslLegalizeTest.FromFile/hlsl_flattenOpaque_frag, where GetParam() = 8-byte object <A8-9F 35-01 10-9A 35-01>
[  FAILED  ] ToSpirv/HlslLegalizeTest.FromFile/hlsl_flattenOpaqueInit_vert, where GetParam() = 8-byte object <C0-9F 35-01 10-9A 35-01>
[  FAILED  ] ToSpirv/HlslLegalizeTest.FromFile/hlsl_flattenOpaqueInitMix_vert, where GetParam() = 8-byte object <DC-9F 35-01 10-9A 35-01>

 4 FAILED TESTS
```

Let's get these fixed quickly, thanks.

antiagainst:
Hmm, that's interesting then. Didn't notice the close/reopen unstableness. Let's wait to see if it happens again in the future.

PR #1071, opened 2017-09-25
Travis: use make instead of ninja

antiagainst:


antiagainst:
Interesting. Using clang 3.9, some of the tests fail: https://travis-ci.org/KhronosGroup/glslang/jobs/279619092.

I haven't checked why that happens. Just switched back to use clang 3.6 now to unblock @greg-lunarg.

PR #1072, opened 2017-09-26
Implement extension GL_NV_shader_atomic_int64

amdrexu:


johnkslang:
The code that is present looks good.

Where is the check that the extension is enabled to allow use of the new functions? (That is, how is the new functionality protected against general use?) If the functions are too common for function-level testing, is something needed in builtInOpCheck()?

amdrexu:
Thank you. I do forget it and update the change.

Issue #1073, opened 2017-09-26
HLSL: invalid SPIR-V generated by SampleCmpLevelZero


danginsburg:
Please see related RADV bug which shows the offending SPIR-V and describes the bug with glslang: https://patchwork.freedesktop.org/patch/178831/

The problem is coming from HLSL that looks like the following:
```
SamplerComparisonState g_tShadowDepthTexture_sampler; 
Texture2D g_tShadowDepthTexture;

float DoGlobalLightShadow_PCF( float3 vPositionWs )
{
	float4 vPositionTextureSpace = mul( float4( vPositionWs.xyz, 1.0 ), g_matShadowWorldToTexture );
	vPositionTextureSpace.xyz /= vPositionTextureSpace.w;
	float2 vTexCoord = vPositionTextureSpace.xy;
	float flDepth = saturate( -vPositionTextureSpace.z ); // Why am I using -z?
	return g_tShadowDepthTexture.SampleCmpLevelZero( g_tShadowDepthTexture_sampler, vTexCoord.xy, flDepth );
}
```

See attached SPIR-V file that contains the SPIR-V showing the problem.

[hero_ps.zip](https://github.com/KhronosGroup/glslang/files/1333789/hero_ps.zip)

It seems the problem is maybe related to g_tShadowDepthTexture not being known to be a depth texture at declaration time but rather when sampling?

Note that the SPIR-V passes spirv-val and works on NVIDIA + AMD closed source drivers.  Should probably catch this in spirv-val if it is invalid spir-v (@dneto0).

johnkslang:
Yes, GLSL requires this in the image type. If HLSL allows a mixture of uses, this may be a problem, and we would want to relax the requirement that the depth information in the image type must match. If HLSL is consistent about this, we could fix the type with analysis. Needs some further investigation.

danginsburg:
Apparently HLSL allows mixed use of texture.  For example:

```
SamplerComparisonState g_tShadowDepthTexture_sampler; 
SamplerState testNotCmpSampler;
Texture2D g_tShadowDepthTexture;

float4 MainPs( void ) : SV_TARGET0
{
	float flShadow = g_tShadowDepthTexture.SampleCmpLevelZero( g_tShadowDepthTexture_sampler, float2( 0, 0 ), 0.5 );
	float flColor = g_tShadowDepthTexture.Sample( testNotCmpSampler, float2( 0, 0 ) ).r;
[...]

PR #1076, opened 2017-09-28
HLSL: force textures to shadow modes from combined samplers

LoopDawg:
Texture shadow mode must match the state of the sampler they are combined with.  This change does that, both for the AST and the symbol table.  Note that the texture cannot easily be *created* the right way, because this may not be known at that time.  Instead, the texture is subsequently patched.

This cannot work if a single texture is used with both a shadow and non-shadow sampler, so that case is detected and generates an error. This is permitted by the HLSL language, however.  See #1073 discussion.

Fixed one test source that was using a texture with both shadow and non-shadow samplers.

Addresses #1073.  (Is not a complete fix, given mismatch between what HLSL allows and what SPIR-V currently requires).


LoopDawg:
There's a loop which could be a range based for, but I don't remember if those are allowed.  I didn't see any others, so I left it explicit with iterators.

johnkslang:
Right, we are still staying away from range-based for loops.

PR #1080, opened 2017-10-02
Add getAsLoopNode() method for TIntermLoop

hedejing:
Every other type of node has a getAsXXXNode() method for them, a getAsLoopNode() for loop node is also necessary.

### DIFF
--- a/glslang/Include/intermediate.h
+++ b/glslang/Include/intermediate.h
@@ -790,6 +790,7 @@ class TIntermBranch;
 class TIntermTyped;
 class TIntermMethod;
 class TIntermSymbol;
+class TIntermLoop;
 
 } // end namespace glslang
 
@@ -817,6 +818,7 @@ public:
     virtual       glslang::TIntermMethod*        getAsMethodNode()          { return 0; }
     virtual       glslang::TIntermSymbol*        getAsSymbolNode()          { return 0; }
     virtual       glslang::TIntermBranch*        getAsBranchNode()          { return 0; }
+	virtual       glslang::TIntermLoop*          getAsLoopNode()            { return 0; }
 
     virtual const glslang::TIntermTyped*         getAsTyped()         const { return 0; }
     virtual const glslang::TIntermOperator*      getAsOperator()      const { return 0; }
@@ -829,6 +831,7 @@ public:
     virtual const glslang::TIntermMethod*        getAsMethodNode()    const { return 0; }
     virtual const glslang::TIntermSymbol*        getAsSymbolNode()    const { return 0; }
     virtual const glslang::TIntermBranch*        getAsBranchNode()    const { return 0; }
+	virtual const glslang::TIntermLoop*          getAsLoopNode()      const { return 0; }
     virtual ~TIntermNode() { }
 
 protected:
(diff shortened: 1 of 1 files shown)


glslang/Include/intermediate.h
```diff
@@ -817,6 +818,7 @@ class TIntermNode {
     virtual       glslang::TIntermMethod*        getAsMethodNode()          { return 0; }
     virtual       glslang::TIntermSymbol*        getAsSymbolNode()          { return 0; }
     virtual       glslang::TIntermBranch*        getAsBranchNode()          { return 0; }
+	virtual       glslang::TIntermLoop*          getAsLoopNode()			{ return 0; }
```
johnkslang: Need to follow local conventions of spaces, not tabs.

johnkslang:
This often comes out through visitLoop(), rather than manual discovery, but if you have a use case this is okay.  Fix the tabs vs. spaces first, thanks.

hedejing:
I am writing a small tool to modify shader and print the variant out. First I implement a class called ShaderPrintTraverser derived from TIntermTraverser and override all the visitXXX() methods to print all kinds of nodes. Then I implement another class called ReplaceNodeShaderPrintTraverser derived from ShaderPrintTraverser with two TIntermNode * members to record which node is modified and the node to replace it. Here are some codes I wrote about this:

```
class ShaderPrintTraverser : public glslang::TIntermTraverser {
public:
    ShaderPrintTraverser(std::ostream &out_, int TabCount_ = 4)
		:out(out_), TabCount(TabCount_){}

	virtual bool visitBinary(glslang::TVisit, glslang::TIntermBinary* node) override;
    //...other visitXXX() methods
};

class ReplaceNodeShaderPrintTraverser : public ShaderPrintTraverser {
public:
	ReplaceNodeShaderPrintTraverser(std::ostream &out_, TIntermNode *target_, TIntermNode *replacement_, int TabCount_ = 4):
		ShaderPrintTraverser(out_, TabCount_), target(target_), replacement(replacement_) {}

	virtual bool visitBinary(glslang::TVisit, glslang::TIntermBinary* node) override;
    //...
private:
	TIntermNode *target;
	TIntermNode *replacement;
};
```

In visitXXX() methods of ReplaceNodeShaderPrintTraverser I check if the current node is the node to be replaced and call ShaderPrintTraverser::visitXXX() to print the right content, just like this:
```
void ReplaceNodeShaderPrintTraverser::visitSymbol(glslang::TIntermSymbol* node) {
	if (node == target)
		ShaderPrintTraverser::visitSymbol(replacement->getAsSymbolNode());
	else
		ShaderPrintTraverser::visitSymbol(node);
}

bool ReplaceNodeShaderPrintTraverser::visitLoop(glslang::TVisit visit, glslang::TIntermLoop* node) {
	if (node == target)
		return ShaderPrintTraverser::visitLoop(visit, replacement->getAsLoopNode());
	else
		return ShaderPrintTraverser::visitLoop(visit, node);
}

//...
```
But currently there isn't a getAsLoopNode() method for loop nodes. Using dynamic_cast<> may do the trick, but it breaks the consistency between codes, so I suggest providing getAsLoopNode() for this case. Thanks for your patience.

johnkslang:
It it okay to add this functionality to glslang.

Proper process is the that these two commits should be squashed into a single commit.

hedejing:
OK I squashed them into a single commit.

PR #1081, opened 2017-10-03
HLSL: Fix crash when flattening both side of assignement simultaneously.

xxxbxxx:


johnkslang:
I have another fix coming that supersets this one, but realize it will also be good to see what happens with test results, so pulling this in anyway.

Issue #1082, opened 2017-10-03
HLSL mixed shadow sample mode conflicts may be avoidable 


daveh-lunarg:
PR #1076 added checks for mixed sampler modes in HLSL shaders.  A particular shader might include a header file which includes conflicting modes, yet only call functions that use a single mode.  This is currently flagged as an error in glslang, even though the conflict is essentially with dead code. 

For example:
```HLSL
struct PS_In 
{ 
    float3 tex0 : TEXCOORD0 ; 
} ; 

Texture2D              shadowTex : register ( t1 ); 
SamplerState           shadowSampler : register ( s2 ) ; 
SamplerComparisonState shadowCompSampler : register ( s3 ) ; 

float ShadowDepth (in float3 tc)
{
    return shadowTex.SampleLevel(shadowSampler, tc.xy, 0).x;
}

float ShadowComp (in float3 tc)
{
    return shadowTex.SampleCmpLevelZero(shadowCompSampler, tc.xy, tc.z);
}

float4 PS_Shadow0 ( in PS_In i ) : SV_TARGET 
{ 
    float value = ShadowDepth(i.tex0);
    float test  = ShadowComp(i.tex0);
    return float4(value, test, 0, 0);
} 

float4 PS_Shadow1 ( in PS_In i ) : SV_TARGET 
{ 
    float value = ShadowDepth(i.tex0);
    return float4(value, 0, 0, 0);
} 

float4 PS_Shadow2 ( in PS_In i ) : SV_TARGET 
{ 
    float test  = ShadowComp(i.tex0);
    return float4(0, test, 0, 0);
} 
 ```

It seems like PS_Shadow0 could never be compiled to SPIR-V, but PS_Shadow1 or PS_Shadow2 should compile without error.

LoopDawg:
I consulted with @johnkslang about this, who opined that the best available way to handle this is to generate the initial module by splitting the texture variable into two texture variables, one shadow and one not, and letting the optimizer eliminate the dead one.  (So, one must be dead).

I can pursue that and come up with a candidate PR.  I'm not equipped to test it under your workload, but I'll report back when there's something you could try.

daveh-lunarg:
Cool.  I'm not entirely sure we don't have any of the PS_Shadow0 cases, but it'll at least be much easier to weed those out once the others are eliminated.  Thanks for digging in!

LoopDawg:
[...]

PR #1089, opened 2017-10-06
HLSL: split textures used for both shadow and non-shadow modes

LoopDawg:
WIP: Still some changes in progress to remove the setId() method in the AST node.  That won't change external behavior.

A single texture can statically appear in code mixed with a shadow sampler and a non-shadow sampler.  This would be create invalid SPIR-V, unless one of them is provably dead.

The previous detection of this happened before DCE, so some shaders would trigger the error even though they wouldn't after DCE.  To handle that case, this PR splits the texture into two: one with each mode.  It sets "needsLegalization" (if that happens for any texture) to warn that this shader will need post-compilation legalization.

If the texture is only used with one of the two modes, behavior is as it was before.

Fixes: #1082


LoopDawg:
Oh crap, I forgot to put a WIP in the PR title :(.  I meant to, since I was going to replace the setId() method with something different if possible.

daveh-lunarg:
@LoopDawg, this PR _seems_ to be working with our HLSL collection, but can't say for 100% yet.  All the compilation failures are now successes, with the "AST, needs legalization" warning appearing in the output from glslangValidator.  It appears that all the affected shaders are being successfully 'legalized'; at least I'm not seeing any run-time shader creation complaints.  I've only tested a small pct of the affected shaders, so far.  

Thanks for the quick resolution!

LoopDawg:
OK, glad to hear it.

PR #1090, opened 2017-10-06
Only track variables in the global scope

### LINKED ISSUE #1084: Segmentation fault when compiling trivial HLSL shader
The latest source version of `glslangValidator` crashes on both Linux/Windows when compiling the following hlsl shader:
```HLSL
RWStructuredBuffer<uint> srt0;
void main() {
    RWStructuredBuffer<uint> srt0Local = srt0;
}
```
The underlying issue seems to be related to link symbol tracking where the symbol for srt0Local is removed during symbol table destruction for main, but the pointer is still left dangling in `linkageSymbols`.

tafuri:
This changes fixes #1084, although I'm not 100% sure this is the right thing to do?


johnkslang:
Blocks were originally only declared at global scope. Is this path being picked up from a parameter declaration or so?

Note that including test result changes is always done with a commit, which serves to answer questions like the above, in addition to protecting against regression.

Also, code changes need to follow local conventions for spaces (no tabs) and curly placement.

johnkslang:
FYI, I see now you are protecting against a local variable declaration. These should get optimized away, and agreed the spirit of your change is good.

Can you fix the formatting and add a test that shows it working correctly?

tafuri:
Yes, it seems like srt0Local gets picked up even though its just a reference to the original buffer (see the sample in the linked issue). Only the buffer counts get optimized away.

Sure thing. will do!

PR #1099, opened 2017-10-10
Set ENABLE_OPT default to OFF

atgoo:
Reason: decision to use ENABLE_OPT should be explicitely
taken by glslang client (such as shaderc).


dneto0:
That was a deliberate choice, as part of the HLSL legalization work.
See https://github.com/KhronosGroup/glslang/pull/1070#issuecomment-332346271

dneto0:
Can you achieve what you want by other means? (e.g. in Shaderc)
Also, I imagine that you would need to update Glslang test base results ?

PR #1105, opened 2017-10-16
Update spirv-tools known-good

greg-lunarg:


johnkslang:
This failed earlier, only because of an interaction with a mistake I made.

johnkslang:
Successfully rebased and merged into master now.

PR #1110, opened 2017-10-17
Nonfunctional: fix build warning

LoopDawg:
Fixes a build warning about unused function parameters.


johnkslang:
Actually, the entire argument should not be there, not just the name.

It is a new feature; we should either assert the stage matches the shader, or have no parameters at all.

LoopDawg:
OK: I'm not plugged into the motivation for that, so I'll withdraw the PR and let the involved parties fix the build warnings.

PR #1111, opened 2017-10-18
Add per-descriptor-set IO mapping shift values.

### LINKED ISSUE #1107: HLSL feature request: per descriptor set binding shifts
The current methods for shifting binding numbers in TShader shifts all descriptor sets with the same offsets:

```
    void setShiftSamplerBinding(unsigned int base);
    void setShiftTextureBinding(unsigned int base);
    void setShiftImageBinding(unsigned int base);
    void setShiftUboBinding(unsigned int base);
    void setShiftUavBinding(unsigned int base);
    void setShiftCbufferBinding(unsigned int base); // synonym for setShiftUboBinding
    void setShiftSsboBinding(unsigned int base);
```

If we could add variants of this accepting a descriptor set and fallback to the generic values if no shift is present, that should work.

LoopDawg:
WIP: still needs ~~interfacing with TProcesses~~, ~~testing~~, ~~documentation of new APIs~~, code cleanup, etc.

This PR adds the ability to provide per-descriptor-set IO mapping shift values.  If a particular binding does not land into a per-set value, then it falls back to the prior behavior (global shifts per resource class).

Because there were already 6 copies of many different methods and internal variables and functions, and this PR would have added 6 more, a new API is introduced to cut down on replication and present a cleaner interface.  For the global (non-set-specific) API, the old entry points still exist for backward compatibility, but are phrased internally in terms of the following.

    // Resource type for IO resolver
    enum TResourceType {
        EResSampler,
        EResTexture,
        EResImage,  
        EResUbo,  
        EResSsbo,
        EResUav, 
        EResCount
    };

Methods on TShader:

    void setShiftBinding(TResourceType res, unsigned int base);
    void setShiftBindingForSet(TResourceType res, unsigned int set, unsigned int base);

The first method replaces the 6 prior entry points of various spellings, which exist now in depreciated form.  The second provides per-resource-set functionality.  Both accept an enum from the list above.  This is also more extensible going forward.

From the command line, the existing options can accept either a single shift value as before, or a series of 1 or more [set offset] pairs.  Both can be provided, as in:

    ... --stb 20 --stb 2 25 3 30 ...

which will use the offset 20 for anything except descriptor set 2 (which uses 25) and 3 (which uses 30).

Fixes #1107

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -55,6 +55,7 @@
 #include <cctype>
 #include <cmath>
 #include <array>
+#include <map>
 #include <memory>
 #include <thread>
 
@@ -157,12 +158,11 @@ int OpenGLClientVersion = 450;           // doesn't influence anything yet, but 
 unsigned int TargetVersion = 0x00001000; // maps to, say, SPIR-V 1.0
 std::vector<std::string> Processes;      // what should be recorded by OpModuleProcessed, or equivalent
 
-std::array<unsigned int, EShLangCount> baseSamplerBinding;
-std::array<unsigned int, EShLangCount> baseTextureBinding;
-std::array<unsigned int, EShLangCount> baseImageBinding;
-std::array<unsigned int, EShLangCount> baseUboBinding;
-std::array<unsigned int, EShLangCount> baseSsboBinding;
-std::array<unsigned int, EShLangCount> baseUavBinding;
+// Per descriptor-set binding base data
+typedef std::map<unsigned int, unsigned int> TPerSetBaseBinding;
+
+std::array<std::array<unsigned int, EShLangCount>, glslang::EResCount> baseBinding;
+std::array<std::array<TPerSetBaseBinding, EShLangCount>, glslang::EResCount> baseBindingForSet;
 std::array<std::vector<std::string>, EShLangCount> baseResourceSetBinding;
 
 // Add things like "#define ..." to a preamble to use in the beginning of the shader.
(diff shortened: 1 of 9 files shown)


LoopDawg:
Repush: fix command line parsing error, add test.  Still WIP pending documentation and TProcesses integration.

LoopDawg:
Repush: add TProcesses support.

johnkslang:
Nice improvement.

chaoticbob:
Quick question on this, the ordering of the command line options, e.g.:
````
--stb <set#> <binding#>
````
...these are flipped from how it's arranged in the HLSL source: ```register(binding#, space#)```  - any particular rationale for this, in case someone asks?

LoopDawg:
> ...these are flipped from how it's arranged in the HLSL source: register(binding#, space#) - any particular rationale for this, in case someone asks?

I'm pretty agnostic about it myself, so if you or anyone else feels strongly it should be the other way about, I'm happy to reverse it (with awareness from @danginsburg, who filed the issue).  The current way I picked for symmetry with the ``-resource-set-binding [stage] name set binding`` option.  It's a good point though, that it doesn't match the register() form any more, which could be confusing.

The TShader API is the same as the command line form, and the opposite of the register form.

johnkslang:
We also set up the attribute syntax as `(binding, set)` because the set has a default, but the binding does not.

antiagainst:
Yep. It seems doing the reverse order gives better consistency.

LoopDawg:
> Yep. It seems doing the reverse order gives better consistency.

OK, thanks.  If everyone is agreed that's better, I can make a PR to reverse it.  I'll leave any such as WIP until explicit synchronization with @danginsburg, to avoid surprises since he may be using or nearly using the current form.

danginsburg:
> OK, thanks. If everyone is agreed that's better, I can make a PR to reverse it. I'll leave any such as WIP until explicit synchronization with @danginsburg, to avoid surprises since he may be using or nearly using the current form.

I don't use the command-line form since we drive glslang from the C++ API, so I am fine with changing the binding/set in the command-line ordering.  Thanks.

LoopDawg:
Note also that #1116 makes everything consistent the same direction, including the TShader API.  So

     TShader::setShiftBindingForSet(TResourceType res, unsigned int set, unsigned int base)
becomes

    TShader::setShiftBindingForSet(TResourceType res, unsigned int base, unsigned int set)

It would be possible to flip only the command line interface, but then there's another kind of inconsistency.

PR #1114, opened 2017-10-19
Add script to crank test shaders through spirv-val

LoopDawg:
**WIP: This has only been tried in a linux environment so far**

This script will crank a supplied set of glslang test shaders through the spirv-val tool, reporting on the results.

There are some important things to note:

* Like 'runtests', this must be invoked from the 'Test' subdirectory.

* This is mostly useful on the hlsl.* tests, although it is not strictly limited to those.  The reason is that most of the glsl tests either contain validation error cases, and so fail to compile, or are not using a #version  compatible with producing SPIR-V modules.

* Some tests, such as negative tests, or many of the glsl tests, have intentional compilation errors.  This script treats that as OK.  A failure is a successfully compiling shader which proceeds to fail spirv-val.

* spirv-val is looked for in either the External directory, or if not found there, in a sibling directory of glslang, and if not found there either, in /usr/local/bin.

* There are a bunch of command line options.  ``./validate-shaders.sh --help`` will describe them.

Some examples to try:

    ./validate-shaders.sh hlsl.*   # exercise all hlsl.* tests.

    ./validate-shaders.sh --terse hlsl.*  # same, but tersely.

    # dump validator results for problems in mybadshader.frag:
    ./validate-shaders.sh --quiet --dump-val mybadshader.frag

### DIFF
--- a/Test/validate-shaders.sh
+++ b/Test/validate-shaders.sh
@@ -1,0 +1,269 @@
+#!/bin/bash
+
+# This script validates shaders (if successfully compiled) using spirv-val.
+# It is not meant to preclude the possible addition of the validator to
+# glslang.
+
+declare -r EXE='../build/install/bin/glslangValidator'
+
+# search common locations for spirv-tools: keep first one
+for toolsdir in '../External/spirv-tools/build/tools' '../../SPIRV-Tools/build/tools' '/usr/local/bin'; do
+    [[ -z "$VAL" && -x "${toolsdir}/spirv-val" ]] && declare -r VAL="${toolsdir}/spirv-val"
+    [[ -z "$DIS" && -x "${toolsdir}/spirv-dis" ]] && declare -r DIS="${toolsdir}/spirv-dis"
+done
+
+declare -r gtests='../gtests/Hlsl.FromFile.cpp ../gtests/Spv.FromFile.cpp'
+
+declare -r targetenv='vulkan1.0'
+
+function fatal() { echo "ERROR: $@"; exit 5; }
+
+function usage
+{
+    echo
+    echo "Usage: $(basename $0) [options...] shaders..."
+    echo
+    echo "   Validates shaders (if successfully compiled) through spirv-val."
+    echo
+    echo "General options:"
+    echo "   --help          prints this text"
+    echo "   --no-color      disables output colorization"
+    echo "   --dump-asm      dumps all successfully compiled shader assemblies"
+    echo "   --dump-val      dumps all validation results"
+    echo "   --dump-comp     dumps all compilation logs"
+    echo "Spam reduction options:"
+    echo "   --no-summary    disables result summaries"
+    echo "   --skip-ok       do not print successful validations"
+    echo "   --skip-comperr  do not print compilation errors"
+    echo "   --skip-valerr   do not print validation errors"
+    echo "   --quiet         synonym for --skip-ok --skip-comperr --skip-valerr --no-summary"
+    echo "   --terse         print terse single line progress summary"
+    echo "Disassembly options:"
+    echo "   --raw-id        uses raw ids for disassembly"
+    echo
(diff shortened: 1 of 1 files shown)


johnkslang:
Thanks.

Is this something on a vector to hook up with integrated (bot) testing?

LoopDawg:
> Is this something on a vector to hook up with integrated (bot) testing?

I would think it could be, perhaps with a strategic choice of initial set, and percolating a few currently failing ones into the good set as it becomes possible.

It would add a little execution time.  It takes about 8 seconds to verify the hlsl.* subset with a release glslang build (Linux, ~3.5 GHz), and the script is single threaded.  It's ~30 seconds with a debug build.  Presumably it would be a little slower under cygwin, but I haven't tried that.

Test/validate-shaders.sh
```diff
@@ -0,0 +1,263 @@
...
+        glsl) ;;
+        bin) continue;;   # skip binaries
+        *) $opt_quiet || status "$test" "${red}UNKNOWN LANGUAGE${no_color}"; continue;;
+    esac
+
+    # compile the test file
+    if compout=$("$EXE" -e "$entry" $langops -V -o "$spvfile" "$test" 2>&1)
+    then
+        # successful compilation: validate
+        if valout=$("$VAL" "$spvfile" 2>&1)
```
dneto0: This should also specify `--target-env vulkan1.0`
That way when the validator starts adding Vulkan-specific checks, then we'll catch them.

dneto0:
This is a step in the right direction from my perspective.

dneto0:
A alternative is to add a --validate flag to glslangValidator to have it run validation itself.  Then we'd need less machinery.  I don't know the about the attitude about code bloat, but for HLSL compilation SPIRV-Tools is already being pulled in for the optimizer.  (Currently SPIR-V validation is in the core SPIRV-Tools library.)

Test/validate-shaders.sh
LoopDawg: Thanks for the review, @dneto0.  Added --target-env.

LoopDawg:
> A alternative is to add a --validate flag to glslangValidator to have it run validation itself.

That's a good point.  Would be interested in @johnkslang's take.

johnkslang:
I'm interested in a glslang --validate, if it is something that can be set, but not cause trouble if the validation code is missing.  This would be similarly to the HLSL legalization.

The critical need is that glslang changes are in fact run against the validator, and since SPIR-V tools are not always present in glslang, it still makes sense for someone to have a script that could use a side validator.

With a glslang --validate, and results captured correctly, it would work for the bot testing, so the only pending question is how a manual local script run compares against what the bots might see (e.g., it could have been a different validator version, etc.)

dneto0:
> I'm interested in a glslang --validate, if it is something that can be set, but not cause trouble if the validation code is missing. This would be similarly to the HLSL legalization.

Agreed.

> With a glslang --validate, and results captured correctly, it would work for the bot testing, so the only pending question is how a manual local script run compares against what the bots might see (e.g., it could have been a different validator version, etc.)

A bot set up for the purpose of gating should use the known-good commit of SPIRV-Tools.  That should always pass.  So then there ought not to be a question about seeing something different from what the bot sees.

However, updating SPIRV-Tools to a newer known-good is its own step.  You'd have to resolve any problems introduced by that refresh.  It could be validator problems or glslang codegen problems.  
To avoid suprises there, for this kind of scenario I typically set up an *advisory* bot somewhere that pulls in latest Glslang and latest SPIRV-Tools.   The GitHub CI setup isn't great about clearly separating gating vs. advisory bots.  So I'm not sure if you want to pay for this cognitive load.

LoopDawg:
I'm thinking that this script (a) exists, (b) doesn't preclude adding glslang --validate, and (c) doesn't pollute glslang's binary with anything; it's just a thing sitting off to the side minding its own business.

If validation is added to the executable, this could either keep being what it is, or get deleted if it serves no more purpose.

Will remove WIP.

PR #1116, opened 2017-10-21
Reverse order of setShiftBindingForSet parameters

LoopDawg:
**WIP until coordination with all involved parties.**

Per feedback on PR #1111, this reverses the order of the parameters for the setShiftBindingForSet API.

It is now:

    void TShader::setShiftBindingForSet(TResourceType res, unsigned int base, unsigned int set);

No test result diffs (but the test source invocation reverses the order).


LoopDawg:
Coordinated as much as possible with consumers.  Removed WIP.

PR #1122, opened 2017-10-23
GLSL: Only parse [-]1.#INF in HLSL mode.

### LINKED ISSUE #1121: 1.#INF is accepted in GLSL shaders.
`1.#INF`/`-1#INF` is illegal in GLSL according to the specification, but is accepted by glslangValidator.

This was introduced by #833 (in response to #672).

xorgy:
Resolves #1121


johnkslang:
Thanks. Would just want to add a negative test in, say, numeral.frag.

johnkslang:
FYI, I added the test.

PR #1124, opened 2017-10-23
GLSL: Check for implicit inner dimension in array of atomic_uint.

### LINKED ISSUE #1123: Implicit-inner array of atomic_uint violates assertion in TArraySizes::getCumulativeSize()
```GLSL
#version 460
layout(binding = 0) uniform atomic_uint c[1][];
```

Crashes `TArraySizes::getCumulativeSize()` because of an insufficient for implicit arrays of atomic_uint. I think this is an oversight from the introduction of arrays of arrays.

xorgy:
Resolves #1123

### DIFF
--- a/Test/baseResults/implicitInnerAtomicUint.frag.out
+++ b/Test/baseResults/implicitInnerAtomicUint.frag.out
@@ -1,0 +1,21 @@
+implicitInnerAtomicUint.frag
+ERROR: 0:2: '[]' : only outermost dimension of an array of arrays can be implicitly sized 
+ERROR: 0:2: 'atomic_uint' : array must be explicitly sized 
+ERROR: 2 compilation errors.  No code generated.
+
+
+Shader version: 460
+ERROR: node is still EOpNull!
+0:?   Linker Objects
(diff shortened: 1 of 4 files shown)


johnkslang:
This is generally not allowed. Only the outer dimension can be implicit (once constructors and initializers are done filling in the details).  Was this a cascading issue, or is a more general fix called for?

xorgy:
@johnkslang I'll take another look once the power comes back on, but I'd guess that it has something to do with ordering. i.e. it checks for implicitly sized arrays of atomic uint before the general check for implicit outer arrays.

xorgy:
@johnkslang So after taking a look at this, it seems there are other ways to trigger this bug. It seems like declareVariable needs to be short-circuited when arraySizesCheck produces an error, like you implied.

xorgy:
#1131 Contains a revert and a fix for the more general problem.

johnkslang:
There was a generic check for implicit sizing already, giving an error message.

I added error recovery, to set the sizes, removing the implicitness, to avoid cascading error problems.

PR #1129, opened 2017-10-25
GLSL: Fold constant SHRT_MIN/INT_MIN/LLONG_MIN % -1 to 0.

### LINKED ISSUE #1128: GLSL: Constant integer INT_MIN or LLONG_MIN modulo -1 causes SIGFPE
```GLSL
#version 460 core
#extension GL_AMD_gpu_shader_int16 : enable
#extension GL_ARB_gpu_shader_int64 : enable

void a() {
    int16_t x = -32768S % -1S; // SHRT_MIN
    int y = -2147483648 % -1; // INT_MIN
    int64_t z = -9223372036854775808L % -1L; // LLONG_MIN
}                                                     
```
This is [handled as a special case for `EOpDiv`](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/Constant.cpp#L174), but [not for `EOpMod`](https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/Constant.cpp#L174).

xorgy:
Resolves #1128

### DIFF
--- a/Test/baseResults/constFoldIntMin.frag.out
+++ b/Test/baseResults/constFoldIntMin.frag.out
@@ -1,0 +1,51 @@
+constFoldIntMin.frag
+Shader version: 460
+Requested GL_AMD_gpu_shader_int16
+Requested GL_ARB_gpu_shader_int64
+0:? Sequence
+0:5  Function Definition: a( ( global void)
+0:5    Function Parameters: 
+0:6    Sequence
+0:6      Sequence
+0:6        move second child to first child ( temp int16_t)
+0:6          'u' ( temp int16_t)
+0:6          Constant:
+0:6            32768 (const int)
+0:7      Sequence
+0:7        move second child to first child ( temp int)
+0:7          'v' ( temp int)
+0:7          Constant:
+0:7            -2147483648 (const int)
+0:8      Sequence
+0:8        move second child to first child ( temp int64_t)
+0:8          'w' ( temp int64_t)
+0:8          Constant:
+0:8            -9223372036854775808 (const int64_t)
+0:9      Sequence
+0:9        move second child to first child ( temp int16_t)
+0:9          'x' ( temp int16_t)
+0:9          Constant:
+0:9            0 (const int)
+0:10      Sequence
+0:10        move second child to first child ( temp int)
+0:10          'y' ( temp int)
+0:10          Constant:
+0:10            0 (const int)
+0:11      Sequence
+0:11        move second child to first child ( temp int64_t)
+0:11          'z' ( temp int64_t)
+0:11          Constant:
+0:11            0 (const int64_t)
+0:?   Linker Objects
+
+
+Linked fragment stage:
+
+ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point
+
+Shader version: 460
+Requested GL_AMD_gpu_shader_int16
+Requested GL_ARB_gpu_shader_int64
+0:? Sequence
+0:?   Linker Objects
+
--- a/Test/constFoldIntMin.frag
+++ b/Test/constFoldIntMin.frag
@@ -1,0 +1,12 @@
+#version 460 core
+#extension GL_AMD_gpu_shader_int16 : enable
+#extension GL_ARB_gpu_shader_int64 : enable
+
+void a(){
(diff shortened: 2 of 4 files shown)


johnkslang:
Where are these semantics specified?

xorgy:
@johnkslang Nowhere, they're undefined.

johnkslang:
With missing semantics, we need some caution in changing existing semantics. But, maybe there is something in specifications that introduce int16?

@amdrexu do you know?

xorgy:
Specifically regarding the change for int16 here, it is going into an int16_t constant, so it can not store the value 32768 anyway. If it tried to store 32768, the actual value would be -32768.

xorgy:
@johnkslang How about I cut the second commit out entirely, the first one is a straightforward fix for a crashing bug, which generates what seem to be the industry standard values for these expressions.

xorgy:
Incidentally, it's a bit weird that the AST readout lists the type of the `int16_t` as `const int`.

johnkslang:
Note that so far, this is just a vendor-specific extension. It does indeed deserve better scrutiny if it becomes more than that.

I agree there is something to fix here. One thing I'm uncomfortable with is moving to abstract macros for min/max/etc. unless they are defined to be the correct bit widths. E.g., is GLSL int16 really defined to be the same thing as C++ short? Better to use explicit 0x.... numbers where we can see the bit pattern matches the GLSL extension. I know we assume int is 32-bits, but short is a bit more of a risk.

johnkslang:
To be clear, I am happy to take any subset of this that is a clear bug fix, and thank you for you effort on this.

xorgy:
Okay then, since we need to understand `int16_t` better in the 32768 vs. -32768 sense, I'll remove that commit from this PR. The first commit just keeps the validator from outright crashing, and only deals with the (well understood) encoding of `0`, not nearly as pesky of a number. :- )

johnkslang:
Is SHRT_MIN guaranteed to be 0x8000?  What we care about is the 0x8000, not the size short happens to be on the host compiler system.

xorgy:
@johnkslang On any system with two's compliment arithmetic, and 16-bit shorts. I don't know of any system with a C++11 compiler and an environment capable of running glslang where short is not 16 bits or not two's compliment. As a result, it seems that SHRT_MIN would be 0x8000 just as surely as INT_MIN is 0x8000000.

xorgy:
It's clear that the AMD_gpu_shader_int16 was written with the assumption that `int16_t` was, for all intents and purposes, `short`; otherwise I don't see why they would use the `s` suffix.

xorgy:
Now that you mention it though, there seems to be a discrepancy between real <limits.h> headers in the wild, and the way the ISO IEC 9899 prints it (in §5.2.4.2.1). The standard says it's -32767, the headers I can find actually say it's -32768,

I am very confused. Maybe SHRT_MIN (and INT_MIN, and LLONG_MIN) can not be trusted. To add insult to injury, the headers actually directly reference that section in the standard, which it appears the authors of the headers never did actually read.

xorgy:
The glibc headers say it's `(-32767 - 1)`, the FreeBSD headers for every platform (`$MACHINE_ARCH/_limits.h`) all define it as `(-0x7fff - 1)` or in one case `(-0x7fff-1)` just for sparc64.

xorgy:
It's up to you, we could remove the use of these `WHATEVER_MIN` macros entirely. In that case we should probably go and replace all of the `int` definitions in glslang with `int32_t`, that's how I do it in all of my greenfield code.

xorgy:
For what it's worth, ISO/IEC 14882 doesn't even bother to specify `SHRT_MIN`, just basically says that they could be anything, and they are intended to be whatever's in limits.h.

johnkslang:
>  In that case we should probably go and replace all of the int definitions in glslang with int32_t,

When I looked into that before, I found `int` was more portable (as 32 bit) than was `int32_t`.  Sounds like you found is equally portable (as 16 bit), but that the macros themselves could be suspect, yet, they also sound pretty portable in practice.

xorgy:
I think that, for better or worse, `int`, `short`, and `long long` have long been standardized to deal with the lack of <cstdint> in C++ (twelve or thirteen years late!). It's not perfect, but is de facto perfect.

Maybe in a few more years, we can count on cstdint?

PR #1131, opened 2017-10-26
Stop declareVariable short on an array with an inner implicit dimension.

xorgy:
`fixOffset()` and `layoutObjectCheck()` should never be reached if the variable in question has an inner implicit dimension.


xorgy:
My commits went somewhere, disregard this.

PR #1139, opened 2017-11-02
Remapper: Fix strip algo when strip ranges overlap

### LINKED ISSUE #1138: HLSL shader fails to compile with "spir instruction terminated too early"
The following HLSL shader is causing some trouble with glslang:

```
struct VSOut
{
    float4 Pos : SV_POSITION;
};


float4 Func1()
{
    float4 OutputColor = (1,0,0,0);
    return OutputColor;
    float4 OutputColor2 = (0,1,0,0);
    return OutputColor2;
}

float4 BugRepro(VSOut In) : SV_TARGET
{
    return Func1();
}
```

```
nsubtil@bender:~$ glslangValidator -D -V -S frag -e BugRepro -o simple.out simple.hlsl
simple.hlsl
spir instruction terminated too early
```

Adding `-Od` causes the problem to go away. Tested with TOT glslang (commit 715c353a1).

greg-lunarg:
Fixes #1138

### DIFF
--- a/SPIRV/SPVRemapper.cpp
+++ b/SPIRV/SPVRemapper.cpp
@@ -1392,7 +1392,7 @@ namespace spv {
 
         int strippedPos = 0;
         for (unsigned word = 0; word < unsigned(spv.size()); ++word) {
-            if (strip_it != stripRange.end() && word >= strip_it->second)
+            while (strip_it != stripRange.end() && word >= strip_it->second)
                 ++strip_it;
 
             if (strip_it == stripRange.end() || word < strip_it->first || word >= strip_it->second)


LoopDawg:
I think this is OK IFF nothing has changed the SPIR-V materially between the creation of the first overlapping range and the second.

That might in fact be the case.  OTOH, if it ever wasn't, it's safer to call strip() after each transform that can insert strip ranges (dceFuncs here).  So while AFAIK there's nothing wrong with the fix, and I don't object to it directly, maybe another PR could and/or should:

(1) Remove the bag of SPIRV from member variable status, and pass it as either a const or non-const reference to the transforms, so it is (somewhat, barring silliness like const_cast) enforced that some transforms don't change it directly, and thus, it's safe to defer the strip() until all those are done.  Or:

(2) Call strip() at the end of each of the transforms, whether or not they happen to modify the spv container, so as to be robust against future changes to the transforms.

Issue #976, opened 2017-07-10
Memory analysis and proposal


johnkslang:
From looking at all the the issues/PRs labeled as [Memory](https://github.com/KhronosGroup/glslang/labels/Memory), and deciding what do actually do.

### Relevant History

- Glslang stopped using its original Win32 DLL model:
  * `DllMain()` originally called to dispatch `DetachProcess()` and `DetachThread()`,
  * This code stopped being called.

- Addition of `TShader::preprocess()`:
  * this only fit the memory model if used as a substitute for `TShader::parse()`
  * no design for both: Externalized preprocessing was never part of the original memory design.

### List of Existing Problems

- Final cleanup is not done by the exposed interfaces of `FinalizeProcess()` or `ShFinalize()`
  * #916
  * #928 
  * #389 

- No scheme to have both `TShader::preprocess()` and `TShader::parse()` in use together.
  * #766 
  * #275 

- Multiple independent uses of glslang within a process are cumbersome because, not knowing about each other, they each want to do one-time initialization and finalization.

- The `TPragmaTable` container itself can leak (the contents are properly pool managed)
  * #916

- Looking inside `inUseList` after executing destructor, before deallocating memory. Works, but technically not proper.
  * #705

 - The original `threadPoolAllocator` is sometimes not freed up? Not yet clear which problem triggers this.
  * #916

### Independent Improvements

- Move all the OS-dependent stuff to use C++11 features.
  * #826: how orthogonal is this to the above?

- Stop worrying about threads altogether in glslang, use an instancing model for the whole thing
  * #389 note, needs a different interface?

## Proposed Plan

1. Fix the two existing interfaces `glslang::FinalizeProcess()` or `ShFinalize()` do to full clean up. The important thing to preserve is multiple independent uses get to share the one-time cost of the shared super-global symbol table.

1. Do 1 with a hidden mutexed reference count, such that multiple calls within a single process are okay.

1. Move `TPragmaTable` container to come from the pool, to avoid need for delete. Subset of #916.

1. Understand the issue if `SetThreadPoolAllocator()` is sometimes losing the original, and that needs to be preserved. (#916)  Fix any underlying problems.

1. Fix the preprocess interface problem. Since this does not make an AST and cannot be linked, and is relatively new, there is probably a simple way of having it clean up it's memory either automatically or through an extra interface call. Don't yet have a specific design; but possibly just having two independent pools that both get deleted.

1. Work around the need to look inside `inUseList` after executing its destructor.

[...]

PR #1147, opened 2017-11-12
Memory fixes, round 1

johnkslang:
First commit: Non-functional; clean-up rationalization.

Second commit:  

> Move to a normal model of ownership of memory pools, for new/delete.

> Addresses step 4 of #976, overlaps #916.

> For each pool, now, it is newed, remembered, and freed by the same entity, rather than having a mix (thread finalize freeing current pool) that could  lead to double freeing of the same pool. It is quite rational and simple now.

> This will enable reinstalling process and thread tear down.

Third commit: On it's way, will actually add tear down logic that's been skipped so far

Main issue: Nothing triggering per-thread tear down of additional threads.

### DIFF
--- a/OGLCompilersDLL/InitializeDll.cpp
+++ b/OGLCompilersDLL/InitializeDll.cpp
@@ -45,6 +45,10 @@ namespace glslang {
 
 OS_TLSIndex ThreadInitializeIndex = OS_INVALID_TLS_INDEX;
 
+// Per-process initialization.
+// Needs to be called at least once before parsing, etc. is done.
+// Will also do thread initialization for the calling thread; other
+// threads will need to do that explicitly.
 bool InitProcess()
 {
     glslang::GetGlobalLock();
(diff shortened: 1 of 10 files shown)


LoopDawg:
I added #1152 relative to this branch, which fleshes out OS_DumpMemoryCounters.

I tried 50 iterations of compiling a randomly selected shader from the test set with a debug build.  There's an initial bump, then it appears stable:

    Working set size: 13373440
    Working set size: 27119616
    Working set size: 27652096
    Working set size: 27652096
    Working set size: 27652096
    Working set size: 27652096
    Working set size: 27652096
    ...

johnkslang:
The need for per-thread tear down has been eliminated.

Pending questions:

- do pools need to be popped before being deleted?
- test multi-threading with the C++ interface; it's only been tested with the old Sh* interface so far

johnkslang:
Have removed popping before deletion, memory use is the same. And, locally wired up a multi-thread test with the new interface that worked.  That deserves a real PR in the future, done right at some point.

PR #1151, opened 2017-11-13
Improve build instructions

### LINKED ISSUE #956: ninja-build is not listed as a build dependency?
Why isn't ninja listed as a build dependency? If one was following the described build process and doesn't have the ninja-build package installed (GNU/LINUX) the build process will complain about ninja is missing. 

-
Linux devuan 3.16.0-4-amd64 #1 SMP Debian 3.16.43-2+deb8u1 (2017-06-18) x86_64 GNU/Linux

cgmb:
Fixed Windows build instructions:
- ./update_glslang_sources.py is not a valid command on Windows.
  Just call python explicitly.
- -DCMAKE_INSTALL_PREFIX=\`pwd\`/install creates a directory named \`pwd\`.
  Replaced with %CD%.

Simplified:
- Remove SSH clone instructions. Instructions on how to clone are for
  those unfamiliar with GitHub, so fewer options are better and SSH
  requires additional configuration to make work.
- Replace ninja with make, because more people are familiar with it and
  we probably don't need to explain how to install it. Fixes #956.
- The section mentioning cygwin is unclear and doesn't seem necessary
  to build.

Expanded:
- Mention minimum language version in the Dependencies section.
  Though perhaps we should elaborate on exactly which compilers are
  supported? Issue #42 discusses what compilers to support, but that
  information doesn't seem to be communicated anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,20 +57,22 @@ branch.
 
 ### Dependencies
 
+* A C++11 compiler
 * [CMake][cmake]: for generating compilation targets.
+* make: _Linux_, ninja is an alternative, if configured.
 * [Python 2.7][python]: for executing SPIRV-Tools scripts. (Optional if not using SPIRV-Tools.)
 * [bison][bison]: _optional_, but needed when changing the grammar (glslang.y).
 * [googletest][googletest]: _optional_, but should use if making any changes to glslang.
 
 ### Build steps
 
+The following steps assume a Bash shell. On Windows, that could be the Git Bash
+shell or some other shell of your choosing.
+
 #### 1) Check-Out this project 
 
 ```bash
 cd <parent of where you want glslang to be>
-# If using SSH
-git clone git@github.com:KhronosGroup/glslang.git
-# Or if using HTTPS
 git clone https://github.com/KhronosGroup/glslang.git
 ```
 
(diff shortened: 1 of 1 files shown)


johnkslang:
Probably, these two comments cancel each other out:

>./update_glslang_sources.py is not a valid command on Windows.

>The section mentioning cygwin is unclear and doesn't seem necessary
to build.

cgmb:
I'm afraid I don't understand.

johnkslang:
The update_glslang_sources.py script works in the cygwin environment.  So, I can understand how these come as a pair or can be eliminated as a pair.

cgmb:
Perhaps I'm missing something, but `python update_glslang_sources.py` seems to run fine without cygwin. Without that step, the `-Os` option causes glslangValidator to complain, but after running it and building, `glslangValidator.exe` accepts `-Os` without complaint...

Though, I'm starting to wonder... Was there an unmentioned assumption that everything on Windows was being run in the Bash shell that comes with Git? Is that cygwin? I had assumed the Windows instructions were for cmd, since they never said otherwise.

johnkslang:
The build instructions are assuming a normal Linux-like shell console, not a Windows-like command/cmd console.

cgmb:
I definitely misunderstood the instructions, then. I'll rework this PR to try to clarify that bash is the expected Windows shell, instead of 'fixing' the commands for cmd.

cgmb:
Actually, before I go do a second draft, were there any other issues? Is there any part of this you'd rather be done some other way?

johnkslang:
No. Generally, I appreciate your refinements here, thanks. Will take what you come up with, and tweak if necessary.

cgmb:
In light of the information about all platforms using the same shell, I wanted to see if there were any further simplifications I could make to the instructions. Ideally, the same set of commands would work on both Windows and Linux. After some investigation, I found a way of doing that but I didn't end up liking it much.

Anyways, it had been a while, so I just wanted to ping in to explain why this PR is still waiting on my changes. I just got too ambitious. :disappointed:

cgmb:
And, there we go.

PR #1152, opened 2017-11-13
Memory1 linux counters

LoopDawg:
This is relative to #1147 (Memory fixes, round 1).  It implements OS_DumpMemoryCounters for Linux, using getrusage().


johnkslang:
Thanks @LoopDawg.

I cherry picked the Linux-counters commit onto the memory1 branch in #1147, which has an intervening commit.

PR #1159, opened 2017-11-20
Fix SampleMask's SampleRateShading dependency

### LINKED ISSUE #1158: SampleMask incorrectly depends on SampleRateShading
CapabilitySampleRateShading for SampleMask should be removed from GlslangToSpv.cpp::TranslateBuiltInDecoration()
```
case glslang::EbvSampleMask:
        builder.addCapability(spv::CapabilitySampleRateShading);
        return spv::BuiltInSampleMask;
```

Spec language:
```
4.13. Changes from Version 1.00, Revision 11
Khronos SPIR-V issue #182: The SampleMask built in does not depend on SampleRateShading, only Shader.
```

iostrowsINTEL:
Fixes #1158


johnkslang:
This would need to come hand-in-hand with test-result changes.

johnkslang:
I'll look at doing this....

PR #1174, opened 2017-12-06
HLSL: add optional position.Y inversion

### LINKED ISSUE #1173: HLSL: Support Y flip for position outputs from VS/GS
This was previously discussed in #494 but I want to bring this up because the solution there is insufficient in practice.

APIs that natively support HLSL all have NDC Y pointing up. Because of this, any HLSL code base would assume this when outputting SV_Position. glslang currently compiles code like this as is, which, when used in Vulkan, would result in flipped rendering.

It's possible to solve this on the shader side, but that requires modifying every single vertex/geometry shader, which is obviously fragile.
The outcome of #494 seems to have been the VK_KHR_maintenance1 extension which allows you to flip the viewport height. Unfortunately, the support of this extensions remains non-existent on Android: https://vulkan.gpuinfo.org/listreports.php?extension=VK_KHR_maintenance1&option=not

An option that seems user-friendly is to flip position Y in glslang. This needs to be flagged in some way so that the caller can optionally specify it (it would also need to be eventually implemented in DXC...).

One alternative is to make a SPIRV-SPIRV pass but that seems somewhat more fragile (?).

#494 has some concern around which stage to flip the position in - I feel like it's sufficient for glslang to always flip gl_Position, and then the caller can appropriately pass this value if necessary (I'm not clear if it's valid to output gl_Position in VS and read it later in GS, which is the only scenario where you'd need to disable Y flip in VS).

LoopDawg:
Adds command line options:

    --invert-y
    --iy

(synonyms) which invert position.Y on vertex shader output.  Handles these cases:   

* Direct single variable return
* Member of direct returned struct
* Single variable output parameter
* Member of struct output parameter

API:

    // Enables position.Y output negation in vertex shader
    void TShader::setInvertY(bool invert);

Fixes #1173


StandAlone/StandAlone.cpp
```diff
@@ -1359,6 +1366,7 @@ void usage()
            "                                       uint32_t array named <name>\n"
            "                                       initialized with the shader binary code.\n"
            "  --vn <name>                          synonym for --variable-name <name>\n"
+           "  --negate-Y | --neg-Y                 negate position.Y output in vertex shader\n"
```
johnkslang: Should be just --ny by precedent.  Or maybe

  --invert-y
  --iy

or

    --neg-y
    --ny

PR #1175, opened 2017-12-07
HLSL: Stop flattening non-IO structs containing opaques.

johnkslang:
This makes struct returns from functions work, but breaks
structs containing arrays, due to limitations in subsequent
transforms in spirv-opt. This is expected to be fixed soon.


johnkslang:
The build failure looks nonsensical.  Will retrigger.

Issue #1176, opened 2017-12-07
HLSL: Right hand side is ignored during 


tafuri:
Declaring a buffer block, for example a RWStructuredBuffer with an assignment, ignores the right hand side. Making the following code compile without errors or warnings.
The assignment should clearly validate the assigned type and also copy over any applicable layout specifications, such as `binding` and `set`.

```hlsl
static const float TestValue = 1;

void main()
{
    RWStructuredBuffer<uint> values = TestValue;
    values[0] = 0;
}
```

johnkslang:
Couple background points:
- Vulkan does not support local buffers, they must be at global scope
- glslang is not an HLSL validator

However, it is okay to catch and report errors when the rules are understood well enough, and if post-front-end legalization eliminates the local buffer, resulting in only global buffers, this can work.

Do we know that if the assignment occurs that legalization passes eliminate it again?

tafuri:
I may have focused on the wrong thing with this issue so lets look at another example.

```hlsl
layout(set=4, binding=5)
RWStructuredBuffer<uint> Input;

void main()
{
    RWStructuredBuffer<uint> values = Input;
    values[0] = 0;
}
```

In HLSL `values` is basically a reference to `Input`. glslang
generates code which behaves in a similar way in the case where Input is bound to (0,0). But in the example above the layout specifications on Input get lost in the assignment, without any warning, and `values` will still be bound to (0,0).

A current workaround is to specify layout on `values` also, but that seems like a less than ideal workaround.

```hlsl
layout(set=4, binding=5)
RWStructuredBuffer<uint> Input;

void main()
{
    layout(set=4, binding=5)
    RWStructuredBuffer<uint> values = Input;
    values[0] = 0;
}
```

johnkslang:
The problem is that assignments to buffers are not yet supported.  This one gives an error message:

```
layout(set=4, binding=5)
RWStructuredBuffer<uint> Input;

void main()
{
    RWStructuredBuffer<uint> values;
    values = Input;
    values[0] = 0;
}
```

I would agree the initializer case should give the same error message, but I think it's not what you really want.  I'm guessing what you really need is:

- `values` is a local variable, not a global uniform
- the assignment/initializer to buffers works
- SPIRV-opt will copy propagate `Input` to where `values` is used, because it is illegal to assign to buffers
- glslang will generate the illegal code (assigns to buffer) and SPIRV-opt will legalize it, per the previous bullet

tafuri:
[...]

PR #1177, opened 2017-12-07
#1176 block initialization

tafuri:
I'm not at all certain this is the correct way to solve this. Since I'm unsure if it should copy the whole type. Some differences may be legal?
It seems to work for my purposes though, and makes it possible to assign buffers using custom layouts.


johnkslang:
Vulkan does not support local buffers.  See https://github.com/KhronosGroup/glslang/issues/1176 for more discussion.

If this is the path we want to take, the test examples need to result in generated code that shows it works; currently the declaration is just dead code.

johnkslang:
This is replaced with 8116cfc9bf1b48f61f85c8d059879b7ef2c34efb.

PR #1184, opened 2017-12-13
Add implementation of SPV_EXT_fragment_fully_covered

pdaniell-nv:
This implementation uses the GLSL extension [GL_NV_conservative_raster_underestimation](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_conservative_raster_underestimation.txt) to generate the new SPIR-V FullyCoveredEXT built in.


SPIRV/spirv.hpp
```diff
@@ -450,6 +450,7 @@ enum BuiltIn {
     BuiltInSecondaryViewportMaskNV = 5258,
     BuiltInPositionPerViewNV = 5261,
     BuiltInViewportMaskPerViewNV = 5262,
+    BuiltInFullyCoveredEXT = 5264,
```
johnkslang: Is this manually entered? We should publish new headers more quickly, so this can come from new headers. I will look at doing that.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5858,6 +5865,11 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
         }
 #endif
 
+#ifdef NV_EXTENSIONS
+        symbolTable.setVariableExtensions("gl_FragFullyCoveredNV", 1, &E_GL_NV_conservative_raster_underestimation);
+        SpecialQualifier("gl_FragFullyCoveredNV", EvqVaryingIn, EbvFragFullyCoveredNV, symbolTable);
```
johnkslang: I'm worried this should not be done.

From the comment above `SpecialQualifier()`:

```
// Only use this for built-in variables that have a special qualifier in TStorageQualifier.
// New built-in variables should use a generic (textually declarable) qualifier in
// TStoraregQualifier and only call BuiltInVariable().
```

Since the text form is able to get the qualifier right, that's the preferred way.  Does it not work to textually decare it? Put the text in the right conditional places earlier in the file, with similar declarations:

```
in bool gl_FragFullyCoveredNV;
```

E.g., if you look for `gl_FrontFacing` in this function:

```
// ... Most built-ins variables can be added as simple text strings... 
void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvVersion)
```

Or, is this context dependent, or otherwise more conditional than just the version/vulkan/GL/stage allowed for expressing it in textually in that function?

johnkslang: Wait, you did have a declaration, sorry I missed that.

I think I understand what happened now. E.g., gl_FrontFacing was an original specially-qualified (EvqFace) variable (the qualifier could not be textually declared) and perhaps following its example lead to the current PR. But, that old design did not scale: we don't do special qualifiers anymore for the plethora of new inputs/outputs, using instead the TBuiltInVariable to carry the extra meaning, and just `in` for the qualifier.

I'll assume this is correct, and fix the PR for you base on it.

johnkslang:
I changed glslang infrastructure to allow a built-in input to be declared as a `bool`, so that the modern approach of avoiding special qualifiers can be followed, and without needing to call `SpecialQualifier`.

The alternative used in the original submission worked, and was okay too, I just think this is slightly better.

So, I took the liberty of tweaking the PR in this way.  If it is okay, and passes tests, there is no request to modify anything.

pdaniell-nv:
Looks good to me. Thanks for making the change.

PR #1186, opened 2017-12-14
HLSL: Implement support for RT/Viewport output from VS

zeux:
In DX10/DX11 you can only output RT/Viewport indices from GS; however,
DX11.4/DX12 add support for outputting these from VS as well.

This is supported by Vulkan if the relevant extension is available, and
by MSL and by MSL (which you can cross-compile to via SPIRV-Cross).


zeux:
Not sure what the build failure is about, seems like possibly an unrelated/spurious issue. Would it be possible to restart the build?

PR #1190, opened 2017-12-15
HLSL: Fix possibly incorrect type conversion in Store2-3-4

jfmarquis:
Set type to r-value resulting from indexing vector, to prevent
float->uint conversion when source is already uint. Resulting
OpConvertFToU would otherwise fail validation because source is
already uint.

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -3302,7 +3302,13 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
                 const TOperator idxOp = (offsetIdx->getQualifier().storage == EvqConst) ? EOpIndexDirect
                                                                                         : EOpIndexIndirect;
 
-                vec = intermediate.growAggregate(vec, intermediate.addIndex(idxOp, argArray, offsetIdx, loc));
+                TIntermTyped* indexVal = intermediate.addIndex(idxOp, argArray, offsetIdx, loc);
+
+                TType derefType(argArray->getType(), 0);
+                derefType.getQualifier().makeTemporary();
+                indexVal->setType(derefType);
+
+                vec = intermediate.growAggregate(vec, indexVal);
             }
 
             vec->setType(TType(argArray->getBasicType(), EvqTemporary, size));
@@ -3366,8 +3372,14 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
                 const TType derefType(argArray->getType(), 0);
                 lValue->setType(derefType);
 
-                TIntermTyped* rValue = (size == 1) ? argValue :
-                    intermediate.addIndex(EOpIndexDirect, argValue, idxConst, loc);
+                TIntermTyped* rValue;
+                if (size == 1) {
+                    rValue = argValue;
+                } else {
+                    rValue = intermediate.addIndex(EOpIndexDirect, argValue, idxConst, loc);
+                    const TType indexType(argValue->getType(), 0);
+                    rValue->setType(indexType);
+                }
                     
                 TIntermTyped* assign = intermediate.addAssign(EOpAssign, lValue, rValue, loc); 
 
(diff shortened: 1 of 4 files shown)


jfmarquis:
First PR to glslang. I could not find a contributing guide, but I did run and update the tests.

johnkslang:
Have you read through all of the project readme?

It looks like not all the tests were updated for the new results.

Also, look around for local context on how curly braces, etc., are placed.

johnkslang:
@LoopDawg any input on this?  Thanks.

LoopDawg:
The change looks correct to me.  While looking it over, I noticed there's a similar issue in the LoadN methods, so maybe that fix can get rolled into this PR too?  It should look something like so, around 3305:

                TIntermTyped* indexVal = intermediate.addIndex(idxOp, argArray, offsetIdx, loc);

                const TType derefType(argArray->getType(), 0);
                indexVal->setType(derefType);

                vec = intermediate.growAggregate(vec, indexVal);

It's possible the load issue is commonly masked by a subsequent vector constructor, but really it should set the type as above.

@johnkslang: would you object to a separate PR that adds a new method which encapsulates the index creation and type setting, and replacement of a bunch of these constructs with that?  I could volunteer to do that.  IMHO, right now it's too easy to accidentally forget to set the return type.  I don't know why any index op wouldn't want to, but if one did for some reason, it could still fall back to the existing method.

johnkslang:
> would you object to a separate PR that adds a new method which encapsulates the index creation and type setting

Sounds good, especially if collects all the places that need it.  Thanks!

jfmarquis:
@LoopDawg I am doing the change you suggested for LoadN. It does remove some odd-looking uint->float->uint constructs from the tests output.

About the inquiry to @johnkslang, shouldn't it just be the addIndex method that sets the type? I really don't know the details in here or the concepts behind such changes to a function. I guess you want to put the least amount of things in TIntermediate.

LoopDawg:
> I am doing the change you suggested for LoadN. It does remove some odd-looking uint->float->uint constructs from the tests output

Cool - sounds right.

About addIndex: after your PR goes in, I was going to make a pass through and add such an encapsulation, and change current call sites to use that.  Since your PR will fix the AST and SPIR-V, at that point the encapsulated form shouldn't cause any more test diffs.  (Unless further defects of this class are lurking).

hlsl/hlslParseHelper.cpp
```diff
@@ -3302,7 +3302,12 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
                 const TOperator idxOp = (offsetIdx->getQualifier().storage == EvqConst) ? EOpIndexDirect
                                                                                         : EOpIndexIndirect;
 
-                vec = intermediate.growAggregate(vec, intermediate.addIndex(idxOp, argArray, offsetIdx, loc));
+                TIntermTyped* indexVal = intermediate.addIndex(idxOp, argArray, offsetIdx, loc);
+
+                const TType derefType(argArray->getType(), 0);
+                indexVal->setType(derefType);
```
johnkslang: Should probably be followed up with a `getQualifier().makeTemporary()`?

hlsl/hlslParseHelper.cpp
```diff
@@ -3366,8 +3371,15 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
                 const TType derefType(argArray->getType(), 0);
                 lValue->setType(derefType);
 
-                TIntermTyped* rValue = (size == 1) ? argValue :
-                    intermediate.addIndex(EOpIndexDirect, argValue, idxConst, loc);
+                TIntermTyped* rValue;
+                if (size == 1) {
+                    rValue = argValue;
+                }
```
johnkslang: Make it `} else`, thanks.

johnkslang:
I gave specific comments, but on a general note: Other than looking pretty good, the main issue is whether intermediate results still have the original base object's I/O qualification, or instead are qualified as temporary objects. The answer might not be clear, but I'd like to error on the side of matching existing behavior, or similar behavior that is not broken.

So, open to any input on that, thanks.

LoopDawg:
Agreed with the makeTemporary addition.  Other sites are doing that too.

PR #1195, opened 2017-12-20
compile error fix for g++ 5.4.0

LoopDawg:
Fix compilation error under g++ 5.4.0.  Possible g++ defect, or possible difference in behavior.  Either way, this compiles now under that version.


LoopDawg:
For some reason the new form failed the multithreading test on the second travis test.  I have no idea why.  Will repush with another experiment.

PR #1198, opened 2017-12-21
HLSL Legalization: Add scalar replacement

greg-lunarg:
This allow for propagation through structs with dynamically indexed
arrays. This supports recent removal of non-io struct flattening.


johnkslang:
I also want to remove the y-flip test from the affected tests, will look at doing that independently.

greg-lunarg:
Add -Od to the glslangValidator calls in runtests.

Issue #1187, opened 2017-12-14
Preprocessor: Per-token extension queries dramatically slow down preprocessor


zeux:
Context: I'm trying to switch to glslang preprocessor from mojoshader preprocessor that we use in our build pipeline; performance is important since full preprocessed source is the compilation cache key, so we preprocess all our shaders to check if they need rebuilding. On a specific shader pack, mojoshader takes ~90ms including various overheads, glslang preprocessor takes ~280ms.

I found that a very significant portion of this time is spent in `TPpContext::tStringInput::scan` (and possibly other callsites) calling `extensionTurnedOn` (which does `std:;string` construction and `std::map` lookup).

Replacing `extensionTurnedOn` body with `return false` drops preprocessing time from 280ms to 160ms which makes me think that all supported extensions should be cached in primitive variables somewhere.

johnkslang:
Good find, thanks. I guess this has been slowly creeping in over time, as more keywords became conditional. Agreed with caching the common queries here.

jfmarquis:
This started being a bottleneck to compile our 400 000+ shaders, so I did something as simple as possible before actually finding out there was an issue on the subject first.
All I did was change the map to an unordered_map and make it contain const char* instead of string, with corresponding custom hash/equal functions.

jfmarquis:
If std::string_view was available it would look much better but wouldn't allow me to compare pointers, considering most (if not all) the elements of the map and the things we compare with come from constants in Versions.h. Comparing pointers gave the biggest gain after moving to unordered_map.

jfmarquis:
See [here](https://github.com/jfmarquis/glslang/tree/getExtensionBehavior-perf) if its worth pulling.

johnkslang:
I did something similar, and got a good speed up. But, I went all the way, fixing the code to use the `const *char const` from the header. Unfortunately, in debug mode on Windows, sometimes this constant changed value, so that experiment failed.

However, the real problem is how often the lookup is called. The actual culprit is this the following, which does 3 lookups *per token*!  This should instead be rare:

```
int TPpContext::tStringInput::scan(TPpToken* ppToken)
{
...
    bool enableInt64 = pp->parseContext.version >= 450 && pp->parseContext.extensionTurnedOn(E_GL_ARB_gpu_shader_int64);
#ifdef AMD_EXTENSIONS
[...]

PR #1209, opened 2018-01-05
GLSL Scan performance: cache allowed literal types based on extension.

johnkslang:
This addresses #1187.

On some whole-stack tests, and whole run tests, I see 10% improvements.  Scanning itself must be much better.


johnkslang:
BTW, I think this PR is a bit klunky, just valuable for performance comparison.

As I think about it, no matter what, it seems worth not having 3 extension checks per token, and will hence find a slicker way of flagging what literals can be accepted, with out all this mechanism.

The question is, once this is removed (and the current PR is a good enough basis for that), is there any point in doing anything else?

PR #1213, opened 2018-01-08
HLSL: Adding the option to append semantic names to input variable names

Cry-Mory:
Adds --append-semantic-names option to the standalone validator which will append '___' and semantic name to all vertex shader input variables' name. Renaming is applied to all entry function input arguments whether they are instances of a basic type or a struct type. When the input parameter is an instance of a struct type, the struct type member names are changed accordingly to reflect semantic type in members' definition and in function bodies.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -271,6 +271,14 @@ void Error(const char* message)
 }
 
 //
+// Give a warning but program continues its execution
+//
+void Warning(const char* message)
+{
+    fprintf(stdout, "%s: Warning %s \n", ExecutableName, message);
+}
+
+//
 // Process an optional binding base of one the forms:
 //   --argname [stage] base            // base for stage (if given) or all stages (if not)
 //   --argname [stage] [base set]...   // set/base pairs: set the base for given binding set.
@@ -668,6 +676,13 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
     if ((Options & EOptionFlattenUniformArrays) != 0 &&
         (Options & EOptionReadHlsl) == 0)
         Error("uniform array flattening only valid when compiling HLSL source.");
+
+    // Show a warning if --append-semantic-names is not used for HLSL vertex shader translation
+    if ((Options & EOptionAppendSemanticToVarName) && (!(Options & EOptionReadHlsl) || strcmp(shaderStageName, "vert") != 0))
+    {
+        Warning("--append-semantic-names arguments can only be used for translating HLSL vertex shaders. The option is disabled automatically.");
+        Options &= ~EOptionAppendSemanticToVarName;
+    }
 }
 
 //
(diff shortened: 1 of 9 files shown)


chaoticbob:
Hi Morteza,

There's currently a work in progress proposal to add a decorator to SPIR-V that enables a string decoration to node for HLSL semantics specifically. Apologies that it's been slow coming. The decoration approach makes the semantic concretely apart of the SPIR-V and it won't be subject to possible removal during optimization. Assuming the proposal passes, the string decoration should be supported by both glslang and DXC.

Cry-Mory:
I agree the decoration approach is a clean solution to get the semantics when parsing SPIR-V shaders.

I am implementing support for glslang and DXC at the same time. DXC is appending the semantics to the variable names and that is how I can get the semantic when reading the SPIR-V code. Equivalently I wanted to add it to glslang that is why I proposed this pull request.

hlsl/hlslParseHelper.h
```diff
@@ -483,6 +494,12 @@ class HlslParseContext : public TParseContextBase {
     };
 
     TMap<int, tShadowTextureSymbols*> textureShadowVariant;
+
+    // Maps the old variable name to the new variable name (semantic appended one) when --append-semantic-name option is used.
+    std::map<TString, TString> semanticAppendedVarMap;
+
+    // Maps the base name and its field name to the new field name (semantic appended one) when --append-semantic-name option is used.
+    std::map<std::pair<TString, TString>, TString> semanticAppendedFieldMap;
```
johnkslang: This seems overall like a lot of mechanism. Maybe that's okay, but if something simpler works, it might be worthwhile, like, for example, just append during SPIR-V output? (I don't know all the goals, so I can't claim it solves the problem.) Is it just for the debug names in the SPIR-V output?

StandAlone/StandAlone.cpp
```diff
@@ -665,6 +668,13 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
     if ((Options & EOptionFlattenUniformArrays) != 0 &&
         (Options & EOptionReadHlsl) == 0)
         Error("uniform array flattening only valid when compiling HLSL source.");
+
+    // Show a warning if --append-semantic-names is not used for HLSL vertex shader translation
+    if ((Options & EOptionAppendSemanticToVarName) && (!(Options & EOptionReadHlsl) || strcmp(shaderStageName, "vert") != 0))
+    {
+        fprintf(stdout, "WARNING: --append-semantic-names arguments can only be used for translating HLSL vertex shaders. The option is disabled automatically. \n");
```
johnkslang: Why not other stages?

StandAlone/StandAlone.cpp
```diff
@@ -1367,6 +1380,8 @@ void usage()
            "                                       initialized with the shader binary code.\n"
            "  --vn <name>                          synonym for --variable-name <name>\n"
            "  --invert-y | --iy                    invert position.Y output in vertex shader\n"
+           "  --append-semantic-names              appends semantic name to HLSL vertex shader input variable names\n"
```
johnkslang: the --invert-y line, I believe basically uses up 80 characters. Nice to keep all the lines fitting in 80.

johnkslang:
Testing is missing.  Simplest is to add a test file, and then test it by adding a block in Test/runtests, which is the easiest way for novel new command-line options.  After executing runtests, execute bump, and then add the new file result from baseResults along with the new file.

PR #1224, opened 2018-01-16
Fix unused function warning.

fjhenigman:
TGlslangToSpvTraverser::getExtBuiltins is only used when AMD_EXTENSIONS
is defined, so only define it in that case to avoid an unused function
warning.


fjhenigman:
I think the build check timed out.  Is there a way to run it again?

johnkslang:
It looks good to me, I'll merge it.

PR #1227, opened 2018-01-25
Update spirv-tools known-good

greg-lunarg:
This adds dead insert elimination. It also adds redundancy elimination to -Os


dneto0:
The uses of constexpr are being removed from SPIRV-Tools.

dneto0:
The  uses of constexpr have been removed from SPIRV-Tools:  https://github.com/KhronosGroup/SPIRV-Tools/pull/1254 
I expect to make a new Glslang PR to update SPIRV-Tools known-good.

PR #1234, opened 2018-01-31
Update SPIRV-Tools known-good

dneto0:
Update SPIRV-Tools.  Relevant functional changes:
 - Optimizer enhancements:
   - ADCE now removes OpSwitch
   - Block merging occurs in more cases
 - Optimizer fixes:
   - Constant propagation (CCP): support matrix constants
   - KhronosGroup/SPIRV-Tools#1199: Optimizer: Fix CCP: don't propagate spec constants.
   - KhronosGroup/SPIRV-Tools#1203: Optimizer: Fix common uniform elim bug introduced by refactoring.
   - KhronosGroup/SPIRV-Tools#1210: Optimizer: Aggressive dead code elimination: Fix 'break' identification.
   - KhronosGroup/SPIRV-Tools#1212: Optimizer: Aggressive dead code elimination: Was skipping too many instructions.
   - KhronosGroup/SPIRV-Tools#1214: Optimizer: Aggressive dead code elimination: Fix infinite loop.
   - KhronosGroup/SPIRV-Tools#1228: Optimizer: Fix CCP: Handling of varying Phi nodes; was resulting in infinite loop.
   - KhronosGroup/SPIRV-Tools#1245: Optimizer: Dead branch elimination: Avoid a null pointer dereference.
   - KhronosGroup/SPIRV-Tools#1250: Optimizer: Dead branch elimination: Avoid spuriously reporting a change.

Update SPIRV-Headers, with "unified1" directory.

Updated one Glslang legalization test base result due to better block merging.

### DIFF
--- a/Test/baseLegalResults/hlsl.partialFlattenLocal.vert.out
+++ b/Test/baseLegalResults/hlsl.partialFlattenLocal.vert.out
@@ -52,11 +52,9 @@ WARNING: AST will form illegal SPIR-V; need to transform to legalize
                               Store 137 38
                               Branch 100
              100:             Label
-             164:     21(int) Phi 25 5 119 102
-                              LoopMerge 101 102 None
-                              Branch 103
-             103:             Label
+             164:     21(int) Phi 25 5 119 106
              105:    54(bool) SLessThan 164 31
+                              LoopMerge 101 106 None
                               BranchConditional 105 106 101
              106:               Label
              138:     39(ptr)   AccessChain 133 164
@@ -69,8 +67,6 @@ WARNING: AST will form illegal SPIR-V; need to transform to legalize
              116:   14(fvec3)   Load 140
              117:   14(fvec3)   VectorShuffle 116 114 3 4 2
                                 Store 140 117
-                                Branch 102
-             102:               Label
              119:     21(int)   IAdd 164 31
                                 Branch 100
              101:             Label
--- a/known_good.json
+++ b/known_good.json
@@ -5,14 +5,14 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit" : "eb0c73dad6102fc0d4f03c62fe910348bae43a11"
+      "commit" : "9e19fc0f31ceaf1f6bc907dbf17dcfded85f2ce8"
     },
     {
       "name" : "spirv-tools/external/spirv-headers",
       "site" : "github",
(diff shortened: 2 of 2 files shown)


dneto0:
Hey @greg-lunarg : This includes the SPIRV-Tools functionality changes from your #1227 but does not update the optimization recipe in Glslang.

This also includes the removal of constexpr that was causing bot failures in your #1227.  I hope this passes VS 2013 builds so that you can then update the recipe.

dneto0:
The build failure on VS 2013 is:

C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xmemory0(593): error C2280: 'std::unique_ptr<spvtools::ir::Loop,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)' : attempting to reference a deleted function [C:\projects\glslang\build\External\spirv-tools\source\opt\SPIRV-Tools-opt.vcxproj]
          with
          [
              _Ty=spvtools::ir::Loop
          ]
          C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\memory(1486) : see declaration of 'std::unique_ptr<spvtools::ir::Loop,std::default_delete<_Ty>>::unique_ptr'

As I understand it, the problem is:

* spvtools::ir::Loop has a data member LoopContainerType loops_
* That's a std::vector<std::unique_ptr<Loop>>  , with LoopContainerType declared at
 https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/loop_descriptor.h#L235
* The trouble is that std::unique_ptr is not copy-constructible, so its copy constructor is 'deleted'.
* But the VS2013 implementation of std::vector doesn't have a specialization for std::unique_vector members, and so fails compilation. 
* I would guess later compilers specialize vectors over unique_ptrs by using move construction instead of copy construction.

Using unique_ptr affords us clarity, simplicity, and safety.  I don't want to change the SPIRV-Tools code to accommodate VS 2013.

johnkslang:
It may be okay to raise the requirement for legalization. Need some time for everyone to update their tools.

Historically, the only features I saw regularly attempted beyond 2010 were range-based  for loops and the "{...}" in place construction. So, it had been a relatively small cost to stay at 2010 level.

Other input on minimum versions?

See #42.

dneto0:
SPIRV-Tools now builds with VisualStudio 2013. 
I've updated this PR with new commit hashes for SPIRV-Tools and SPIRV-Headers

PR #1237, opened 2018-02-01
WIP: AppVeyor: Run ENABLE_OPT=off build with VS 2010, optimizing build with VS 2015

dneto0:
Also deploy only from VS 2015 build.


dneto0:
This is WIP.  I have not tested it with AppVeyor yet.

PR #1244, opened 2018-02-09
Update SPIRV-Tools known-good

dneto0:
One legalization case updated for a renumbered ID


dneto0:
cc: @greg-lunarg 

The change is due to one of these optimizer changes:

ca4457b4 SROA: Do replacement on structs with no partial references.
06cdb969 Make use of the instruction folder.
87102277 Registering a type now rebuilds it out of memory owned by the manager.
860b2ee5 ADCE: Fix combinator initialization

Issue #1162, opened 2017-11-22
GLSL: interpolateAtVertexAMD accepts temporary without generating proper error


Themaister:
```
    vec4 pos =  interpolateAtVertexAMD(vec4(0.0), 0);
```
Fails in an assertion rather than an error.

johnkslang:
@amdrexu any input on this?

johnkslang:
Maybe something is needed in `builtInOpCheck()`?

amdrexu:
Sorry, I miss this issue. I will reproduce it and give you feedback in these two days.

amdrexu:
@Themaister Thank you for catching this issue. We make a fix here: https://github.com/KhronosGroup/glslang/pull/1246.

Please note the first argument of interpolateAtVertexAMD is interpolant. It should be an input with "in" qualified. In your case, it is a vec4 constant. So the shader is invalid. This rule is similar to other interpolation functions like interpolateAtXXX() in GLSL.

amdrexu:
The issue is fixed. Thank you.

PR #1246, opened 2018-02-14
Fix an issue of interpolateAtVertexAMD().

amdrexu:
An error message should be reported if the interpolant is not an input.


amdrexu:
This is to fix the issue: https://github.com/KhronosGroup/glslang/issues/1162

PR #1265, opened 2018-02-28
EShLangFragment falls through to EShLangCompute, only #ifndef AMD_EXTENSIONS.

davidhubbard:
EShLangFragment falls through to EShLangCompute, only #ifndef AMD_EXTENSIONS.
This is probably a bug.


davidhubbard:
#ifndef AMD_EXTENSIONS is poor word choice. (It's too easy to think #ifdef instead.)

I meant to say, the "break" is inside the #ifdef AMD_EXTENSIONS. I'm almost 100% sure that's not what was meant.

I did go find where the break line and the #ifdef lines were added using git blame, and the break was there first, then the #ifdef was added in the wrong place.

Hope this clarifies things a bit.

PR #1266, opened 2018-03-02
Implement the extension GL_AMD_gpu_shader_half_float_fetch

amdrexu:
- Support new opaque types: f16sampler*, f16image*, f16subpassInput*.
- Add new built-in GLSL texture/image functions.


amdrexu:
The GLSL spec is here: https://github.com/KhronosGroup/OpenGL-Registry/blob/master/extensions/AMD/AMD_gpu_shader_half_float_fetch.txt

The SPIR-V spec is here: http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/AMD/SPV_AMD_gpu_shader_half_float_fetch.html

Please reference them.

oscarbg:
Hi @amdrexu,
was going to open an issue asking for existing min16float HLSL support mapping to Float16 SPIR-V capability.. right now it maps to float type with no option to override it.. 
anyway after seeing your new "half float" related commit and thinking, that really what I'm asking for, is support for SPV_AMD_gpu_shader_half_float.. as Float16 SPIR-V Vulkan (outside OpenCL) seems offered right now (i.e. requires) SPV_AMD_gpu_shader_half_float..
as you implemented support for that for GLSL I would ask if you are insterested on implementing for HLSL too..
note also that right now DirectXShaderCompiler provides this support in his SPIR-V support
(i.e. dxc -spirv) by enabling 16 bit types (-enable-16bit-types) it generates SPIR-V using SPV_AMD_gpu_shader_half_float..
sadly right now I'm hitting an issue but will be solved any time right now as developers are very supportive..
https://github.com/Microsoft/DirectXShaderCompiler/issues/1095
anyway glslang support will be nice too..

amdrexu:
@oscarbg Hi, I take a look at what glslang has implemented for min16float. This basic type is mapped to float. I think we could make a change and map this type to float16. But we have a problem. General use of float16 requires the extension SPV_AMD_gpu_shader_half_float. When glslang does translation for HLSL, it is inappropriate to specify the use of SPV_AMD_gpu_shader_half_float. This is because the HLSL codes are vendor-dependent while the generated SPIR-V codes are vendor-dependent. Thus, the HLSL shader could not run on each vendor's hardware because corresponding SPIR-V codes introduce vendor-dependent SPIR-V extension (the portability is broken). I think there must be a KHR/EXT Vulkan extension fully supporting float16 operations. After that, we could do a decent translation.

oscarbg:
Hi @amdrexu,
thanks for your response and evaluation..
I understand the portability problem but then IMHO by the same reasoning, AMD should not provide patches for glsang GLSL support for mapping GL_AMD_gpu_shader_half_float to SPIR-V using SPV_AMD_gpu_shader_half_float as this SPIR-V code is not vendor independent also..
Anyway I think that glslang users are not blind users, are developers that perhaps want to have HLSL as input to generate SPIR-V code for his Vulkan needs and as I said in original request this should be exposed via optional argument to glslang (-enable-16bit-types?) and leave the default usage to generate floats..
even with  optional argument (say -enable-16bit-types) perhaps a warning should generated ("Warning generated Spirv requires SPV_AMD_gpu_shader_half_float as you request SPIR-V float16 cap.")
IMHO I think if AMD is supporting generating SPIR-V codes with Float16 from GLSL (that is for me talking clearly:  promoting taking advantage of his Rapid Packed Math in Vega GPUs) it could do the same for HLSL..
In fact is surprising to me that for having "rapid packed math" SPIR-V from HLSL I have to use DirectXShaderCompiler SPIR-V backend with  SPV_AMD_gpu_shader_half_float support which Google devs are exposing..
also perhaps would be better to open a separate issue for this if you want, and if I'm asking for this directly to you is because you seem to be commiting a lot of this Rapid Packed Math features in glslang in behalf of AMD so I think you are the right person to ask..

antiagainst:
Apologies first before chiming in to muddle the water. ;-P

I'd like to provide the reasoning behind our decisions to emit `Float16` support using `SPV_AMD_gpu_shader_half_float` in DXC SPIR-V CodeGen:
* [Starting from SM6.2, DXC introduces 16bit types and `-enable-16bit-types` to turn them on](https://github.com/Microsoft/DirectXShaderCompiler/wiki/16-Bit-Scalar-Types). The SPIR-V CodeGen in it will need to support that, eventually.
* [The only extension we've spotted ATM that we can leverage to do the CodeGen is `SPV_AMD_gpu_shader_half_float`](https://github.com/KhronosGroup/SPIRV-Registry). (Please definitely correct me here if I'm wrong.)
* Based on the above, if a developer truly wants 16bit floats, he/she can only get `SPV_AMD_gpu_shader_half_float` ATM. It's vendor-specific, but I'd think it's a step forward than not CodeGen'ing, where the developer cannot use it at all.

Note that we haven't finalized our extension stories in DXC; how to properly control SPIR-V extensions  emitted by DXC for HLSL is still under consideration. The above is experimental and subject to change.

With that said, it seems a KHR extension would be much more preferable.

But I'd suggest to open new issues accordingly on the proper projects to track the request. This pull request has fulfilled its purpose and we should avoid putting more discussions here. (Sorry I actually did that.)

amdrexu:
Thanks for your feedback. I open an issue ticket here: https://github.com/KhronosGroup/glslang/issues/1271. We can track this requirement and maybe continue discussion in this ticket. From internal evaluation, AMD believes it is OK to do such translation and expose the float16 capability conditionally when converting such shaders from HLSL to SPIR-V binary. We will do further scoping.

oscarbg:
thanks for opening issue and sharing AMD thoughts..

PR #1267, opened 2018-03-02
SPV: The result type of OpArrayLength should be UINT

amdrexu:


amdrexu:
In SPIR-V, it says: OpArrayLength, Result Type must be an OpTypeInt with 32-bit Width and 0 Signedness.

johnkslang:
Since this changes the way existing code in generated, we need to bump up the generator number.

Add a line to:

```
{
    // return 1; // start
       return 4; // some deeper access chains: for dynamic vector component, and local Boolean component
}
```

Something like


```
return 5; // make OpArrayLength result type be an int with signedness of 0
```

amdrexu:
Thank you. I make a update. The test result files are changed as well and the amount seems large :)

johnkslang:
Yes, now I think I should do this as an independent commit.

I think you forgot to update a few tests, but I'll try doing all this locally.

amdrexu:
Thanks for your kind help.

johnkslang:
This was reverted by ac3707921ed313eaba1c206e8113b85fef054c34.

Sorry about this, but the fix seems incomplete, as it can move the type issue to the subsequent instruction. It may need to be combined with an OpBitCast.

PR #1269, opened 2018-03-03
Fixed GCC+Clang errors and warnings on Linux and OSX.

bkaradzic:


glslang/Include/Common.h
```diff
@@ -37,7 +37,7 @@
 #ifndef _COMMON_INCLUDED_
 #define _COMMON_INCLUDED_
 
-#if (defined(_MSC_VER) && _MSC_VER < 1900 /*vs2015*/) || defined MINGW_HAS_SECURE_API
+#if (defined(_MSC_VER) && _MSC_VER < 1900 /*vs2015*/) // || defined MINGW_HAS_SECURE_API
```
johnkslang: Why does this remove the MINGW path?

bkaradzic: No clue, I'll fix this.

Issue #1271, opened 2018-03-06
HLSL: Map min16float/min16int/min16uint to float16_t/int16_t/uint16_t


amdrexu:
Maybe add an option to conditionally map those types in HLSL to AMD float16_t/int16_t/uint16_t.  Reference the requirement https://github.com/KhronosGroup/glslang/pull/1266.

johnkslang:
Perhaps options like these should be part of the --target-env options.

amdrexu:
I make a pull request to address this requirement. I add an option "--hlsl-enable-16bit-types". I think it is inappropriate to add this to --target-env. The option is better to be an independent one. I want to enlarge the enum TOptions but it is full. Therefore, I have to add a boolean variable to represent this option.

amdrexu:
See the pull request: https://github.com/KhronosGroup/glslang/pull/1275

oscarbg:
many thanks for your (fast) support!
is really appreciated..

oscarbg:
Hi @amdrexu ,
testing it still seems has some issues :
I'm attaching test.hlsl and all generated outputs here:
DOWNLOAD->
[tests.zip](https://github.com/KhronosGroup/glslang/files/1800468/tests.zip)

1)it works correctly on DXC  -enable-16bit-types 

> dxc.exe -E CSMandelJulia_scalarFloat test.hlsl -enable-16bit-types -T cs_6_2 -spirv -O0 -Vd >test.txt

2) of course works with glslangValidator default option of translating to float

> glslangValidator -D -H -e CSMandelJulia_scalarFloat -S comp test.hlsl >testglslc.txt

3)with -enable-16bit-types fails:

> glslangValidator -D -H -e CSMandelJulia_scalarFloat -S comp --hlsl-enable-16bit-types test.hlsl 

>testglslcfp16.txt
returns:
```
test.hlsl
ERROR: test.hlsl:32: '' : unsupported construction 
ERROR: test.hlsl:32: ' temp mediump float16_t' : cannot construct with these arguments 
ERROR: test.hlsl:32: 'assignment expression' : Expected 
test.hlsl(32): error at column 34, HLSL parsing failed.
ERROR: 4 compilation errors.  No code generated.
SPIR-V is not generated for failed compile or link
```

amdrexu:
Reproduced. I will have a look and give you update.

oscarbg:
Ok thanks..
tested already int16 shader:
[...]

PR #1275, opened 2018-03-07
HLSL: Map min types to GLSL 16-bit types

amdrexu:


amdrexu:
This is to address the requirement: https://github.com/KhronosGroup/glslang/issues/1271

PR #1277, opened 2018-03-07
SPV: Implement Vulkan 1.1 features and extensions.

johnkslang:


johnkslang:
Merging this. The base version is tagged as 5.1, in case there are any troubles with this.

antiagainst:
I haven't checked all intrinsic functions. But it seems Microsoft intend to use `WaveReadLaneFirst` instead of `WaveReadFirstLane`:

https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics#type-wavereadlanefirst-type-expr-

sheredom:
@antiagainst that was a typo on my part, I'll get it fixed thanks!

PR #1282, opened 2018-03-08
WIP: Implement SPV_GOOGLE_hlsl_functionality1.

johnkslang:
Implement subset of SPV_GOOGLE_hlsl_functionality1: semantics on (non-member) variables.

Intention is to get early feedback on the following, and fully implement before removing WIP:

- command-line enable is `--target-env hlsl_functionality1`
- API enable is `shader->setEnvTargetHlslFunctionality1()`

So far, it is quite trivial to do this (e.g., compared to #1213).


dneto0:
I'd follow GCC's lead here on option naming:   Machine-independent options that affect code generation normally start with -f<blah>   
So I suggest -fhlsl_functionality1      (or as @antiagainst suggests an umbrella option that includes future HLSL special functionality upgrades: -fhlsl_functionality)

johnkslang:
Agreed, if it is machine independent. I was worried about drivers rejecting instructions they don't support. Is the model here that an offline tool always transforms this away before a driver sees it?

Cry-Mory:
@johnkslang I tried this branch and replaced "vulkan1.0" target environment with "hls_functionality1" in the compilation command, I am getting the following error while the spir-v file is being generated:

error: input:0:0:605: Invalid opcode: 5632

Then, I used spirv-cross to parse the generated binary SPIR-V but that ends up with the following exception in spirv-cross:

_"Currently no block to insert opcode."_ 

For clarification, this error does not appear if I use "vulkan1.0" target environment. 

In case you want to try the shader, you can find it in the attachment to this comment. I compile it with the following command:

**.\glslangValidator.exe -D .\testShader.txt -o testShader.spv --target-env hlsl_funct
ionality1 -S vert -e AuxGeomTextureVS -V100 --invert-y**

johnkslang:
From off-line feedback, and going with `-fhlsl_functionality1`, following gcc use of `-f`.

@Cry-Mory the following works for me:

`./glslangValidator -D testShader.txt -o testShader.spv -fhlsl_functionality1 -S vert -e AuxGeomTextureVS -V100 --invert-y -H`

That works both with/without `--target-env vulkan1.0` and `--target-env vulkan1.1`.

Make sure you have the latest version of SPIRV-Tools (see `glslang/../update_glslang_sources.py`).

PR #1293, opened 2018-03-13
Replace en-dash with simple ASCII hyphen

baldurk:
Currently there's a few en-dashes in a comment in `iomapper.cpp`, I think because of copy-paste from [this MSDN page](https://msdn.microsoft.com/en-us/library/windows/desktop/dn899207%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396). Without a UTF-8 BOM if the system codepage doesn't include the character this can lead to compile warnings - see baldurk/renderdoc#915. I think this change is preferable to adding a BOM and potential headaches from that.

If this change isn't desired then feel free to reject the PR, but I think it's a good idea.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -598,7 +598,7 @@ protected:
 /********************************************************************************
 The following IO resolver maps types in HLSL register space, as follows:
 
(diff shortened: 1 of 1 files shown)


johnkslang:
Yes, it's desired. I've done a few removal passes of similar issues in the past.  Thanks.

Issue #1296, opened 2018-03-14
Shifts should be allowed for base/shift operands of different bit-width and sign


infapi00:
Since commit 66011cb2c22a19c2fc18885d6a9f2a8fee191ce8 the following shader would fail to be converted to SPIR-V under OpenGL semantics (and I assume under Vulkan too):

> #extension GL_ARB_gpu_shader_int64 : require
> writeonly uniform image2D tex;
> layout(local_size_x = 54) in;
> uniform int64_t arg0;
> uniform uint arg1;
> uniform int64_t expected;
> 
> void main()
> {
>   vec4 tmp_color;
>   int64_t result = (arg0 << arg1);
>   tmp_color = result == expected ? vec4(0.0, 1.0, 0.0, 1.0) : vec4(1.0, 0.0, 0.0, 1.0);
>   ivec2 coord = ivec2(gl_GlobalInvocationID.xy);
>   imageStore(tex, coord, tmp_color);
> }

With the following error:

> '<<' :  wrong operand types: no operation '<<' exists that takes a left-hand operand of type 'layout( location=1) uniform int64_t' and a right operand of type 'layout( location=2) uniform uint' (or there is no acceptable conversion)

The mentioned commit changed the checks at TIntermediate::addConversion, forcing base/shift to be of the same sign and same rank. 

But from GLSL spec:

> The shift operators (<<) and (>>). For both operators, the operands must be signed or unsigned
> integers or integer vectors. One operand can be signed while the other is unsigned. In all cases, the
> resulting type will be the same type as the left operand. If the first operand is a scalar, the second
> operand has to be a scalar as well. If the first operand is a vector, the second operand must be a scalar
> or a vector, and the result is computed component-wise. The result is undefined if the right operand is
> negative, or greater than or equal to the number of bits in the left expression's base type.

so base/shift can be of different sign

And from SPIR-V spec:

> The type of each Base and Shift must be a scalar or vector of integer type. Base and Shift must have the same number
> of components. The number of components and bit width of the type of Base must be the same as in Result Type.
> Shift is treated as unsigned. The result is undefined if Shift is greater than the bit width of the components of Base.

it asks Base and Result to be the same bit-width, not Base and Shift operands.

This was found while working on improving ARB_gl_spirv support for the mesa driver. We use glslang to convert GLSL shaders to SPIR-V. After the mentioned commit, a test that was working properly using the generated SPIR-V started to fail as glslang rejected to generate such SPIR-V binary.

[...]

PR #1297, opened 2018-03-14
SPV/GL: allow shift operation for int<<uint of different rank

infapi00:
From GLSL spec:
"The shift operators (<<) and (>>). For both operators, the operands
must be signed or unsigned integers or integer vectors. One operand
can be signed while the other is unsigned. In all cases, the resulting
type will be the same type as the left operand. If the first operand
is a scalar, the second operand has to be a scalar as well. If the
first operand is a vector, the second operand must be a scalar or a
vector, and the result is computed component-wise. The result is
undefined if the right operand is negative, or greater than or equal
to the number of bits in the left expression's base type"

It clearly points that int/uint combinations are allowed.

From SPIR-V spec:
"The type of each Base and Shift must be a scalar or vector of integer
type. Base and Shift must have the same number of components. The
number of components and bit width of the type of Base must be the
same as in Result Type.  Shift is treated as unsigned. The result is
undefined if Shift is greater than the bit width of the components of
Base."

It clearly points that the requirement for rank is for the Base and
Result, not on the Base and Shift.

Fixes:
https://github.com/KhronosGroup/glslang/issues/1296


glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -816,17 +816,17 @@ TIntermediate::addConversion(TOperator op, TIntermTyped* node0, TIntermTyped* no
...
-    // with assign shift ops either.
+    // with assign shift ops either. Result and base needs to be the same rank.
     case EOpLeftShift:
     case EOpRightShift:
     case EOpLeftShiftAssign:
     case EOpRightShiftAssign:
 
-        if (isTypeInt(type0) && isTypeInt(type1)) {
-            if (getTypeRank(type0) == getTypeRank(type1)) {
+        if ((isTypeInt(type0)|| isTypeUnsignedInt(type0)) &&
```
johnkslang: isTypeInt() includes both signed and unsigned already.

johnkslang:
Would want to see a test change (or new test) come along with this.

(I haven't investigated this issue yet myself though; will do that soon as well.)

johnkslang:
The comment also says

```
    // Shifts can have mixed types as long as they are integer and of the same rank,
    // without converting.
```

What does the extension say about this? The core GLSL spec. wouldn't comment on it.

I agree it seems like different ranks should be okay, but given the alignment of the comment and the code, I think we need to be sure what is specified.

infapi00:
I have just pushed a new version of the branch, with the simplified code, plus a test.

johnkslang:
But, this change effects multiple existing tests, which need updating.

See `install/bin/glslangtests.exe --update-mode`.

Also, it is much preferred to add tests like this to `gtests/Spv.FromFile.cpp`.

johnkslang:
@alegal-arm do you have concerns with relaxing shifts to work with any mixed integers? It seems correct to relax to me.

alegal-arm:
>  do you have concerns

I do. Such relaxation would imply implicit conversion from uint32 to (signed) int64. Such *implicit* conversion is not allowed by GLSL/ARB_gpu_shader_int64 or the explicit types extension. I believe that the original glslang code was not entirely correct.

johnkslang:
> I do. Such relaxation would imply implicit conversion from uint32 to (signed) int64.

No, that's not true, unless we're missing some detail. This is how shifts are different from everything else. Consider:

`Result = Base << Shift`

In this, we generally say, in GLSL and in SPIR-V, that `Base` and `Shift` can be different types. Different in that no conversion ever takes place. They never need to be made to match, they just stay different.  What must match is the type of `Result` and `Base`.  This is because it is perfectly reasonable, say, to shift a 64-bit number by an amount of shift that fits in an 8-bit number. The two sizes stay as two sizes through the entire stack, no conversions needed or done.

The only conversion to consider is in converting the type of `Base` to the type of `Result`.

GLSL and SPIR-V have both always been this way.

> Such implicit conversion is not allowed by GLSL/ARB_gpu_shader_int64 or the explicit types extension.

Yes, I agree with that. The thing odd about the shift situation is that the conversion is not needed to begin with, so does not play into this rule.

alegal-arm:
Ok, I'm fine with this as long as there is no intermediate conversion generated.

johnkslang:
I'm doing a broader fix for this. The code is partly replicated in both addConversion() functions, but should not be.

johnkslang:
This will now be replaced by #1299.

@alegal-arm see some key points in #1299, and see if the changes look okay.

PR #1299, opened 2018-03-16
Fix conversions

### LINKED ISSUE #1296: Shifts should be allowed for base/shift operands of different bit-width and sign

johnkslang:
This fixes #1296 and replaces #1297.

The above uncovered some subtle and complex interactions with the history of adding multiple-width types, while splitting addConversion(), in parallel with HLSL getting different treatment from GLSL.

This set of commits rationalizes all of this.

Of key importance is that types of binary operations are not simply symmetrical things to convert to each other the same way for all operations.  E.g., logical ops want bools, not matching conversions, and shifts want ints, not matches.

Also important is the split of addConversion() was incomplete, in that both copies still supported ops from the original, even though they didn't quite make sense in the copies, except that the unconverted HLSL path still exercised them (now inappropriately).


johnkslang:
Any input here @alegal-arm?

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -3087,8 +3007,10 @@ bool TIntermediate::promoteBinary(TIntermBinary& node)
         case EOpSub:
         case EOpDiv:
         case EOpMul:
-            left = addConversion(op, TType(EbtInt, EvqTemporary, left->getVectorSize()), left);
-            right = addConversion(op, TType(EbtInt, EvqTemporary, right->getVectorSize()), right);
+            if (left->getBasicType() == EbtBool)
```
alegal-arm: how can we get bool operand here?

johnkslang: This is an HLSL-specific path, which allows numeric operations on Boolean types.

johnkslang: Ah, this is the only reason addConversion was called, was to change int -> bool, and it was calling the split out unary/conversion addConversion, but with binary ops.  This helps clean all that up. (It is at result-promote time, not at operand convert time, even though it goes back and adjusts the operands.) It is done in two phases: first operand-operand matching, as the current binary addConversion does, and then promotion to rationalize all three types (operand, operand, and result type).

PR #1302, opened 2018-03-19
PP: don't give errors on some tokens under #if 0 (or similar).

### LINKED ISSUE #1295: GL_AMD_gpu_shader_half_float is erroneously reported as required if half-float variables are used inside an #ifdef block which is removed at preprocessing time
Consider the following GLSL source code:

```
#ifdef SOME_DEFINE

   float16_t sth = 1.0hf;

#endif 
```
Let SOME_DEFINE be NOT defined in the shader. Under such assumption, GL_AMD_gpu_shader_half_float should *not* be required. However, glslang doesn't seem to notice that the half-floats are used inside an #ifdef block which gets removed at pre-processing stage and believes the corresponding float16 extension is needed. This leads to an error reported at compilation time, which is incorrect.

johnkslang:
Fixes #1295.

Tokens that are accepted by any version of HLSL or GLSL should be
allowed when #ifdef'd off, such that errors are not reported.


amdrexu:
Thank you. The change is reasonable.

PR #1303, opened 2018-03-20
Improve --auto-map-locations for uniforms

bpeel:
These patches make --auto-map-locations take into account the size of uniforms (structs and arrays occupy more than one location) and also assign locations to opaque types.


bpeel:
I’ve sent a v2 of the pull request.

PR #1321, opened 2018-03-29
Allowed for LIB_TYPE to be set to switch between static and shared libs for glslang, SPIRV, and HLSL

mattparks:
This adds the ability to switch the projects glslang, SPIRV, and HLSL from static to shared libraries by setting ${BUILD_SHARED_LIBS} to ON. ${BUILD_SHARED_LIBS} is set to OFF by default.


johnkslang:
I think the PR contains more than the commit you intend.

There should just be the one commit, not the uniform block bindings or merge.

mattparks:
It appears I managed to pull this wrong, will resubmit

PR #1322, opened 2018-03-30
Implement GL_NV_shader_subgroup_partitioned

jeffbolznv:


SPIRV/doc.cpp
```diff
@@ -876,6 +881,7 @@ const char* CapabilityString(int info)
     case 5255: return "ShaderViewportMaskNV";
     case 5259: return "ShaderStereoViewNV";
     case 5260: return "PerViewAttributesNV";
+    case 5297: return "GroupNonUniformPartitionedNV";
```
johnkslang: `case CapabilityGroupNonUniformPartitionedNV: ...`

But, I'll merge and do a set of these.  We didn't used to have the enumerants, as this is how they were made, but now we have them through the JSON grammar generation of the header files.

SPIRV/doc.cpp
```diff
@@ -1272,6 +1278,9 @@ const char* OpcodeString(int op)
     case OpDecorateStringGOOGLE:       return "OpDecorateStringGOOGLE";
     case OpMemberDecorateStringGOOGLE: return "OpMemberDecorateStringGOOGLE";
 
+#ifdef NV_EXTENSIONS
+    case 5296: return "OpGroupNonUniformPartitionNV";
```
johnkslang: `case OpGroupNonUniformPartitionNV:...`

johnkslang:
Long term, we should switch over to the SPIRV-Tools disassembler.  Near-term, I greatly reduced the camouflage in doc.cpp so its easier to see what needs to get added when adding new functionality.

See #1325.

PR #1338, opened 2018-04-09
Minor cleanup in ShaderLang.cpp

cgmb:
`CreateParseContext` can return null, so I added a check before dereferencing its returned pointer. Returning early after failure would require adding a bunch of explicit delete statements in `ProcessDeferred`, so I used `std::unique_ptr` to handle that automatically. There was also a really weird ternary expression for `versionNotFirst`.

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -744,9 +744,9 @@ bool ProcessDeferred(
     const int numPre = 2;
     const int numPost = requireNonempty? 1 : 0;
     const int numTotal = numPre + numStrings + numPost;
-    size_t* lengths = new size_t[numTotal];
-    const char** strings = new const char*[numTotal];
-    const char** names = new const char*[numTotal];
+    std::unique_ptr<size_t[]> lengths(new size_t[numTotal]);
+    std::unique_ptr<const char*[]> strings(new const char*[numTotal]);
+    std::unique_ptr<const char*[]> names(new const char*[numTotal]);
     for (int s = 0; s < numStrings; ++s) {
         strings[s + numPre] = shaderStrings[s];
         if (inputLengths == nullptr || inputLengths[s] < 0)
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -778,9 +779,8 @@ bool ProcessDeferred(
     int version = 0;
     EProfile profile = ENoProfile;
     bool versionNotFirstToken = false;
-    bool versionNotFirst = (source == EShSourceHlsl)
-                                ? true
-                                : userInput.scanVersion(version, profile, versionNotFirstToken);
+    bool versionNotFirst = source == EShSourceHlsl ||
```
johnkslang: This seems to obfuscate the result a bit. The original way explicitly says what computes the result, as a function of the source language.

johnkslang: Note that, currently, if HLSL evolved to a more complex rule, the `true` would simply get replaced by the more complex rule.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -236,7 +236,8 @@ bool InitializeSymbolTable(const TString& builtIns, int version, EProfile profil
     std::unique_ptr<TParseContextBase> parseContext(CreateParseContext(symbolTable, intermediate, version, profile, source,
                                                                        language, infoSink, spvVersion, true, EShMsgDefault,
                                                                        true));
-
+    if (parseContext == nullptr)
```
johnkslang: This was an internal error only, which would require a fix elsewhere. If you are hitting this, is there not a bug to fix somewhere else?

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -744,9 +745,9 @@ bool ProcessDeferred(
     const int numPre = 2;
     const int numPost = requireNonempty? 1 : 0;
     const int numTotal = numPre + numStrings + numPost;
-    size_t* lengths = new size_t[numTotal];
-    const char** strings = new const char*[numTotal];
-    const char** names = new const char*[numTotal];
+    std::unique_ptr<size_t[]> lengths(new size_t[numTotal]);
```
johnkslang: The `unique_ptr` parts look good.

glslang/MachineIndependent/ShaderLang.cpp
cgmb: Perhaps `assert` would be preferable?

glslang/MachineIndependent/ShaderLang.cpp
cgmb: I still think it's a little weird, but I'll put it back how I found it.

cgmb:
I wasn't hitting any errors. The change was motivated by the memory leak in `ProcessDeferred`. IIRC, it was picked up by static analysis when I was looking at RenderDoc.

This patch was written in November, but I got wrapped up in other things and never submitted it. The leak was apparently fixed in b502a80, but I figured it was worth submitting for the ancillary stuff.

I'll be stripping those ancillary changes out, so perhaps it wasn't... but, at least it's negative lines of code.

johnkslang:
Agreed that it is worth moving to `unique_ptr`, especially given the multiple exits.

glslang/MachineIndependent/ShaderLang.cpp
cgmb: Good point.

cgmb:
I've stripped out the various other changes. Now it's just switching stuff over to `std::unique_ptr`.

PR #1341, opened 2018-04-09
fix VS2012 compile issues

St0fF-NPL-ToM:
small fixes needed to compile under Win8.1/VS2012/static_lib

**settings**
ENABLE_AMD_EXTENSIONS: yes
ENABLE_GLSLANG_BINARIES: no
ENABLE_HLSL: no
ENABLE_NV_EXTENSIONS: yes
ENABLE_OPT: no

left out: SPIRV and SPVRemapper

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -3638,7 +3638,7 @@ spv::Id TGlslangToSpvTraverser::createImageTextureFunctionCall(glslang::TIntermO
             if (builder.getImageTypeFormat(builder.getImageType(operands.front())) == spv::ImageFormatUnknown)
                 builder.addCapability(spv::CapabilityStorageImageReadWithoutFormat);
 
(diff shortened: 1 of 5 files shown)


johnkslang:
This seems like it is mixing commits for multiple different purposes. Is this accidentally picking up history for something else?

johnkslang:
No problem.

Use a different branch for each proposed fix.  Git branch management is super light-weight and easy.

Try this for each independent thing you want to juggle:

```
git checkout master # or whatever your main branch is corresponding to glslang master
git checkout -b <new-branch-name>
```

Then, trivially switch between your branches with `git checkout <branch-name>`.

St0fF-NPL-ToM:
Thanks.  Gotta get used to it, SVN and branching are things not to be taken that easily :)  But sounds like the right approach GIT takes, there.

Anyhow, I think all those commits ain't that bad.  It's just a question if best possible compatibility or newest possible code style should be preferred.  I'm kind of bound to using VS2012, still.  There it's a fact that C++11 support is heavily lacking (like initializer lists and such).

PR #1342, opened 2018-04-10
fixes #1334 - "Named defines"

### LINKED ISSUE #1334: Make "tested" and "statically declared" preprocessor symbols visible to the caller
Some kinds of uber-shader editors have means to enable or disable specific named defines via Checkboxes.  For that it would be good to be able to access all used named defines after parsing a shader.

Definition "static named define":

    #define SOME_NAME

Definition "checked named define":

    #ifdef ANOTHER_NAME
      [...shader code...]
    #elif not defined A_THIRD_NAME
      [...more shader code...]
    #endif

The simplest solution to this would be to add two public sets to TIntermediate:

    public:
        std::unordered_set< std::string > checkedDefines, staticDefines;

and inserting the required code to add symbols to those sets everywhere necessary in the preprocessor code (Pp.cpp - it's only very few lines).  Though this approach looks more like a hack to me, thus it should be discussed in depth to find a maybe more sophisticated solution.

About the insertions within Pp.cpp: they should actively check if the Pp has already reached user code, thus it is not within a manually given preamble.  The thought behind this is, that "-D" compiler flags could be prepared as a preamble of a bunch of `#define symbol`-lines.

St0fF-NPL-ToM:
implement outside-accessibility of "Named defines" (accessed and statically defined)


St0fF-NPL-ToM:
P.S.: Although the revert only lists the first "VS2012 compile issues", it actually reverts both of those commits.  Just check the modified files.

johnkslang:
Oh, read about git squash.  Maybe make a fresh branch, and merge this branch into it with --squash.  It will make a clean single commit.

PR #1345, opened 2018-04-11
Add AMD float16/int16 SPV extension string when GLSL specifies them

amdrexu:


johnkslang:
I don't understand this yet.

SPIR-V extensions are be triggered by needing to use the SPIR-V feature and what version of Vulkan/SPIR-V needs the extension or has it in core. It should be decoupled from what extensions the high-level language declared.

SPIRV/GlslangToSpv.cpp
```diff
@@ -2578,8 +2578,11 @@ spv::Id TGlslangToSpvTraverser::convertGlslangToSpvType(const glslang::TType& ty
     case glslang::EbtFloat16:
         builder.addCapability(spv::CapabilityFloat16);
 #if AMD_EXTENSIONS
-        if (builder.getSpvVersion() < glslang::EShTargetSpv_1_3)
-            builder.addExtension(spv::E_SPV_AMD_gpu_shader_half_float);
+        {
```
johnkslang: not sure you the extra variable, and hence the braces, etc. here, why not just

```
if (glslangIntermediate->getRequestedExtensions().find(glslang::E_GL_AMD_gpu_shader_half_float) != sourceExtensions.end())
    builder.addExtension(spv::E_SPV_AMD_gpu_shader_half_float);

```

amdrexu:
This is because we have an issue of this: https://github.com/KhronosGroup/SPIRV-Tools/pull/1474
In SPV_AMD_gpu_shader_int16, we have such function prototype f16vec frexp(f16vec, i16vec). The validation rules should be changed based on the enablement of SPV_AMD_gpu_shader_int16. For glslang, when SPV environment is specified as SPIR-V 1.3. This extension string will not be generated so I have no way detecting this. According to your suggestion, I might add the SPV extension string when handling frexp() translation particularly. Am I right?

johnkslang:
I think so... you want to test in the paths that emit the SPIR-V code that needs the extension.

SPIRV/GlslangToSpv.cpp
johnkslang: Need to trigger by use of feature, target specific.

amdrexu:
Thank you. I modify the change.

amdrexu:
@johnkslang Any feedback about this commit?

johnkslang:
Doesn't the code need to be protected by AMD_EXTENSIONS?

amdrexu:
Sorry. My carelessness. Corrected.

PR #1352, opened 2018-04-12
Make glslang validator support files ending in .glsl

nicebyte:
This patch makes  the validator accept *.stage_name.[g/h]lsl pattern
for file names.

This patch preserves previous behavior (i.e. *.vert/*.frag/etc. in file
names still work).


nicebyte:
Ah, sorry about that. Thanks for reviewing!

johnkslang:
Hmmm, it's also missing updates to the usage and test results. Am in the middle of it, so will look at that too.  Next time...

PR #1356, opened 2018-04-13
PP: lexing: use "C" locale for strtod() and atof().

### LINKED ISSUE #234: Decimal mark dependency
Currently the compile time string to binary number parsing depends on the current active locale.

```
#version 450 core
layout(set = 1, binding = 0) uniform sampler2D tex;
layout(location = 0) in vec2 texcoord;
layout(location = 0) out vec4 outColor;

const float OFFSET_X = 0.5f;

void main()  {
    outColor = texture(tex, vec2(texcoord.x + OFFSET_X, texcoord.y));
}
```

Decimal mark set to "," (e.g. setlocale(LC_ALL, "German_Germany")) produces:

```
...
23:    6(float) Constant 0
...
```

Default Decimal mark "." produces the correct output:

```
...
23:    6(float) Constant 1056964608
...
```

### LINKED ISSUE #1228: Weird locale dependence when parsing float literals
Compiling the following program succeeds when `LC_NUMERIC` is `C` (or `en_US.UTF-8`), but fails when `LC_NUMERIC` is `de_DE.UTF-8`.

```
#version 450 
#extension GL_ARB_compute_shader : enable 
layout (local_size_x = 8, local_size_y = 8) in; 
 
void main() { 
    float foo = 2.2; 
} 
```

To demonstrate this, I wrote a small proof of concept:

```
#include <stdio.h>
#include <string.h>
#include <locale.h>

#include <glslang/Include/ResourceLimits.h>
#include <glslang/Public/ShaderLang.h>
#include <SPIRV/GlslangToSpv.h>
using namespace glslang;

extern const TBuiltInResource DefaultTBuiltInResource;

static const char *glsl =
    "#version 450 \n"
    "#extension GL_ARB_compute_shader : enable \n"
    "layout (local_size_x = 8, local_size_y = 8) in; \n"
    " \n"
    "void main() { \n"
    "    float foo = 2.2; \n"
    "} \n";

int main(int argc, char *argv[])
{
    setlocale(LC_NUMERIC, "de_DE.UTF-8");
    InitializeProcess();

    TShader *shader = new TShader(EShLangCompute);
    shader->setEnvClient(EShClientVulkan, 450);
    shader->setEnvInput(EShSourceGlsl, EShLangCompute, EShClientVulkan, 450);
    shader->setEnvTarget(EshTargetSpv, 0x00010000);
    shader->setStrings(&glsl, 1);
    if (!shader->parse(&DefaultTBuiltInResource, 450, true, EShMsgDefault)) {
        fprintf(stderr, "glslang log:\n%s\n", shader->getInfoLog());
        return 1;
    }

    TProgram *prog = new TProgram();
    prog->addShader(shader);
    if (!prog->link(EShMsgDefault)) {
        fprintf(stderr, "glslang log:\n%s\n", prog->getInfoLog());
        return 2;
    }

    std::vector<unsigned int> spirv;
    GlslangToSpv(*prog->getIntermediate(EShLangCompute), spirv);
    OutputSpvBin(spirv, "output");

    delete prog;
    delete shader;
    FinalizeProcess();
    return 0;
}

// copy pasted from StandAlone.h
const TBuiltInResource DefaultTBuiltInResource = {
    /* .MaxLights = */ 32,
    /* .MaxClipPlanes = */ 6,
    /* .MaxTextureUnits = */ 32,
    /* .MaxTextureCoords = */ 32,
    /* .MaxVertexAttribs = */ 64,
    /* .MaxVertexUniformComponents = */ 4096,
    /* .MaxVaryingFloats = */ 64,
    /* .MaxVertexTextureImageUnits = */ 32,
    /* .MaxCombinedTextureImageUnits = */ 80,
    /* .MaxTextureImageUnits = */ 32,
    /* .MaxFragmentUniformComponents = */ 4096,
    /* .MaxDrawBuffers = */ 32,
    /* .MaxVertexUniformVectors = */ 128,
    /* .MaxVaryingVectors = */ 8,
    /* .MaxFragmentUniformVectors = */ 16,
    /* .MaxVertexOutputVectors = */ 16,
    /* .MaxFragmentInputVectors = */ 15,
    /* .MinProgramTexelOffset = */ -8,
    /* .MaxProgramTexelOffset = */ 7,
    /* .MaxClipDistances = */ 8,
    /* .MaxComputeWorkGroupCountX = */ 65535,
    /* .MaxComputeWorkGroupCountY = */ 65535,
    /* .MaxComputeWorkGroupCountZ = */ 65535,
    /* .MaxComputeWorkGroupSizeX = */ 1024,
    /* .MaxComputeWorkGroupSizeY = */ 1024,
    /* .MaxComputeWorkGroupSizeZ = */ 64,
    /* .MaxComputeUniformComponents = */ 1024,
    /* .MaxComputeTextureImageUnits = */ 16,
    /* .MaxComputeImageUniforms = */ 8,
    /* .MaxComputeAtomicCounters = */ 8,
    /* .MaxComputeAtomicCounterBuffers = */ 1,
    /* .MaxVaryingComponents = */ 60,
    /* .MaxVertexOutputComponents = */ 64,
    /* .MaxGeometryInputComponents = */ 64,
    /* .MaxGeometryOutputComponents = */ 128,
    /* .MaxFragmentInputComponents = */ 128,
    /* .MaxImageUnits = */ 8,
    /* .MaxCombinedImageUnitsAndFragmentOutputs = */ 8,
    /* .MaxCombinedShaderOutputResources = */ 8,
    /* .MaxImageSamples = */ 0,
    /* .MaxVertexImageUniforms = */ 0,
    /* .MaxTessControlImageUniforms = */ 0,
    /* .MaxTessEvaluationImageUniforms = */ 0,
    /* .MaxGeometryImageUniforms = */ 0,
    /* .MaxFragmentImageUniforms = */ 8,
    /* .MaxCombinedImageUniforms = */ 8,
    /* .MaxGeometryTextureImageUnits = */ 16,
    /* .MaxGeometryOutputVertices = */ 256,
    /* .MaxGeometryTotalOutputComponents = */ 1024,
    /* .MaxGeometryUniformComponents = */ 1024,
    /* .MaxGeometryVaryingComponents = */ 64,
    /* .MaxTessControlInputComponents = */ 128,
    /* .MaxTessControlOutputComponents = */ 128,
    /* .MaxTessControlTextureImageUnits = */ 16,
    /* .MaxTessControlUniformComponents = */ 1024,
    /* .MaxTessControlTotalOutputComponents = */ 4096,
    /* .MaxTessEvaluationInputComponents = */ 128,
    /* .MaxTessEvaluationOutputComponents = */ 128,
    /* .MaxTessEvaluationTextureImageUnits = */ 16,
    /* .MaxTessEvaluationUniformComponents = */ 1024,
    /* .MaxTessPatchComponents = */ 120,
    /* .MaxPatchVertices = */ 32,
    /* .MaxTessGenLevel = */ 64,
    /* .MaxViewports = */ 16,
    /* .MaxVertexAtomicCounters = */ 0,
    /* .MaxTessControlAtomicCounters = */ 0,
    /* .MaxTessEvaluationAtomicCounters = */ 0,
    /* .MaxGeometryAtomicCounters = */ 0,
    /* .MaxFragmentAtomicCounters = */ 8,
    /* .MaxCombinedAtomicCounters = */ 8,
    /* .MaxAtomicCounterBindings = */ 1,
    /* .MaxVertexAtomicCounterBuffers = */ 0,
    /* .MaxTessControlAtomicCounterBuffers = */ 0,
    /* .MaxTessEvaluationAtomicCounterBuffers = */ 0,
    /* .MaxGeometryAtomicCounterBuffers = */ 0,
    /* .MaxFragmentAtomicCounterBuffers = */ 1,
    /* .MaxCombinedAtomicCounterBuffers = */ 1,
    /* .MaxAtomicCounterBufferSize = */ 16384,
    /* .MaxTransformFeedbackBuffers = */ 4,
    /* .MaxTransformFeedbackInterleavedComponents = */ 64,
    /* .MaxCullDistances = */ 8,
    /* .MaxCombinedClipAndCullDistances = */ 8,
    /* .MaxSamples = */ 4,
    /* .limits = */ {
        /* .nonInductiveForLoops = */ 1,
        /* .whileLoops = */ 1,
        /* .doWhileLoops = */ 1,
        /* .generalUniformIndexing = */ 1,
        /* .generalAttributeMatrixVectorIndexing = */ 1,
        /* .generalVaryingIndexing = */ 1,
        /* .generalSamplerIndexing = */ 1,
        /* .generalVariableIndexing = */ 1,
        /* .generalConstantMatrixVectorIndexing = */ 1,
    }};
```

The resulting SPIRV looks like this:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 3
; Bound: 15
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 8 8 1
               OpSource GLSL 450
               OpSourceExtension "GL_ARB_compute_shader"
               OpName %main "main"
               OpName %foo "foo"
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%_ptr_Function_float = OpTypePointer Function %float
    %float_2 = OpConstant %float 2
       %uint = OpTypeInt 32 0
     %uint_8 = OpConstant %uint 8
     %uint_1 = OpConstant %uint 1
%gl_WorkGroupSize = OpConstantComposite %v3uint %uint_8 %uint_8 %uint_1
          %5 = OpLabel
        %foo = OpVariable %_ptr_Function_float Function
               OpStore %foo %float_2
               OpReturn
               OpFunctionEnd
```

Commenting out the `setlocale` call gives this instead:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 3
; Bound: 15
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 8 8 1
               OpSource GLSL 450
               OpSourceExtension "GL_ARB_compute_shader"
               OpName %main "main"
               OpName %foo "foo"
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%_ptr_Function_float = OpTypePointer Function %float
  %float_2_2 = OpConstant %float 2.2
       %uint = OpTypeInt 32 0
     %uint_8 = OpConstant %uint 8
     %uint_1 = OpConstant %uint 1
%gl_WorkGroupSize = OpConstantComposite %v3uint %uint_8 %uint_8 %uint_1
          %5 = OpLabel
        %foo = OpVariable %_ptr_Function_float Function
               OpStore %foo %float_2_2
               OpReturn
               OpFunctionEnd
```

Note the diff between the two:
```
--- english-glslang.dis	2018-01-26 11:35:47.603583350 +0100
+++ german-glslang.dis	2018-01-26 11:35:48.850574035 +0100
@@ -17,7 +17,7 @@
           %3 = OpTypeFunction %void
       %float = OpTypeFloat 32
 %_ptr_Function_float = OpTypePointer Function %float
-  %float_2_2 = OpConstant %float 2.2
+    %float_2 = OpConstant %float 2
        %uint = OpTypeInt 32 0
      %uint_8 = OpConstant %uint 8
@@ -26,6 +26,6 @@
           %5 = OpLabel
         %foo = OpVariable %_ptr_Function_float Function
-               OpStore %foo %float_2_2
+               OpStore %foo %float_2
                OpReturn
                OpFunctionEnd
```

I'm not entirely sure why parsing the GLSL depends on the locale of the executing program, but this is very clearly a bug since it produces radically wrong results when using glslang on a system with german locale.

I suggest you rewrite your parser to be independent of the system locale.

johnkslang:
Fixes #234. Fixes #1228.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -41,6 +41,7 @@
 
 #include "../OSDependent/osinclude.h"
 #include <algorithm>
+#include <clocale>
 
 #include "preprocessor/PpContext.h"
 
@@ -188,7 +189,20 @@ bool TParseContext::parseShaderStrings(TPpContext& ppContext, TInputScanner& inp
 {
     currentScanner = &input;
     ppContext.setInput(input, versionWillBeError);
+
+    // Push the locale to "C", for numeric standard processing,
+    // where lexical analysis depends on library functions for
+    // "C" style locale.
+    TString startLocale(setlocale(LC_NUMERIC, nullptr));
+    if (startLocale != "C")
+        setlocale(LC_NUMERIC, "C");
+
+    // parse
     yyparse(this);
+
+    // pop the locale
+    if (startLocale != "C")
+        setlocale(LC_NUMERIC, startLocale.c_str());
 
     finish();
 
--- a/glslang/MachineIndependent/preprocessor/PpScanner.cpp
+++ b/glslang/MachineIndependent/preprocessor/PpScanner.cpp
@@ -231,6 +231,7 @@ int TPpContext::lFloatConst(int len, int ch, TPpToken* ppToken)
     ppToken->name[len] = '\0';
 
     // Get the numerical value
+    // Depends on setlocale(LC_NUMERIC, "C") for correct behavior.
     ppToken->dval = strtod(ppToken->name, nullptr);
 
     // Return the right token type
--- a/glslang/MachineIndependent/preprocessor/PpTokens.cpp
+++ b/glslang/MachineIndependent/preprocessor/PpTokens.cpp
@@ -216,6 +216,7 @@ int TPpContext::TokenStream::getToken(TParseContextBase& parseContext, TPpToken 
         case PpAtomConstFloat:
         case PpAtomConstDouble:
         case PpAtomConstFloat16:
+            // Depends on setlocale(LC_NUMERIC, "C") for correct behavior.
             ppToken->dval = atof(ppToken->name);
             break;
         case PpAtomConstInt:


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -188,8 +189,21 @@ bool TParseContext::parseShaderStrings(TPpContext& ppContext, TInputScanner& inp
 {
     currentScanner = &input;
     ppContext.setInput(input, versionWillBeError);
+
+    // Push the locale to "C", for numeric standard processing,
+    // where lexical analysis depends on library functions for
+    // "C" style locale.
+    TString startLocale(setlocale(LC_NUMERIC, nullptr));
+    if (startLocale != "C")
+        setlocale(LC_NUMERIC, "C");
```
dneto0: setlocale is not necessarily thread-safe.

dneto0: It looks like a strictly correct and thread safe method for this is to use an istringstream and use imbue to set the locale in a thread-safe way.
http://en.cppreference.com/w/cpp/io/ios_base/imbue 
But boy that seems heavyweight.

dneto0: Also, the alternative strategy whereby this grabs a mutex is still not sufficient because the locale state is global to the process, and not all threads might be going through this codepath.

johnkslang: Good points.

I wonder if we should require the consumer of libraries to only call with the "C" locale.

Or, just write/get a parser.  I should do the latter, it would be faster than all this (faster in multiple senses). I think someone offered one, but don't recall actually seeing it.

johnkslang:
Labeling WIP, because this is not a robust solution.

johnkslang:
@dneto0 wrote

> It looks like a strictly correct and thread safe method for this is to use an istringstream and use imbue to set the locale in a thread-safe way.
http://en.cppreference.com/w/cpp/io/ios_base/imbue
But boy that seems heavyweight.

Yes, it is much much slower than strtod(). I have implemented using it though, without loss of performance. It is quite complex to work around the differences across platforms and also the differences between strtod and sstream, which uses a different algorithm. Especially when trying to have tests that get the same result across platforms.

Summary: 
- A "fast path" that avoids library calls, for 15-digit mantissas and exponents up to 22, and caching the imbued sstream give a net gain in performance gain per accuracy (new results are more accurate on Windows).
- Printing with reduced precision, and doing some string manipulation gets matching results across platforms.
- Used a new binary printing scheme, locally (turned off in master) to prove correctness of fast path.

johnkslang:
See https://github.com/KhronosGroup/glslang/pull/1385.

Issue #1324, opened 2018-03-30
Missing get vertex attribute location (binding)


mattparks:
This is a getter that appears to be missing from TProgram (a function named something along the lines of getAttributeLocation), essentially it would get the location given to a vertex layout.

Also missing: getAttributeSize, getUniformSize (these would return the size of the attribute/uniform and support array sizing for uniforms).

Here is a glsl snippet used in my example:
```
#version 450

layout(location = 0) in vec3 vertexPosition;
layout(location = 1) in vec2 vertexUv;
layout(location = 2) in vec3 vertexNormal;

void main() 
{
}
```

A expected result might be as follows:
```
program.getAttributeName(0) = "vertexPosition"
program.getAttributeLocation(0) = 0

program.getAttributeName(1) = "vertexNormal"
program.getAttributeLocation(1) = 2

program.getAttributeName(2) = "vertexUv"
program.getAttributeLocation(2) = 1
```

Edited to put code markers on their own lines

johnkslang:
We do have

```
TProgram::getAttributeName
TProgram::getAttributeTType
```

The last of which returns a `TType*` which lets you query everything about a type.  Does that help?  I'm unclear about duplicating everything the the `TType` path already provides.

mattparks:
Using getAttributeTType does work ok for finding the size in my situation; however, the location given though glsl does not match the index provided to C++. Unlike indices from uniform blocks that are just reverse ordered, attribute locations are given in a random order by index. At location 2 I expect "vertexUv", but get "vertexNormal".

johnkslang:
I'm not sure I follow. Indexes are not locations. I'll have to look deeper I guess, but off the top of my head, I'd think you'd go through the indexes, and for each one query the name and the location, which would go together, regardless of the index used to do the query.  Sorry if I remember incorrectly, let me know.

PR #1357, opened 2018-04-13
Added attribute location getter.

mattparks:
This resolves issue https://github.com/KhronosGroup/glslang/issues/1324, gives access to layoutLocation from a shader attribute qualifier for reflections.


johnkslang:
Does the existing entry point to return the TType work for getting access to this information?

Closing, because the tests failed, but also unclear we need an endless stream of new getters for things already available from the TType (if that is what's going on).

PR #1366, opened 2018-05-02
Added a [[dynamically_uniform]] variable attribute.

sheredom:
Developers are constantly asking for a way to make `Uniform` decorations
appear in the output SPIR-V. I've taken a stab at reusing the attribute
mechanism introduced in the control flow GLSL extension to add a variable
attribute [[dynamically_uniform]] that can apply to variables, function
parameters, and function return types - which results in the corresponding
variables having the `Uniform` decoration.

I haven't drafted any GLSL extension for a couple of reasons:

1. Do we actually need one for functionality that is unconditionally in
SPIR-V anyway?
2. Even if I do require an extension, it'll be trivial to draft it after
the technical approach I've taken is approved.

I've also added extensive warnings for cases when a user might be moving
a value of unknown uniformity into a value that is denoted as dynamically
uniform - a normal cause of errors.


johnkslang:
> Do we actually need one for functionality that is unconditionally in SPIR-V anyway?

Three reasons to do so:

1. Other front ends (and consumers) need a specification to know what rules to implement. Consumers benefit as well from having documentation. E.g., this issue should not be the documentation.

2. It is effectively still defining a standard; there should be some formalism in establishing semantics.

3. Can get input on the spec. from others. It can happen in parallel to implementation; we don't need a heavyweight/blocking process.

A quick vendor or EXT added to the GLSL repository would be worthwhile.  It does not need to introduce a stall in implementation.

sheredom:
Done here https://github.com/KhronosGroup/GLSL/pull/17

johnkslang:
Ah, I was thinking that would only happen if the behavior was requested. But, I see the point either way. I don't have a clear opinion right now.

PR #1378, opened 2018-05-15
HLSL: Allow stream output Append() method after entry point.

### LINKED ISSUE #1217: HLSL: Trying to call Append on a TriangleStream argument causes compile error
Compiling the following valid HLSL shader

```hlsl
struct GSPS_INPUT
{
};

void EmitVertex(in GSPS_INPUT output, inout TriangleStream<GSPS_INPUT> TriStream)
{
    TriStream.Append( output );
}

[maxvertexcount(3)]
void main( triangle GSPS_INPUT input[3], inout TriangleStream<GSPS_INPUT> TriStream )
{
    EmitVertex(input[0], TriStream);
    EmitVertex(input[1], TriStream);
    EmitVertex(input[2], TriStream);
}
```
generates the following error:
```
ERROR: bug.gp:7: '' : unable to find output symbol for Append()
bug.gp(7): error at column 31, HLSL parsing failed.
```

LoopDawg:
The HLSL Append() method and associated stream output symbol is special: it does not copy some temporary data to the output symbol in the entry point epilogue, but rather uses an EmitVertex builtin after each write to the output stream.

This had been handled by remembering the special output symbol for the stream as it was declared in the shader entry point before symbol sanitization.  However the prior code was too simple and only handled cases where the Append() method happened after the entry point, so that the output symbol had been seen.

This PR adds a patching step so that the Append()s may appear in any order WRT the entry point.  They are patched in an epilogue, whereupon it is guaranteed in a well formed shader that we have seen the appropriate declaration.

Fixes #1217.


johnkslang:
Looks good.

PR #1380, opened 2018-05-16
Print advanced type info in human readable output

krOoze:
- add `void`
- <s>print specific integer and fp types (based on width and signedness)</s>
Refine changes by the duplicate PR #1387
- print vector and matrix types similarly to GLSL (incl width and signedness of components)
- print arrays and pointers with underlying type
- print arrays with size (if compile time)


johnkslang:
This should cause a massive change to all test files.

From a purist perspective, we should be standardizing on the SPIRV-Tools disassembler. There have been a few practical concerns (and of course, original history) that has glslang using this original disassembler, rather than pulling in the one from SPIRV-Tools.

It might make sense though, to enhance more the SPIRV-Tools disassembler, if it is lacking the features you'd like, and then executing it on the SPIR-V binaries that glslang produces.

krOoze:
Ups; pushed an empty branch...

Right. Added the tests and some tweaks.
Take it or leave it. Ping if you need rebase.

`spirv-dis` does seem somewhat better with its "friendly" result id names. Would it make sense to advertise it in the `glslangValidator -h` in the meantime?

I was also thinking this should actually be the job for the `OpName`. The user could then get familiar typenames of their original shader language (where applicable).

krOoze:
I am not currently available, but probably neither are you on weekends. 
Should I remove the tests when I get the chance?

'vec' is borrowed from GLSL, but AIS 'OpName' could be better solution than 
hardcoding.

spirv-dis has unfriendly names mode, which might be preferrable for the 
tests when the switch is made...

---------- Původní zpráva ----------
Od: John Kessenich 
Datum: 24. 5. 2018 v 17:48:06
Předmět: Re: [KhronosGroup/glslang] Print advanced type info in human 
readable output (#1380)


If we decide to pull this in, I will regenerate the test results myself, at 
the moment we bring it in.

I'm still reluctant to cause such a massive change. Perhaps some is 
unnecessary; e.g., the change of vec to fvec. But, this will impact many 
projects.

Maybe it should be tied to a command-line option? Even better is if an 
optional presence of the SPIRV-Tools disassembler is present, and that is 
selected via command-line option, that there is a coherent option for all 
this.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
(https://github.com/KhronosGroup/glslang/pull/1380#issuecomment-391764237), 
or mute the thread

LoopDawg:
>  Even better is if an optional presence of the SPIRV-Tools disassembler is present, and that is selected via command-line option, that there is a coherent option for all this.

I like that idea.  It also relates to a prior suggestion somewhere to adapt to the optional presence of spirv-val to enable an automatic post-compilation validation step.

johnkslang:
The standard SPIRV-Tools disassembler is now available through the `--spirv-dis` option. I'd like to start with that, and see what needs improved.

johnkslang:
Closing, as the --spirv-dis option is working. I've put in a request for better color coding, and other enhancements can be handled in SPIRV-Tools, rather than here.

Issue #1376, opened 2018-05-15
16 bit float use never emits SPV_AMD_gpu_shader_half_float if target-env is vulkan1.1


krOoze:
Test shader:
```
#version 450
#extension GL_AMD_gpu_shader_half_float : require

layout (location = 0) in float16_t interpolant;

void main(){
 	// forces SPV_AMD_gpu_shader_half_float to be necessary
	float16_t ac = interpolateAtCentroid(interpolant);
}
```

Output:
```
>glslangValidator.exe -V --target-env vulkan1.1 -H float16.frag
float16.frag
// Module Version 10300
// Generated by (magic number): 80006
// Id's are bound by 12

                              Capability Shader
                              Capability Float16
                              Capability InterpolationFunction
                              Capability StorageInputOutput16
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main" 10
                              ExecutionMode 4 OriginUpperLeft
                              Source GLSL 450
                              SourceExtension  "GL_AMD_gpu_shader_half_float"
                              Name 4  "main"
                              Name 8  "ac"
                              Name 10  "interpolant"
                              Decorate 10(interpolant) Location 0
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 16
               7:             TypePointer Function 6(float)
               9:             TypePointer Input 6(float)
 10(interpolant):      9(ptr) Variable Input
         4(main):           2 Function None 3
               5:             Label
           8(ac):      7(ptr) Variable Function
              11:    6(float) ExtInst 1(GLSL.std.450) 76(InterpolateAtCentroid) 10(interpolant)
                              Store 8(ac) 11
                              Return
                              FunctionEnd
```

It does not generate `Extension  "SPV_AMD_gpu_shader_half_float"`, which is necessary for `InterpolateAtCentroid` to accept 16-bit floats.

johnkslang:
@amdrexu it seems like the implementation here needs to be finer grained, intercepting operators, not just based on types.

krOoze:
@johnkslang That too (or perhaps the unextended `GLSL.std.450` spec updated -- for some reason it accepts 16-bit for some ops but not others; which I find weird).

[...]

PR #1383, opened 2018-05-17
Fix issue #1376: SPV_AMD_gpu_shader_half_float is not emitted under vulkan1.1

amdrexu:
Interpolation functions, using 16-bit interpolants, should always emit this extension string regardless of vulkan enviornment version.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -4630,6 +4630,10 @@ spv::Id TGlslangToSpvTraverser::createUnaryOperation(glslang::TOperator op, OpDe
         unaryOp = spv::OpFwidthCoarse;
         break;
     case glslang::EOpInterpolateAtCentroid:
+#ifdef AMD_EXTENSIONS
+        if (typeProxy == glslang::EbtFloat16)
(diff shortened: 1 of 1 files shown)


SPIRV/GlslangToSpv.cpp
```diff
@@ -4630,6 +4630,10 @@ spv::Id TGlslangToSpvTraverser::createUnaryOperation(glslang::TOperator op, OpDe
         unaryOp = spv::OpFwidthCoarse;
         break;
     case glslang::EOpInterpolateAtCentroid:
+#ifdef AMD_EXTENSIONS
+        if (typeProxy == glslang::EbtFloat16)
+            builder.addExtension(spv::E_SPV_AMD_gpu_shader_half_float);
+#endif
```
krOoze: Would that work without `AMD_EXTENSIONS`?
Should throw or log an error, if `AMD_EXTENSIONS` is OFF.

amdrexu: If AMD_EXTENSIONS is OFF, you can not use interpolateXXX() functions with float16_t type. The function prototypes are unavailable. Compilation error will be reported.

PR #1387, opened 2018-05-23
Add bit width of types to disassembleInstruction

jeffbolznv:
Without this, it is very hard to read disassembly of shaders that use explicitly-sized types. I left int/float as-is to avoid perturbing hundreds more generated files.

disassemble.cpp is the only non-generated file that changed.


jeffbolznv:
I was unaware of the other change. That one doesn't handle explicitly-sized types, which is really what I'm interested in. IIRC the SPIRV-Tools disassembly doesn't print short names for the basic types, it just uses whatever is in the OpName.

johnkslang:
Okay. I'm more comfortable with a change that only affects non-32-bit types.

The result of running `runtests` must also be updated.

johnkslang:
I should also point out that using `bump` in the `Test` directory updates the results of `runtests`.

jeffbolznv:
I've regenerated the file from runtests.

PR #1391, opened 2018-05-26
Removed usage of strnlen in floating-point tests

mattparks:
`strnlen` is defined as `size_t strnlen(const char *s, size_t maxlen);` on most systems; however, it is not a standard C++ function. Some compilers like MinGW fail to compile after https://github.com/KhronosGroup/glslang/commit/1ea1b13f388a06cbfa363585657eec7f4fc189a7

### DIFF
--- a/glslang/MachineIndependent/intermOut.cpp
+++ b/glslang/MachineIndependent/intermOut.cpp
@@ -1115,7 +1115,7 @@ static void OutputDouble(TInfoSink& out, double value, TOutputTraverser::EExtraO
 
         // remove a leading zero in the 100s slot in exponent; it is not portable
         // pattern:   XX...XXXe+0XX or XX...XXXe-0XX
-        int len = (int)strnlen(buf, maxSize);
+        int len = std::min((int)strlen(buf), maxSize);
         if (len > 5) {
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/intermOut.cpp
```diff
@@ -1115,7 +1115,7 @@ static void OutputDouble(TInfoSink& out, double value, TOutputTraverser::EExtraO
 
         // remove a leading zero in the 100s slot in exponent; it is not portable
         // pattern:   XX...XXXe+0XX or XX...XXXe-0XX
-        int len = (int)strnlen(buf, maxSize);
+        int len = std::min((int)strlen(buf), maxSize);
```
johnkslang: The purpose of `strnlen` is to prevent overrun. If we are not going to do that, there is no purpose in `min`. It should probably just be a straight used of `strlen`.

mattparks: The way `strnlen` works is it gos though the buffer until the maxSize is reached or '\0' is found. `strlen` basicly does the same thing without checking if it has passed a maxSize, so what I did is added a cap to the `strlen` length so it works just as before. I still had the min check if the double copies in the buffer with the defined format and exceeds the largest floating point of 300. If that can never happen than the min part is just a optional check.

PR #1393, opened 2018-05-30
Re-arranged usage message.

Strum355:
The message should be more organized than what it was before. Grouped the synonyms where there was enough line space, grouped similar length options


johnkslang:
I'm not sure about this:

- the diff shows diffs that don't look different: did line endings or other white space change?
- something like this should be at most one commit
- it moves from one way of organizing to another ("should" is a tricky word)

Strum355:
1) The lines that dont look different in the diff have a different space count.
2) You can always squash the commits when youre merging.
3) It is definitely better than what was there before. Notable with the various synonyms that were split up (such as `-v` and `--version`), there was little/no consistency. `-v` was right in the middle while `--version` was the second from the bottom. By grouping synonyms together, adding some more whitespace and grouping options of equal length, it makes the entire message overall a lot more readable.  

I could probably group the options to be a bit more legible again, however I thought that this would suffice.

Old:  
image  

New:  
image

johnkslang:
Just so a I understand:

For order, it was three parts, each part alphabetical within it: Single letter capital, single letter lower, and multiple letters.

What defines the order now?

johnkslang:
Oh, this also used to fit in 80 character terminals, but now it does not.

Strum355:
Theres a few lines that were over 80 characters in the original one, admittedly the line length has increased, which I will work on.
I hadnt noticed that they were previously sorted alphabetically, something which I will make my edit follow as well. My largest issue was with the inconsistencies, where for some, the synonyms would be together on the same line, while for others they are basically on opposite ends

Strum355:
Heres another revision. I'd be happy to get your feedback on this one :) The image was taken on a terminal at 80 characters wide

Ordering:
- single letter lowercase alphabetical
- single letter uppercase alphabetical
- three flags for creating SPIR-V
- multi letter alphabetical
- mutli letter varying parameters alphabetical

image

johnkslang:
I appreciate improvements. Still have feedback though, given the large amount of text involved:

I'd like to see a single commit that only changes what needs to change, so that it can be reviewed from a "what changed" perspective, and justifies the change everyone will see. That would mean not needlessly swapping of the upper case and lower case sections (existing commits didn't do that, just your image) and not changing the spacing from 10 to 11 spaces.

Bigger things though are 
- all the new `-` which are not actually option-selecting; that's a bit obfuscating
- if one wondered what, say, `-w`, meant, one must now search for it, because it's no longer where you'd expect to see it (almost like it is missing)

Maybe things like `-w` should be in both places?

Strum355:
I've locally changed the spacing to 10 as well as reversed the upper case and lower case sections as you have requested. 

Regarding the other two points:
- I can see what you mean by it being misleading. What character(s) would you prefer to replace them?
- regarding this point, personally I would be of the opinion that duplicate/separated entries should be avoided. Taking a page from the linux man pages, they have flags in the form of `-short, --long`. Maybe we could simply swap the synonymous flags, such as `-w`, to be in the form `-w | --suppress-warnings` rather than `--supress-warnings | -w` and then group them with the single character flags?

johnkslang:
Respectively:

- just indent and don't use a bullet, or keep it in sentence form, separated by ";"
- following the current prevailing standard is okay with me

PR #1405, opened 2018-06-13
Take into account arrays of opaque types with --auto-map-bindings

bpeel:
Previously if you had code like the following:

```
uniform r_values {
        float val;
} r_arr[5];

uniform g_values {
        float val;
} g_arr[5];
```

it would assign binding 0 to r_arr and binding 1 to g_arr. This means that the binding for g_arr collides with r_arr because r_arr uses 5 consecutive bindings. This pull request tries to fix that.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -383,29 +383,34 @@ struct TDefaultIoResolverBase : public glslang::TIoMapResolver
         return !(at != slots[set].end() && *at == slot);
     }
 
-    int reserveSlot(int set, int slot)
+    int reserveSlot(int set, int slot, int size = 1)
     {
         TSlotSet::iterator at = findSlot(set, slot);
 
         // tolerate aliasing, by not double-recording aliases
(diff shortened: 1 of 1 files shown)


johnkslang:
Is this a Vulkan versus OpenGL difference?  Are doing this for OpenGL?

bpeel:
Ah right, yes, this issue only applies to OpenGL because under GL each element of the array of a buffer or opaque type gets a separate binding whereas in Vulkan they are all combined into one binding and each binding can have multiple descriptors. I didn’t realise this at first so I guess I should modify the patch to only do this for GL.

Just for reference, this is the relevant quote from the GL 4.6 spec section 4.4.5:

> If the binding identifier is used with a uniform or shader storage block instanced as an array, the first element of the array takes the specified block binding and each subsequent element takes the next consecutive uniform block binding point. For an array of arrays, each element (e.g., 6 elements for a[2] [3]) gets a binding point, and they are ordered per the array of array ordering described in section 4.1.9 “Arrays.”

bpeel:
I’ve updated the branch so that it only affects OpenGL.

PR #1407, opened 2018-06-14
TProgram exposes layout manipulation functions

de2nis:
New functions allow the manipulation of an already built reflection.
User can query and change the location/binding of the shader's uniforms,
vertex attributes and varyings.
Changes will be reflected in the generated spirv (if requested).

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -687,40 +687,40 @@ public:
     int layoutOffset;
     int layoutAlign;
 
-                 unsigned int layoutLocation            :12;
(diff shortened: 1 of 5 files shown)


johnkslang:
Isn't all this already available by querying the type? Or, are we missing a writable type query? I don't think we need reencapsulate every way of modifying a type, or risk breaking the readonly/writable design of TType with mutable, which is critical to the thread safety of sharing symbol tables.

de2nis:
That's correct, thank you for your comment. We will try to reimplement our approach and come back if something needed.

Issue #1408, opened 2018-06-19
SPVRemapper does not handle OpConstantNull


danginsburg:
SPVRemapper.cpp gives the error "unimplemented constant type" for OpConstantNull:

https://github.com/KhronosGroup/glslang/blob/bea08fe02ac5715efe77ce1e559c0c3a208863ab/SPIRV/SPVRemapper.cpp#L223

Our compile pipeline halts when SPVRemapper calls the error callback so this issue is blocking me with one of our latest shaders.  

Is there something SPVRemapper needs to do special to handle OpConstantNull?

greg-lunarg:
I will take a look.

greg-lunarg:
It looks like we are using isConstOp to guard logic assuming OpConstantNull will not be seen. I will look to add additional logic to handle it.

danginsburg:
The attached SPIR-V reproduces the issue.

[test_309.zip](https://github.com/KhronosGroup/glslang/files/2116036/test_309.zip)

greg-lunarg:
I threw in OpConstantSampler while I was in the neighborhood.

PR #1409, opened 2018-06-19
Add support for OpConstantNull and OpConstantSampler to spirv-remap

greg-lunarg:
Fixes issue #1408


johnkslang:
Note: `Fixes issue #1408`. If it had said `Fixes #1408`, the issue would have automatically closed.

PR #1416, opened 2018-06-25
Add support for GL_EXT_samplerless_texture_functions

aejsmith:
This adds support for GL_EXT_samplerless_texture_functions, the specification for which can be found at https://github.com/KhronosGroup/GLSL/pull/18


aejsmith:
I see the spv.specConstant.vert test is now failing with this difference, but I don't really see why that would happen as a result of this change:
{code}
With diff:
@@ -12,5 +12,5 @@
                               Name 4  \"main\"
                               Name 9  \"arraySize\"
-                              Name 14  \"foo(vf4[s2468];\"
+                              Name 14  \"foo(vf4[s2543];\"
                               Name 13  \"p\"
                               Name 17  \"builtin_spec_constant(\"
@@ -103,8 +103,8 @@
               48:          10 Load 22(ucol)
                               Store 47(param) 48
-              49:           2 FunctionCall 14(foo(vf4[s2468];) 47(param)
+              49:           2 FunctionCall 14(foo(vf4[s2543];) 47(param)
                               Return
                               FunctionEnd
-14(foo(vf4[s2468];):           2 Function None 12
+14(foo(vf4[s2543];):           2 Function None 12
            13(p):     11(ptr) FunctionParameter
               15:             Label
{code}

johnkslang:
That sometimes changes when the source changes, but it has always been stable across platforms, for a given source base.

aejsmith:
OK, can I just include a change to that value with this commit? It does appear to be stable at the new value.

johnkslang:
> OK, can I just include a change to that value with this commit? It does appear to be stable at the new value.

Yes.

Though, I have not fully reviewed the commit, because of the tests not passing.

aejsmith:
I've pushed an updated commit that fixes the test.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5914,15 +5914,19 @@ void TBuiltIns::add2ndGenerationSamplingImaging(int version, EProfile profile, c
...
+                                    //
+                                    // GL_EXT_samplerless_texture_functions
+                                    // allows texelFetch() and query functions
+                                    // (other than textureQueryLod()) for all
+                                    // texture types.
                                     sampler.setTexture(sampler.type, sampler.dim, sampler.arrayed, sampler.shadow,
                                                        sampler.ms);
                                     TString textureTypeName = sampler.getString();
                                     addSamplingFunctions(sampler, textureTypeName, version, profile);
+                                    addQueryFunctions(sampler, textureTypeName, version, profile);
```
johnkslang: Where are the semantic checks that these are not used unless the extension is currently enabled by the shader?

johnkslang: We need to see added positive and negative tests that show correct functioning of the new feature when the extension is enabled, and rejection of the new feature when the extension is disabled.

aejsmith: I had thought about adding checks for the extension there but I see no way to do it currently. The builtins are all added before we process any #extension directives. setFunctionExtensions applies to all overloads of a function, so I can't just use that on the affected functions.

Could possibly add a new version of setFunctionExtensions that applies to only specific overloads? Looks like at that point it'd have to deal with mangled symbol names though which doesn't seem particularly tidy.

Do you have any other suggestions?

johnkslang: Check out `TParseContext::builtInOpCheck()`:

```
//
// Do additional checking of built-in function calls that is not caught
// by normal semantic checks on argument type, extension tagging, etc.
//
// Assumes there has been a semantically correct match to a built-in function prototype.
//
void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCandidate, TIntermOperator& callNode)
```

aejsmith:
Thanks for the pointer to builtInOpCheck. I've pushed a new version which adds the extension checks there, and adds tests for code generation and rejection based on whether the extension is enabled.

PR #1419, opened 2018-06-27
Include spirv-remap.exe in AppVeyor artifacts

tgjones:
This would help me downstream in [Shader Playground](http://shader-playground.timjones.io/), where I have a build script that downloads a number of compilers, and it's easier to work with existing CLI binaries.

Of course, if there's a reason why `spirv-remap.exe` isn't (yet) included in the AppVeyor artifacts, I can always build from source in my build script.


antiagainst:
Just wanted to say awesome work on shader playground! Kudos, @tgjones! :)

tgjones:
Thanks, much appreciated :)

tgjones:
Thanks for merging this! And I've just added it to Shader Playground:

spirv-remap

johnkslang:
Nice!

PR #1420, opened 2018-06-29
SPV: Add option to print disassembly in standard form using SPIRV-Tools.

johnkslang:
This allows printing SPIR-V in the standard form, as per SPIR-V tools.  Can be used, for example, like:

```
glslangValidator -V --spirv-dis t.frag  # compile for Vulkan, print SPIR-V assembly via SPIRV-Tools
```

The existing and default uses of glslang are all still as they were.

@dneto0 and @greg-lunarg, or anyone, a few design questions:

- Recommendations for the  option spelling and interactions? Note, I am not yet intending to switch how the regression suites work.
- I reused ENABLE_OPT to know if SPIRV-Tools was present or not, and it seems it had some use like that already. Should we change the name to generalize, or have a two-level hierarchy? Seems the hard part of this change will be managing CMake options and compiler flags. Simple is to just keep using ENABLE_OPT for this, despite being misnamed. I did not have to include an extra library to get the functionality.
- Opened related issue https://github.com/KhronosGroup/SPIRV-Tools/issues/1652 to request standalone disassembly.
- Should I do anything with the SPIRV-Tools `spv_diagnostic`?

### DIFF
--- a/SPIRV/CMakeLists.txt
+++ b/SPIRV/CMakeLists.txt
@@ -60,6 +60,7 @@ if(ENABLE_OPT)
         PRIVATE ${spirv-tools_SOURCE_DIR}/source
     )
     target_link_libraries(SPIRV glslang SPIRV-Tools-opt)
+    target_include_directories(SPIRV PUBLIC ../External)
 else()
     target_link_libraries(SPIRV glslang)
 endif(ENABLE_OPT)
--- a/SPIRV/disassemble.cpp
+++ b/SPIRV/disassemble.cpp
@@ -716,4 +716,25 @@ void Disassemble(std::ostream& out, const std::vector<unsigned int>& stream)
     SpirvStream.processInstructions();
 }
 
+#if ENABLE_OPT
+
+#include "spirv-tools/source/disassemble.h"
+
+// Use the SPIRV-Tools disassembler to print SPIR-V.
+void SpirvToolsDisassemble(std::ostream& out, const std::vector<unsigned int>& spirv)
+{
+    spv_context context = spvContextCreate(SPV_ENV_UNIVERSAL_1_3);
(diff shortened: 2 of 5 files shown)


johnkslang:
This also relates to https://github.com/KhronosGroup/glslang/pull/1380, as trying to shift to SPIRV-Tools as the place to invest in better printing.

johnkslang:
Regarding the design of the command-line option, we probably want room for sub-options to control how the disassembly is generated.  E.g., naming methods, indentation, etc.

dneto0:
About the spv_diagnostic:  If it's null then no diagnostics will be generated.
A diagnostic would only be generated if SPIRV-Tools couldn't parse the binary.  In this use, that will happen if someone adds a new instruction or token to Glslang but not the grammar that SPIRV-Tools builds from.  That might be good to do from a defensive perspective to protect people extending Glslang.

dneto0:
About the option name.  --spirv-dis seems fine to me.  It matches the spirv-dis executable name.

About options on the options.  The one I use most commonly use is --raw-id to turn off the friendly-name generation.  

Maybe the nicest way to map parameters is something like this:
   --spirv-dis             # the usual case


Optional string with comma-separated options matching spirv-dis executable options

   --spirv-dis=raw-id,no-indent

dneto0:
About CMake options.   I think it's fine to keep life simple with a single ENABLE_OPT option.

The disassembler is in the core SPIRV-Tools library of about 1.7MB.  The optimizer code is in an extra library of 4.4MB that requires the first).    But I assume you'll only use the SPIRV-Tools disassembler in the standalone executable? So code size isn't much of an issue?

johnkslang:
I added handling of the diagnostic.

PR #1422, opened 2018-06-29
Constant.cpp Floating point divide by zero

HaydnTrigg:
Constant.cpp will throw a floating point divide by zero if floating point exceptions are enabled in Win32 causing the program to crash. This fix manually checks the right-hand argument of the division and sets appropriate Infinity, Negative Infinity, or NAN as if the floating point exceptions were disabled.


johnkslang:
Following the local convention, this would take 8 lines of code instead 21 (mostly empty) lines of code.

Fixing locally.

johnkslang:
This is also missing tests. Adding.

PR #1438, opened 2018-07-10
Reflection exposes the Shader Stages where a Uniform is present

de2nis:
This pull request adds the exposition of the shader stages on uniforms reflection.
This approach seems the only way to get this information.
Is there any other possible implementation?


johnkslang:
Are you using this in a mode of compiling multiple stages simultaneously (or just validating)?

de2nis:
Every stage is compiled separately, but the reflection building is done after the program linking. At this point no such information is available.

PS: Travis CI is facing an issue at Android compilation on Linux

PR #1442, opened 2018-07-12
NDK build: Use NDK r17b and its own CMake toolchain file

dneto0:
More recent NDK releases have their own CMake toolchain file.
Use it.

Also, download the NDK from github.com:dneto0/android-ndk.
That is a fork of the repo we used to use, but we have more
control over how long it stays stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,20 +48,16 @@ install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$CC" == "clang" ]]; then
       export CC=clang-3.6 CXX=clang++-3.6;
     fi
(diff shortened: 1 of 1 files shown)


dneto0:
The urho3d repo is deprecated.  I've forked it so we can keep its contents around for as long as we like.

I also looked into using Travis-CI's native support for the Android SDK.  That is supported via "language: android".  See https://docs.travis-ci.com/user/languages/android/ 
You can configure it to install "ndk-bundle" which is a full installation of the NDK.

However that language tag would appear to conflict with the existing "language: cpp" setting in .travis.yml.

dneto0:
The GCC bot failure is an infra problem: can't resolve github.com (!)

PR #1443, opened 2018-07-12
Fix warnings: Remove name of unused parameter.

dneto0:


johnkslang:
FYI, that is quite an information loss... the string should have been passed through to the error message, giving more detailed error messages.

Before I saw this, I independently fixed this in e7f9caeac4ba65c6e64c50ca3647e9e31923f3db.

PR #1444, opened 2018-07-12
Add missing OpExtension SPV_EXT_descriptor_indexing for dynamic indexing capabilities

jeffbolznv:


johnkslang:
Why doesn't this change test results?  (I merged it locally to check.)

jeffbolznv:
There aren't any glslang tests that do dynamic indexing of these slightly less common descriptor types. This arose in a new test I was writing, when the spirv-tools validator complained that they were missing,

PR #1465, opened 2018-08-06
Remove execute permissions

otakuto:


johnkslang:
I see these appear pointless, but what is the problem being caused? Do we know root cause of having execute permission on test results? My main concern is to avoid a constant battle of going back and forth.

otakuto:
> Do we know root cause of having execute permission on test results?
Sorry, I don't know. But other files (e.g., .gitattributes, CMakeLists.txt) doesn't need execute permission.

PR #1466, opened 2018-08-08
Fix -Wignored-qualifier and -Wunused-variable warnings

Kangz:
PTAL @johnkslang


SPIRV/GlslangToSpv.cpp
```diff
@@ -5385,6 +5385,9 @@ spv::Id TGlslangToSpvTraverser::createInvocationsOperation(glslang::TOperator op
 {
     bool isUnsigned = isTypeUnsignedInt(typeProxy);
     bool isFloat = isTypeFloat(typeProxy);
+    // These variables are unused when !defined(AMD_EXTENSIONS)
+    (void) isUnsigned;
```
johnkslang: Wouldn't using #ifdef be better ?

Kangz: Done

PR #1472, opened 2018-08-10
Don’t print the source filename if the log is empty

bpeel:
VkRunner invokes glslangValidator as an external process to generate SPIR-V from the GLSL shaders. Whenever it does this even if there are no errors glslang needlessly prints the filenames of the temporary files used for the GLSL shaders like this:

```
$ vkrunner indices.shader_test 
/tmp/vkrunner-8nEFZP
/tmp/vkrunner-8LTVsa
PIGLIT: {"result": "pass" }
```

I imagine this could be annoying for other uses too such as using glslangValidator in the build scripts for a project because the output just clutters the build log.


johnkslang:
Closing, as #2314 addressed this with a `--quiet` option.

PR #1474, opened 2018-08-15
SPV: only declare the pure 8/16-bit capabilities when needed.

johnkslang:
Only when operations stray outside the 8/16-bit storage-capabilities
are the general (pure 8/16-bit) capabilities needed.

### DIFF
--- a/SPIRV/SpvPostProcess.cpp
+++ b/SPIRV/SpvPostProcess.cpp
@@ -61,8 +61,78 @@ namespace spv {
 
 namespace spv {
 
+// Hook to visit each operand type and result type of an instruction.
+// Will be called multiple times for one instruction, once for each typed
+// operand and the result.
+void Builder::postProcessType(const Instruction& inst, Id typeId)
+{
+    // Characterize the type being questioned
+    Id basicTypeOp = getMostBasicTypeClass(typeId);
+    int width = 0;
+    if (basicTypeOp == OpTypeFloat || basicTypeOp == OpTypeInt)
+        width = getScalarTypeWidth(typeId);
+
+    // Do opcode-specific checks
+    switch (inst.getOpCode()) {
+    case OpLoad:
+    case OpStore:
+        if (basicTypeOp == OpTypeStruct) {
+            if (containsType(typeId, OpTypeInt, 8))
+                addCapability(CapabilityInt8);
+            if (containsType(typeId, OpTypeInt, 16))
+                addCapability(CapabilityInt16);
+            if (containsType(typeId, OpTypeFloat, 16))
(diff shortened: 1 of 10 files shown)


SPIRV/SpvPostProcess.cpp
```diff
@@ -61,8 +61,82 @@ namespace spv {
...
+            }
+        }
+        break;
+    case OpAccessChain:
+    case OpPtrAccessChain:
+    case OpCopyObject:
+    case OpConvertFToU:
+    case OpConvertFToS:
+    case OpConvertSToF:
+    case OpConvertUToF:
```
jeffbolznv: I don't think FTo{U,S} or {S,U}ToF are supported, just {F,S,U}Convert.

johnkslang: Right, I discovered this testing. I think the extension is wrong. I uncovered quite a few problems with the extension, which I go into deeper in GitLab.

For this one, I think it is perfectly reasonable to do the following, and it generates one of the additional instructions I subtracted:

```
#version 460
#extension GL_KHX_shader_explicit_arithmetic_types_int8: require

struct S1 {
    bool b;
    int8_t i8;
};

struct S2 {
    int i;
    S1 s1;
};

buffer ub {
    S2 s2;
} buff;

void main()
{
    float f = buff.s2.s1.i8;
}
```

johnkslang: My example is meant as pseudo SPIR-V, don't get hung up on the #extension, etc.

The intention is that a SPIR-V shader that does that only needs the int8-storage capability, not the int8 capability.

If it was the only added instruction, maybe I shouldn't do it, but there were many instructions I needed to add to make things actually work.

jeffbolznv: For better or worse, the SPV extension has always been clear that the width-only conversions (F,S,U) are supported and the others are not. I think it's much too late to change that. GL_EXT_shader_16bit_storage avoids needing these instructions by forbidding implicit conversions and limiting the constructors that are supported for these types. So I don't think there is any actual demand for them coming from GLSL.

johnkslang: Okay, I can live with that.

I think OpAccessChain will be another story.

jeffbolznv: I agree that the spec is unclear/misleading about access chains' usage of these pointer types (and other things...). But there is CTS test coverage that uses them, so it is OK to clarify the spec to explicitly allow those.

johnkslang: Extra conversions are now removed.

PR #1476, opened 2018-08-17
Update SPIRV-Tools known good.

greg-lunarg:
Includes:

    Update OpPhi instructions after splitting block. (#1783)
    Don't change decorations and names in merge return. (#1777)
    Transform to combine consecutive access chains
    Handle undef literal value in vector shuffle
    Fix block ordering in dead branch elim
    Fix finding constant with particular type. (#1724)
    Fix infinite loop while folding OpVectorShuffle (#1722)
    Fix size error when folding vector shuffle. (#1721)
    Layout validation: Permit {vec3; float} tight packing

### DIFF
--- a/known_good.json
+++ b/known_good.json
@@ -5,7 +5,7 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit" : "9ecbcf5fc87db00d3d6275522c735b5667007647"
+      "commit" : "714bf84e58abd9573488fc365707fb8f288ca73c"
     },
     {
       "name" : "spirv-tools/external/spirv-headers",


johnkslang:
> Layout validation: Permit {vec3; float} tight packing

How does this one get triggered?

I can see the others as being part of normal legalization/optimization, but how is validation getting used?

johnkslang:
Are we missing the test cases to verify the validator functionality is working correctly?  These should be added, as part of the PR.

greg-lunarg:
Actually, I don't think this gets triggered directly from glslang.

When spirv-opt is called via command line, it runs spirv-val first, but this does not happen when spirv-opt is called via the library path from glslang.

However, the SDK uses the same version of SPIRV-Tools that glslang uses, so in order to "fix" the SDK I need to update glslang's SPIRV-Tools known good.

greg-lunarg:
> Are we missing the test cases to verify the validator functionality is working correctly? These should be added, as part of the PR.

As I mentioned in the previous comment, the validator is not called when spirv-opt is invoked from glslang. So a test case in glslang is probably not warranted. And David added a test in SPIRV-Tools when he fixed the problem there. So that should be sufficient.

johnkslang:
So, currently, that aspect of that file is effectively part of a particular SDK, rather than getting tested as part of glslang proper.

We have a side script to validate (ad hoc) generated SPIR-V, which is not the best approach. However, with the validator code living in glslang, it could be used instead, verifying it works correctly for all consumers of glslang, and preventing accidental updates to this file that include a broken SPIR-V validator.

PR #1480, opened 2018-08-20
Use our own SPIRV-Tools message stringifier

### LINKED ISSUE #1477: Compile error: message.h file not found
SPIRV-Tools remove the 'message.h' file recently
https://github.com/KhronosGroup/SPIRV-Tools/pull/1838

So the compile failed with "SPIRV/GlslangToSpv.cpp:59:14: fatal error: message.h: No such file or directory"

dneto0:
Stop including an internal SPIRV-Tools header. It has been
deleted.

Fixes #1477


johnkslang:
Thanks @dneto0.

I still suspect (as in #1477) something is odd about needing this, but it looks more correct than including the internal header.

PR #1482, opened 2018-08-22
SPV: Add SPIRV-Tools validator.

johnkslang:


antiagainst:
Awesome stuff!

johnkslang:
Is it possible that the validator is not deterministic? Or uses a data structure that is not necessarily in the same order on all platforms?  Apple builds see this difference relative to the others:

```
-error: Input variable id <20> is used by entry point 'main' id <4>, but is not listed as an interface
-  %i_1 = OpVariable %_ptr_Input_ps_in Input
+error: Output variable id <23> is used by entry point 'main' id <4>, but is not listed as an interface
+  %_entryPointOutput = OpVariable %_ptr_Output_ps_out Output
```

johnkslang:
I see use of `unordered_set` and `unordered_map` in the tools.  I haven't dug any deeper yet, but that could be a source of platform-specific behavior in what order things are seen/found/etc.

SPIRV/GlslangToSpv.cpp
```diff
@@ -7097,6 +7140,8 @@ void GlslangToSpv(const glslang::TIntermediate& intermediate, std::vector<unsign
     it.dumpSpv(spirv);
 
 #if ENABLE_OPT
+    SpirvToolsValidate(intermediate, spirv, logger);
```
dneto0: Ah.  I didn't expect you to add it to the main flow, i.e. always-on.   I was requesting that the *test* flows turn on validation when SPIRV-Tools was available.
I'd be happy/more comfortable if this were guarded by a non-default option.

antiagainst: FWIW, we run validation **by default** in DXC. It might be painful sometimes, but I view it as a nice way to cross-check and improve tooling for a greater good. When things break, people will notice immediately instead of needing to debug and then find out, and we can have a bug report for it quicker. Gradually we'll see less and less breaks.

But since DXC has much a smaller user base and younger than Glslang, I cannot see the same for Glslang. :)

antiagainst: sorry, s/see/say/

johnkslang: Agreed. I'm settling how to switch it in parallel with sorting out other aspects; one reason it is WIP.

It can't be on by default, due to the presence of false positives (which are not necessarily bugs in the validator; it's sometimes more complicated than that).

johnkslang: I didn't see @antiagainst's comment when I replied, only @dneto0's, which my reply was to.

Replying to @antiagainst, there are other points to keep in mind:

1. This is not a correctness test, it is a regression test, meaning it's job is to flag up changes, not to enforce correctness. It is actually nice to have failures included in the results, for a variety of reasons.
2. Currently, some of the best ways to test some things includes using tests that are not conforming to all the specifications, meaning there will be validation failures, and we might want to keep the test the way it is because that's how it tests something important.
3. There might be something we want to relax somewhere, and seeing these failures is a reminder of that.

antiagainst: Yep. Agreed. ;-P

johnkslang: This is now called optionally, off by default, but always requested by the google-test suite.

For the command line, it uses `--spirv-val`, in parallel with the syntax for the SPIRV-Tools disassembler being `--spirv-dis`.

johnkslang:
@greg-lunarg this is an important quality improvement, but requires updating the known good. Do you have any concerns about this?

greg-lunarg:
I usually update known good to a "stable" commit ie the latest commit on the SPIRV-Tools "stable" branch. The latest is 87f363, which just got posted and I haven't updated glslang to it yet. If that commit is good for you I have no concerns. If you need a commit beyond that, it might be good to get Steven Perron to put forward a new "canary" for Lei and I to test (I for one have some trace-based testing that is difficult to automate).

johnkslang:
SPIRV-Tools stable now has the required commit, so I can merge now.

PR #1489, opened 2018-09-07
GL_KHR_memory_scope_semantics

jeffbolznv:
Implementation of https://github.com/KhronosGroup/GLSL/pull/23.


johnkslang:
Two questions about this:

1. It updates the SPIR-V header. Ideally, we'd do a commit now that brings in the new header from SPIR-V Headers repository, and then this PR would go on top of it, without changing the header. This would be to avoid any accidental header conflicts, especially numerical value differences, from the two different paths taken to get to the header.  How does this header compare?  Should we go directly with this MR, or first commit the latest headers?  (I can do the latter, and resolve the resulting conflicts.)

2. Is there a short description of what the change to known_good for SPIRV-Tools does?  Is it to something more recent, like the latest one, which also supports this new functionality, or ...?

Thanks.

jeffbolznv:
It would be fine to split out the SPIR-V header change into a separate PR. Do you want me to do that or is it easier to do it yourself?

Right, the updates to SPIRV-{Headers, Tools} are to update to the most recent commits that have SPV_KHR_vulkan_memory_model support.

johnkslang:
I verified the header differences are just line ending differences (Windows vs. Linux: this MR is mixing them) in the lines:

```
     ScopeQueueFamilyKHR = 5,
     CapabilityVulkanMemoryModelDeviceScopeKHR = 5346,
```

Fixing locally.

PR #1497, opened 2018-09-19
Implement GL_EXT_subgroupuniform_qualifier.

sheredom:
Implement the extension which introduces a subgroupuniformEXT qualifier to allow the generation of the Uniform decoration in SPIR-V.


sheredom:
Extension specification https://github.com/KhronosGroup/GLSL/pull/17

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3319,7 +3322,8 @@ void TParseContext::mergeQualifiers(const TSourceLoc& loc, TQualifier& dst, cons
     MERGE_SINGLETON(readonly);
     MERGE_SINGLETON(writeonly);
     MERGE_SINGLETON(specConstant);
-    MERGE_SINGLETON(nonUniform);
+#undef MERGE_SINGLETON
+    dst.propagateUniformity(src);
```
johnkslang: Just coming up to speed on this one:

The normal thing here is if either qualifier added a keyword, we say together, the keyword is present. I'm concerned that at this point in the code it should be a mechanical thing about what the shader said, versus a deductive thing with complex semantics (and if it is the latter, whether that aspect of declarations is explained in the specification).

Contradictory ones should show up in the checks a bit earlier in the function.

johnkslang: The alternative design I'm evaluating is to have two bits, one for each of the keywords. Here, each would be independently merged with `MERGE_SINGLETON`. A bit earlier in this function, would be the semantic checks regarding whether they contradict each other, or contradict other qualifiers like **uniform**, etc.

Then, for the rest of the code, I'm thinking the abstraction used is good, and hoping the bottom level function can be reimplemented in terms of the two bits instead of the enum.

johnkslang: Or, the enum is a separate thing, layered on top of all uniform-related keywords.

sheredom: Have you had any other brain cycles on this? We discussed at the F2F last week, and if I'm understanding correctly, I need to check that I'm not putting two conflicting qualifiers on the same thing (which is what this MERGE_SINGLETON did previously).

johnkslang: Yes, I meant to capture that in https://github.com/KhronosGroup/glslang/pull/1497#discussion_r219163216, which I posted at that time, but recapping more completely:

- have one bit per keyword, then MERGE_SINGLETON does its job for that, and it is never lost how the user expressed the syntax
- about from the function start to 40 lines earlier, do any semantic checking regarding mixing of keywords
- change/complete the logic of the encapsulating functions (encapsulation interface looks good) to be based on the keyword bits (probably remove the new enum), assuming that the semantic checking weeded out combinations you don't want to consider

The key point that might not be obvious is that if there is a declaration like this:

```
uniform nonuniform subgroup uniform in out vec4 v;
```

It is this function we are discussing (`mergeQualifiers`) that is called to combine those qualifiers into a single type. The keywords are recognized one at a time, and then merged together.  So, for example, `uniform in` is a bad combination, and that is caught in this function.

When you do the semantic checking, you can sanitize the keyword-bit state, if that helps future code not blow up. But, if they were inconsistent, by default parsing will stop (a testing option lets it continue, which is why I protect against crashes for the normal error cases).

sheredom: I added a check (not based on MERGE_SINGLETON) that will error on the subgroupuniform/nonuniform case you described in mergeQualifiers https://github.com/KhronosGroup/glslang/pull/1497/files#diff-b8c32f56c79e0d9c624d0b69bc6a822eR33

glslang/Include/Types.h
```diff
@@ -526,7 +532,7 @@ class TQualifier {
     bool readonly     : 1;
     bool writeonly    : 1;
     bool specConstant : 1;  // having a constant_id is not sufficient: expressions have no id, but are still specConstant
-    bool nonUniform   : 1;
+    TUniformity uniformity : 3;
```
johnkslang: This should not be tristate. It is two binary states, even though one seems invalid. Then, you get to do a lot of things exactly they way they are already done for whether or not a keyword was present.  Note this is the design throughout, and shows up, for example, in that we know whether a keyword was present even if it was the default, and we know in general if contradictory sets of keywords were present.

glslang/Include/Types.h
```diff
@@ -926,7 +932,10 @@ class TQualifier {
     }
     bool isNonUniform() const
     {
-        return nonUniform;
+        return EUniformityIsNonUniform == uniformity;
```
johnkslang: Here, you can keep the desired encapsulation, but rewrite the internal to be based on the combination of two binary states. You can assume semantic checking culled the invalid one when writing these functions.

hlsl/hlslParseHelper.cpp
```diff
@@ -6518,7 +6518,8 @@ void HlslParseContext::mergeQualifiers(TQualifier& dst, const TQualifier& src)
     MERGE_SINGLETON(readonly);
     MERGE_SINGLETON(writeonly);
     MERGE_SINGLETON(specConstant);
-    MERGE_SINGLETON(nonUniform);
+#undef MERGE_SINGLETON
```
johnkslang: Are you adding this to the HLSL path? Well, same comment as for the GLSL path, your right it's probably safest to have it here too.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3351,7 +3360,8 @@ void TParseContext::mergeQualifiers(const TSourceLoc& loc, TQualifier& dst, cons
     MERGE_SINGLETON(readonly);
     MERGE_SINGLETON(writeonly);
     MERGE_SINGLETON(specConstant);
-    MERGE_SINGLETON(nonUniform);
+#undef MERGE_SINGLETON
```
johnkslang: Here, you get to do things the same way it is done for all the others.

PR #1505, opened 2018-09-24
Add BUILD.gn configuration.

dj2:
This CL adds the necessary configuration to build glslang inside a
Chromium checkout. Two build warnings were fixed in the process to
make things compile.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,157 @@
+# Copyright (C) 2018 Google, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#    Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#
+#    Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#
+#    Neither the name of Google Inc. nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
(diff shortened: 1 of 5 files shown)


dj2:
@johnkslang @Kangz

BUILD.gn
```diff
@@ -0,0 +1,157 @@
...
+
+import("//build_overrides/glslang.gni")
+
+spirv_tools_dir = glslang_spirv_tools_dir
+
+config("glslang_public") {
+  include_dirs = [ "." ]
+}
+
+static_library("glslang_static") {
```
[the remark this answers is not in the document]
dj2: Done.

SPIRV/SpvBuilder.cpp
```diff
@@ -631,7 +631,7 @@ Id Builder::getContainedTypeId(Id typeId) const
 // Returns true if 'typeId' is or contains a scalar type declared with 'typeOp'
 // of width 'width'. The 'width' is only consumed for int and float types.
 // Returns false otherwise.
-bool Builder::containsType(Id typeId, spv::Op typeOp, int width) const
+bool Builder::containsType(Id typeId, spv::Op typeOp, uint32_t width) const
```
johnkslang: Despite what the purists might desire, `int` and `unsigned` were still more portable than `uint32_t` when I looked into it a while back. Is the issue that this needs to be unsigned? And that's tied to adding the BUILD.gn file?

dj2: The build issued a warning (which became an error) about the signed vs unsigned comparison that's happening. I just made them both unsigned in the simplest way I could, the inputs all appeared to be 16 or 32.

BUILD.gn
```diff
@@ -0,0 +1,162 @@
...
+    "glslang/MachineIndependent/preprocessor/PpContext.h",
+    "glslang/MachineIndependent/preprocessor/PpScanner.cpp",
+    "glslang/MachineIndependent/preprocessor/PpTokens.cpp",
+    "glslang/MachineIndependent/preprocessor/PpTokens.h",
+    "glslang/MachineIndependent/propagateNoContraction.cpp",
+    "glslang/MachineIndependent/propagateNoContraction.h",
+    "glslang/MachineIndependent/reflection.cpp",
+    "glslang/MachineIndependent/reflection.h",
+    "glslang/OSDependent/osinclude.h",
+    "glslang/Public/ShaderLang.h",
```
johnkslang: What about HLSL?

johnkslang:
Is anything needed to indicate compiling with Nvidia/AMD extensions and HLSL?

BUILD.gn
dj2: This was copied from Dawn which was the minimum it needed. There is more stuff in the Chrome build file which I can also pull into here if we want it to be 'complete'?

dj2:
As discussed elsewhere, we'll handle Nvidia/AMD extensions and HLSL as needed.

SPIRV/SpvBuilder.cpp
dj2: Did you want this changed to unsigned int? Do we support platforms where uint32_t isn't available?

johnkslang: Either `unsigned` or `unsigned int`, thanks. Both work and are maximally portable.

Glslang has been embedded in a wide range of platforms, not just as an offline tool. It is exercised in a wider range of environments than the bots check. I won't intentionally limit its portability over small things.

dj2: done

PR #1507, opened 2018-09-25
Add ES 320 support and additional error checks for SPV_NV_mesh_shader

sparmarNV:
- Add ES 320 support
- Error out use of perprimitiveNV for non mesh/fragment shaders
- Error out use of mesh/task shaders w/o use of NV_mesh_shader
- Error out use of NV_mesh_shader for non task/mesh shaders
- Error out use of perviewNV for non mesh shaders
- Error out use of taskNV for non mesh/task shaders
- Add test case for mesh shader with ES 320 profile


sparmarNV:
Adding @dgkoch @johnkslang for visibility

Test/spv.meshShaderUserDefined.mesh
```diff
@@ -1,4 +1,4 @@
-#version 450
+#version 320 es
```
dgkoch: Should we have a separate test shader for ES and GL profiles?

sparmarNV: We already have 6 other tests with GL profiles. So we should be fine.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4983,7 +4998,7 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
         if (id.compare(0, 11, "local_size_") == 0) {
 #ifdef NV_EXTENSIONS
             if (language == EShLangMeshNV || language == EShLangTaskNV) {
-                profileRequires(loc, ~EEsProfile, 450, E_GL_NV_mesh_shader, "gl_WorkGroupSize");
+                requireExtensions(loc, 1, &E_GL_NV_mesh_shader, "gl_WorkGroupSize");
```
johnkslang: This looks like it dropped the 450 requirement(?).

You can call profileRequires() multiple times, once for each set of requirements:

```
profileRequires(loc, ~EEsProfile, ...);
profileRequires(loc, EEsProfile, ...);
```

glslang/MachineIndependent/parseVersions.h
```diff
@@ -103,6 +103,9 @@ class TParseVersions {
     virtual void requireSpv(const TSourceLoc&, const char* op);
     virtual bool checkExtensionsRequested(const TSourceLoc&, int numExtensions, const char* const extensions[], const char* featureDesc);
     virtual void updateExtensionBehavior(const char* const extension, TExtensionBehavior);
+#ifdef NV_EXTENSIONS
```
johnkslang: I'm taking these ifdef out of this header.

glslang/MachineIndependent/ParseHelper.cpp
sparmarNV: I added requireExtensions check which will ensure we have the extension.
The extension in turn has the profile GL450+/ES320+ requirement.
So I removed the redundant profileRequires() check from here.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5350,15 +5341,28 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "in highp uvec3 gl_GlobalInvocationID;"
             "in highp uint gl_LocalInvocationIndex;"
 
+            "\n");
+    }
+
+    if (profile != EEsProfile && version >= 450) {
```
johnkslang: The interface here is changing; if the test @dgkoch pointed out could easily be duplicated, it should be, unless you are sure all these interfaces changes are tested in the other files.

sparmarNV: Okay, I will add another es 320 test.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3600,6 +3600,13 @@ void TParseContext::arraySizesCheck(const TSourceLoc& loc, const TQualifier& qua
                 extensionsTurnedOn(Num_AEP_tessellation_shader, AEP_tessellation_shader))
                 return;
         break;
+#ifdef NV_EXTENSIONS
+    case EShLangMeshNV:
+        if (qualifier.storage == EvqVaryingOut)
+            if ((profile == EEsProfile && version >= 320) ||
+                extensionTurnedOn(E_GL_NV_mesh_shader))
+                return;
+#endif
```
johnkslang: Shouldn't this case be fully self contained, and not rely on a fall through?

glslang/MachineIndependent/Versions.cpp
```diff
@@ -834,6 +834,23 @@ void TParseVersions::updateExtensionBehavior(const char* extension, TExtensionBe
     }
 }
 
+#ifdef NV_EXTENSIONS
+// Validate if extension name is used with correct shader stage.
+bool TParseVersions::validateExtensionName(const TSourceLoc& loc, const char * const extension)
```
johnkslang: I don't think I agree with adding this function; it does not know what is going on.

There is a requireStage() function, to call from the place that does know what is going on. Does that work?

Or, maybe you want to call this one with a more specific name that says what it actually does, like check mesh stages.

glslang/MachineIndependent/parseVersions.h
johnkslang: I am removing these #ifdefs from this header.

glslang/MachineIndependent/ParseHelper.cpp
sparmarNV: Yes, I forgot the break statement. Thanks for the catch.

glslang/MachineIndependent/Versions.cpp
sparmarNV: Okay, I will inline the error check in Pp.cpp

glslang/MachineIndependent/parseVersions.h
sparmarNV: Okay.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: Ah, that means it was a bug before(?), allowing 450 to support it without an extension, at least locally. If an extension must be present, the 450 should be a 0.

I think things are slightly flexible right when a brand new stage comes in, due to an extension, because their use is 100% correlated.

sparmarNV: Yes, this was a bug before. I thought profileRequires would mandates both 450 && NV_mesh_shader. But since that is not the case, I had to change it to requireExtensions.

johnkslang:
I think this failed from testing infrastructure problems, not glslang itself.

sparmarNV:
> I think this failed from testing infrastructure problems, not glslang itself.

Is it possible to rerun the tests?

glslang/MachineIndependent/preprocessor/Pp.cpp
```diff
@@ -874,6 +874,11 @@ int TPpContext::CPPextension(TPpToken* ppToken)
         return token;
     }
 
+#ifdef NV_EXTENSIONS
+    if (!parseContext.checkShaderStageForNVExtensions(ppToken->loc, extensionName))
```
johnkslang: Incorrect extension semantics are not preprocessor errors.

johnkslang: Maybe you want to do something in 

```
void TParseVersions::updateExtensionBehavior(int line, const char* extension, const char* behaviorString)
```

glslang/MachineIndependent/Versions.cpp
```diff
@@ -834,6 +834,25 @@ void TParseVersions::updateExtensionBehavior(const char* extension, TExtensionBe
     }
 }
 
+#ifdef NV_EXTENSIONS
+// Check if extension is used with correct shader stage.
+bool TParseVersions::checkShaderStageForNVExtensions(const TSourceLoc& loc, const char * const extension)
```
johnkslang: Do we even need #extensions to turn on a stage that can only exist if the #extension is there? Or is this incremental functionality? The system already has to buy into the fact that the stage is present to even get this far, it seems a bit circular.

johnkslang: See also the stage logic recently added for these stages in ShaderLang.cpp around line 568:

```
    // Correct for stage type...
    switch (stage) {
```

johnkslang:
I think the preprocessor stuff needs fixing anyway; I don't think anything here should be changing the preprocessor.  GLSL proper has all the semantic helpers needed; per my recent comments.

glslang/MachineIndependent/Versions.cpp
sparmarNV: This is to error out the case where an extension is incorrectly used in some other shader stage.
For e.g. using GL_ARB_compute_shader extension inside a non-compute shader.
I believe glslang doesn't error out such use cases.
I agree I should move the error check in updateExtensionBehavior( ).

johnkslang: Okay, I understand. I keep thinking it will disappear. But, it's still here, so more issues:

- it should be not be called an NV function
- it should not do odd checks for error reporting; it can allow errors to continue
- basically, it only needs to be the one if-test and call to requireStage
- (implying this is a void, not bool)

It would be fine with me if that one statement was inlined and the rest disappeared. But, to keep the function is also okay, for all checks like this, in which case it won't be NV-specific and should only do the checking.

johnkslang: Something like:

```
void TParseVersions::checkShaderStageForNVExtensions(const TSourceLoc& loc, const char * const extension)
{
    // GL_NV_mesh_shader extension is only allowed in task/mesh shaders
    if (strcmp(extension, "GL_NV_mesh_shader") == 0)
        requireStage(loc, (EShLanguageMask)(EShLangTaskNVMask | EShLangMeshNVMask),
                     "#extension GL_NV_mesh_shader");
}
```

johnkslang: More background: there is already a generic mechanism for stopping parsing around the first error (unless cascading errors are requested), so individual error checks do not need to worry about termination, unless the next lines of code would crash glslang.

sparmarNV: Thanks for the detailed feedback.
I have done the changes to the function as suggested above.
Only thing which I added was the #ifdef NV_EXTENSIONS in function body, because the underlying code used EShLang*NVMask which is also guarded by same #ifdef.

sparmarNV:
Also let me know if you want me to squash all commits into 1.

PR #1531, opened 2018-10-16
Fix flags.nonprivate to be true for workgroup memory, which is implicitly workgroupcoherent/nonprivate

jeffbolznv:
This was found by updated spirv-tools validation running against Vulkan CTS.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -554,11 +554,11 @@ spv::Builder::AccessChain::CoherentFlags TGlslangToSpvTraverser::TranslateCohere
     flags.subgroupcoherent = type.getQualifier().subgroupcoherent;
     // *coherent variables are implicitly nonprivate in GLSL
     flags.nonprivate = type.getQualifier().nonprivate ||
-                       type.getQualifier().subgroupcoherent ||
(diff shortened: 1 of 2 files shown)


jeffbolznv:
cc @dneto0

SPIRV/GlslangToSpv.cpp
```diff
@@ -554,11 +554,11 @@ spv::Builder::AccessChain::CoherentFlags TGlslangToSpvTraverser::TranslateCohere
     flags.subgroupcoherent = type.getQualifier().subgroupcoherent;
     // *coherent variables are implicitly nonprivate in GLSL
     flags.nonprivate = type.getQualifier().nonprivate ||
-                       type.getQualifier().subgroupcoherent ||
-                       type.getQualifier().workgroupcoherent ||
-                       type.getQualifier().queuefamilycoherent ||
-                       type.getQualifier().devicecoherent ||
-                       type.getQualifier().coherent;
+                       flags.subgroupcoherent ||
+                       flags.workgroupcoherent ||
```
jeffbolznv: github hides it by default, but the reason to use flags.workgroupcoherent is at line 553, where it includes shared memory which is implicitly workgroupcoherent.

dneto0:
Yay. Thanks.

PR #1544, opened 2018-10-24
ParseHelper: assign global XfbBuffer to a block missing it

### LINKED ISSUE #1535: xfb offset overlapping checks fails when combining block and per-member xfb_offset qualifier
So the following test.vert file:

```
#version 450

layout(location=5, xfb_offset=0) out block2 {
   float y1_out;
   layout(xfb_offset = 16) vec4 y2_out;
};

void main() {
   y1_out = 7.0;
   y2_out = vec4(1.0, 0.0, 0.0, 1.0);
   gl_Position = vec4(0.0);
}
```
Will fail with the following error:

> $ glslangValidator -G test-member-offset.vert -H
> test-member-offset.vert
> ERROR: test-member-offset.vert:3: 'xfb_offset' : overlapping offsets at offset 16 in buffer 0
> ERROR: 1 compilation errors.  No code generated.

After checking a little the code, this assert is triggered at ParseHelper.cpp, line 5329, method TParseContext::layoutTypeCheck

What it is happening is that this method adds the range defined by the type/offset on its xfb_buffer as it iterates on the type. So it first add the range for y2_out ([16, 16+size(vec4)]). And then the method is called for the full block2, and it throws an error because part of the range needed for the block2 [0, size(block)] is already marked as used.

So there are two problems here:
1. When checking if there is an overlap, it doesn't differentiate members and it containers
2. TIntermediate::computeTypeXfbSize doesn't take into account that one of the members could have a custom offset. So it returns the minimum size needed for the block without custom offsets.

FWIW, when I was checking if fixing by myself those issues were assumable, I started to find how the xfb offsets are dealt not really consistent with other qualifiers. And I think that it is partly due xfb being underspecified on the spec. Specifically with output blocks. So using ARB_gl_spirv spec as reference, quoting:

>     "If the shader used to record output variables for transform feedback
>     varyings uses the xfb_buffer, xfb_offset, or xfb_stride layout qualifiers,
>     or its SPIR-V entry point declares the *Xfb* Execution Mode, the values
>     specified by TransformFeedbackVaryings are ignored, and the set of
>     variables captured for transform feedback is instead derived from the
>     specified layout qualifiers or SPIR-V decorations: outputs specifying both
>     an *XfbBuffer* and an *Offset* are captured, while outputs not specifying
>     both of these are not captured. Values are captured each time the shader
>     writes to such a decorated object."

Note that the last sentence, points a "decorated object", so that could mean a individual variable or a full block. As nothing else is specified, we would need to rely on GLSL spec. So from GLSL 4.60 spec, section "4.4.2 Output Layout Qualifiers", subsection "Transform Feedback Layout Qualifiers":

> If a block is qualified with xfb_offset, all its members are assigned transform feedback buffer offsets. If a block is not qualified with xfb_offset, any members of that block not qualified with an xfb_offset will not be assigned transform feedback buffer offsets. Only variables and block members that are assigned offsets will be captured (thus, a proper subset of a block can be captured).

Right now glslang is just adding on the SPIR-V the xfb offset (just Offset) if it is sets explicitly on the GLSL shader (so on the example used on top, 0 for the block, 16 for y2_out, no offset for y1_out). And from the spec, that's correct. But this seems somewhat inconsistent to what happens with other qualifiers, like bindings, location, offset (on a ubo/ssbo) etc, were it is expected that SPIR-V would include a explicit value for all variables. Even if on GLSL the driver has rules to automatically set all those if the explicit values are missing, when the driver consumes SPIR-V it is expected to being already provided.

In other words, if a explicit xfb_offset is set for a block, I would expect that the SPIR-V generated included a xfb_offset for each member. But I would also find inconsistent to force glslang to be able to assign automatically those offsets. So, on the case of a block being assigned a xfb offset, I would expect the following behaviour:

1. Raise an error if any of the members lack a xfb_offset. Something like "Setting xfb_offset to a block mark all members to be captured. All members needs a xfb_offset"
2. Optionally add a command line option so glslang would automatically assign it (equivalent to aml, something like --auto-map-xfb-offset)

Having said so, all this is a kind of my opinion based on how this work with other qualifiers, with the feeling that this should be more clear on spec. So perhaps it would be needed to open a spec issue first?

infapi00:
If a block has assigned a XfbOffset it is assumed that it would
inherit the current global XfbBuffer. This commit fixes two use cases:

1) Getting the members of a Block with a XfbOffset to be assigned an
   offset, as explained on GLSL 4.60 spec, section "4.4.2 Output
   Layout Qualifiers", subsection "Transform Feedback Layout
   Qualifiers".

2) Compute properly an error on overlapping ranges if a block is
   assigned a XfbOffset and one of it members is assigned a explicit
   one. This gets working because when the members of a block get
   assigned a Offset/Buffer at fixBlockXfbOffsets, then the block is
   deassigned the Offsets, so ranges are computed only with the block
   members.

BTW, this is already done when redeclaring block builtins.

Fixes #1535


infapi00:
This pull request touches the same files that pull request #1547 is modifying. For simplicity, I will close this one until #1547 get resolved, and then send a new pull request.

PR #1547, opened 2018-10-25
Apinheiro/xfb struct assignment

infapi00:
As they are related, I kept the fix for issue #1535 on this pull request too. 

In relation to the new commit, it is true that it is somewhat counter-intuitive to use fixBlockXfbOffsets for a struct, but as mentioned on the commit message, that method just uses the qualifier and the member list, so it is independent of being a block or a struct. I was tempted to rename the method, but seemed like an overkill that will need too many changes. But it is not a strong opinion, so if you prefer to rename the method I can provide a new version of the patch.


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5685,6 +5685,24 @@ void TParseContext::fixOffset(const TSourceLoc& loc, TSymbol& symbol)
...
+    //    to d the next available offset aligned to a multiple of that
+    //    component’s size. Aggregate types are flattened down to the
+    //    component level to get this sequence of components."
+    //
+    // Here we are using the method fixBlockXfbOffsets. Although that
+    // one was initially intended for blocks, it is implemented based
+    // on qualifiers and ttypelists, so it should be working here.
+
+    if (symbol.getType().getBasicType() == EbtStruct &&
+        qualifier.hasXfbOffset()) {
```
johnkslang: This is already tested for inside `fixBlockXfbOffsets`.

I'm thinking this should be set up differently.  Like, just calling it directly from `declareVariable`.

Change `fixBlockXfbOffsets()` to `fixXfbOffsets()`, have it handle all types, and have that called from `declareVariable`.

infapi00:
> I'm thinking this should be set up differently. Like, just calling it directly from declareVariable.

I have just pushed a new version of the branch, that moves the call to declareVariable, plus the style fixes you mentioned before.

> Change fixBlockXfbOffsets() to fixXfbOffsets(), 

Done too.

> have it handle all types, and have that called from declareVariable.

Could you ellaborate "handle all types"? Not sure what I should do there. From the GLSL spec, it is mentioned that this assignment is done for aggregates. So we have structs (that this pull request cover) or arrays. But as far as I see, arrays are declared on the final shader as just one object. We can't request and set properties for individual array members. In fact, checking glslang code base, arrays are represented as TArraySizes, that quoting 

> "Represent an array, or array of arrays, to arbitrary depth.  This is not done through a hierarchy of types in a type tree"

Additionally, the GLSL spec doesn't allow to specify a custom xfb_offset for a given array element, making the xfb_offset easier (contiguous).

BUILD.gn
```diff
@@ -152,18 +152,3 @@ source_set("glslang_sources") {
     "${spirv_tools_dir}:spvtools_opt",
   ]
 }
-
```
johnkslang: I don't think this should be part of this PR.

Test/runtests
```diff
@@ -222,6 +222,17 @@ $EXE -D -E hlsl.pp.expand.frag > $TARGETDIR/hlsl.pp.expand.frag.out 2> $TARGETDI
 diff -b $BASEDIR/hlsl.pp.expand.frag.out $TARGETDIR/hlsl.pp.expand.frag.out || HASERROR=1
 diff -b $BASEDIR/hlsl.pp.expand.frag.err $TARGETDIR/hlsl.pp.expand.frag.err || HASERROR=1
 
+#
+# Testing xfb tests
+#
+echo "Testing XFB tests"
+$EXE -G -H spv.xfbOffsetOnBlockMembersAssignment.vert > $TARGETDIR/spv.xfbOffsetOnBlockMembersAssignment.vert.out
```
johnkslang: These all look like they would be much better as part of the googletests.  See `CompileOpenGLToSpirvTest` for OpenGL tests.

`runtests` should now be more for one-off complex testing of command-line options.

Can you transfer these to be in, for example, `Spv.FromFile.cpp`.

johnkslang:
Please make this just a single commit, which is only about the changes needed for this PR.

infapi00:
> Please make this just a single commit, which is only about the changes needed for this PR.

Done. And I also removed from this PR the commit that fixed issue 1535, that was originally included on PR #1544. As it touches the same files, for simplicity, I closed the latter. When this PR gets merged, I will send a new PR with an updated commit.

PR #1550, opened 2018-10-26
[HLSL] Fix for edge case, where vector truncation was not considered to find viable overloads

### LINKED ISSUE #1548: HLSL: overload matching with vector truncation not working
Hi,

we have this kind of code in our code base which compiles fine with FXC but fails with glslang:

```
float randPos(float3 p) { ... }
float rand(float p) { ... }

float2 ssgi_current_frame = ...
float3 worldPos = ...

float randVal = rand(randPos(worldPos)+sin(ssgi_current_frame*0.1317));
```

Error message is: 

> 'rand' : no matching overloaded function found

TiemoJung:
This fixes #1548


TiemoJung:
this is insufficient and needs a proper rework of implicit type cast rules.

PR #1557, opened 2018-10-30
Add names for composite spec constants in SPIR-V

nicebyte:
Consider the following code:

    layout(constant_id=0) const int Y = 1;
    layout(constant_id=1) const int Z = 2;
    layout(constant_id=3) const int X = Y + Z;

Previously, it would produce SPIR-V decorations like this:

    Decorate 21(Y) SpecId 1
    Decorate 22 SpecId 3
    Decorate 33(Z) SpecId 0

This seems inaccurate, since the spec constant `X` that is dependent on
the two others did not get a name in the SPIR-V decorations. This behavior
may potentially negatively affect shader introspection capabilities.

This change alters the behavior to always add a name, which results in the code
above producing the following decorations:

    Decorate 21(Y) SpecId 1
    Decorate 22(X) SpecId 3
    Decorate 33(Z) SpecId 0


nicebyte:
The tests are failing right now because names are being emitted even for constants that don't have a `layout` qualifier with `constant_id`, like `const bool i64_to_b = bool(si64);` (where `si64` is a specialization constant). If this behavior is acceptable, I can update the tests.

nicebyte:
All tests pass now, however it appears that SPIR-V validation tools think that a constant that is derived from other spec constants cannot be decorated with a `constant_id`.  See the following portion of the output:

```
error: SPIRV-Tools Validation Errors
error: OpDecorate SpecId decoration target <id> '1' is not a scalar specialization constant.
  OpDecorate %su64inc SpecId 105
```

Above, `su64inc` is decorated with a `constant_id`, and is a constant whose default value depends on another specialization constant. 

This ties back to the discussion here: https://github.com/KhronosGroup/SPIRV-Cross/pull/733#discussion_r229229653

Depending on the final decision on how the spec should be interpreted, either the validation tools need to be modified to accept this as valid SPIR-V, or glslang should start rejecting code that was shown in the commit message.

nicebyte:
Now that I think about it, the change doesn't affect just the constants decorated with `constant_id`, it affects any constant derived from spec constants. I.e. names would be generated for constants like:

`const int C = A+B;` 

where `A` and `B` are spec constants, for example.

Previously names for such constants were not available, which resulted in SPIRV-Cross emitting something along the lines of:

`const int _30 = ...`

With this change, I'm seeing actual names instead of IDs in SPIRV-Cross output, so it helps improve readability. I think the change should be considered for merging regardless of how the spec should be interpreted for `OpSpecConstantOp`

SPIRV/GlslangToSpv.cpp
```diff
@@ -7276,19 +7276,20 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
 
     // An AST node labelled as specialization constant should be a symbol node.
     // Its initializer should either be a sub tree with constant nodes, or a constant union array.
-    if (auto* sn = node.getAsSymbolNode()) {
-        if (auto* sub_tree = sn->getConstSubtree()) {
+    if (auto *sn = node.getAsSymbolNode()) {
```
johnkslang: Why are these lines changing?

I'd rather not use this PR as a forum for changing the coding convention of tying * and & to the type instead of the name, which is done throughout the entire code base.

johnkslang:
Agreed it looks like a nice improvement.  Address the unnecessary changes (still staying within one commit), and we can merge it.

SPIRV/GlslangToSpv.cpp
nicebyte: might have been IDE reformatting automatically

nicebyte:
I have fixed the pointer syntax

PR #1558, opened 2018-10-30
BUILD.gn: remove reference to non-existent header

Kangz:
PTAL @johnkslang (FYI @dj2).

Glslang was never build in Chrome so some sanity check for GN files were never run causing this to go unnoticed. After this is landed we can build glslang in chrome in the "all" target (as a dependency of Dawn)


dj2:
lgtm

PR #1559, opened 2018-10-31
Use precompiled headers for some glslang projects (MSVC-only)

jeffbolznv:
This is about a 2x speedup in incremental build time when I touch Types.h and rebuild the files that depend on it (i.e. most of the glslang code, but not the spirv-tools libraries). From around 1 minute 30 seconds to 45 seconds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,22 @@ if(USE_CCACHE)
     endif(CCACHE_FOUND)
 endif()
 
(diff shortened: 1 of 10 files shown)


johnkslang:
Nice. Does it hold up well when switch back and forth between different git branches? I do that often, and only recompile the necessary files.

jeffbolznv:
It should handle that fine, dependencies still work as usual.

PR #1571, opened 2018-11-12
Handle potentially uninitialized result

dj2:
In the case where the if and else if are both false the spv::Id result will be used uninitialized in the builder.addName call. If this happens, return an error instead of attempting to use result.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2775,7 +2775,9 @@ spv::Id TGlslangToSpvTraverser::createSpvVariable(const glslang::TIntermSymbol* 
     // can still have a mapping to a SPIR-V Id.
     // This includes specialization constants.
     if (node->getQualifier().isConstant()) {
-        return createSpvConstant(*node);
+        spv::Id result = createSpvConstant(*node);
+        if (result != spv::NoResult)
+            return result;
     }
 
     // Now, handle actual variables
@@ -7304,7 +7306,6 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
             result = createSpvConstantFromConstUnionArray(sn->getType(), *const_union_array, nextConst, true);
         } else {
             logger->missingFunctionality("Invalid initializer for spec onstant.");
-            exit(1);
             return spv::NoResult;
         }
         builder.addName(result, sn->getName().c_str());
@@ -7314,7 +7315,6 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
     // Neither a front-end constant node, nor a specialization constant node with constant union array or
     // constant sub tree as initializer.
     logger->missingFunctionality("Neither a front-end constant nor a spec constant.");
-    exit(1);
     return spv::NoResult;
 }
 
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -7302,6 +7302,10 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
         } else if (auto* const_union_array = &sn->getConstArray()) {
             int nextConst = 0;
             result = createSpvConstantFromConstUnionArray(sn->getType(), *const_union_array, nextConst, true);
+        } else {
+            logger->missingFunctionality("Invalid initializer for spec onstant.");
+            exit(1);
+            return spv::NoResult;
         }
         builder.addName(result, sn->getName().c_str());
         return result;


SPIRV/GlslangToSpv.cpp
```diff
@@ -7302,6 +7302,10 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
         } else if (auto* const_union_array = &sn->getConstArray()) {
             int nextConst = 0;
             result = createSpvConstantFromConstUnionArray(sn->getType(), *const_union_array, nextConst, true);
+        } else {
+            logger->missingFunctionality("Invalid initializer for spec onstant.");
+            exit(1);
```
johnkslang: That's too big a hammer. Generally, glslang-proper does not exit, it could be embedded. I see the only other exit is also in this function, and I must have missed it when it was added as a side contribution.

Can you expand your commit to remove both exits? Log the problem with missingFunctionality, but allow partial enabling of turn on of new systems rather than killing them.

SPIRV/GlslangToSpv.cpp
```diff
@@ -7302,6 +7302,10 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
         } else if (auto* const_union_array = &sn->getConstArray()) {
```
johnkslang: Someone added `if` to the `else` after the fact. Did you see why?

SPIRV/GlslangToSpv.cpp
```diff
@@ -7302,6 +7302,10 @@ spv::Id TGlslangToSpvTraverser::createSpvConstant(const glslang::TIntermTyped& n
...
             int nextConst = 0;
             result = createSpvConstantFromConstUnionArray(sn->getType(), *const_union_array, nextConst, true);
+        } else {
+            logger->missingFunctionality("Invalid initializer for spec onstant.");
+            exit(1);
+            return spv::NoResult;
```
johnkslang: It in just as questionable to return an unexpected 0 to callers not prepared for that, so this statement does not improve things much.

Was this triggered by a real life example?  That is, can it actually happen? Or, are we focused on the wrong thing here?

johnkslang:
> If this happens, return an error instead of attempting to use result.

Actually, you exited the entire controlling process. Whether there is a way to handle an "error result", which typically don't exist in this layer, depends on the actual situation triggering it, which I haven't tried to figure out yet.

SPIRV/GlslangToSpv.cpp
dj2: exit seemed to be how errors were handled in this method, so I followed what existed previously. I can remove the exit, but I need to know what the default value for result should be in that case. Currently it's whatever happens to be in memory ....

SPIRV/GlslangToSpv.cpp
dj2: I didn't look.

SPIRV/GlslangToSpv.cpp
dj2: The existing code returns spv::NoResult, so I assume it's handled?

It generates a compiler error because Result can be used uninitialized. I can disable that warning when building glslang if needed.

dj2:
Updated the commit message to match the current functionality.

SPIRV/GlslangToSpv.cpp
johnkslang: It was changed just recently by 4c9876b34 which basically undid what your are doing.

johnkslang: cc @nicebyte

johnkslang: It looks like the front end guarantees one or the other paths, so this is just logging that broke and doing error recovery that "should" never happen.

johnkslang:
It looks like, after logging the missing functionality, the beginning of `createSpvVariable` could handle a `NoResult` by just creating a variable instead:

E.g., something like changing:

```
    if (node->getQualifier().isConstant()) {
        return createSpvConstant(*node);
    }
```

to:

```
    if (node->getQualifier().isConstant()) {
        spv::Id result = createSpvConstant(*node);
        if (result != spv::NoResult)
            return result;
    }
```

SPIRV/GlslangToSpv.cpp
dj2: Setting an initial value seems like a good safety check in case the frontend changes to allow other paths through.

johnkslang: Right, the point of the `missingFunctionality` is to catch changes and real missing functionality. In this case the "initial value" can instead be a normal variable (per suggestion below).

dj2:
Updated to remove both exits and handle the NoResult return.

johnkslang:
Thanks.

PR #1577, opened 2018-11-13
Updated type conversion rules for HLSL frontend

TiemoJung:
This fixes some of the type conversion gaps of scalar, vector and matrix type of the HLSL frontend.

All cases of the tests where compiled with D3dcompiler_47.dll v 10.0.17.134.1 to verify which conversions are invalid and which are not (comments after each line contain the compile log of fxc).


johnkslang:
Thanks, nice contribution. Just needs tidied up.

TiemoJung:
@johnkslang I think i addressed all your issues, please have a look.

Issue #1576, opened 2018-11-13
Raytracing shader tests use gl_InstanceID but that's invalid in Vulkan


dneto0:
The validator recently started checking the rule that InstanceID is not allowed in Vulkan shaders. (Thank you @sheredom )

But Tests/ spv.AnyHitShader.rahit, spv.ClosestHitShader.rchit, spv.IntersectShader.rint  use gl_InstanceID.
That is translated to an access of BuiltIn InstanceID, which is not permitted in Vulkan.

The gl_InstanceIndex is not defined for those shader types.  It's unclear to me what is intended here.  The SPV_NV_ray_tracing extension does not specifically enable InstanceID, so it appears that the tests are invalid, and glslang is incorrectly accepting gl_InstanceID for such shaders.

Cc: @chaoc @dgkoch 

(Edited to fix VertexID -> InstanceID)

dneto0:
This is causing failures in Shaderc bots because they always pull top-of-tree SPIRV-Tools

dgkoch:
We're discussing what to do here. Our implementation is expecting InstanceId, but the SPV and VK specs say InstanceIndex.

dneto0:
@dgkoch Thanks for the update.  I'd be ok with whatever resolution makes sense for you.   I just need the specs and tools to line up somehow.  If it isn't a speedy resolution please consider blessing #1578 to temporarily disable the offending portions of the test.

dgkoch:
@dneto0 unfortunately @alelenv is OOTO this week and we'd like to confer with him, but our inclination is to update the SPV and Vulkan specs so match the implementation (meaning that InstanceId instead of InstanceIndex is allowed in intersection, anyhit, and closest hit shaders). 

Can we at least temporarily whitelist that in spirv-tools while we figure that out?

johnkslang:
It is okay to track failures in the regression results.  See https://github.com/KhronosGroup/glslang/pull/1578#issuecomment-438441961.

dneto0:
@dgkoch 

> Can we at least temporarily whitelist that in spirv-tools while we figure that out?

Yeah.  Please file an issue against SPIRV-Tools to disable the check for the 3 raytracing shader types.  We can do that before spec changes land.

dgkoch:
@dneto0 I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/2046 for this.

dgkoch:
[...]

PR #1578, opened 2018-11-13
Disable gl_InstanceID in raytracing tests

dneto0:
See #1576 

Cc: @chaoc @dgkoch


dneto0:
cc: @alelenv

johnkslang:
I'd prefer to leave these failures as "data". The regression suite captures the validation output, good or bad, and shows the differences made by new commits.

E.g., if a new PR updates the known_good SPIR-V tools, and that adds validation failures to the glslang regression suite, and they are verified to be about this, we can accept as a package the new SPIR-V tools plus the captured failures.  Later, if we decide to spell built-ins differently, or tools change, then the we'll see the regression suite track the disappearance of those failures. That's actually worth seeing.

PR #1587, opened 2018-11-16
Update to newer SPIRV-Tools, including VK_EXT_scalar_block_layout

dneto0:
Validator has more refined messages about what kind of block layout
rules have been applied.


dneto0:
This is somewhat related to https://github.com/KhronosGroup/Vulkan-Docs/issues/854

Issue #1259, opened 2018-02-26
Uniform initializers are silently ignored in the generated SPIR-V


bpeel:
We created [a test for Piglit](https://github.com/Igalia/piglit/blob/igalia/arb_gl_spirv/tests/spec/arb_gl_spirv/execution/uniform-initializer.shader_test) to test uniform initializers in GL_ARB_gl_spirv as follows:
```
#version 450

layout (location = 0) out vec4 color;

layout (location = 0) uniform vec4 in_color = vec4(0.0, 1.0, 0.0, 1.0);

void main()
{
	color = in_color;
}
```

glslangValidate generates the following output:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %color
               OpExecutionMode %main OriginLowerLeft
               OpSource GLSL 450
               OpName %main "main"
               OpName %color "color"
               OpName %in_color "in_color"
               OpDecorate %color Location 0
               OpDecorate %in_color Location 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
    %v4float = OpTypeVector %float 4
%_ptr_Output_v4float = OpTypePointer Output %v4float
      %color = OpVariable %_ptr_Output_v4float Output
%_ptr_UniformConstant_v4float = OpTypePointer UniformConstant %v4float
   %in_color = OpVariable %_ptr_UniformConstant_v4float UniformConstant
       %main = OpFunction %void None %3
          %5 = OpLabel
         %12 = OpLoad %v4float %in_color
               OpStore %color %12
               OpReturn
               OpFunctionEnd
```

Notice that the vec4 constant does not appear. I think it should be possible to add the constant as the optional third argument to OpVariable. The SPIR-V spec has this for the UniformConstant storage class:

> **UniformConstant** […] They may have initializers, as allowed by the client API.

johnkslang:
But, the Vulkan client API does not allow them.

Perhaps an error/warning is needed.

bpeel:
Sorry for not making it very clear, but here I am targeting OpenGL with GL_ARB_gl_spirv and the `-G` option of glslangValidator.

PR #1588, opened 2018-11-19
spirv: Support initializers on uniforms

infapi00:
If a uniform has an initializer it will now be given as the optional
initializer operand to the OpVariable instruction.

Fixes: https://github.com/KhronosGroup/glslang/issues/1259

Signed-off-by: Neil Roberts <nroberts@igalia.com> (the code)
Signed-off-by: Alejandro Piñeiro <apinheiro@igalia.com> (the tests)


johnkslang:
What is the specification language backing this change?

Note, you can't really initialize uniforms from a shader; at best it is side-band information, but where does it say this is the way to do that?

infapi00:
> What is the specification language backing this change?
> 
> Note, you can't really initialize uniforms from a shader; at best it is side-band information, but where does it say this is the way to do that?

Note that as the original issue #1259 already clarified, we are talking about OpenGL/ARB_gl_spirv here. So, from GLSL 4.60 spec, section 4.3.5, "Uniform Variables":

> The link-time initial value is either the value of the variable’s
> initializer, if present, or 0 if no initializer is present. Opaque types cannot have initializers, or a
> compile-time error results.
> Example declarations are:

```
uniform vec4 lightPosition;
uniform vec3 color = vec3(0.7, 0.7, 0.2); // value assigned at link time
```

Do you want me to update the commit on expilcitly mention that this would be only for OpenGL, and/or include the spec quote?

johnkslang:
What's missing is how a SPIR-V OpVariable initializer turns into pre-initialized memory.

Are you assuming that drivers will scan the SPIR-V looking for these?  Or, something else?

johnkslang:
Note that it GLSL, these to not become run-time assignments in the shader, they become side-band data to the driver, and not part of the shader.

infapi00:
> What's missing is how a SPIR-V OpVariable initializer turns into pre-initialized memory.
> 
> Are you assuming that drivers will scan the SPIR-V looking for these? Or, something else?

The patch is assuming a runtime assignment. But as you are saying below ...

> Note that it GLSL, these to not become run-time assignments in the shader, they become side-band data to the driver, and not part of the shader.

... that's true, and a good point. 

I get now your original concern, sorry the misunderstanding. After checking the different specs (ARB_gl_spirv/SPIR-V) it is right that there isn't anything supporting this fix. 

But in general, there isn't anything about how uniform initializers should be mapped on SPIR-V. This fix is doing a runtime assignment, basically because as far as I see, there is not a clear way to represent it in other way. And if we try to mimic GLSL/OpenGL behaviour, that is link assignment, as you are saying before that would mean scan the SPIR-V and do the uniform assignment, that seems to go somewhat in the opposite direction of how SPIR-V shaders should be linked.

So concluding: I think that it would be better to create a spec issue to try to clarify this.

infapi00:
> So concluding: I think that it would be better to create a spec issue to try to clarify this.

In fact, before doing this. While researching for this spec issue, I found the following from the SPIR-V shader, section 3.7 Storage Class, UniformConstant;

> UniformConstant Shared externally, visible across all functions in
> all invocations in all work groups. Graphics uniform memory.  OpenCL
> constant memory. Variables declared with this storage class are
> read-only. They may have initializers, as allowed by the client API

And from 3.32.8 Memory Instructions, OpVariable:

> Initializer is optional. If Initializer is present, it will be the
> initial value of the variable’s memory content. Initializer must be an
> <id> from a constant instruction or a global (module scope) OpVariable
> instruction. Initializer must have the same type as the type pointed
> to by Result Type.

So those two spec quotes justifies the implementation. What it is still pending is if we need to clarify how that is used. Looking at the Mesa driver, initializers are included on the internal IR while processing the SPIR-V shader, and then lowered to load a constant. That is basically the same final outcome with a GLSL shader.

Should I still create a spec issue to clarify the GLSL->SPIR-V mapping, or it is enough with those spec quotes?

johnkslang:
Okay, I didn't see an API protection in the code, but I guess you pick that up automatically, because only OpenGL allows default uniforms to begin with.

In the specification, the closest I see is this:

>     Corresponding features that GLSL keeps, despite GL_KHR_vulkan_glsl removal:
>      . default uniforms (those not inside a uniform block)

I think it should be made more clear whether that includes initializers, and ask whether drivers are supporting that.

infapi00:
> Okay, I didn't see an API protection in the code, but I guess you pick that up automatically, because only OpenGL allows default uniforms to begin with.

True. We could add some OpenGL checks if you prefer.

> In the specification, the closest I see is this:
> 
> > ```
> > Corresponding features that GLSL keeps, despite GL_KHR_vulkan_glsl removal:
> >  . default uniforms (those not inside a uniform block)
> > ```
> 
> I think it should be made more clear whether that includes initializers, and ask whether drivers are supporting that.

Ok, it makes sense. I think that at this point it is better to clarify this, instead of guessing if it needs to be supported or not. I will create a spec issue soon.

infapi00:
> Ok, it makes sense. I think that at this point it is better to clarify this, instead of guessing if it needs to be supported or not. I will create a spec issue soon.

The spec has been updated recently, in order to explicitly allow uniform initializers. Could it be possible to revisit this pull request now?

johnkslang:
Yes, of course.

Can you sign the CLA?

infapi00:
> Can you sign the CLA?

Done. It also seams that now there are conflicts with this branch. Will check if it is easy fix and update it now.

infapi00:
> Done. It also seams that now there are conflicts with this branch. Will check if it is easy fix and update it now.

Done too.

infapi00:
Hmm, it is strange. I have just realized that it appears as I didn't sign the LCA. But if I click again the link in order to do that (as I did in the past), I arrive to a page that says that I already signed the agreement. Not sure if Im missing something.

infapi00:
> Not sure if Im missing something.

Well, what I was missing is that although I made the pull request in behalf of @bpeel, as he is the author of the patch, he is the one that needs to sign the cla. Silly me. Sorry for that.

@bpeel has just signed it. Sorry for the delay.

johnkslang:
We need more tests that show, for example:

- what happens in the cases where initializers were already handled by another path (make sure we don't initialize twice?)
- what happens with specialization constants
- what happens with more complex composites
- ...

crzwdjk:
Hi, I'm taking over this merge request. I added some simple tests, is there anything else I should be testing?

crzwdjk:
@johnkslang ping

crzwdjk:
@johnkslang ping

PR #1589, opened 2018-11-20
HLSL: fix for overload and intrinsic rules

TiemoJung:
- added FmaD, as 'mad' and 'fma' are not the
  same on HLSL, mad is non double, fma is
  double exclusive. Boils down to same fma
  instruction in spir-v.
- following intrinsics forbid type conversion
  of their inputs (truncation is still allowed)
  - asdouble
  - fma
  - CheckAccessFullyMapped
- added opcode for CheckAccessFullyMapped
  to support proper intrinsics handling,
  still not implemented though.
- Changed overload system to properly handle
  HLSL rules. Changes how selectFunction works.
  - better is now a function that returns the
    relative distance between two param comparisons.
  - now sums all relative distances up to decide
    if two possible overloads are equal.
- Added more complex test to validate overload
  rules.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -6721,6 +6721,7 @@ spv::Id TGlslangToSpvTraverser::createMiscOperation(glslang::TOperator op, spv::
         opCode = spv::OpBitFieldInsert;
         break;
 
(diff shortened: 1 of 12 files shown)


TiemoJung:
will be replaced by a different changeset

PR #1592, opened 2018-11-21
GLSL: Fix #1591: Require the id in subgroupBroadcast to be constant.

### LINKED ISSUE #1591: No compiler error thrown when a non-constant ID is passed to subgroupBroadcast
From the GL_KHR_shader_subgroup spec:

    The function subgroupBroadcast() returns the <value> from the invocation
    whose <gl_SubgroupInvocationID> is equal to <id>.  <id> must be an integral
    constant expression.

However, putting in a non-constant expression works fine in GLSLang.  Incidentally, it also works fine in our driver so I could easily see shader authors (apparently myself included) mistaking `subgroupBroadcast` for `subgroupShuffle` and a compile error would significantly help.

johnkslang:


johnkslang:
Would like approval from @sheredom .

sheredom:
Looks good - it does work with specialization constants though, right?

johnkslang:
Yes. The constants passed in are just constants and work as specialization constants the same way any other constant does.

Issue #1574, opened 2018-11-13
OpLines are incorrect for HLSL with #include


greg-lunarg:
Execute the following on the files in the attached .zip:

glslangValidator.exe -D -V -e MainPs -g -Od -o foo4.frag.spv foo4.frag

Notice that the OpLines in the resulting SPIR-V indicate that all source came from foo4.frag, missing the fact that one statement came from the include file u1.h.

[i1.zip](https://github.com/KhronosGroup/glslang/files/2574145/i1.zip)

greg-lunarg:
Fixed by #1593

Issue #1575, opened 2018-11-13
OpLines incorrect for HLSL with #lines from multiple files


greg-lunarg:
Execute the following on the files in the attached .zip:

glslangValidator.exe -D -V -e MainPs -g -o foo4.i.frag.spv foo4.i.frag

Notice that while the #line directives indicate that the source came from multiple files, the OpLines specify that all source came from foo4.frag.

[i2.zip](https://github.com/KhronosGroup/glslang/files/2574153/i2.zip)

greg-lunarg:
@johnkslang  I am going to start looking at this.

johnkslang:
Okay, thanks.

Note that each node tracks its location, including string number or compilation unit, but that no work has been done yet to map that through to OpLine support.

PR #1593, opened 2018-11-21
Fix filename-based #line handling (#1575)

greg-lunarg:
SPIR-V OpLines were not capturing filenames on HLSL-style #lines.


greg-lunarg:
Changed to WIP. This change does not correctly handle #lines that have no filename attached to them which are preceeded by a #line which does have a filename attached to it. Such a #line refers to the same file as the preceeding #line.

greg-lunarg:
Ready to merge.

greg-lunarg:
Sorry. Need to go back to WIP. The filename in the OpSource is not right, and I fear that I may have caused GLSL line info to regress. I need to add a GLSL-style test into the regression tests.

johnkslang:
Also, if you are making the location significantly larger, I will ask you to not do that.

greg-lunarg:
> Also, if you are making the location significantly larger, I will ask you to not do that.

I should not need more than the TString* I already added.

greg-lunarg:
Ready for merge when CI passes.

johnkslang:
I see you distinguishing between "GLSL style" vs. "HLSL style" naming. I don't understand yet what the point is, yet the comments are in terms of this distinction.

I could see an "OpenGL API style" vs. "Vulkan style", where the OpenGL style is an array of strings from the online API entry point, while the Vulkan style is (in practice) a set of file names from the offline tool processing. Yet, that can't be it, because the existing code already uses the existing name field to store the file name holding the shader.

Note that HLSL only has one compilation unit anyway, while it is GLSL that supports multiple compilation units.  Both support #include, reasonably symmetrically.

I can dig deeper, but would be better to understand the high-level design first, if you can address the above.

greg-lunarg:
My references to GLSL-style vs HLSL-style is to #line directives. Particularly, when I say HLSL-style #line I mean #line with a filename specified.

When I first started this, I didn't know about the (Google) extension that allowed filename-based #line in GLSL.

I will go through and change all references of HLSL-style #line to filename-base #line.

glslang/Include/Common.h
```diff
@@ -238,7 +241,14 @@ struct TSourceLoc {
             return quoteStringName ? ("\"" + std::string(name) + "\"") : name;
         return std::to_string((long long)string);
     }
+    const char* getFilename() const
+    {
+        if (filename == nullptr)
+            return nullptr;
+        return filename->c_str();
+    }
     const char* name; // descriptive name for this string
```
johnkslang: Filenames are also being stored in `name`, aren't they? I don't yet see why we need both `name` and `filename`.

glslang/MachineIndependent/Scan.h
```diff
@@ -170,8 +170,11 @@ class TInputScanner {
     // for #line override in filename based parsing
     void setFile(const char* filename)
     {
+        TString* fn_tstr = NewPoolTString(filename);
         logicalSourceLoc.name = filename;
+        logicalSourceLoc.filename = fn_tstr;
         loc[getLastValidSourceIndex()].name = filename;
+        loc[getLastValidSourceIndex()].filename = fn_tstr;
```
johnkslang: These TSourceLocation are everywhere, making them 20% bigger might impact something, and yet this is doing that and then being redundant (or apparently so). Is there a way to avoid this, like merging the concepts of "descriptive name" and "file name"?

greg-lunarg: OK, I got rid of the new member of TSourceLoc and am using the existing name member.

glslang/Include/Common.h
greg-lunarg: OK. Will combine them...

greg-lunarg:
This also fixes #1574. I have just added a regression test for that issue.

glslang/MachineIndependent/preprocessor/PpContext.h
```diff
@@ -533,7 +533,7 @@ class TPpContext {
               prologue_(prologue),
               epilogue_(epilogue),
               includedFile_(includedFile),
-              scanner(3, strings, lengths, names, 0, 0, true),
+              scanner(3, strings, lengths, nullptr, 0, 0, true),
```
johnkslang: Does this mean the `names` member is dead? If so, remove it?

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -881,6 +881,8 @@ bool ProcessDeferred(
         intermediate.setSourceFile(names[numPre]);
         for (int s = 0; s < numStrings; ++s)
             intermediate.addSourceText(strings[numPre + s]);
+        // Filter carriage returns which manifest during testing
+        intermediate.filterSourceText('\r');
```
johnkslang: Does that potentially filter something else too? E.g., if an HLL has strings or characters or literals that happen to have this value, do they get deleted? Is it illegal to ever use this character except for the end of a line?

glslang/MachineIndependent/Scan.h
```diff
@@ -65,7 +65,7 @@ class TInputScanner {
         }
         if (names != nullptr) {
             for (int i = 0; i < numSources; ++i)
-                loc[i].name = names[i];
+                loc[i].name = names[i] ? NewPoolTString(names[i]) : nullptr;
```
johnkslang: `names[i] != nullptr ? ...`

glslang/Include/Common.h
```diff
@@ -229,16 +229,28 @@ inline const TString String(const int i, const int /*base*/ = 10)
...
+        name = nullptr; string = 0; line = 0; column = 0;
+    }
     void init(int stringNum) { init(); string = stringNum; }
     // Returns the name if it exists. Otherwise, returns the string number.
     std::string getStringNameOrNum(bool quoteStringName = true) const
     {
-        if (name != nullptr)
-            return quoteStringName ? ("\"" + std::string(name) + "\"") : name;
+        if (name != nullptr) {
+          TString qstr = quoteStringName ? ("\"" + *name + "\"") : *name;
```
johnkslang: Make indentation consistent.

greg-lunarg:
> Can you also reduce this to a small number of rational commits (1 is okay, if you want).

To clarify, you are asking that I squash all the commits on this branch to one commit and force push to this branch?

SPIRV/SpvBuilder.cpp
```diff
@@ -98,6 +100,38 @@ void Builder::setLine(int lineNum)
...
+            // If filename previously seen, use its id, else create a string
+            // and put it in the map.
+            auto sItr = stringIds.find(filename);
+            if (sItr != stringIds.end()) {
+                addLine(sItr->second, currentLine, 0);
+            } else {
+                Instruction* fileString =
+                      new Instruction(getUniqueId(), NoType, OpString);
+                fileString->addStringOperand(filename);
+                unsigned int stringId = fileString->getResultId();
```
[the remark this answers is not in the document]
greg-lunarg: Done.

glslang/MachineIndependent/preprocessor/PpContext.h
greg-lunarg: Yep. Done.

johnkslang:
Yes. There is nothing wrong with force pushing here.

One commit is okay.

If you also wanted a few rational commits that is fine too, like one that changes const* to TString, another that adds the functionality, etc. The point is, the master branch history does not need things like individual bug fixes from your development.

glslang/Include/Common.h
greg-lunarg: Done.

glslang/MachineIndependent/Scan.h
greg-lunarg: Done.

greg-lunarg:
> The point is, the master branch history does not need things like individual bug fixes from your development.

Agreed. Once we are a point that you believe that no further changes are required, let me know and I will do the squash.

glslang/MachineIndependent/ShaderLang.cpp
greg-lunarg: This just filters out true ascii CR.

I think a true ascii CR is only interpreted as whitespace as it appears in GLSL or HLSL source, so I think we are ok filtering it out.

An HLSL string such as "foo\r\n" would actually contain the characters f, o, o, \, r, \, n, and so would not be subject to filtering.

greg-lunarg:
@johnkslang I think I have addressed everything in your latest review.

SPIRV/SpvBuilder.cpp
```diff
@@ -98,6 +100,38 @@ void Builder::setLine(int lineNum)
...
+// Emit OpLine if we've been asked to emit OpLines and the line number or filename
+// has changed since the last time, and line number is valid.
+void Builder::setLine(int lineNum, const char* filename)
+{
+    if (filename == nullptr) {
+        setLine(lineNum);
+        return;
+    }
+    if ((lineNum != 0 && lineNum != currentLine) || currentFile == nullptr ||
+            strcmp(filename, currentFile) != 0) {
```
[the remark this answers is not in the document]
johnkslang: We do have many uses of `strncmp`.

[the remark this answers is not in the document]
greg-lunarg: OK. Using strncmp.

johnkslang:
Okay, thanks. Go ahead and rationalize the commits, and then I'll merge.

glslang/MachineIndependent/ShaderLang.cpp
johnkslang: Wait, didn't Mac used to use just CR alone as the end-of-line? I think glslang handles that in shader source.

Glslang was treating any of CR, LF, or CR+LF as an end of line? Maybe need to double check what the intent is here.

greg-lunarg: I think you are right about Mac.

I will look at coming up with a more comprehensive solution.

greg-lunarg: Somehow my new test source files got corrupted with \r  :P

Fixed those and pulled out filterSourceText.

PR #1595, opened 2018-11-22
Implement GL_EXT_fragment_invocation_density

dgkoch:
including SPV generation using SPV_EXT_fragment_invocation_density.
This is an alias of the functionality in SPV_NV_shading_rate, and thus in some
cases we can only have one set of the tokens present (switch statements), so we have picked the EXT version. This required updating the expected test results for SPV_NV_shading_rate.

Also update known_good.json so that the validator knows about the new extension/tokens.


dgkoch:
cc @johnkslang

PR #1597, opened 2018-11-26
Add conversion folding when the source is a constant.

sheredom:
This change adds unary conversion folding when the source is a constant.

This fixes an ISV issue whereby:

```
const float16_t f = float16_t(42.0);
```

Wouldn't compile because the conversion operator would always produce an EvqTemporary when it could have produced an EvqConst.

I've also added a test case that proves out that all basic-type to basic-type conversions work.


glslang/MachineIndependent/Constant.cpp
```diff
@@ -670,6 +670,279 @@ TIntermTyped* TIntermConstantUnion::fold(TOperator op, const TType& returnType)
...
+        case EOpConvUint16ToBool:
+            newConstArray[i].setBConst(unionArray[i].getU16Const()); break;
+        case EOpConvIntToBool:
+            newConstArray[i].setBConst(unionArray[i].getIConst()); break;
+        case EOpConvUintToBool:
+            newConstArray[i].setBConst(unionArray[i].getUConst()); break;
+        case EOpConvInt64ToBool:
+            newConstArray[i].setBConst(unionArray[i].getI64Const()); break;
+        case EOpConvUint64ToBool:
+            newConstArray[i].setBConst(unionArray[i].getI64Const()); break;
```
Anteru: This should be `getU64Const`.

sheredom: Good catch!

PR #1598, opened 2018-11-26
Update known good SPIRV tools and headers

alan-baker:
* Fixes use of optimizer
* Updated test outputs
 * GeometryStreams and TransformFeedback no longer fail validation

SPIRV-Tools no longer flags GeometryStreams and TransformFeedback capabilities as invalid in Vulkan. This PR updates SPIRV-Tools with those commits and updates expected test results.


johnkslang:
After syncing with this change, I can no longer build with MSVC 2015:

```
44>  CMake Error at External/spirv-tools/source/reduce/cmake_install.cmake:39 (file):
44>    file INSTALL cannot find
44>    "C:/cygwin64/home/johnkessenich/GitHub/glslang/build/External/spirv-tools/source/reduce/Release/SPIRV-Tools-reduce.lib".
```

Any ideas?

dneto0:
What build tool are you using? msbuild visualstudio?  What -G flag did you pass to cmake?

johnkslang:
That was the clue I needed... Virtually always, visual studio itself reconfigures from the CMake changes it sees. But, this change required regeneration externally (visual studio reconfigured, but silently failed).  Thank you.

PR #1599, opened 2018-11-26
HLSL: overload and intrinsic fixes

TiemoJung:
Changes overload handling and intrinsic
information to match fxc more closely.

Changes are:
- Added new intrinsic arg order key '=',
  which is a postfix modifier for M to
  limit matrices to NxN and exclude all
  other.
- vector and matrix intrinsics for N and/or
  M is 1 enabled for HLSL.
- Updated some intrinsic definitions:
  - abs only for float and signed int
  - asdouble only for unsigned int
  - clip only for float and signed int, spec
    says that kill is called for values less
    than zero, unsigned int can never be less
    than zero.
  - countbits only for unsigned int
  - determinant only for matrices with NxN
  - reversebits only for unsigned int
- Added intrinsic representation of
  CheckAccessFullyMapped, still not implemented.
- Added FmaD to represent fma for double and changed
  fma to use FmaD instead. HLSL fma is limited to
  double, mad is fma for everything else.
- Overhauled overload handling.
  - new selectFunction for HLSL as the rules are
    very different from GLSL. This allows also a
    more efficient implementation.
  - Conversion is now turned into a cost factor.
  - After cost calculation, the set is filtered
    for to boolean conversion and int to float
    conversions as both are considered worse even
    on same cost factors than others.
  - After filtering the, cheapest is chosen and all
    more expensive calls are removed from the set.
    Should more than one remain, the call is ambiguous.
  - Expanded error messages on overload errors.
    On ambiguous calls or calls where no matching
    overload was found the possible candidates are
    reported with the error message.
- Added one more test case. Combination of expansion
  truncation and overload search.
- Fixed incorrect hlsl.attributeC11.frag test,
  load for 2d texture requires 3 coordinates,
  uv and the lod level.
- hlsl.overalod.frag now produces errors like fxc does,
  some conversions are not clear for user functions.
- some test results have slightly different AST but represent
  basically the same structure. This is due to some of the
  intrinsic changes.
- hlsl.intrinsics.negative.frag now produces less errors, like
  fxc does.

NOTE: this needs a followup fix to handle the templated resource
      object types correctly, currently they result in incorrect
      return types of some resource methods (like load, sample)


arcady-lunarg:
Closing as this has been inactive for many years. Feel free to open a new PR if you still want to make these changes.

PR #1602, opened 2018-11-27
GlslangToSpv: prevent add XfbStride twice

infapi00:
Without this commit, if the XfbStride was explicitly set, the
decoration was added twice on the shader. This commit prevents that
and add a sanity check, as at that point the explicit stride and the
stride for the given buffer should be the same.

Note that AFAIK this is not causing any bug. It is just removing a superfluous XfbStride.


SPIRV/GlslangToSpv.cpp
```diff
@@ -7168,8 +7168,16 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
...
-            if (stride != glslang::TQualifier::layoutXfbStrideEnd)
-                builder.addDecoration(id, spv::DecorationXfbStride, stride);
+            if (stride != glslang::TQualifier::layoutXfbStrideEnd) {
+               if (!symbol->getQualifier().hasXfbStride()) {
+                  builder.addDecoration(id, spv::DecorationXfbStride, stride);
+               } else {
+                  // If the qualifier has an stride it was already
+                  // added above. We just check that the stride is the
+                  // same.
+                  assert(stride == symbol->getQualifier().layoutXfbStride);
```
johnkslang: This is not an inconsistency coding error, it is decision made by the front end, based on HLL semantics. I agree that today they should probably match, but not that execution should stop if, say, there are ever HLL semantics that don't require a match.

SPIRV/GlslangToSpv.cpp
```diff
@@ -7168,8 +7168,16 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
         if (symbol->getQualifier().hasXfbBuffer()) {
             builder.addDecoration(id, spv::DecorationXfbBuffer, symbol->getQualifier().layoutXfbBuffer);
             unsigned stride = glslangIntermediate->getXfbStride(symbol->getQualifier().layoutXfbBuffer);
-            if (stride != glslang::TQualifier::layoutXfbStrideEnd)
-                builder.addDecoration(id, spv::DecorationXfbStride, stride);
+            if (stride != glslang::TQualifier::layoutXfbStrideEnd) {
+               if (!symbol->getQualifier().hasXfbStride()) {
+                  builder.addDecoration(id, spv::DecorationXfbStride, stride);
```
johnkslang: Indentation is 4, then 3? Let's stick to just 4.

infapi00:
> Now that we've pinned down semantics more, I suspect the better fix is to delete the code just above:
> 
>          if (symbol->getQualifier().hasXfbStride())
>              builder.addDecoration(id, spv::DecorationXfbStride, symbol->getQualifier().layoutXfbStride);
> 
> Does that work?

Seems so. I have tested with the test included on the first version, and also without setting it, and everything keeps working. So I just removed those two lines.

> This is not an inconsistency coding error, it is decision made by the front end, based on HLL semantics. I agree that today they should probably match, but not that execution should stop if, say, there are ever HLL semantics that don't require a match.

Ok, assert removed.

> Indentation is 4, then 3? Let's stick to just 4.

Sorry for that. In any case, with the new version of the fix, that is not required.

PR #1603, opened 2018-11-27
SPV: Generate OpArrayLength with no signedness.

johnkslang:
GLSL says `.length()` is an `int`, while SPIR-V says `OpArrayLength` is uint.

This PR does the nominal thing to get the SPIR-V type correct, but it does not cure all validator errors for storing types, due to the semantic mismatch between the languages.

I suspect that it is correct to make the change to the type of the result of `OpArrayLength`, while we should relax SPIR-V rules to allow storing/loading to mix signedness.


dneto0:
The code change looks good to me.

I prefer not to change the spec for loading/storing mixed signedness. It's extra complexity on consumers that I think should be absorbed on generators.

johnkslang:
> The code change looks good to me.

But, it just does the "inner circle" type change, bumping up to the next data consumer where the inconsistency lies:

```
int L = s.a.length();  // no mismatch before, mismatch now
uint L = s.a.length(); // mismatch before, no mismatch now
```

Something more is needed still to fully emulate one set of semantics with the other (I'm hoping arithmetic vs. logical shifts are not a problem).

alegal-arm:
A very similar change got reveted by https://github.com/KhronosGroup/glslang/commit/ac3707921ed313eaba1c206e8113b85fef054c34. How is this different?

johnkslang:
Yes, we need caution here.  I think it might work with full emulation, which means a bunch of type casting will show up.

From old comments:

```
```

johnkslang:
Update: I added the bitcasts to take the uint result back to int again. HLSL was generating the AST as if the length was uint, so only GLSL gets the extra bitcast. That solves the store-type wack-o-mole aspect of this specific issue.

So, now the question is, does the PR in its current form have a problem with CTS?

alegal-arm:
> So, now the question is, does the PR in its current form have a problem with CTS?

It works fine for the CTS.

PR #1606, opened 2018-11-27
Add NDK build files

dj2:
This CL adds Android.mk files to simplify NDK builds for downstream projects. The necessary
files are added so an NDK build can be done inside glslang to test it is functional.

### DIFF
--- a/ndk_test/jni/Application.mk
+++ b/ndk_test/jni/Application.mk
@@ -1,0 +1,5 @@
+APP_ABI := all
+APP_BUILD_SCRIPT := Android.mk
+APP_STL := gnustl_static
+APP_PLATFORM := android-9
+NDK_TOOLCHAIN_VERSION := 4.9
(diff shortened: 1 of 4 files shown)


dj2:
@dneto0

johnkslang:
Seems like a bot failure, not a commit quality failure. These files should be completely benign, yes?

ndk_test/jni/Application.mk
```diff
@@ -0,0 +1,5 @@
+APP_ABI := all
+APP_BUILD_SCRIPT := Android.mk
+APP_STL := gnustl_static
```
dneto0: This is fine, but I think this is now deprecated.  I think libc++ is recommended for very recent NDKs.   A future MR might want to upgrade this.

dneto0:
> These files should be completely benign, yes?

Yes, adding these files is benign until some Glslang bot or external build starts using them.  Our downstream projects intend to do so, and we'll handle the integration costs.

PR #1608, opened 2018-11-30
DX9 sampler compatibility 

ClemensRognerSD:
Added DX9 compatible sambler decleration and fixed functions for it

*) added a trigger that iinterprets sampler declarations as a texture/sampler combo like DirectX9 would: "--hlsl-dx9-compatible". This should also serve as a trigger to handle more dx9 features in the future
*)  added an acceptType function in accordance to the above
*) fixed texXDlod functions not being implemented
*) fixed tex2Dlod function not being visible in the vert stage
*) added tests for dx9 compatible mode (2 test files)


johnkslang:
I manually did this, locally, to fix conflicts and tabs, and accidentally changed branches at the wrong time, which made git lose track of this as the origination of the change. But, it's all pushed and working.

PR #1614, opened 2018-12-04
Update test infrastructure

### LINKED ISSUE #1610: Vulkan memory model scope test fails validation
The test Glsl/CompileVulkanToSpirvTest.FromFile/spv_memoryScopeSemantics_comp fails validation using SPIRV-Tools (https://github.com/KhronosGroup/SPIRV-Tools/pull/2149)

```
 spv.memoryScopeSemantics.comp
+error: SPIRV-Tools Validation Errors
+error: Load: in Vulkan 1.0 environment Memory Scope is limited to Device, Workgroup and Invocation
+  %89 = OpLoad %uint %88 Volatile|MakePointerVisibleKHR|NonPrivatePointerKHR %uint_3
```

According to Appendix A of the Vulkan spec, this is error correct (though it omits QueueFamilyKHR as permissible in the text it does accept it).

This test should probably be targeting Vulkan 1.1. The SPV_KHR_vulkan_memory_model extension lists SPIR-V 1.3 as a required version.

@jeffbolznv should all the memory model tests generate SPIR-V 1.3?

alan-baker:
Modify testing to only record validation pass/fail

* Changed unit tests to only record known the validation pass/fail
status
 * errors are output as part of the failure message if the result is
 unexpected
 * can turn off validation for each test individually
* Moved some SPV_KHR_vulkan_memory_model tests to be compiled for Vulkan
1.1

I've gone through the test updates and they all seem reasonable to me. The result of this is that the base results only need updated if the validation state changes (e.g. fail -> pass or pass -> fail).

The validation can be set on a per-test basis now. This may be useful to disable validation for some tests that are never expected to be valid.

I use the presence of message from the logger as a proxy for the validation result. I did not want to plumb changing the return of GlslangToSpv without first checking with you.

Most of the new validation failures are a result of missing binding decorations in the Vulkan environment. I can provide a full test log if necessary.


Test/baseResults/spv.memoryScopeSemantics.comp.out
```diff
@@ -1,7 +1,7 @@
 spv.memoryScopeSemantics.comp
-// Module Version 10000
```
alan-baker: Addresses #1610

alan-baker:
If the validation state changes, the test will output the diff first then the errors and warnings it recorded. As an example:
```
The following warnings/errors occurred: 
error: SPIRV-Tools Validation Errors 
error: Uniform id '22' is missing Binding decoration. 
From Vulkan spec, section 14.5.2: 
These variables must have DescriptorSet and Binding decorations specified 
   %_ = OpVariable %_ptr_Uniform__Global Uniform 
```

gtests/Spv.FromFile.cpp
```diff
@@ -401,6 +399,8 @@ INSTANTIATE_TEST_CASE_P(
         "spv.subgroupQuad.comp",
         "spv.subgroupVote.comp",
         "spv.vulkan110.storageBuffer.vert",
+        "spv.memoryScopeSemantics.comp",
```
johnkslang: These moved from alphabetical order to not alphabetical order. It does help some to keep order when managing test issues.

gtests/TestFixture.h
```diff
@@ -111,7 +111,10 @@ class GlslangTest : public GT {
         : defaultVersion(100),
           defaultProfile(ENoProfile),
           forceVersionProfile(false),
-          isForwardCompatible(false) {}
+          isForwardCompatible(false) {
+        // Perform validation by default.
+        options_.validate = true;
```
johnkslang: Note that glslang is not using the "_" notation, except where a few have slipped it in, but conventions are getting mixed up and I'd rather have one and not two. Can you use local conventions?

known_good.json
```diff
@@ -5,14 +5,14 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit" : "d543f7dfed9ba02910996121375e57fff92c3c93"
+      "commit" : "a1439604eac013b2ad08c54783ec54cb75cc52ca"
```
johnkslang: Still want to do one thing per PR, or maybe one thing per commit.  Maybe this gets undone later?

johnkslang:
Still want to do one thing per PR, or possibly one thing per commit, but I'm not successfully teasing that out here. It seems some of the 7 commits are bug fixes to earlier commits, and some still do multiple independent things, and the PR as a whole is both changing test results and doing the change in representation of how test results are captured.

Can you change it to one thing per PR (or commit, if the commit set is clean)?  Thanks.

gtests/TestFixture.h
alan-baker: I did this to avoid shadowing the options() accessor. Is there another name you'd prefer (e.g. validatorOptions())?

known_good.json
alan-baker: This was undone in 1627b2f83c0942f21b1144db000776c92b84ffa5.

Test/baseResults/spv.memoryScopeSemantics.comp.out
johnkslang: I want do be able to see that one PR (or commit) changed only the formatting.

(It is much easier to see if any regressions occurred if a massive formatting change doesn't also change results.)

alan-baker:
I've squashed all the commits into a single commit.

gtests/Spv.FromFile.cpp
alan-baker: Done.

gtests/TestFixture.h
alan-baker: Renamed the variable to validatorOptions.

alan-baker:
Undid the change for #1610. Will address that after this PR.

PR #1622, opened 2018-12-10
Fix executable paths in test suite

akien-mga:
Fix executable paths in test suite

There were several outdated hardcoded reference to the glslangValidator build path, now all rely on the 'glslangValidator' shell script.
Also add '/usr/bin' as search path for spirv tools.


akien-mga:
I now see that those paths are actually used in the CI, so they shouldn't be changed.

It would be good to add support for injecting the `CMAKE_INSTALL_PREFIX` directly in the tests to be able to run them regardless of the chosen prefix, but that's beyond the scope of my small packaging improvements :) I'll keep my small patch locally for now.

Issue #1484, opened 2018-08-25
Shared library use not first-class functionality, not tested


jengelh:
Compilation fails in commit e7df8e0b76af2fb9f58124be6b95688e5dbcdb7a; it used to previously work in a51525059f3f264c2337ae72997ca4e7ad7e17d0 (where essentially no libHLSL.so was generated).
Please specify the dependencies of HLSL properly.
```
/usr/bin/cmake . '-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--no-undefined' -DBUILD_SHARED_LIBS:BOOL=ON

-- The C compiler identification is GNU 8.1.1
-- The CXX compiler identification is GNU 8.1.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Google Mock was not found - tests based on that will not build
-- spirv-tools not linked - illegal SPIRV may be generated for HLSL
-- Configuring done
-- Generating done

$ make VERBOSE=1
[...]
/usr/bin/c++ -fPIC  -Wl,--no-undefined -shared -Wl,-soname,libHLSL.so -o libHLSL.so CMakeFiles/HLSL.dir/hlslAttributes.cpp.o CMakeFiles/HLSL.dir/hlslParseHelper.cpp.o CMakeFiles/HLSL.dir/hlslScanContext.cpp.o CMakeFiles/HLSL.dir/hlslOpMap.cpp.o CMakeFiles/HLSL.dir/hlslTokenStream.cpp.o CMakeFiles/HLSL.dir/hlslGrammar.cpp.o CMakeFiles/HLSL.dir/hlslParseables.cpp.o 
/usr/lib64/gcc/x86_64-suse-linux/8/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/HLSL.dir/hlslParseHelper.cpp.o: in function `glslang::HlslParseContext::initializeExtensionBehavior()':
[...]

PR #1623, opened 2018-12-10
WIP: CMake: Allow shared build of OGLCompiler and OSDependent

akien-mga:


akien-mga:
Don't merge for now, I think those might add some more cyclic dependencies (see #1484), which could be avoided by linking statically to `glslang`.

akien-mga:
This actually seems to be a bad idea as mentioned above. The linker is happier if we simply build those as static libraries and then collate into the `libglslang.so`, than if we try to keep track of their cyclic dependencies.

This means that those libraries can't be provided as shared libraries system-wide, but I guess they might never be needed directly by `glslang` users?

PR #1626, opened 2018-12-12
Infrastructure: Generalize and broaden per-variable extension checking.

johnkslang:
Glslang has infrastructure to declare what extensions a symbol needs (in Initialize.cpp) and then parsing automatically checks and issues errors when the declared extensions are not present.

However, this infrastructure did not cover members of blocks; e.g., gl_PointSize in an anonymous block or gl_out[] block, etc.

This PR adds that infrastructure, which has multiple uses that should make use of it, as many block members have been added without sufficient checks that the right extensions are declared.

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -7737,11 +7737,12 @@ void TBuiltIns::initialize(const TBuiltInResource &resources, int version, EProf
 static void SpecialQualifier(const char* name, TStorageQualifier qualifier, TBuiltInVariable builtIn, TSymbolTable& symbolTable)
 {
     TSymbol* symbol = symbolTable.find(name);
-    if (symbol) {
-        TQualifier& symQualifier = symbol->getWritableType().getQualifier();
-        symQualifier.storage = qualifier;
-        symQualifier.builtIn = builtIn;
-    }
+    if (symbol == nullptr)
+        return;
+
+    TQualifier& symQualifier = symbol->getWritableType().getQualifier();
+    symQualifier.storage = qualifier;
+    symQualifier.builtIn = builtIn;
 }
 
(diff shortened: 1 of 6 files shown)


sparmarNV:
John,
Thanks for adding the infrastructure to allow error checks for builtin members.
Overall the change looks good. I haven't looked at your gl_PointSize issue though.

I had a more generic question regarding future error checks to be added for builtins which are incorrectly defined in an anonymous block. Adding these error checks might break existing shaders who have been using these loose builtins.
Do you think we should bind these error checks to a newer SPV version OR just add these checks with current SPV version and hope that nothing breaks?

johnkslang:
Yes, when we fix all the incorrect declarations (per internal Khronos discussion), we want to use these checks.

These checks are in some ways independent of such changes. All they do is make sure you declared the right extension. I don't see that we should support using the extension-specific built-ins without declaring the extension.

There is another infrastructure problem, in translation to SPIR-V, in keeping track of what subset of these have been redeclared by the shader (as I recall). That also needs a fix, but I think we can start with getting the pre-SPIR-V stuff done, if still only in this branch.

sparmarNV:
>>These checks are in some ways independent of such changes. All they do is make sure you declared the right extension. I don't see that we should support using the extension-specific built-ins without declaring the extension.

Agree that these are 2 different set of changes.
Let's focus on adding the extension checks first (using your framework) and later we can remove any incorrect declarations (based on latest revised specs).

Also were you able to fix the gl_PointSize issue in 310.tesc?
I was hoping to add similar extension checks for NV extensions once this PR is merged.

johnkslang:
Yes, the test result changes are gone now, and it's just that the declarations are not yet correct. I thought maybe you could to that on top of this PR, but I guess I can merge it in first.

PR #1624, opened 2018-12-10
CMake: Link OSDependent with Threads::Threads

akien-mga:
Would typically be pthreads.
Without this, '-Wl,--no-undefined' would output these errors:
```
/usr/bin/ld: OSDependent/Unix/libOSDependent.a(ossource.cpp.o): in function `glslang::OS_CleanupThreadData()':
/home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:83: undefined reference to `__pthread_register_cancel'
/usr/bin/ld: /home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:93: undefined reference to `__pthread_unregister_cancel'
/usr/bin/ld: OSDependent/Unix/libOSDependent.a(ossource.cpp.o): in function `glslang::OS_AllocTLSIndex()':
/home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:122: undefined reference to `pthread_key_create'
/usr/bin/ld: OSDependent/Unix/libOSDependent.a(ossource.cpp.o): in function `glslang::OS_SetTLSValue(void*, void*)':
/home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:137: undefined reference to `pthread_setspecific'
/usr/bin/ld: OSDependent/Unix/libOSDependent.a(ossource.cpp.o): in function `glslang::OS_FreeTLSIndex(void*)':
/home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:162: undefined reference to `pthread_key_delete'
/usr/bin/ld: OSDependent/Unix/libOSDependent.a(ossource.cpp.o): in function `glslang::InitGlobalLock()':
/home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:175: undefined reference to `pthread_mutexattr_init'
/usr/bin/ld: /home/akien/Mageia/Sandbox/_rpm/BUILD/glslang-7.10.2984/glslang/OSDependent/Unix/ossource.cpp:176: undefined reference to `pthread_mutexattr_settype'
[...]

PR #1633, opened 2018-12-18
CMake: Fix linking pthread of CMake < 3.1

akien-mga:
As reported in #1624, Ubuntu 14.04 LTS still uses CMake 2.8.12 which
does not support the Threads::Threads target (added in CMake 3.1).

This could be reverted once the required CMake version is bumped to 3.1+.

Not tested against CMake < 3.1, could you try it @archimedus?


archimedus:
Will do soon

akien-mga:
(Sorry for the various force pushes, fixed a few style issues - it should now be ready.)

akien-mga:
Fixed a typo in a comment.

PR #1641, opened 2019-01-03
Make sure source strings are terminated

dj2:
The source strings may or may not have a null terminator. We need to
make sure we add one before outputting the source strings as we iterate
over the c-str looking for the null terminator.

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -880,10 +880,9 @@ bool ProcessDeferred(
     if (messages & EShMsgDebugInfo) {
         intermediate.setSourceFile(names[numPre]);
         for (int s = 0; s < numStrings; ++s) {
-            // The string may be null-terminated, so make sure we create one
-            // with the given length.
-            std::string tmp(strings[numPre + s], lengths[numPre + s]);
-            intermediate.addSourceText(tmp.c_str());
+            // The string may not be null-terminated, so make sure we provide
+            // the length along with the string.
+            intermediate.addSourceText(strings[numPre + s], lengths[numPre + s]);
         }
     }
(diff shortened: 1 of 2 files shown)


dneto0:
Looks right to me.

cc: @greg-lunarg This needs to be fixed before the next SDK release.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -879,8 +879,12 @@ bool ProcessDeferred(
         intermediate.setHlslOffsets();
     if (messages & EShMsgDebugInfo) {
         intermediate.setSourceFile(names[numPre]);
-        for (int s = 0; s < numStrings; ++s)
-            intermediate.addSourceText(strings[numPre + s]);
+        for (int s = 0; s < numStrings; ++s) {
+            // The string may be null-terminated, so make sure we create one
```
johnkslang: Do you mean might not be?

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -879,8 +879,12 @@ bool ProcessDeferred(
...
+            // with the given length.
+            std::string tmp(strings[numPre + s], lengths[numPre + s]);
+            intermediate.addSourceText(tmp.c_str());
```
johnkslang: addSourceText makes a copy anyway.  See `addIncludeText` for the analogous fix for headers.... Just pass the length in. We get one less copy of the entire source, and symmetry in the intermediate API for setting these.

glslang/MachineIndependent/ShaderLang.cpp
dneto0: Yes.

glslang/MachineIndependent/ShaderLang.cpp
dj2: done

Issue #1642, opened 2019-01-03
OpSpecConstantOp with UConvert is generated for test spv.int16.amd.frag


dneto0:
This fails a new validation check I'm adding to SPIRV-Tools. (Bot run https://source.cloud.google.com/results/invocations/91de4ece-7661-479a-817c-47ea8c58bd2f/log)

I think the relevant source text is:

```
layout(constant_id = 106) const uint16_t su16 = 4US;
...
const uint u16_to_u = uint(su16);
```
There are several instances.

The offending code in the base results file is:

```
   541(u16_to_u):     18(int) SpecConstantOp 113 528(su16)    
```

sfricke-samsung:
ping

```glsl
#extension GL_EXT_shader_explicit_arithmetic_types_int8 : enable
#extension GL_EXT_shader_explicit_arithmetic_types_int16 : enable
layout (constant_id = 0) const uint8_t a = uint8_t(0);
layout (constant_id = 1) const uint16_t b = 0us;
```

also generates 

```swift               
          %a = OpSpecConstant %uchar 0
         %16 = OpSpecConstantOp %uint UConvert %a
          %b = OpSpecConstant %ushort 0
         %22 = OpSpecConstantOp %uint UConvert %b
```

so is the work around just to make sure to `--target-env vulkan1.2` in order to not get the `spirv-val` error?

arcady-lunarg:
I believe the underlying issue has been fixed in a spirv-val change here: https://github.com/KhronosGroup/SPIRV-Tools/pull/2264 where UConvert was permitted in this case.

PR #1643, opened 2019-01-03
Mark spv.int16.amd.frag.out as failing validation

dneto0:
See #1642


dneto0:
Cc: @greg-lunarg : this needs to be fixed before the next SDK release, since the SPIRV-Tools validation check has been merged into master.

PR #1646, opened 2019-01-08
Parenthesize to avoid GCC warning

dneto0:


dneto0:
Without this, I get the following warning, and it's elevated to an error:

```
glslang/MachineIndependent/../Include/../Include/Types.h:2006:44: error: suggest parentheses around ‘&&’ within ‘||’ [-Werror=parentheses]
             isStruct() && right.isStruct() && structure == right.structure)
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Issue #1654, opened 2019-01-11
Is xfb_stride must be multiple of 4


amdrexu:
I use int8 or int16 to test a transform feedback shader like this:

`layout(location = 0, xfb_buffer=0, xfb_stride = 2, xfb_offset = 30) out uint16_t u16;`

An error is reported concerning xfb_stride incorrectness.

Apparently, the xfb stride computation does not take 8-bit/16-bit into account because the standard xfb does not support such data types. But we do have 8-bit/16-bit types introduced by EXT_shader_explicit_arithmetic_types. For AMD, 16-bit data exported to xfb buffer is allowed by GL_AMD_gpu_shader_half_float. So I wonder if we can introduce such computation as a general solution. If you have concerns, I could separate the codes with AMD_EXTENSIONS.

johnkslang:
I am okay with an unprotected computation that works for multiple sizes.

pdaniell-nv:
As per the VK_EXT_transform_feedback spec, elements have to be 32-bit or 64-bit when decorating for transform feedback. Non-Vulkan SPIR-V environments might allow 8-bit and 16-bit, but not Vulkan.

DominikWitczakAMD:
As a temporary measure, would it work if we added an interaction with VK_EXT_TF to:

* VK_AMD_gpu_shader_half_float ext spec, lifting this restriction for fp16s
* VK_AMD_gpu_shader_int16 ext spec, this time for iu16s

We wouldn't be able to do that for IU8s since we don't have an AMD extension for these, but at least expanding our vendor-specific extensions would allow us to partially resolve the problem.

One way or another, this feels like something that we're going to need to raise with Khronos WG.

pdaniell-nv:
VK_KHR_shader_float16_int8 adds 8-bit, so you could use that too.

amdrexu:
Thank you for feedbacks. I will make the change.

PR #1656, opened 2019-01-14
Fix xfb_stride incorrectness(#1654)

amdrexu:
Add int64 support in XFB. Change containsDouble to contains64BitType. Make it more general.


amdrexu:
I add int64 support in this change. Will make new changes to add float16/int16 and int8 support.

amdrexu:
Any comment? I will continue my work based on it once it is accepted and is merged.

PR #1664, opened 2019-01-20
Fix a few GN build issues.

null77:
* adds a source set for default resource limits to mirror CMake
 * adds a target executable for the standalone validator
 * fixes a missing warning

Allows ANGLE to use the integrated BUILD.gn instead of a custom one.

ANGLE bug: 3088

@johnkslang PTAL


null77:
Thanks!

PR #1666, opened 2019-01-21
Fix crashes in HLSL parsing.

hodgman:
I'm running glslang over a large commercial HLSL code base for the first time, to convert it to SPIR-V. Unfortunately I can't post the offending HLSL code, but I can try to make a minimum reproducible shader if that would help.

The changes in this commit were required to prevent glslValidator from crashing when trying to parse these shaders: https://github.com/hodgman/glslang/commit/7ee47fee095fbba6db62d00300eb64a026ec6120

The offending code is based around practices such as this:
`struct TextureSampler2D
{
	Texture2D    tex;
	SamplerState texSampler;
};
TextureSampler2D CreateTexSampler2D(Texture2D tex, SamplerState texSampler)
{
	TextureSampler2D textureSampler;
	textureSampler.tex = tex;
	textureSampler.texSampler = texSampler;
	return textureSampler;
}`

### DIFF
--- a/hlsl/hlslParseHelper.cpp
+++ b/hlsl/hlslParseHelper.cpp
@@ -3256,7 +3256,7 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
     if (argAggregate) {
         if (argAggregate->getSequence().empty())
             return;
-		if( argAggregate->getSequence()[0] )
+		if (argAggregate->getSequence()[0])
 	        bufferObj = argAggregate->getSequence()[0]->getAsTyped();
     } else {
         bufferObj = arguments->getAsSymbolNode();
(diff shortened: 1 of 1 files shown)


hlsl/hlslParseHelper.cpp
```diff
@@ -3256,7 +3256,8 @@ void HlslParseContext::decomposeStructBufferMethods(const TSourceLoc& loc, TInte
     if (argAggregate) {
         if (argAggregate->getSequence().empty())
             return;
-        bufferObj = argAggregate->getSequence()[0]->getAsTyped();
+		if( argAggregate->getSequence()[0] )
```
johnkslang: Fix the style of these `if` parenthesis to match the rest, and I think this PR looks good to add.

hodgman: Ok, I've pushed another commit that fixes my parenthesis style to match the surrounding code :)

PR #1667, opened 2019-01-21
Add 8-bit/16-bit transform feedback support for future use

amdrexu:
I change a change to fix this: https://github.com/KhronosGroup/glslang/issues/1654.


amdrexu:
As we have discussed, it is for future use. And currently, AMD float16/int16 extension supports 16-bit data capture. To make it a general solution, I do not use AMD_EXTENSIONS. It migt be controversal. We can discuss it if anyone is unhappy with this.

amdrexu:
Any feedback?

johnkslang:
So, this generically computes/allows natural alignment in XFB: 2-byte things need 2-byte alignment, and 8-bit integers need no special alignment.

I assume that is okay because other extensions had to be declared and checked against in order to get that far. Yes?

Is that okay with everyone? @pdaniell-nv @DominikWitczakAMD @jekstrand

DominikWitczakAMD:
Yes, that works for us just fine.

jekstrand:
The VK_EXT_transform_feedback extension explicitly doesn't allow XFB of 8 and 16-bit things.  If we were to implement 8 or 16-bit XFB on Intel, we would need more than scalar alignment.  That said, that's in the very hypothetical future and this is now.  As long as this is properly gated behind an AMD extension, it's probably fine.

amdrexu:
Thanks for all your feedbacks. I understand the concerns. I will add AMD_EXTENSIONS to separate the codes from normal path to make 8-bit/16-bit XFB export as AMD-specific. I will update the commit.

amdrexu:
I update the change (use AMD_EXTENSIONS).

johnkslang:
appveyor has a bogus failure on this, unrelated to the changes.

PR #1668, opened 2019-01-22
CTC++ skipping

twiddletec:
Source code instrumentation is not possible on that overloaded '&&' and '||' operators,


johnkslang:
Needing 7 lines of code instead of one line of code fully undermines any advantage of using operator overloading as a concise notation.

We should either not do this, or eliminate the operator overloading. I'm unclear though why we should do this.

PR #1675, opened 2019-01-24
Add support for GL_EXT_YUV_target

nicebyte:
This PR adds support for the [GL_EXT_YUV_target](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_YUV_target.txt) extension.


glslang/MachineIndependent/Versions.cpp
```diff
@@ -157,6 +157,7 @@ void TParseVersions::initializeExtensionBehavior()
     extensionBehavior[E_GL_EXT_frag_depth]                   = EBhDisable;
     extensionBehavior[E_GL_OES_EGL_image_external]           = EBhDisable;
     extensionBehavior[E_GL_OES_EGL_image_external_essl3]     = EBhDisable;
+    extensionBehavior[E_GL_EXT_YUV_target]                   = EBhDisable;
```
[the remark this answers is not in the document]
nicebyte: Added

glslang/MachineIndependent/Versions.cpp
```diff
@@ -307,6 +308,7 @@ void TParseVersions::getPreamble(std::string& preamble)
             "#define GL_EXT_frag_depth 1\n"
             "#define GL_OES_EGL_image_external 1\n"
             "#define GL_OES_EGL_image_external_essl3 1\n"
+            "#define GL_EXT_YUV_target\n"
```
johnkslang: The `1` is missing? Is that expected?

Maybe this is missing in the extension specification, and should be added to it, probably in the way all the others are done, unless the extension specification somehow does say it's just a simple `#define`.

glslang/Include/Types.h
```diff
@@ -233,6 +236,9 @@ struct TSampler {   // misnomer now; includes images, textures without sampler,
             s.append("ExternalOES");
             return s;
         }
+        if (yuv) {
+          return "__" + s + "External2DY2YEXT";
```
johnkslang: 4 spaces (fix if the other comment below needs a fix).

glslang/MachineIndependent/Versions.cpp
nicebyte: nope, i just missed it. fixed.

PR #1676, opened 2019-01-25
build: Do not cache debug library suffix

nsubtil:
For nested project builds, writing CMAKE_DEBUG_POSTFIX into the cache
ends up affecting other projects. Caching this value doesn't seem to be
required in practice, so this change removes the cache tag.


johnkslang:
Yes, we had an open issue about this, was about to test it, thanks for the bootstrap.  Will accept this when bots are happy.

PR #1681, opened 2019-01-29
Reflection consistency changes & bugfixes

baldurk:
This is a couple commits with one bugfix and some changes to the reflection code that I think are the least controversial.

7e500c7 is a simple bugfix, some members were uninitialised in the default `TObjectReflection()` constructor that's used for `badReflection()`.

25217e4 updates the calculation of `offset` for reflection variables. At the moment the offset is only calculated based on direct struct-to-member indexing, which means the offset can be wrong in some cases, as well as duplicated among array members in array cases.

I've updated all the code paths so that indexing through arrays (either directly or indirectly) updates the offset according to the array's stride, and when we explode a struct's members we update the offset for each member before recursing.

a0e16bd is hard to split up any further, even though it's a couple of logical changes. The problem it's solving is that currently arrays can get the `[x]` suffix dropped if you access through the array directly, which leads to them looking like structs instead of arrays e.g. `t.v.position` instead of `t[0].v[0].position`. There are weird inconsistencies like that where depending on how an array is accessed it either lists all array members, just the directly indexed one, or the suffix gets dropped. Meaning you can end up with false duplicates like `deepB.d2.d1.va` in `hlsl.reflection.vert.out`.

I've changed the behaviour so that it should follow the rules in `ARB_program_interface_query`. Uniform blocks always include the array index but they don't exhaustively expand unless there's a dynamic index that forces it. Storage blocks are the same but only after the first array member - a top-level array is only ever expanded as `[0]` no matter what. That stops the reflection blowing up with dynamic indexing into very large SSBO members.

I think this is generally a straight improvement especially given the old codepath is kind of inconsistent depending on what the actual shader source does, but it is a behaviour change so I can defer this and put it behind an option in a later PR.


johnkslang:
Thank you for the fixes.

> a0e16bd is hard to split up any further, even though it's a couple of logical changes. The problem it's solving is that currently arrays can get the [x] suffix dropped if you access through the array directly, which leads to them looking like structs instead of arrays e.g. t.v.position instead of t[0].v[0].position. There are weird inconsistencies like that where depending on how an array is accessed it either lists all array members, just the directly indexed one, or the suffix gets dropped. Meaning you can end up with false duplicates like deepB.d2.d1.va in hlsl.reflection.vert.out.

Have you cross-checked the reflected syntax with the OpenGL API rules for the syntax? I haven't looked myself for some time, but there were some odd things about arrays.

johnkslang:
The merge to master failed, but apparently for unrelated reasons:

```
/home/travis/build/KhronosGroup/glslang/gtests/Config.FromFile.cpp:98:1: error: no viable conversion from 'const ::testing::TestParamInfo<ConfigTest::ParamType>' to '::std::string' (aka 'basic_string<char>')
INSTANTIATE_TEST_CASE_P(
^
/home/travis/build/KhronosGroup/glslang/External/googletest/googletest/include/gtest/gtest-param-test.h:590:33: note: expanded from macro 'INSTANTIATE_TEST_CASE_P'
#define INSTANTIATE_TEST_CASE_P INSTANTIATE_TEST_SUITE_P
                                ^
/home/travis/build/KhronosGroup/glslang/External/googletest/googletest/include/gtest/gtest-param-test.h:574:23: note: expanded from macro 'INSTANTIATE_TEST_SUITE_P'
        DUMMY_PARAM_))(info);                                                 \
                      ^~~~~~
```

Looking into it.

PR #1683, opened 2019-01-30
Disable Clang extra semicolon warning.

null77:
Usually this warning is disabled by default. But when turned on Clang
complains about extra semicolons in Glslang headers. Turn this off for
now. See http://crbug.com/926235

@johnkslang PTAL


johnkslang:
It seems the bots failed for unrelated reasons.

PR #1684, opened 2019-01-30
Opt-in improvements to reflection interface

baldurk:
As a follow-up to #1681, these are changes to the reflection I found necessary to implement ARB_program_interface_query. I've constructed them all as either strictly _new_ interfaces/reflection - so any use of the original interface will not notice the additional information, or as opt-in behaviour changes behind a flag passed to `buildReflection()`.

I've tried to keep each change incremental so that you can review commit-by-commit and get a clearer picture of each. That includes updates to test results which should show the rough effect of each - I added a couple of new test files that are built with all of the options enabled.

1. I moved `TObjectReflection` up into `ShaderLang.h` and the reflection functions do the same as before - fetching this object and looking up properties - just now it's publicly accessible. This makes adding more interfaces (e.g. buffer blocks) and more reflection properties (e.g. stride) easier without adding lots of new functions for each combination.

2. `buildReflection()` now takes a parameter to control its behaviour. I added a default value to it so that code build against the headers would have no changes required at all but there didn't seem to be precedent for that - I can drop that and let users pass `EShReflectionDefault` manually if you prefer.

3. The change I dropped from the last PR to follow `ARB_program_interface_query`'s rules for arrays-of-struct naming is now behind the `EShReflectionStrictArraySuffix` option, available on the command line via `--reflect-strict-array-suffix`. This will always list an array suffix for arrays-of-structs, except for root members in SSBOs which always expand only `[0]`.

4. According to `ARB_program_interface_query`, plain arrays of base types should always be reflected with a trailing `[0]` to indicate that it's an array, so added `EShReflectionBasicArraySuffix` which enforces that.

5. Added reflection of pipeline outputs as well as inputs (aka vertex attributes). This always happens, but if `EShReflectionIntermediateIO` is set then the inputs and outputs are reflected from the first/last stage available rather than hardcoded from Vertex/Fragment stages. The easiest example is that without the option a program with only a VS won't have any outputs listed, whereas with the option it will.

6. Added reflection of the number of members in a block variable, the array stride of arrays, and the 'top level array stride' for any SSBO member - going hand in hand with the policy of not expanding out any root SSBO member arrays.

7. Added an option `EShReflectionSeparateBuffers` to reflect buffer variables and buffer blocks into separate lists. I think you could implement this just with always in one list and multiple views onto it, but since with the option you want to _exclude_ buffer variables from the uniform list, it becomes more complex than it's worth in my view.

8. Lastly I added `EShReflectionAllBlockVariables` which will always reflect out all variables in a block, regardless of whether they are active/used or not. The variables will have finer-grained masks indicating whether or not they're used in a given stage.

Hopefully it's a reasonable set of changes, the worst of it is handling the weird rules of ARB_program_interface_query in the first place. I'm pretty confident that nothing is changed in a backwards compatible way until you change the `buildReflection()` flags.


johnkslang:
Thanks. Historically, the reflection interface was an add on, somewhat second-class, so I think you're basically making it first class.

> I'm pretty confident that nothing is changed in a backwards compatible way...

> ... opt in ...

If your judgment and legacy testing results show this is okay, maybe we don't need the opt-in aspect. (That is, I'm okay with a taking a small risk here, if you want 😄).

baldurk:
As much as the reflection results are more consistent, it is quite possible that code has been written relying on the format that was coming back in some cases. So far as I know the legacy testing has only shown this to be safe with the opt-out, it hasn't been tested with the changes enabled.

Since it ended up being minimally intrusive to keep them behind options it's probably the safer option for now and I'd prefer to leave the opt-in. It would be easy to remove the options and have some or all of these as opt-in down the road, if that makes sense. You could remove the legacy wrapper functions that I've kept at the same time, since it's easy to rewrite those calls but it would force consumers to look at what has actually changed.

A lot of the awkwardness comes from the way that the reflection only enumerates leaf basic types, with a convention on naming scheme for all the expansions along the way. Since we're not limited to plain GL function calls it might make more sense in future to add more information to `TObjectReflection` so that users can walk the type/object tree themselves. This would be similar to what you can get today with `TType` but it's missing a few key bits of information that the reflection currently calculates as it goes.

baldurk:
Rebased my branch onto latest master, resolved up against #1688 changes which were mostly trivial to account for.

danginsburg:
I regression tested this MR with our shader compilation tool that relies on glslang reflection.  There were no regression with the MR.  If I toggle on the new reflection options, it would require code changes on my side to handle them, so it seems to make sense to put them behind an opt-in to avoid breaking other clients.

That being said, I'd be willing to update my code to conform to the new reflection results if anyone wanted to turn this to the the default  behavior so that glslang more closely matched what drivers do in ARB_program_interface_query.

baldurk:
Yeh I also think in this scenario having different spellings is likely just to cause confusion for anyone coming to the interface new. In the end having flags on `buildReflection` and using them internally wasn't onerous so I think it's a good trade-off.

glslang/Public/ShaderLang.h
```diff
@@ -599,6 +611,41 @@ class TShader {
...
+    int glDefineType;
+    int size;                   // data size in bytes for a block, array size for a (non-block) object that's an array
+    int index;
+    int counterIndex;
+    int numMembers;
+    int arrayStride;            // stride of an array variable
+    int topLevelArrayStride;    // stride of the top-level variable in a storage buffer member
+    EShLanguageMask stages;
+
+protected:
```
johnkslang: Can copies of this be made, or should there also be a protected copy constructor and `operator=`? Some compilers will warn about this (unless copies work and are intended).

StandAlone/StandAlone.cpp
```diff
@@ -1518,6 +1529,12 @@ void usage()
            "  --invert-y | --iy                 invert position.Y output in vertex shader\n"
            "  --keep-uncalled | --ku            don't eliminate uncalled functions\n"
            "  --no-storage-format | --nsf       use Unknown image format\n"
+           "  --reflect-strict-array-suffix     use strict array suffix rules when reflecting\n"
+           "  --reflect-basic-array-suffix      arrays of basic types will have trailing [0]\n"
+           "  --reflect-intermediate-io         reflection includes inputs/outputs of linked shaders\n"
+           "                                    rather than just vertex/fragment\n"
+           "  --reflect-separate-buffers        reflect buffer variables and blocks separately to uniforms\n"
+           "  --reflect-all-block-variables     reflect all variables in blocks, whether inactive or active\n"
```
johnkslang: I think this one is 96 columns wide, but the existing usage (barring defects) has been kept to fit within 80 columns without wrapping.

We should decide if the 80 should be made bigger, or just fix the new lines to stay within 80.

glslang/MachineIndependent/reflection.cpp
```diff
@@ -107,8 +108,36 @@ class TReflectionTraverser : public TLiveTraverser {
 
             TReflection::TNameToIndex::const_iterator it = reflection.nameToIndex.find(name.c_str());
             if (it == reflection.nameToIndex.end()) {
-                reflection.nameToIndex[name.c_str()] = (int)reflection.indexToAttribute.size();
-                reflection.indexToAttribute.push_back(TObjectReflection(name.c_str(), type, 0, mapToGlType(type), 0, 0));
+                reflection.nameToIndex[name.c_str()] = (int)reflection.indexToPipeInput.size();
+                reflection.indexToPipeInput.push_back(TObjectReflection(name.c_str(), type, 0, mapToGlType(type), 0, 0));
```
johnkslang: I can't see what changed in these lines.

glslang/Public/ShaderLang.h
```diff
@@ -689,27 +718,117 @@ class TProgram {
...
+    int getNumAttributes() const;
+    const TObjectReflection& getAttribute(int index) const;
+
+    // Legacy Reflection Interface - expressed in terms of above interface
+    int getNumLiveUniformVariables() const                 // can be used for glGetProgramiv(GL_ACTIVE_UNIFORMS)
+    {
+        return getNumUniformVariables();
+    }
+
+    int getNumLiveUniformBlocks() const                    // can be used for glGetProgramiv(GL_ACTIVE_UNIFORM_BLOCKS)
```
johnkslang: This is a 5X expansion over the normal style: only going to a multi-line block when there is more than one line of code (or the line is too long).  I see the comments are in the way for that, but since it's our interface file, I'd still like to see less expansion here.

How about the following style?

```
// comment...
int name() const { return implementation(); }

// comment
int name() const { return implementation(); }

...
```

johnkslang: I think that also makes it easier "to see" the mapping getting done, via visually seeing the columns.

johnkslang:
This looks really good. I had just a few superficial things to address (some on the commits, which did not show up here). Let's do one round of fixing those and then I'll merge. Thanks.

StandAlone/StandAlone.cpp
baldurk: I will re-wrap these at 80, there's no reason for them to go over. I think I was thinking of the source column limit and not of how the actual terminal output would look.

glslang/MachineIndependent/reflection.cpp
baldurk: `indexToAttribute` -> `indexToPipeInput` purely a naming change for better symmetry/readability (since technically I don't think anything except a vertex shader input is an 'attribute'.

glslang/Public/ShaderLang.h
baldurk: That's a good question/ I'm not sure, I moved this as-is from `reflection.h`. Looking at it closer, there's no reason in principle why you couldn't copy it, as it would all work. I think you could make the argument either way.

One thing I did notice while looking though that's independent of the resotution of the above - the constructor takes a copy of the input type via `pType.clone()` but doesn't ever seem to delete it. This is true on HEAD code as well before my PR. Is there some way those `TType`s are pooled and released? Or is this a leak and I should add a destructor.

glslang/Public/ShaderLang.h
baldurk: That's fine with me, I think I tried to err on the side of following existing style / clang-format rules for new code.

glslang/MachineIndependent/reflection.cpp
johnkslang: Ah, sorry I missed that. No problem.

glslang/Public/ShaderLang.h
johnkslang: Ah, if it's already been there without warnings, it's okay. It does depend on what the compiler thinks it can generate.

Yes, TType has the POOL_ALLOCATOR_NEW_DELETE in it, so it all gets freed on a single pop later on.

baldurk:
Pushed a new commit with the reflow of parameter help lines and the legacy wrapping public interface. If you prefer I can rebase & squash those changes into the original commits before you merge, I figured I'd leave it separate at first.

johnkslang:
Looks good thanks.

I wouldn't want to squash this myself, as having multiple commits is rational.  Fewer might be nice, but I think this is okay.

PR #1686, opened 2019-01-30
Port Unix/ossource.cpp to Fuchsia

mspang:
Fuchsia has pthreads but doesn't have sys/resource.h and doesn't
implement pthread_cleanup_push().

ANGLE bug: 2475


null77:
@johnkslang can you PTAL?

PR #1697, opened 2019-02-11
Fix resizing of gl_PrimitiveIndicesNV[] to max_primitives*geomSize

sparmarNV:
- This change also allows redeclaration of gl_PrimitiveIndicesNV and
  adds error checks against incorrect explicit array size.
- Also modifies gtests to check array bound limits and redeclare gl_PrimitiveIndicesNV[].

This fixes issue #1697


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3783,7 +3792,7 @@ void TParseContext::declareArray(const TSourceLoc& loc, const TString& identifie
     existingType.updateArraySizes(type);
 
     if (isIoResizeArray(type))
-        checkIoArraysConsistency(loc, false, type.getQualifier().isPerPrimitive());
+        checkIoArraysConsistency(loc, symbol->getType().getQualifier(), true);
```
sparmarNV: Note: I have changed the value of "tailOnly" to true, so that it only checks the latest array declaration.

johnkslang: Why is this necessary?

The first half of the function declares a new variable, and just checks the tail. But this half of the function is for a redeclaration of something, and maybe needs to recheck some earlier ones? I haven't (re)figured it all out yet, but it feels slightly dangerous, so want to know what motivates the change.

johnkslang:
Generally looks okay, other than that one change about checking the tail only.

glslang/MachineIndependent/ParseHelper.cpp
sparmarNV: Let me try to explain with a shader example.
Consider below mesh shader with I/O array re-declarations -

out gl_MeshPerVertexNV {
    vec4  gl_Position;
} gl_MeshVerticesNV[MAX_VERT];

perprimitiveNV out gl_MeshPerPrimitiveNV {
    int gl_Layer;
} gl_MeshPrimitivesNV[MAX_PRIM];

out uint gl_PrimitiveIndicesNV[MAX_PRIM*3];

- All 3 I/O arrays have different sizes.

- For the first 2 block arrays gl_MeshVerticesNV and gl_MeshPrimitivesNV, we call redeclareBuiltinBlock() which in turn calls checkIoArraysConsistency() with tailOnly=true.
So we correctly check only 1 block at a time with requiredSize = MAX_VERT and MAX_PRIM respectively.

- For uint array gl_PrimitiveIndicesNV, declareVariable() first calls redeclareBuiltinVariable() which adds the arraySym to ioArraySymbolResizeList and then calls declareArray() which in turn calls checkIoArraysConsistency() with tailOnly = false (before my change).
As a result we check all the 3 I/O arrays (parsed so far) with "requiredSize = MAX_PRIM * 3" and incorrectly error out the first 2 block declarations with incorrect sizes.

- This is not seen in any other shader stages because sizes for all I/O arrays within a shader stage are same.
- I can change the tailOnly value back to false and special handle this code for mesh shaders. However, as declareVariable()/declareArray() only handles 1 non-block variable at a time, checking for the tail (latest) symbol makes perfect sense.

Let me know if this all makes sense.

johnkslang: Per class of autosizing (vertex and primitive?), the sizes need to be consistent, so a redeclaration requires revisiting whatever had once been consistency-checked before the redeclaration.

One thing that happens with earlier arrays that are declared unsized, is that they are assigned a size once a size is known. A redeclaration will go back and resize them to the new size, though I suspect that stops happening with this change of visiting only the tail.

sparmarNV: >Per class of autosizing (vertex and primitive?), the sizes need to be consistent, so a redeclaration requires revisiting whatever had once been consistency-checked before the redeclaration.

This is not true for mesh shaders, where we have different sizes for per-vertex/per-primitive out block arrays. getIoArrayImplicitSize() takes care of assigning correct array sizes for autosizing.

>One thing that happens with earlier arrays that are declared unsized, is that they are assigned a size once a size is known. A redeclaration will go back and resize them to the new size, though I suspect that stops happening with this change of visiting only the tail.

Shouldn't each I/O array redeclaration take care of its own resizing? Why should declareArray(), which is suppose to (re)declare 1 array symbol, resize all other arrays?
Nonetheless, I have reset the value of tailOnly to false.

Also I have special handled the mesh shader case in checkIoArraysConsistency() and added code in updateStandaloneQualifierDefaults() to update all the array sizes when all the necessary qualifiers are defined.

johnkslang: I recall some implicitly-sized arrays were implicitly sized based on (or consistency checked against) the size of others. A declaration affects that, and not just the last one, but potentially others.

See the comment for `ioArraySymbolResizeList` which mentions retroactive array resizing.  Maybe you found an efficiency, but I'm not sure yet. For now, let's go the conservative route.

PR #1698, opened 2019-02-18
Add Float16/Int8/Int16 capabilities for private variables and function parameters

jeffbolznv:


johnkslang:
How does doing this here contrast to using the post process, and/or distinguishing from load/store versus declaration versus use? I think there are some pending discussions about aspects of this, which are fine to continue/resolve here.

jeffbolznv:
For function parameters, it is probably equally simple to put it in the postprocess or during traversal.

For the private storage class variables, the postprocess code currently only catches load/store instructions, and this test shader has no load/store. I figured it's easy and consistent to catch those at the time they are created, similarly to how other storage classes are handled.

johnkslang:
I think the trouble comes from not wanting to catch declarations, but to catch actual use. There is some question here about a declaration only, with no other use, turning on the right (or none) capability, which is why the post-process model was switched to, including possibly giving a chance to optimize some things before making the decision.

jeffbolznv:
I'm happy to move it to the post process if that's what you prefer. But do you agree that these unused variables should require the capabilities to be set?

johnkslang:
We had some inversion in the rules, where first we said a declaration is all that's needed to require a non-storage-only capability, but then if the only use was to load/store, only a storage capability was needed, and if a further use beyond load/store was done, then you need the non-storage-only capability. Something is logically inconsistent about that, somewhat arguing that with a declaration only, you don't know what capability to declare.

Maybe adding the subtlety of which storage class is involved resolves it, but I think the spec should get fixed here.

jeffbolznv:
I don't know what spec change you would want. The spec currently says things like "StorageBuffer16BitAccess: Allows 16-bit OpTypeFloat and OpTypeInt for the sole purpose of creating an OpTypePointer to a 16-bit floating-point or 16-bit integer member of an object. The object must be in the StorageBuffer Storage Class, or be in the Uniform storage class and have the BufferBlock decoration". It's clear that Private variables are outside of that, and so they need Int16/Float16 capabilities.

PR #1699, opened 2019-02-18
Handle buffer references vs 'const'

jeffbolznv:
Allow constructors to and from references to be constant folded. Section 4.3.3
says constructors whose arguments are all constant expressions must fold.

Disallow 'const' on buffer reference types. It is not a 'non-void transparent
basic data type' (it is not considered 'basic').

Handle buffer reference constants (which can be assigned to a non-const reference,
or can be further folded to another type of constant) by converting to
'constructor(uint64_t constant)' in addConversion.

Disallow == and != operators on reference types.

See also https://github.com/KhronosGroup/GLSL/pull/54

(set WIP just to give time for review)


johnkslang:
One of the Travis-CI builds hung, hopefully unrelated?

Looks okay.  (I normally don't review while something is still WIP, unless pinged about it.)

Let me know when it is de-WIP.

jeffbolznv:
Rebased and removed WIP (@johnkslang)

PR #1702, opened 2019-02-20
Update SPIRV-Tools known good

greg-lunarg:


greg-lunarg:
> Are we aiming to include SPV_NV_cooperative_matrix or not? It is in the middle of the process of getting into all the components.

It doesn't look like its in spirv-tools stable. I also didn't see it in TOT glslang just now (or did I miss something). So it seems the answer would be that it will not make it into this SDK. Is that going to be a problem?

johnkslang:
It's a pending PR; I don't know if you want TOT to stay stable and free of it, or will you grab the right commit you want, independent of whether it gets de-WIP and merged?

greg-lunarg:
At this point we are not looking to support it in this SDK and probably would prefer to be free of it. So we will probably grab a version of glslang that does not have it yet.

PR #1705, opened 2019-02-21
Fix type recursion with EOpIndexIndirect dereferences

baldurk:
I found this because I changed `isReflectionGranularity` to correctly count arrays-of-arrays as not yet fully reflected, however it can happen in specific other cases with arrays-of-structs (as I've added a test for).

In `blowUpActiveAggregate` when walking the list of dereferences in reflection to determine what active uniforms to reflect, the `terminalType` starts at the passed in `baseType` and then is updated with `visitNode->getType()` for each index. This means the dereference from the indirection is already accounted for. For direct indexes the type doesn't change, all that happens is the `offset` and `name` are adjusted to account for the index. For an indirect indexing into an array we go wide and recurse for the remaining dereferences over each possible index that could be.

The problem is that when we do this recursion, there's an incorrect extra dereference `TType derefType(*terminalType, 0);` before recursing, even though `visitNode->getType()` has already accounted for it. On multi-dimensional arrays this means one level is skipped, for arrays-of-structs it can mean we accidentally step into the first child instead of ending up at the struct.

Most of the time this doesn't cause problems, because if we continue to walk a list of dereferences the problem is hidden/corrected by the subsequent `visitNode->getType()` getting things back on track. However if the _last_ dereference is `EOpIndexIndirect` then the recursive call will have `baseType` be wrong for the rest of the processing down to reflection granularity.

This is easiest to see with multi-dimensional arrays, but you can also reflect this shader:

```
#version 440 core
struct foo {
    float a;
    vec2 b;
};
struct bar {
    foo x[4];
};
uniform bar root;
void main() {
    foo tmp = root.x[gl_InstanceID];
}
```

And without this PR you will get reflection such as this:

```
Uniform reflection:
root.x[0]: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[1]: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[2]: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[3]: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
```

Because the extra dereference has already stepped into the first float child of the `struct foo`, before we decide to process the struct copy itself. With this change, we get more correctly:

```
Uniform reflection:
root.x[0].a: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[0].b: offset -1, type 8b50, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[1].a: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[1].b: offset -1, type 8b50, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[2].a: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[2].b: offset -1, type 8b50, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[3].a: offset -1, type 1406, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
root.x[3].b: offset -1, type 8b50, size 1, index -1, binding -1, stages 1, topLevelArrayStride 16
```

The test change in `reflection.vert` exercises this aspect, where the test change in `reflection.options.vert` tests the multi-dimensional array case.

### DIFF
--- a/Test/baseResults/reflection.options.vert.out
+++ b/Test/baseResults/reflection.options.vert.out
@@ -6,9 +6,33 @@ UBO.verts[1].position[0]: offset 24, type 1406, size 3, index 0, binding -1, sta
 UBO.verts[1].normal[0]: offset 36, type 1406, size 3, index 0, binding -1, stages 0, arrayStride 4, topLevelArrayStride 24
 UBO.flt[0]: offset 48, type 1406, size 8, index 0, binding -1, stages 1, arrayStride 4, topLevelArrayStride 4
 UBO.unused: offset 80, type 8dc8, size 1, index 0, binding -1, stages 0
+UBO.uniform_multi[0][0][0]: offset 96, type 1406, size 2, index 0, binding -1, stages 1, arrayStride 4, topLevelArrayStride 24
+UBO.uniform_multi[0][1][0]: offset 104, type 1406, size 2, index 0, binding -1, stages 0, arrayStride 4, topLevelArrayStride 24
+UBO.uniform_multi[0][2][0]: offset 112, type 1406, size 2, index 0, binding -1, stages 0, arrayStride 4, topLevelArrayStride 24
+UBO.uniform_multi[1][0][0]: offset 120, type 1406, size 2, index 0, binding -1, stages 0, arrayStride 4, topLevelArrayStride 24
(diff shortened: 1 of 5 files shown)


johnkslang:
Thanks for figuring out these tough details, @baldurk.

PR #1707, opened 2019-02-21
Remove references to clang-format.

dj2:
The glslang code is not formatted by clang. This CL removes references
to clang-format to reduce confusion.


johnkslang:
I thought we'd made a pretty good clang format to reasonably match glslang code, which people could use on new code they wrote, or areas of heavy modification, just not unthinkingly wholesale to the entire code base.

I don't think we need to lose this information.

dj2:
ok. closing.

PR #1723, opened 2019-03-11
GN: Make glslang_validator an executable.

null77:
The target was accidentally listed as a source_set.

@johnkslang PTAL


null77:
Thanks!

PR #1728, opened 2019-03-14
Error out writes to shaderRecordNV buffer blocks.

alelenv:
In reference to spec clarification
https://github.com/KhronosGroup/GLSL/pull/60

@johnkslang for review and merge
Thanks!


johnkslang:
Looks good.

PR #1730, opened 2019-03-16
Fix tessellation control shader bounding box support.

laurie-hedge:
Prior to this change, OES_primitive_bounding_box and EXT_primitive_bounding_box were both recognised as extensions, but only the name gl_BoundingBoxOES could be used. However the EXT version uses the name gl_BoundingBoxEXT instead. In addition, since GLES 3.2, the extension has been included in the core standard and the name gl_BoundingBox may be used instead. This change aims to make both extensions and the 3.2 core version all work.

All my real changes are in Initialize.cpp. The rest is just updating the test framework to cover uses of all three variable names.

One thing I'd appreciate reviewing is whether my change to replace
`symbolTable.setVariableExtensions("gl_BoundingBoxOES", Num_AEP_primitive_bounding_box, AEP_primitive_bounding_box);`
with
`symbolTable.setVariableExtensions("gl_BoundingBoxOES", 1, &E_GL_OES_primitive_bounding_box);`
is correct. As far as I can see, there is nothing in the spec that suggests requesting EXT_primitive_bounding_box should allow usage of gl_BoundingBoxOES etc., but I'm not certain what the convention is about duplicate extensions like this.

The issue was first noticed trying to compile shaders from the gfxbench test gl_4, which utilises gl_BoundingBoxEXT.

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -5794,7 +5794,14 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "patch out highp float gl_TessLevelOuter[4];"
             "patch out highp float gl_TessLevelInner[2];"
             "patch out highp vec4 gl_BoundingBoxOES[2];"
+            "patch out highp vec4 gl_BoundingBoxEXT[2];"
             "\n");
+        if (profile == EEsProfile && version >= 320) {
+            stageBuiltins[EShLangTessControl].append(
+                "patch out highp vec4 gl_BoundingBox[2];"
+                "\n"
+            );
+        }
     }
 
     if ((profile != EEsProfile && version >= 140) ||
@@ -8014,10 +8021,18 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
 
     case EShLangTessControl:
         if (profile == EEsProfile && version >= 310) {
+            BuiltInVariable("gl_BoundingBoxEXT", EbvBoundingBox, symbolTable);
+            if (version < 320)
+                symbolTable.setVariableExtensions("gl_BoundingBoxEXT", 1,
+                                                  &E_GL_EXT_primitive_bounding_box);
             BuiltInVariable("gl_BoundingBoxOES", EbvBoundingBox, symbolTable);
             if (version < 320)
-                symbolTable.setVariableExtensions("gl_BoundingBoxOES", Num_AEP_primitive_bounding_box,
-                                                  AEP_primitive_bounding_box);
+                symbolTable.setVariableExtensions("gl_BoundingBoxOES", 1,
+                                                  &E_GL_OES_primitive_bounding_box);
+
+            if (version >= 320) {
+                BuiltInVariable("gl_BoundingBox", EbvBoundingBox, symbolTable);
+            }
         }
 
         // Fall through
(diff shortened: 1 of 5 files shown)


johnkslang:
>  As far as I can see, there is nothing in the spec that suggests requesting EXT_primitive_bounding_box should allow usage of gl_BoundingBoxOES...

I don't follow. EXT_primitive_bounding_box clearly discusses, in many places, gl_BoundingBoxEXT, as does OES_primitive_bounding_box.

When multiple extensions enable the same feature, you only need one of them, and that is how the list mechanism is getting used in glslang.

Thank you though for catching the missing spelling.  Can you update based on the above?

laurie-hedge:
> I don't follow. EXT_primitive_bounding_box clearly discusses, in many places, gl_BoundingBoxEXT, as does OES_primitive_bounding_box.
I think I must be looking at the wrong spec. The ones I've been referring to are:

- EXT_primitive_bounding_box: https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_primitive_bounding_box.txt
- OES_primitive_bounding_box: https://www.khronos.org/registry/OpenGL/extensions/OES/OES_primitive_bounding_box.txt

As far as I can see, each spec only refers to its own spelling of the builtin variable (gl_BoundingBoxEXT and gl_BoundingBoxOES respectively), making no reference to the other. Is it implicit that enabling either extension should expose both spellings? Or is there a more up to date spec?

johnkslang:
I understand now. No, normally enabling one should not introduce the spellings of the other, unless with a promotion from EXT to OES we are trying to include the previous? Do all the OES promotions have this issue, or is it really isolated to just primitive-bounding-box?

laurie-hedge:
From inspecting the code, I don't see any other similar examples where alternate spellings exist. The more common pattern for extensions seems to be to omit the suffix entirely (e.g. EXT_geometry_shader and OES_geometry_shader both introduce their builtins by the same spelling - gl_PrimitiveID and gl_Layer).

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8014,10 +8021,18 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
         if (profile == EEsProfile && version >= 310) {
+            BuiltInVariable("gl_BoundingBoxEXT", EbvBoundingBox, symbolTable);
+            if (version < 320)
+                symbolTable.setVariableExtensions("gl_BoundingBoxEXT", 1,
+                                                  &E_GL_EXT_primitive_bounding_box);
             BuiltInVariable("gl_BoundingBoxOES", EbvBoundingBox, symbolTable);
             if (version < 320)
-                symbolTable.setVariableExtensions("gl_BoundingBoxOES", Num_AEP_primitive_bounding_box,
-                                                  AEP_primitive_bounding_box);
+                symbolTable.setVariableExtensions("gl_BoundingBoxOES", 1,
```
johnkslang: Did 320 make it illegal to use the extension?

laurie-hedge: I don't believe the 3.2 spec makes any mention of the extensions at all. My understanding is that this check means that the EXT and OES spellings can be used without the extension in 320. It makes some intuitive sense, but I can't see any spec language to justify this. Do you think it should be changed?

johnkslang: 320 does not define such spellings. You still need the extension to access those spellings.

As you noted earlier, the template for all these extensions seems to be for suffix-less spellings, and that's why this one is broken in general, because it did not following the template.

I think the if tests should be for `>=` the minimum supported version.

laurie-hedge: Makes sense. I've included the fix in a second commit to the same branch, which should now be part of this pull request. Again most of the changes are in the unit tests to verify the behaviour. The real change is just removing a couple of lines from Initialize.cpp.

PR #1741, opened 2019-03-26
Enable GL_KHR_shader_subgroup properly

dgkoch:
(and GL_NV_shader_subgroup_partitioned) based on GL/ES version
instead of predicating it on vulkan SPV generation

Also add AST testing.
The glsl.450.subgroup* files are largely the same as the spv.subgroup*
The glsl.es320.subgroup* files are the same as the 450 versions, but modified to be ES compatible.


dgkoch:
cc @johnkslang

PR #1748, opened 2019-04-03
Casts to enable capabilities in spir-v output

jsmall-nvidia:
First to explain the origin of the need for this fix, when compiling a particular shader the spir-v produced would produce validation errors

```
[ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object: VK_NULL_HANDLE (Type = 0) | SPIR-V module not valid: Using a 16-bit floating point type requires the Float16 or Float16Buffer capability, or an extension that explicitly enables 16-bit floating point.
 %half = OpTypeFloat 16
```

An example shader that produces this would be

```#version 450
layout(row_major) uniform;
layout(row_major) buffer;
#extension GL_EXT_shader_16bit_storage : require
#extension GL_KHX_shader_explicit_arithmetic_types : require

layout(rgba16f)
layout(binding = 1, set = 7)
uniform image2D gOutput;

layout(local_size_x = 16, local_size_y = 16, local_size_z = 1) in;
void main()
{
    ivec2 pos = ivec2(gl_GlobalInvocationID);
    vec4 v = ivec4(pos.x, pos.y, pos.x, pos.y) * (1.0f / 16.0f);
    imageStore(gOutput, ivec2(pos), f16vec4(v));
} 
```
Other shaders that use float16_t do not have the problem. On looking into the spir-v assembler the difference is the failing shader does not have `Capability Float16` in it's spir-v code.

The only float16_t instruction is the f16vec4 cast. On looking at the glslang source the method `void spv::Builder::postProcessType(const Instruction& inst, Id typeId)` looks at types on instructions to see about adding capabilities such as with addCapability(CapabilityFloat16);

That in that code we have the sequence
```
    case OpFConvert:
    case OpSConvert:
    case OpUConvert:
        break;
```
 
Which leads to the float16_t cast being ignored from the point of view of enabling capabilities. 

The PR removes these 3 cases, such that the `default:` path is taken - and this path does notice the float16_t usage and the `Capability Float16` line appears in the spir-v for the above shader.

### DIFF
--- a/SPIRV/SpvPostProcess.cpp
+++ b/SPIRV/SpvPostProcess.cpp
@@ -118,9 +118,6 @@ void Builder::postProcessType(const Instruction& inst, Id typeId)
     case OpAccessChain:
     case OpPtrAccessChain:
     case OpCopyObject:
-    case OpFConvert:
-    case OpSConvert:
-    case OpUConvert:
         break;
     case OpExtInst:
 #if AMD_EXTENSIONS
--- a/Test/baseResults/spv.1.3.8bitstorage-ssbo.vert.out
+++ b/Test/baseResults/spv.1.3.8bitstorage-ssbo.vert.out
@@ -4,6 +4,7 @@ spv.1.3.8bitstorage-ssbo.vert
 // Id's are bound by 28
 
                               Capability Shader
+                              Capability Int8
                               Capability CapabilityStorageBuffer8BitAccess
                               Extension  "SPV_KHR_8bit_storage"
                1:             ExtInstImport  "GLSL.std.450"
--- a/Test/baseResults/spv.1.3.8bitstorage-ubo.vert.out
+++ b/Test/baseResults/spv.1.3.8bitstorage-ubo.vert.out
@@ -4,6 +4,7 @@ spv.1.3.8bitstorage-ubo.vert
 // Id's are bound by 29
 
                               Capability Shader
+                              Capability Int8
                               Capability CapabilityUniformAndStorageBuffer8BitAccess
                               Extension  "SPV_KHR_8bit_storage"
                1:             ExtInstImport  "GLSL.std.450"
--- a/Test/baseResults/spv.16bitstorage-int.frag.out
+++ b/Test/baseResults/spv.16bitstorage-int.frag.out
@@ -4,6 +4,7 @@ spv.16bitstorage-int.frag
 // Id's are bound by 172
 
                               Capability Shader
+                              Capability Int16
                               Capability StorageUniformBufferBlock16
                               Capability StorageUniform16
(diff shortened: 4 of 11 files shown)


johnkslang:
When only casting, for load and store to memory, a "weaker" capability is needed.  We need to be careful to not require a stronger capability than required.

tfoleyNV:
Hi @johnkslang.

> When only casting, for load and store to memory, a "weaker" capability is needed. We need to be careful to not require a stronger capability than required.

So is it the `Float16Buffer` capability that is needed in this case? That is the other capability mentioned in the SPIR-V validation error that motivated the PR.

Also, can you clarify what you mean by "when only casting, for load and store to memory"? My reading is that you mean that an `Op*Convert` where the source operand is a load from memory and/or where the only use is a store to memory should *not* add the stronger `Float16` capability. Is that correct?

jeffbolznv:
John is talking about the capabilities like StorageBuffer16BitAccess. However, those are only for buffer accesses, if this is being used with OpImageWrite then it would be correct to require the Float16 capability.

Given that this change is adding the capability to the existing storage tests, I think it's probably too aggressive about adding the capability. The 8/16-bit storage extension behavior is really unfortunate IMO, it requires so much careful analysis to set the right capability.

It would help to post the whole SPIR-V blob, to see what's going on. I'm surprised that the default case in the switch didn't add the Float16 capability, if it is used by OpImageWrite. I'm guessing glslang is converting it back to float32, so the only use of float16 is to convert back and forth, and OpImageWrite uses float32?

jeffbolznv:
Yeah, I compiled the shader and it's only used by converts:

```
         %51 = OpFConvert %v4half %48
         %52 = OpFConvert %v4float %51
               OpImageWrite %43 %47 %52
```

So this shader would be valid with either Float16 or any of the SPV_KHR_16bit_storage capabilities, and it's unclear which one it should enable.

FWIW, for the original shader, the easiest way to avoid this issue would be to remove the f16vec4 cast.

dneto0:
> So this shader would be valid with either Float16 or any of the SPV_KHR_16bit_storage capabilities, and it's unclear which one it should enable.

Yes, it would be valid with the full Float16 capability.  The SPV_KHR_16bit_storage capabilities would be sufficient if the conversion *to* a 16bit scalar or vector type is only consumed by a store to one of the listed (buffer) storage classes.  But in this case the conversion is being consumed by at second conversion back to vec4.

dneto0:
Oh dear, I was remembering incorrectly.  Those restrictions I was pointing out is only where the rounding modes were enabled by the extensions.  Whoops.

johnkslang:
Closing, in favor of #1758.

PR #1749, opened 2019-04-03
Check if _MSC_VER is defined before using.

dgkoch:
Fixes compilation errors on some platforms.


jeffbolznv:
There is a similar line in SPVRemapper.h that should be changed:

```
#if __cplusplus >= 201103L || _MSC_VER >= 1700
```

dgkoch:
> There is a similar line in SPVRemapper.h that should be changed:

Done.

dgkoch:
cc @johnkslang

PR #1751, opened 2019-04-05
Unconditionally enable to_string WAR

dgkoch:
unconditionally define to_string() as not all "\<string\>" libraries have it enabled by default. 

It turns out that prior to https://github.com/KhronosGroup/glslang/pull/1749 some platforms (notably QNX at a minimum) were only working by virtue of _MSC_VER being treated as 0 by the preprocessor when not defined.

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -37,8 +37,7 @@
 #ifndef _COMMON_INCLUDED_
 #define _COMMON_INCLUDED_
 
-
-#if defined(__ANDROID__) || (defined(_MSC_VER) && _MSC_VER < 1700)
+// hack to define to_string as not all <string> libraries have it enabled by default
 #include <sstream>
 namespace std {
 template<typename T>
@@ -48,7 +47,6 @@ std::string to_string(const T& val) {
   return os.str();
 }
 }
-#endif
 
 #if (defined(_MSC_VER) && _MSC_VER < 1900 /*vs2015*/) || defined MINGW_HAS_SECURE_API
     #include <basetsd.h>


johnkslang:
But, this was working on platforms that already have it included?

Does this make it always replace what's there, when something is there?

dgkoch:
(updated description)

We tested it on a modern VS (VS2017), and it doesn't seem to hurt to have this there unconditionally. 
I'm still awaiting full result from our internal build farm, which is why it's still WIP.  The previous version failed on some platforms.

dgkoch:
de-WIPing it.  This passes on our build farm.

dgkoch:
> Does this make it always replace what's there, when something is there?

Before the change in #1749 the to_string override was already (inadvertently?) being enabled on all non-MSVC platforms and as far as I can tell it wasn't doing any harm. Instead of adding more OSes to the exception it seemed cleaner just to enable it all the time. If you want something different I'm open to suggestions.  

The specific failure I'm seeing with ToT is: 
```
glslang/MachineIndependent/iomapper.cpp: In member function 'virtual int glslang::TDefaultIoResolverBase::resolveSet(EShLanguage, const char*, const glslang::TType&, bool)':
glslang/MachineIndependent/iomapper.cpp:433:59: error: 'atoi' was not declared in this scope
              return atoi(getResourceSetBinding()[0].c_str());
```
And seems to be prompted and fixed by tweaking this to_string override (possibly as a side effect of including \<sstream\>?).

turol:
This is [forbidden by the C++ standard](https://stackoverflow.com/questions/36533199/is-specialization-of-stdto-string-for-custom-types-allowed-by-the-c-standard/36533238)  and will likely cause problems in the future.

dgkoch:
But that only seems to apply if adding it in the std:: namespace.

dgkoch:
My alternate solution in #1752 was taken, so I'm going to close this one. 

I still think this should be cleaned up, but I'm not planning to push it forward. 

The alternative is that this overload just be completely removed. I suspect that the problematic versions of Android and MSVC aren't even supported at this point.

PR #1752, opened 2019-04-06
Add cstdlib include

dgkoch:
On some platforms some of the includers of this header don't have
'atoi' defined after the change in #1749


dgkoch:
this one should be less controversial than #1751

dgkoch:
cc @johnkslang

PR #1753, opened 2019-04-16
introduce ability to dump builtin symbol table

pixeljetstream:
introduces EShMsgBuiltinSymbolTable and corresponding --dump-builtin-symbols command line to output the builtin symbols with all details. The intention is for tools that may want to use the data for auto-completion and tooltips or create the ability to provide overloaded functions.

TSymbol::dump functions have option to do "complete" print
bugfix in TType::getCompleteString, structure can be null for block

### DIFF
--- a/glslang/MachineIndependent/SymbolTable.cpp
+++ b/glslang/MachineIndependent/SymbolTable.cpp
@@ -176,37 +176,78 @@ void TType::buildMangledName(TString& mangledName) const
 // Dump functions.
 //
 
-void TVariable::dump(TInfoSink& infoSink) const
+void TSymbol::dumpExtensions(TInfoSink &infoSink) const
 {
-    infoSink.debug << getName().c_str() << ": " << type.getStorageQualifierString() << " " << type.getBasicTypeString();
-    if (type.isArray()) {
-        infoSink.debug << "[0]";
+    int numExtensions = getNumExtensions();
+    if (numExtensions)
+    {
+        infoSink.debug << " <";
+        for (int i = 0; i < numExtensions; i++)
+        {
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/SymbolTable.cpp
```diff
@@ -176,37 +176,79 @@ void TType::buildMangledName(TString& mangledName) const
...
+
+        if (type.isArray())
+            infoSink.debug << "[0]";
+    }
+
     infoSink.debug << "\n";
 }
 
-void TFunction::dump(TInfoSink& infoSink) const
+void TFunction::dump(TInfoSink &infoSink, bool complete) const
```
johnkslang: The convention is to keep the reference (&) with the type, not the variable.

glslang/MachineIndependent/SymbolTable.cpp
```diff
@@ -176,37 +176,79 @@ void TType::buildMangledName(TString& mangledName) const
...
+        for (int i = 0; i < numParams; i++) {
+            const TParameter &param = parameters[i];
+            infoSink.debug << param.type->getCompleteString() << " "
+                           << (param.type->isStruct() ? "of " + param.type->getTypeName() + " " : "")
+                           << (param.name ? *param.name : "") << (i < numParams - 1 ? "," : "");
+        }
+
+        infoSink.debug << ")";
+        dumpExtensions(infoSink);
+    } else
```
johnkslang: Convention is `} else {`.

johnkslang:
Looks good. Just need a couple tweaks to match coding convention.

PR #1755, opened 2019-04-25
Fix issues of explicit conversions.

amdrexu:
For certain int8/int16/float16 conversions,they are conditionally allowed when extensions defining explicit arithmetic types are enabled.

### DIFF
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -498,6 +498,58 @@ TIntermTyped* TIntermediate::createConversion(TBasicType convertTo, TIntermTyped
 
     TOperator newOp = EOpNull;
 
+    // Certain explicit conversions are allowed conditionally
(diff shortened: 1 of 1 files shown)


amdrexu:
I attach a shader for this issue:

```
#version 450 core
#extension GL_EXT_shader_8bit_storage: enable

layout (set = 0, binding = 0) buffer _ {
    float x;
    float y;
    int8_t z;
};

void main()
{
    x = float(z); // Bad
    y = float(int(z)); // Good
}
```
Type conversions are limited to bit width widening and truncating. Mixed type conversions, like integer <-> floating-point, are not allowed by the extension 8bit_storage or 16bit_storage.

amdrexu:
Are there any comments?

johnkslang:
Does this interact with #1758?

amdrexu:
No, they are not the same issues. This issue is caused by doing fp <-> int conversion. We should disallow this unless extensions that introduce explicit arithmetic types are enabled. Otherwise, this is treated as parse error and the compilation is terminated (reporting grammatical error).

PR #1758, opened 2019-05-01
If a shader's only use of float16 is for FConvert, add the Float16 capability

jeffbolznv:
When visiting instructions, check that there are no 16-bit storage capabilities and add the Float16 capability. Same for int8/int16.

This is an alternative solution to https://github.com/KhronosGroup/glslang/pull/1748.


johnkslang:
I'm unclear yet if this will be the final solution, but it does seem better than #1748, so will merge this one and close that one.

Stay tuned for possible further refinements in this space.

PR #1768, opened 2019-05-17
Build.gn: allow optimization in glslang lib and standalone

ShabbyX:


ShabbyX:
@null77 Please take a look as well. You originally added the `glslang_validator` target with `ENABLE_OPT=0`.

ShabbyX:
That was fast, thanks!

null77:
@ShabbyX LGTM, don't specifically recall why I used ENABLE_OPT=0.

PR #1776, opened 2019-05-28
Misc fixes from dgkoch

dgkoch:
1) Fix include guard for GL_EXT_multiview

2) Return consistent names from CapabilityString
 - Don't prefix with "Capability" since the majority of them don't.
- Also add missing CapabilityImageFootprintNV


dgkoch:
cc @johnkslang

Issue #1747, opened 2019-04-03
GLSL vertex attribute overlap


Roy-AMD:
Hi Kessenich,
This issue about GLSL vertex attribute overlap.

Source string see below:
```glsl
#version 430 core
layout(location = 0) in vec4 vs_in_attrib[8][2];
in vec4 vs_in_attrib2[8][1];
out StageData {
  vec4 attrib[16];
} vs_out;
void main() {
  for (int i = 0; i < vs_in_attrib.length(); ++i) {
    vs_out.attrib[i] = vs_in_attrib[i][0] + vs_in_attrib2[i][0];
  }
}
```
The location of vs_in_attrib and vs_in_attrib2 are all be set 0, and It's seem no flag to mark the location qualifier whether be specified. 
In this case, the location of vs_in_attrib2 should be 16, If we have a flag (like ```qualifierSpecMask```) to mark it or set ```location = -1``` to indicate it was not be specified.
that can help us handling the location. If we do not do this, it will be very difficult to deal with locations in this case.
I suggest to use a mask to hold this information, it can help us keep compatibility and retaining more qualifier information.
If you think this scenario works, I'll change the code and request a merge.

Thanks
Roy 
form AMD

Canonzki:
It's seem

Canonzki:
I try to use 
```shell
glslangValidator --auto-map-locations -G test.vert 
spirv-cross .\vert.spv --reflect
``` 
but int result, the location still 0:
```json
{
    "entryPoints" : [
        {
            "name" : "main",
            "mode" : "vert"
        }
    ],
    "types" : {
        "_24" : {
            "name" : "StageData",
            "members" : [
                {
                    "name" : "attrib",
                    "type" : "vec4",
                    "array" : [
                        16
                    ]
                }
            ]
        }
    },
    "inputs" : [
        {
            "type" : "vec4",
            "name" : "vs_in_attrib",
            "array" : [
                2,
                8
            ],
            "location" : 0
        },
        {
            "type" : "vec4",
            "name" : "vs_in_attrib2",
            "array" : [
                1,
                8
            ],
            "location" : 0
        }
    ],
    "outputs" : [
        {
            "type" : "_24",
            "name" : "StageData",
            "location" : 0
        }
    ]
}
```

johnkslang:
Because this has to do with inter-stage linking, it is difficult to get all the location stuff automatically, hence that functionality is not fully available.  It is best to use explicit locations in the shader.

[...]

PR #1779, opened 2019-05-29
Automapping opengl location

Roy-AMD:
## 1. Background.

## 2. New Implementation.
## 3. Subsequent


johnkslang:
A lot of this is changes to formatting, spacing, and coding conventions. Can you factor that out, and make this just about the functional change you want to make?

Roy-AMD:
Hi Kessenich,
Sorry, I do not have a independent change on the format, so  it's difficultt to factor out all the format changes, then I manually changed back most format change. iomapper.cpp's changes are very , and some of the code has been moved to iomapper.h, so the results of the diff appear to be a bit confusing, I am sorry, this change will cause a lot of trouble for the review.

Roy-AMD:
Hi Kessenich,
If it possible, can you accelerated code review? Or If you want me make a new clear branch for review? If I can do anything for accelerated the code review, please let me know.
Thanks
Ruoyu

PR #1786, opened 2019-06-03
Support GL_ARB_fragment_shader_interlock

jeffbolznv:
Temporarily WIP while I do some final testing.

Issue #1503, opened 2018-09-21
Delete #ifdefs for NV and AMD extensions


jekstrand:
Having piles of `#ifdef`s in the code for AMD and NV extensions is probably more trouble than it's worth.  It's lots of compile testing and mental "is this actually compiling" overhead for probably very little actual benefit.

johnkslang:
Original intended benefits were:

1. vendor assuming quality of what was inside of #ifdef
1. footprint, for someone wanting the smallest embedded glslang
1. vendor-neutral glslang was buildable

Reason 1 is no longer applicable, due to interactions, and needed high quality independent of the vendor.

Reasons 2 and 3 may warrant community input, but also agreed the relative size of this is somewhat small (though, the symbol table does get quite a bit bigger).

Agreed they should not be part of interfaces to libraries, per #1502.

johnkslang:
In experimenting just now, I was noticing how much data and if-testing is inside these #ifdef, so am sympathetic to consumers wanting a small footprint.

jeffbolznv:
Did you compare the binary size with and without them enabled? Can you share the data?

dgkoch:
ping @johnkslang to provide some data to back up his earlier comments on size. 

(This came up again in the review of #1787)

johnkslang:
A while back, I saw on the order of 10% for static executable size.  I will be trying again soon.  These are growing faster than core.

dgkoch:
I don't doubt that there is a lot of extension functionality, but what should be the definition of "core" for glslang? It supports multiple versions of GLSL, ESSL, and HLSL (and Vulkan semantics) on the front-end and produces multiple versions of SPV on the backend, so I don't know what you would propose be in the "minimal" version of glslang. Presumably different customers would have different needs and trying to meet all of these through ifdefs seems like it will result in a complex mess of unreadable and unmaintainable code.

johnkslang:
So, I dug deeper in this and have an idea to pursue that will address both:

- this issue, reducing/removing #ifdef
- the quality/maintainability issue in Initialize.cpp, whose old design did not scale to the current usage

What I'd like to do is:
- reduce down to one macro name that identifies (still a bit loosely defined) the subset needed for web use; you're either in that small subset, or not
- table the declarations in Initialize.cpp; this helps with both size and maintainability

I spent a day investigating the last thing.  It'll be a bit of work to get it correct.  See following statistics.

[...]

PR #1787, opened 2019-06-03
Support GL_NV_shader_sm_builtins

dgkoch:
GLSL Spec: https://github.com/KhronosGroup/GLSL/pull/73
SPIR-V Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/31


dgkoch:
The first two commits are duplicates of the changes in #1786 so that the build actually works. Once that PR has been merged, we can drop those commits.

dgkoch:
@johnkslang I don't understand why the AppVeyor builds are failing. 

I'm seeing diffs that look like: 
````
With diff:
@@ -216,8 +216,5 @@
 0:24                Constant:
 0:24                  0 (const uint)
-0:24              right-shift ( temp int)
-0:24                'pos' ( in uint)
-0:24                Constant:
-0:24                  2 (const int)
+0:24              'pos' ( in uint)
 0:5  Function Definition: main( ( temp void)
 0:5    Function Parameters: 
````
So it looks like a `>> 2` has gone missing, but I don't know where that could be happening. 
I'm not seeing that in my local builds, and it also didn't happen in Jeff's #1786 change which is using the same version of SPIRV-tools and headers.

**Comment by jeffbolznv** [comment]
> So it looks like a `>> 2` has gone missing, but I don't know where that could be happening.
> I'm not seeing that in my local builds, and it also didn't happen in Jeff's #1786 change which is using the same version of SPIRV-tools and headers.

I debugged this and it's because the new builtin enums make there be more than 128 values in this enum, and they are stored in an 8-bit bitfield and get sign extended and misbehave. Changing these to 9 bits fixes it:

```
    TBuiltInVariable    builtIn   : 8;
    TBuiltInVariable    declaredBuiltIn : 8;
```

**Comment by dgkoch** [comment]
🤦‍♂ 
Thanks Jeff. 

My Mac compiler could handle them just fine it seems :-/
I pushed a fix, and added a static_assert for when we exceed 255 :-0.

**Comment by dgkoch** [comment]
rebased and removed duplicate commits. 
This is ready to be merged now.

**Comment by dgkoch** [comment]
updated to pull the spirv-tools to a version that fully validates SPV_NV_shader_sm_builtins.

**Comment by dgkoch** [comment]
@johnkslang please review/merge.

File: glslang/MachineIndependent/Initialize.cpp [hunk 1]
```diff
@@ -9024,6 +9083,16 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
             BuiltInVariable("gl_SubgroupLtMask",       EbvSubgroupLtMask2,     symbolTable);
 
             symbolTable.setFunctionExtensions("subgroupMemoryBarrierShared", 1, &E_GL_KHR_shader_subgroup_basic);
+
```
johnkslang: This part is inside `#ifdef NV_EXTENSIONS`, which seems correct. But most the rest of it is outside.

I'm still investigating a possible "small" use of glslang for GLSL -> SPIR-V, and after removing all the side projects/paths, the extensions are a noticeable percentage of what's left.  It could be I'll give up soon, or that we need a more generic `#ifdef ...` to separate two purposes, but for the moment, I'd still like to follow the established precedent for protecting vendor-specific code with `#ifdef`.

johnkslang:
I finished my review. It looks fine except for:

- inconsistently (at least in appearance) of protection with `#ifdef NV_EXTENSIONS`
- which known_good for SPIRV-Tools should get used?

glslang/MachineIndependent/Initialize.cpp
dgkoch: This is only inside the #ifdef NV_EXTENSION block, because all the mesh shader stuff is inside the NV_EXTENSION ifdef. This extension is intentionally outside the NV_EXTENSIONS block, similarly to NV_cooperative_matrix, as we'd rather see the vendor specific ifdefs removed completely and are attempting to stop propagating them (per #1503)

dgkoch:
> which known_good for SPIRV-Tools should get used?
Ideally the one listed in this MR or something newer (I know lunarg snuck an update in first...)

dgkoch:
updated to resolve known_good conflict.

glslang/MachineIndependent/Initialize.cpp
johnkslang: But the `#ifdef ...` allow for both uses, while removing them only allows for one use.

dgkoch: Both of what uses?
Right now it’s inconsistent. Do you really want AMD, NV, Intel (plus every other vendor?) with an ifdef? What about EXT? 
What would be the goal of the smaller version? Vulkan only? what version of Vulkan?

johnkslang: A small mobile embedded footprint of just HLL -> SPIR-V versus the current full everything.

I agree with your concerns about a large set of macro names is not ideal, and that there is a continuum, not just two extremes.  Not sure what next step here; maybe bring in others.

We did break someone the last time a change was made inconsistently, so there are consumers compiling with the flags off.

dgkoch: I don’t think anything in this change is self-inconsistent, but if you insist I can put NV_EXTENSIONS ifdef around it. Let me know.

johnkslang: Okay, if it compiles both ways, we can proceed as is.  It is two subjects, whether the compile is broken with two flags versus how much is under `#ifdef`.  Let's leave the `#ifdef` question as still in motion though.

dgkoch: Thanks - I'd recommend we continue the discussion in #1503. That one never really went anywhere.

PR #1792, opened 2019-06-06
Automapping opengl location

Roy-AMD:
## 1. Background.

## 2. New Implementation.
## 3. Subsequent


Roy-AMD:
Hi @johnkslang 
I refine the code and made a cleaner pull request, I hoped  it more easily to review.
Thanks
Ruoyu

johnkslang:
Thank you.  I had already partially reviewed the first one, but I don't mind switching to this one.

It is a very large change/addition.  Can you identify what existing functionality might change (or is at risk of changing if there is a bug)?

Roy-AMD:
Hi @johnkslang 
Thank you very much, 
This change changes the interface and data structure, Although I want to keep all the existing functionality remain, there may still be some small issue. I have used gtest to check and fixed two issues, so I think there may have mistake, but the risk is low.
Thanks
Ruoyu

Roy-AMD:
Hi @johnkslang ,
Is here any update or question?
if you need I do something to speed up code reviews, please let me know.
Thanks
Ruoyu

Roy-AMD:
Hi @johnkslang，
We have some new change base on this PR need submit, can you speed up code review ?
Thanks
Ruoyu

johnkslang:
Yes, will do that today or tomorrow.

johnkslang:
This changes a publicly exposed interface.  Who is at risk of breaking from changing that interface?

Can it be done in a backward compatible way?  Wrapper functions, default parameters, or it really is compatible and I don't see it?

Roy-AMD:
Hi @johnkslang ,

I think no one will have risk of changing that interface, Although the interface has changed, but it's easy to do compatibility at client,  if they use default resolver, they even need do nothing.
And it's not a good idea to done in a backward compatible at TIoMapResolver, The new interface just helps us deliver more information and enriches the extensibility of the class. It does not bring about two distinct ways of working. so, If we make a interface backward compatible, it would reduce the readability of the code and bring a lot of redundant code, which is not a good choice.

Thanks
Ruoyu

Roy-AMD:
Hi @johnkslang 
Can we merge this code now?
Thanks
Ruoyu

johnkslang:
It seems unstable WRT location numbers in the tests.

Reverting.

johnkslang:
See logs for test failures in the commit to master.

johnkslang:
Also see #1850 for an additional thing to fix.

PR #1793, opened 2019-06-06
Uppdate spirv-tools known-good

greg-lunarg:


dgkoch:
Can we abandon this in favour of #1787 which includes an even newer spirv-tools?

greg-lunarg:
This version of spirv-tools has been vetted as "stable" for the purposes of the next SDK release. 

I propose the following compromise:

Merge this first so the SDK team has a version with a "stable" SPIRV-Tools. Then merge @dgkoch 's #1787.

Would that work for you Daniel?

I realize your feature will not make the next SDK with this compromise, but unfortunately the train has already left the station and I would prefer not to start everything over. Thanks for your understanding!

dgkoch:
well I'd rather get my change in.. it's been sitting there for 3 days already :-/
(technically it doesn't need the newer spirv-tools, but it does provide better validation with those)

greg-lunarg:
FWIW, it looks like all the changes to SPIRV-Tools from the last "stable" to @dgkoch 's commit should not destabilize it, so as a thought-experiment, taking #1787 for the SDK would probably be fine.

I would still rather sneak my change in first as a fall-back :)

PR #1796, opened 2019-06-13
Add missing GL_ARB_shader_ballot builtins to Geometry and Tessellatio…

samuelig:
…n shaders

When processing Geometry and Tessellation Control/Evaluation shaders,
the builtins defined by GL_ARB_shader_ballot were not processed by
glslang. This specially a problem for gl_SubGroupSizeARB because, it is
treated as shader input instead of uniform on Vulkan.

Signed-off-by: Samuel Iglesias Gonsálvez <siglesias@igalia.com>

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -7885,19 +7885,6 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
             symbolTable.setFunctionExtensions("readInvocationARB",      1, &E_GL_ARB_shader_ballot);
             symbolTable.setFunctionExtensions("readFirstInvocationARB", 1, &E_GL_ARB_shader_ballot);
 
-            BuiltInVariable("gl_SubGroupInvocationARB", EbvSubGroupInvocation, symbolTable);
-            BuiltInVariable("gl_SubGroupEqMaskARB",     EbvSubGroupEqMask,     symbolTable);
-            BuiltInVariable("gl_SubGroupGeMaskARB",     EbvSubGroupGeMask,     symbolTable);
-            BuiltInVariable("gl_SubGroupGtMaskARB",     EbvSubGroupGtMask,     symbolTable);
-            BuiltInVariable("gl_SubGroupLeMaskARB",     EbvSubGroupLeMask,     symbolTable);
-            BuiltInVariable("gl_SubGroupLtMaskARB",     EbvSubGroupLtMask,     symbolTable);
-
-            if (spvVersion.vulkan > 0)
-                // Treat "gl_SubGroupSizeARB" as shader input instead of uniform for Vulkan
-                SpecialQualifier("gl_SubGroupSizeARB", EvqVaryingIn, EbvSubGroupSize, symbolTable);
-            else
-                BuiltInVariable("gl_SubGroupSizeARB", EbvSubGroupSize, symbolTable);
-
             if (version >= 430) {
                 symbolTable.setFunctionExtensions("anyInvocationARB",       1, &E_GL_ARB_shader_group_vote);
                 symbolTable.setFunctionExtensions("allInvocationsARB",      1, &E_GL_ARB_shader_group_vote);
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8158,7 +8158,21 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
+        BuiltInVariable("gl_SubGroupGtMaskARB",     EbvSubGroupGtMask,     symbolTable);
+        BuiltInVariable("gl_SubGroupLeMaskARB",     EbvSubGroupLeMask,     symbolTable);
+        BuiltInVariable("gl_SubGroupLtMaskARB",     EbvSubGroupLtMask,     symbolTable);
+
+        if (spvVersion.vulkan > 0)
+            // Treat "gl_SubGroupSizeARB" as shader input instead of uniform for Vulkan
+            SpecialQualifier("gl_SubGroupSizeARB", EvqVaryingIn, EbvSubGroupSize, symbolTable);
+	else
+            BuiltInVariable("gl_SubGroupSizeARB", EbvSubGroupSize, symbolTable);
+
```
dgkoch: I think we should move the whole chuck of code from lines 7876-7900 down to here, since it's not vertex shader specific at all. 

(the whole design of this function is pretty horrible, IMO)

johnkslang: Agreed. I really want to refactor this to have a clear high-level control over what stages built-ins are in.  Possibly tabled instead of procedural, but definitely more compact and easier to visually see what's going on.

johnkslang: For the moment, this comment should be followed:

```
    // N.B.: a symbol should only be tagged once, and this function is called multiple times, once
    // per stage that's used for this profile.  So
    //  - generally, stick common ones in the fragment stage to ensure they are tagged exactly once
    //  - for ES, which has different precisions for different stages, the coarsest-grained tagging
    //    for a built-in used in many stages needs to be once for the fragment stage and once for
    //    the vertex stage
```

samuelig: Thanks for the feedback, I am going to send a new version with the suggested changes.

samuelig: > I think we should move the whole chuck of code from lines 7876-7900 down to here, since it's not vertex shader specific at all.
> 
> (the whole design of this function is pretty horrible, IMO)

Actually I cannot do that. When adding the function extensions, `symbolTable.setFunctionExtensions()` breaks an assert if they were already defined. If I move all the lines to here, it is called for 3 different stages. Therefore, in the second stage (Tess control), the `assert(extensions == 0)` in SymbolTable.h breaks. I am just move the the BuiltInVariables() lines to here and that works.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8158,7 +8145,28 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
+                SpecialQualifier("gl_SubGroupSizeARB", EvqVaryingIn, EbvSubGroupSize, symbolTable);
+            else
+                BuiltInVariable("gl_SubGroupSizeARB", EbvSubGroupSize, symbolTable);
+        }
+
+        if (spvVersion.vulkan > 0)
+            // Treat "gl_SubGroupSizeARB" as shader input instead of uniform for Vulkan
+            SpecialQualifier("gl_SubGroupSizeARB", EvqVaryingIn, EbvSubGroupSize, symbolTable);
+	else
+            BuiltInVariable("gl_SubGroupSizeARB", EbvSubGroupSize, symbolTable);
```
dgkoch: These 5 lines look like (incorrect - since they are not guarded by the !EesProfile check) duplicate of the 5 lines above.

samuelig: Right. I sent PR #1803 as it seems that github doesn't allow to update this after merge.

PR #1808, opened 2019-06-22
Handle all basic scalar types in promoteConstantUnion

jeffbolznv:
I ran across a case where doing "uint8_t(1.0)" generated a garbage constant because converting float to uint8_t wasn't handled in TIntermediate::promoteConstantUnion. This change adds all the missing conversions and simplifies the code by stamping out all combinations with macros.


johnkslang:
Looks like a nice simplification.  I have a minor worry: Does this replace a jump table with a long sequence of if tests?

jeffbolznv:
I think I can convert it back into a nested switch statement. I'll push another version later tonight.

johnkslang:
Nice, thanks.

PR #1812, opened 2019-06-25
Remove __DATE__ and __TIME__ usage

ShabbyX:
These macros result in a non-deterministic build.  In chromium, these
macros are specifically set to empty, resulting in a compile error.

Revealed by this change: 1f5799c155d30b90871c79d4ca92f8da4d3f3872


ShabbyX:
@digit-android

PR #1832, opened 2019-07-10
Avoid generating 8/16-bit constants when 8/16-bit arithmetic extensions aren't enabled

### LINKED ISSUE #1828: 8- and 16-bit storage extensions generate invalid constants
#1827 updates SPIRV-Tools to include validation of 8- and 16-bit storage extensions. The validator flags glslang's constant generation for these tests as generating constants of the incorrect types. It should instead generate 32-bit type constants and use conversion instructions.

jeffbolznv:
Fixes #1828.

This changes how constructors are implemented to construct a 32-bit type and then convert to the 8/16-bit type.

I'm aware of one hole with this. It makes it an error to do something like "const int x = int(int_8(1));" that would have folded away the intermediate 8-bit constant, because the intermediate result won't be folded. But I can live with that...


jeffbolznv:
@alan-baker FYI

alan-baker:
I think you need to undo the test update I made when pulling in the new SPIRV-Tools.

jeffbolznv:
Ah, !1827 got merged after I branched. I've rebased and regenerated.

glslang/Include/intermediate.h
```diff
@@ -657,6 +657,15 @@ enum TOperator {
     EOpConstructVec2,
     EOpConstructVec3,
     EOpConstructVec4,
+    EOpConstructMat2x2,
```
johnkslang: Is this being done to support math on the op enum-value itself?  If so, I think we need a healthy comment saying so.

(On a side branch, I was rearranging these for a different purpose, and it will be very confusing debugging if the conflict is resolved incorrectly.)

johnkslang:
It is comforting to see the validation errors go away, so it's a nice order to do things in.

glslang/Include/intermediate.h
jeffbolznv: I've added a comment. I'm also willing to add a function with a big switch statement to convert, if you'd rather.

johnkslang: Given that we have some "guard" regions for fast checking, I think it is okay this way.

I was kind of wondering if the C++ optimizer would see the reduction in the switch statement, but I think this is okay.  Thanks.

PR #1836, opened 2019-07-15
OpIsHelperInvocationEXT should declare relevant SPV extension and capability

amdrexu:


johnkslang:
Shouldn't this change be visible in test results?

amdrexu:
Currently, the test uses "demote". "demote" will declare the SPV extension and capability. Thus, the problem is hidden. If I only use "IsHelperInvocationEXT" without "demote" appearing. The problem occurs. So currently, there is no test covering this issue. I can add one if you believe it is necessary.

PR #1849, opened 2019-07-23
Add ability to record and retrieve named defines to TIntermediate and the preprocessor.

St0fF-NPL-ToM:
see #1341: finally made a pull request for that enhancement
- TIntermediate gets 2 unordered_sets to record named defines
- TIntermediate gets accessor functions
`	void insertAccessedNamedDefine(const char* defName) { accessedNamedDefines.emplace(defName); }
    void insertDeclaredNamedDefine(const char* defName) { declaredNamedDefines.emplace(defName); }
    void getAddAccessedNamedDefines(std::unordered_set<std::string>& targetSet) const { getAddDefines(targetSet, accessedNamedDefines); }
    void getAddDeclaredNamedDefines(std::unordered_set<std::string>& targetSet) const { getAddDefines(targetSet, declaredNamedDefines); }`
- TPpContext enhanced to add the defines to TIntermediate


St0fF-NPL-ToM:
Sorry - this refers to Issue #1334

glslang/MachineIndependent/localintermediate.h
```diff
@@ -801,6 +801,11 @@ class TIntermediate {
                extensionRequested(E_GL_EXT_shader_explicit_arithmetic_types_float16);
     }
 
+	void insertAccessedNamedDefine(const char* defName) { accessedNamedDefines.emplace(defName); }
```
johnkslang: Mixing tabs and spaces.  Use only 4-space indentation.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -801,6 +801,11 @@ class TIntermediate {
...
+    void insertDeclaredNamedDefine(const char* defName) { declaredNamedDefines.emplace(defName); }
+    void getAddAccessedNamedDefines(std::unordered_set<std::string>& targetSet) const { getAddDefines(targetSet, accessedNamedDefines); }
```
johnkslang: This is basically dead code.

Do you want to add a test that shows how it works and protects against regression?

glslang/MachineIndependent/localintermediate.h
St0fF-NPL-ToM: Ok, I'll do.  Differences in IDE settings.

glslang/MachineIndependent/localintermediate.h
St0fF-NPL-ToM: Lines 804 and 805 are not dead at all, these are what Pp uses to add symbols.  They're the "setters".
For the others I would need some help.  I've never written any tests, I could only say "it works correctly for me since about a year", which doesn't help at all.
Are there any docs explaining on how to proceed in adding a test?

johnkslang: Well, if they create a data structure that is not read, it depends how smart your compiler is whether it is dead.

One reason I ask is wondering if this extra code should be protected by a command-line option (or API option), in which case the logic is missing to turn it on.

If it is managed by a novel command-line option, the easiest thing to do would be to look at Test/runtests.  It is quite straight forward, and you can easily add a new section to test new things.

Alternatively (especially if not through a command-line option) it is a bit better to use the Google tests.

See the "Contributing tests" section of the README.

St0fF-NPL-ToM: To me it feels like it should be an API-option.  I'll have a look into the CMakeLists to see how to add such an API-option.

Please don't expect me to do this shortly.  As google-tests are new to me, I need to get my head free to do this.

johnkslang: I kind of said part of that poorly: if it is an API option, it is also straightforward to expose it through the command line, and then pretty easy to write a test using that command-line option, in Tests/runtest.

glslang/MachineIndependent/localintermediate.h
St0fF-NPL-ToM: Done.

arcady-lunarg:
Closing this as the review has remained unaddressed for several years now.

PR #1855, opened 2019-07-28
Web:  Start creating a very small-footprint glslang for web use.

johnkslang:
There will be a few design techniques to achieve this goal:

1. Remove the vendor-specific *_EXTENSION #ifdef'ing, to reduce errors arising from their use.
2. Add a GLSLANG_WEB #ifdef for the bulk of turning off unneeded code.  Goal is a net reduction in number of #ifdef.  By keeping these as high-level as possible in the function-call tree, and relying on DCE of the underlying C++ compiler, we can reduce somewhat the number that might appear to be needed: code that does not look #ifndef might really be off anyway, because it is called from code that is.
3. Tighten up use of ENABLE_HLSL and make the rest of HLSL usage be C++ compile-time selections.
4. Move to a tabled form of built-in function declaration while making it less error prone to add new ones.
5. Use m4 to have a clean way of doing conditional grammar-rule inclusion; adds a .m4 file that m4 processes to be a .y file for bison.
6. Using methods hard coded to return false, triggering more DCE than accessing a member variable.

These first four commits complete 3, show examples of 1 and 2, and do not yet do anything about 4.


johnkslang:
Some more detail, from offline experiments:  I expect the executable size floor to be around 300K (100K compressed).  That's for a fully linked standalone executable with command-line parsing and `GLSL -> AST -> SPIR-V` translation.

So, the size savings noted in the first 3 commits that impact size correspond to savings of about 1%, 2%, and 33% of the size floor.

Virtually all active development will be in code inside an already existing (at some level) `#ifndef GLSLANG_WEB` and hence not subject to fragility of getting the #ifdef'ing correct.  Again, there will be heavy reliance on the DCE of the underlying C++ compiler to avoid too many #ifdefs.

Also see discussion in #1503, which this branch will eventually close.

johnkslang:
Part 4 of the OP is now addressed by the 5th commit:

> 4. Move to a tabled form of built-in function declaration while making it less error prone to add new ones.

It only did approximately the ES 3.1 minus matrix functions. 

There is a trade-off between a tabled form that can represent anything arbitrarily and how big the table is.  For this first round, aiming for the "small wins over expressiveness" end of the spectrum.  For large complex signatures (e.g. memory model atomics), I think having an extra field for a form that allows arbitrary signatures will be best.  (There is a more expressive form in the HLSL code, but it takes over twice the space.)

johnkslang:
Getting google-tests is periodically failing with the bots.  The content of glslang is testing correctly.

johnkslang:
Adding a fifth technique to the OP, about conditional grammar-rule inclusion: https://github.com/KhronosGroup/glslang/pull/1855#issue-301814637.

johnkslang:
@jeffbolznv @Roy-AMD any issues with this?  (From you or someone in your company.) It is evolving from first showing how this can be done to doing a fair amount now.

From my perspective, this could become part of master at any time.  Key points:
- there are no changes to test results
- NV_EXTENSIONS and AMD_EXTENSIONS will no longer be needed; mostly new code will already naturally fall within an existing condition (there is only one: GLSLANG_WEB)
- otherwise, things are about the same, just a bit simpler with the coarser grained new conditional compilation

johnkslang:
So far, this has cut in half the number conditional code-inclusion tests:

Grammar **56** #if... -> **26** GLSLANG_WEB_EXCLUDE_ON
C++ **430** #if... -> **220** GLSLANG_WEB

The Grammar is done now, however more space savings will come by adding more #if... GLSLANG_WEB.  These will appear more rational than the vendor extensions did, so there is a net improvement in manageability here.

jeffbolznv:
This seems like a good reduction in the number of ifdefs, though I am a bit concerned that it's less obvious where to put ifdefs in the new model. Is it documented somewhere exactly what functionality GLSLANG_WEB excludes? Are we going to have CI test both WEB and !WEB builds on every commit and keep golds for both?

People seem to fairly often have mismatching ifdefs for the NV/AMD_EXTENSIONS defines in their projects. Has anything structurally changed that will improve this with the new ifdef?

I'm curious how much of the savings is from removing strings (e.g. from Initialize.cpp and AST/SPIR-V dumping) and how much is from the finer-grained ifdefing.

johnkslang:
Scope: Right now, I'm basically aiming for OpenGL ES 3.0, but Vulkan capable.  That could expand a bit in the future (say, by adding the compute stage).

#ifdef: Normally, any new development would have nothing to do with static web functionality, it would all be in the same bucket, and typically I not involve adding new #ifdef, because the neighborhood of the change should already be in the right bucket.

Mixing: You're either trying to deal with the tiny web version, or not, yes?  I'm mostly thinking the "default" of everything is the big version.  It's only specialized side projects that have a special tool chain that can approach the ~130K distribution size.

Strings: They were huge.  The strings themselves blow the whole budget, so tabling and turning off is critical.  And, then, collectively, the scattered #ifdefs are also huge, removing at least half the remaining code.  Sometimes this is not obvious, but using a method like `bool isXXXXX() const { return false; }` in turn makes things like `if (isXXXXX() && a > b) { ...; ...; <lots of code>; }` become dead, no #ifdef required.  So, there is a lot of dead code now due to that, which the tool chain removes to save space.

I would like the bots to keep an eye on both configurations, but haven't scripted that yet.

zoddicus:
> Scope: Right now, I'm basically aiming for OpenGL ES 3.0, but Vulkan capable. That could expand a bit in the future (say, by adding the compute stage).
> 
>
> Mixing: You're either trying to deal with the tiny web version, or not, yes? I'm mostly thinking the "default" of everything is the big version. It's only specialized side projects that have a special tool chain that can approach the ~130K distribution size. Everyone else would always be dealing only with the big one.
> 
For more context, the toolchain that we are targeting is emscripten/emsdk, with the goal of being able to produce a WASM binary that can be included as part of a webpage payload. This is why our target is sub 150kB, and it only needs to be able to ingest shaders that are WebGL compatible.

johnkslang:
The most recent commit https://github.com/KhronosGroup/glslang/pull/1855/commits/72d24a0b0ee6e9c8f81f7c4991cba634dae73892 shows the main reason for having multiple `#ifdef`; to subset switch statements.  I think it is pretty clear though, when adding new code, that it would go inside the existing big path, where it would naturally land anyway when locating it near related things.

(This commit saved another 40K of x86 binary, which is aiming to get close to 300K, so even this one commit is an important savings in comparison with the target size.  A 300+K size corresponds to a 100+K size in the tool chain discussed in https://github.com/KhronosGroup/glslang/pull/1855#issuecomment-519564958.)

johnkslang:
This branch is now rebased with master, including the new Emscription-web build.

CMakeLists.txt
```diff
@@ -26,18 +26,17 @@ if(NOT ${SKIP_GLSLANG_INSTALL})
...
-option(ENABLE_NV_EXTENSIONS "Enables support of Nvidia-specific extensions" ON)
-
 option(ENABLE_GLSLANG_WEB "Reduces glslang to minumum needed for web use" OFF)
 option(ENABLE_EMSCRIPTEN_SINGLE_FILE "If using emscripten, enables SINGLE_FILE build" OFF)
 
 CMAKE_DEPENDENT_OPTION(ENABLE_HLSL "Enables HLSL input support" ON "NOT ENABLE_GLSLANG_WEB" OFF)
 
 option(ENABLE_OPT "Enables spirv-opt capability if present" ON)
 
+option(ENABLE_GLSLANG_WEB "Reduces glslang to minumum needed for web use" OFF)
```
zoddicus: ENABLE_GLSLANG_WEB is defined above now on master, so it doesn't need to be added here.

johnkslang: Fixed.  I reedited the history for this, as it came from a rebase.  Also squashed a few commits into a bigger logical commit.

johnkslang:
This is ready to go into master.  Are the any objections or further concerns?  The main impacts are:
- the vender *_EXTENSIONS built switches and #ifdefs will be gone
- the 3.0 built-in functions will be in a table instead of the combinatoric text strings

I think new PRs can be done the relatively obvious way, from how things are done locally in the code.  Ideally, new features end up inside `#ifndef GLSLANG_WEB` but, then, most the compiler is already inside that, and I think this will naturally happen without much education about it.  It is also a fuzzy thing; many details are not inside `#ifndef GLSLANG_WEB` because it's not worth having too many of those for small things.

Until we have bots doing size testing, I will take responsibility for new PRs' impact on `GLSLANG_WEB` size.

Status: Size has shrunk to ~370KB of x86 binary.  The wrapped emscription build, total size, compressed, is 134KB.

The main thing not completed, and this can still happen in steps, is the full tabling of all the built-ins, rather than just those needed for 3.0/3.1.

(It might actually be good to roll this into master in multiple steps anyway.)

Issue #1850, opened 2019-07-23
glslang -> Chromium roll fails with compiler warning


null77:
See for example https://chromium-review.googlesource.com/c/chromium/src/+/1714279

 [376/6278] CXX obj/third_party/glslang/src/glslang_sources/iomapper.o
    FAILED: obj/third_party/glslang/src/glslang_sources/iomapper.o
    /b/s/w/ir/cache/goma/client/gomacc ../../third_par...
    ../../third_party/glslang/src/glslang/MachineIndependent/iomapper.cpp:294:11: error: field 'infoSink' will be initialized after field 'currentStage' [-Werror,-Wreorder]
    , infoSink(i)
    ^
    ../../third_party/glslang/src/glslang/MachineIndependent/iomapper.cpp:295:11: error: field 'currentStage' will be initialized after field 'preStage' [-Werror,-Wreorder]
    , currentStage(EShLangCount)
    ^
    2 errors generated.

Would it be possible to either suppress these warnings in the GN file or to add the warning to the CMake build and fix it? This is currently blocking the roller.

null77:
@dneto0 fyi

johnkslang:
This complaint is about code that was just reverted for test-case failures.  Will ask for this to be fixed as well.

PR #1857, opened 2019-07-28
Automapping opengl location

Roy-AMD:
## 1. Background.

## 2. New Implementation.
## 3. Subsequent


Roy-AMD:
Hi @johnkslang ,
I make a new pull request with the fix code.

Unstable WRT location numbers is fixed in commit [c24033a](https://github.com/KhronosGroup/glslang/pull/1857/commits/c24033afc3f9249a239574c6037093c63771b40b)
This issue are cause by order different in map and vector.

[issue][#1850](https://github.com/KhronosGroup/glslang/issues/1850) is fixed in commit [8a5824f](https://github.com/KhronosGroup/glslang/pull/1857/commits/8a5824f5bfe3d193dbd8d9328c49ac3ec19af344)

Commit [03a93ae](https://github.com/KhronosGroup/glslang/pull/1857/commits/03a93ae153dcb7bcb85cfb4b670da2085f1efdf0) fixed some clang compiler warning.

Thanks
Ruoyu

johnkslang:
Yes, this looks like it needs a set of tests both showing it working correctly and for future regressions.  Marking it WIP in the meantime.

PR #1860, opened 2019-08-01
Add in header for uint32_t definition

zoddicus:
This is needed for stricter/newer MSVC builds of downstream users.

### DIFF
--- a/glslang/MachineIndependent/iomapper.h
+++ b/glslang/MachineIndependent/iomapper.h
@@ -36,6 +36,7 @@
 #ifndef _IOMAPPER_INCLUDED
 #define _IOMAPPER_INCLUDED
 
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


zoddicus:
An example failure that we are seeing in shaderc when attempting to integrate newer versions of glslang:
https://source.cloud.google.com/results/invocations/692927f5-8df1-4050-9265-d11c319efd62/targets/graphics_shader_compiler%2Fshaderc%2Fwindows%2Fpresubmit_release_2013/log

johnkslang:
Note: Core glslang had not been using uint32_t because of its related portability issues (in practice, `unsigned int` is the same thing, yet more portable, and without suffering from being a non-native type).

johnkslang:
This is failing tests due to a failed clone, not glslang itself.

Issue #1852, opened 2019-07-24
Getting BUFFER_DATA_SIZE for unsized arrays


tsuoranta:
I had to make the following modification in order to be able to get buffer data size reflection for unsized arrays:

```
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -1527,7 +1527,7 @@ int TIntermediate::getBaseAlignment(const TType& type, int& size, int& stride, T
         RoundToPow2(size, alignment);
         stride = size;  // uses full matrix size for stride of an array of matrices (not quite what rule 6/8, but what's expected)
                         // uses the assumption for rule 10 in the comment above
-        size = stride * type.getOuterArraySize();
+        size = type.isUnsizedArray() ? stride : stride * type.getOuterArraySize();
         return alignment;
     }
```
For testing, I used https://github.com/ARM-software/opengl-es-sdk-for-android/blob/master/samples/advanced_samples/ComputeParticles/assets/spawn.cs

Test command:
```
glslangValidator --reflect-strict-array-suffix --reflect-basic-array-suffix --reflect-intermediate-io --reflect-separate-buffers --reflect-all-block-variables --reflect-unwrap-io-blocks -l -q -Od -g -S comp ~/git/opengl-es-sdk-for-android/samples/advanced_samples/ComputeParticles/assets/spawn.cs
```

Output without the modification:
```
Buffer block reflection:
SpawnBuffer: offset -1, type ffffffff, size 0, index -1, binding 0, stages 32, numMembers 1
```

I did not see other way to get data size. Is this modification correct, or is there a better way?

johnkslang:
On the surface, does it change an obviously wrong result to a less obviously wrong result?

That is, it would seem ill-defined to ask the size of something that is not yet sized.  Should there be something more specific about what is being asked that is well-defined?

tsuoranta:
Right, so yes, I do agree this is borderline useless. The only use for this is to fulfill what OpenGL specification requires. Would that count for enough justification?

arcady-lunarg:
This was fixed as part of PR #2231.

PR #1861, opened 2019-08-03
Fix unsized array buffer data size

tsuoranta:
A fix for #1852.

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -1527,7 +1527,7 @@ int TIntermediate::getBaseAlignment(const TType& type, int& size, int& stride, T
         RoundToPow2(size, alignment);
         stride = size;  // uses full matrix size for stride of an array of matrices (not quite what rule 6/8, but what's expected)
                         // uses the assumption for rule 10 in the comment above
-        size = stride * type.getOuterArraySize();
+        size = type.isUnsizedArray() ? stride : stride * type.getOuterArraySize();
         return alignment;
     }
 


tsuoranta:
I am examining the test failures related to reflection. While doing this, I noticed that some of the shaders do not meet minimum GL requirements:
GL 4.5 says 14 for MAX_GEOMETRY_UNIFORM_BLOCKS.
Output from my reflection program using OpenGL and minimal fragment shader to match reflection.vert:
```
GL vendor    : Intel Open Source Technology Center
GL renderer  : Mesa DRI Intel(R) UHD Graphics 620 (Kabylake GT2) 
GL version   : 4.5 (Core Profile) Mesa 19.0.2
GLSL version : 4.50
MAX_COMPUTE_WORK_GROUP_COUNT[0]     = 65535
MAX_COMPUTE_WORK_GROUP_COUNT[1]     = 65535
MAX_COMPUTE_WORK_GROUP_COUNT[2]     = 65535
MAX_COMPUTE_WORK_GROUP_SIZE[0]      = 1792
MAX_COMPUTE_WORK_GROUP_SIZE[1]      = 1792
MAX_COMPUTE_WORK_GROUP_SIZE[2]      = 1792
MAX_COMPUTE_WORK_GROUP_INVOCATIONS  = 1792
MAX_COMPUTE_SHARED_MEMORY_SIZE      = 65536
MAX_VERTEX_UNIFORM_BLOCKS           = 14
MAX_GEOMETRY_UNIFORM_BLOCKS         = 14
MAX_TESS_CONTROL_UNIFORM_BLOCKS     = 14
MAX_TESS_EVALUATION_UNIFORM_BLOCKS  = 14
MAX_FRAGMENT_UNIFORM_BLOCKS         = 14
MAX_COMBINED_UNIFORM_BLOCKS         = 84
Loading GL_VERTEX_SHADER from '/home/timo/git/glslang/Test/reflection.vert'
Loading GL_FRAGMENT_SHADER from '/home/timo/git/glslang/Test/reflection.vert.frag'
Linking program:
Could not link program:
Info Log: '''error: Too many vertex uniform blocks (15/14)
'''
```

tsuoranta:
Here is an example of difference in reflection.vert:
* buf1 size before change = 4
* buf1 size with change = 8
* buf1 Intel driver reports GL_BUFFER_DATA_SIZE = 16
```
buffer buf1 {
    float scalar;
    float runtimeArray[];
} buf1i;
```
OpenGL 7.3.1 says on BUFFER_DATA_SIZE:

>   If the final member of an active shader storage block is an array with no declared size, the minimum buffer size is computed assuming the array was declared as an array with one element.

This is the motivation behind the change, and reason why I believe 4 is wrong for buf1 size. With this change, size 8 comes without applying std140 rule 4. Since buf1 does not have standard layout, I suppose GL driver can return anything >= 8, and in case of my laptop with Intel driver, it seems rounding up with base alignment of a vec4 like std140 rule 4 has been applied.

Other differences (buf2, buf3, buf4) are similar.

johnkslang:
The change is simple and probably okay, but the remaining part is to include the spec. support for the change and test results that show the result of the change.

arcady-lunarg:
This was done as part of PR #2231.

PR #1866, opened 2019-08-06
Export glslang targets on installation

rumblehhh:
This allows the targets to be used in other cmake projects. See the following for more details:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages
https://foonathan.net/blog/2016/07/07/cmake-dependency-handling.html


jamesRumbleh:
Looks like I need to make the same changes for the external SPIRV-Tools project.

johnkslang:
Do you want to re-push here and trigger the bots again?

rumblehhh:
Sorry for the delay I was OoO the last couple of weeks. Everything looks good now after kicking the bots again.

PR #1872, opened 2019-08-20
make glslang.js easy to use and work on node, and related changes

kainino0x:


maierfelix:
@kainino0x Is there a chance that support for shader stages like `geometry`, `mesh`, `rgen`  etc. can be added? I need it for [this](https://github.com/maierfelix/nvk) project

kainino0x:
We're still in the middle of some major changes with regard to how WASM builds work, but it should be pretty easy to make this work. Try something like this: https://github.com/KhronosGroup/glslang/compare/master...kainino0x:allow-other-stages
(I haven't tested it, but I think(??) it should work.)

Be aware that this patch is likely to break in the future as we land more support for reducing the size of the WASM build for web. However I think we can make it work for the long run once we have landed those changes.

For reference on how to build, see build.sh in https://github.com/kainino0x/-webgpu-glslang
(it requires you to have sourced emsdk_env.sh).

maierfelix:
Awesome, thanks a bunch!

PR #1880, opened 2019-08-24
Add support for GL_ARB_explicit_uniform_location

georgeouzou:
This enables usage of explicit location on uniforms without the requirement of GLSL 430 version
Also added tests

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -5101,7 +5101,8 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
         return;
     } else if (id == "location") {
         profileRequires(loc, EEsProfile, 300, nullptr, "location");
-        const char* exts[2] = { E_GL_ARB_separate_shader_objects, E_GL_ARB_explicit_attrib_location };
+        const char* exts[2] = { E_GL_ARB_separate_shader_objects, E_GL_ARB_explicit_attrib_location }; 
+        // GL_ARB_explicit_uniform_location requires 330 or GL_ARB_explicit_attrib_location we do not need to add it here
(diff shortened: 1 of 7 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5858,8 +5859,11 @@ void TParseContext::layoutQualifierCheck(const TSourceLoc& loc, const TQualifier
         case EvqBuffer:
         {
             const char* feature = "location qualifier on uniform or buffer";
-            requireProfile(loc, EEsProfile | ECoreProfile | ECompatibilityProfile, feature);
-            profileRequires(loc, ECoreProfile | ECompatibilityProfile, 430, nullptr, feature);
+            requireProfile(loc, EEsProfile | ECoreProfile | ECompatibilityProfile | ENoProfile, feature);
+            profileRequires(loc, ~EEsProfile, 330, E_GL_ARB_explicit_attrib_location, feature);
+            if (!isEsProfile() && (extensionTurnedOn(E_GL_ARB_explicit_attrib_location) || version >= 330)) {
```
johnkslang: `profileRequires()` already has checks against the passed-in profile, the state of the passed-in extension, and the version number.  Can most (or all) of this if-test be removed?

georgeouzou: @johnkslang  i wanted to do it without using an if. the case here is that GL_ARB_explicit_uniform_location is valid if:
1 - No profile and GL_ARB_explicit_attrib_location and GL_ARB_explicit_uniform_location are enabled.
2 - profile >= 330 and GL_ARB_explicit_uniform_location is enabled
3 - profile >= 430
So we need something more complex here.

georgeouzou: Ok, it seems that the if was redundant after all :)

PR #1888, opened 2019-09-02
Adjusting code interface

Roy-AMD:
glslang/include/intermediate.h -> Add a new interface to set TIntermBranch's expression.

glslang/include/Types.h -> Add interface to set Type's basicType and add interface to get basicType form a TSampler.

glslang/MachineIndependent/intermediate.cpp -> Part of the code in createConversion been encapsulating as a new function called buildConvertOp

glslang/MachineIndependent/localintermediate.h -> Export createConversion and
buildConvertOp as a public function, and add a new param to control whether force update xbfBufferStride.

glslang/Public/ShaderLang.h -> Add interface to get shader object and shader source.

This change are use for shader dump and Patch AST, we want do some patch on AST when after parser and before SPVGen, add more interface and export some interface to public, can make code scalability better.

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -135,6 +135,8 @@ struct TSampler {   // misnomer now; includes images, textures without sampler, 
     bool isYuv()         const { return yuv; }
 #endif
     void setCombined(bool c) { combined = c; }
+    void setBasicType(TBasicType t) { type = t; };
+    TBasicType getBasicType()  const { return type; };
     bool isShadow()      const { return shadow; }
     bool isArrayed()     const { return arrayed; }
 
@@ -2184,7 +2186,8 @@ public:
     const TTypeList* getStruct() const { assert(isStruct()); return structure; }
     void setStruct(TTypeList* s) { assert(isStruct()); structure = s; }
     TTypeList* getWritableStruct() const { assert(isStruct()); return structure; }  // This should only be used when known to not be sharing with other threads
-
+    void setBasicType(const TBasicType& t) { basicType = t; }
+    
     int computeNumComponents() const
     {
         int components = 0;
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -562,6 +562,237 @@ bool TIntermediate::isConversionAllowed(TOperator op, TIntermTyped* node) const
     return true;
 }
 
+bool TIntermediate::buildConvertOp(TBasicType dst, TBasicType src, TOperator& newOp) const
+{
+    switch (dst) {
+#ifndef GLSLANG_WEB
+    case EbtDouble:
+        switch (src) {
+        case EbtUint:    newOp = EOpConvUintToDouble;    break;
+        case EbtBool:    newOp = EOpConvBoolToDouble;    break;
+        case EbtFloat:   newOp = EOpConvFloatToDouble;   break;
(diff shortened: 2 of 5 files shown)


glslang/Include/Types.h
```diff
@@ -2184,7 +2186,8 @@ class TType {
     const TTypeList* getStruct() const { assert(isStruct()); return structure; }
     void setStruct(TTypeList* s) { assert(isStruct()); structure = s; }
     TTypeList* getWritableStruct() const { assert(isStruct()); return structure; }  // This should only be used when known to not be sharing with other threads
-
+    void setBasicType(const TBasicType& t) { basicType = t; }
```
johnkslang: I also don't see any calls to this, and it seems potentially a bit dangerous to expose this like this.

Is it getting used somewhere?

(The one in TSampler should probably also be protected; that struct has evolved a lot to the point it should probably be a class.  But, first would like to understand more what's going on here.)

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -827,11 +783,64 @@ TIntermTyped* TIntermediate::createConversion(TBasicType convertTo, TIntermTyped
...
+    }
+
+    if (! getArithemeticFloat16Enabled()) {
+        if ((convertTo == EbtFloat16 && ! convertFromFloatTypes) ||
+            (node->getBasicType() == EbtFloat16 && ! convertToFloatTypes))
+            return nullptr;
+    }
+
+    if (!buildConvertOp(convertTo, node->getBasicType(), newOp))
+    {
```
johnkslang: Inconsistent curly braces.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -693,9 +693,9 @@ class TIntermediate {
     bool getPixelCenterInteger() const { return pixelCenterInteger; }
     void addBlendEquation(TBlendEquationShift b) { blendEquations |= (1 << b); }
     unsigned int getBlendEquations() const { return blendEquations; }
-    bool setXfbBufferStride(int buffer, unsigned stride)
+    bool setXfbBufferStride(int buffer, unsigned stride, bool force = false)
```
johnkslang: It almost seems like you want a changeXfbBufferStride().  Perhaps a changeBasicType() in TType as well?

Why would these things be changing in the middle of something?  It seems contrary to proper semantic checking, and fragile WRT all class invariants that are dependent of the member's value.

glslang/Include/Types.h
Roy-AMD: It's for some external code. we try to patch on AST and modify it's basic type, such as modify a lowp float to float16_t,  or a sampler to f16sampler, I think it's not a general feature so I didn't upstream this code. open this interface can help us easy to do local patch on AST.

glslang/MachineIndependent/localintermediate.h
Roy-AMD: The parameter are thoughtless, we will try to found a better way to fix it in local, I will revert it.

glslang/MachineIndependent/Intermediate.cpp
Roy-AMD: fixed

Roy-AMD:
Hi @johnkslang , code is updated, please take a look.

Roy-AMD:
Hi @johnkslang ,

If there are no more questions, please merge this code.

Thanks
Roy

PR #1894, opened 2019-09-09
Add flags for local size values ( compute shader )

ShchchowAMD:
Purpose :

According to GLSL SPEC 4.6 ( 4.4.1.4 Compute Shader Inputs), for compute shader input qualifiers, we should declare such qualifiers with same values in the same shader (local_size_x, y and z).
"If such a layout qualifier is declared more than once in the same shader, all those declarations must set the same set of local work-group sizes and set them to the same values; otherwise a compile-time error results."

Why this fix:

If we manually set "local_size_x = 1" and directly following a declaration like "local_size_x = 2", this would not be detected. That is because currently we treat all the '1' as default value and could not restrictly detect whether those are default values.

Test case:
......
layout(local_size_x=1) in;
layout(local_size_x=2) in;
......

### DIFF
--- a/Test/430.comp
+++ b/Test/430.comp
@@ -48,6 +48,9 @@ shared vec4 s;
 layout(location = 2) shared vec4 sl;  // ERROR
 shared float fs = 4.2;                // ERROR
 
+layout(local_size_y = 1) in;
+layout(local_size_y = 2) in;     // ERROR, changing
+layout(local_size_y = 1) in;
 layout(local_size_x = 2, local_size_y = 3, local_size_z = 4) out;  // ERROR
 
 int arrX[gl_WorkGroupSize.x];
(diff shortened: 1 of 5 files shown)


ShchchowAMD:
Types.h ===> add flags "localSizeNotDefault" for local sizes to record whether they are default values.

ParseHelper.cpp ===> use above flags to decide whether those values are from source strings or default.

localintermediate.h ===> add flags and use them to decide whether 'setLocalSize' should return a compile-time error. ( now we could know when 'local_size_(x | y | z )== 1' is default or not ).

ShchchowAMD:
Hi, sorry this is my first time to contribute to glslang.

I just can't sign the CLA as the page said "There is no CLA to sign for KhronosGroup/glslang",

should I try again please?

johnkslang:
It makes sense.  Glslang mostly uses an invalid value for the default still being present, but this is okay too.

Can you add the error case to the test suite?

ShchchowAMD:
I agree. Those default values will be used here if there's no 'local_size' inputs. I will check and add tests for this, no worries.

ShchchowAMD:
@johnkslang  Hi, I've added test cases for this commit, please have a review. Have a nice Monday.

PR #1895, opened 2019-09-09
Fix incorrect function prototypes of 64-bit findLSB/findMSB

amdrexu:
According to the 32-bit counterparts, their forms should be
genI64Type findLSB(genI64Type value)
genI64Type findLSB(genU64Type value)
genI64Type findMSB(genI64Type value)
genI64Type findMSB(genU64Type value)


johnkslang:
Where is this specified?

amdrexu:
This is in the spec: GL_EXT_shader_explicit_arithmetic_types.txt
https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GL_EXT_shader_explicit_arithmetic_types.txt

We have this statement:
```
Changes to Chapter 8 of the OpenGL Shading Language Specification:

Where the input arguments (and corresponding output) can be genI64Type, genI32Type, genI16Type, genI8Type, genIType is used as the argument.
```

So genIType in those built-in functions is extended to genI64Type, genI32Type, genI16Type, genI8Type. Thus, findLSB and findMSB could be used to support extended integer types. I see we have already added I8 and I16 support with correct function prototypes. But the I64 forms are incorrect all the time. This is a latent issue existing for a long period of time. The I64 version of these two functions are seldomly used, so we don't find the issue. Recently, we find the SPIR-V opcode is mistakenly generated with source and destination operands have mismatch types. We then find this is a glslang bug.

johnkslang:
Since it is rational to have a small-width return type from these functions, specifically, I wondered it it was intentional.

If this is not well exercised, I wonder what the best way actually is?

We would at least want alignment between SPIR-V and GLSL.

amdrexu:
Original 64-bit forms of findLSB and findMSB are wrong, not defined in ARB_gpu_shader_int64 and violating the spec of EXT_shader_explicit_arithmetic_types. The generated SPIR-V op FindILsb, FindSMsb, FindUMsb are invalid, also violating SPIR-V spec. The issue is not found because SPIRV-tools validation does not catch this all the time, I guess.

PR #1897, opened 2019-09-10
Remove unused params; Fix type warning.

dj2:
This CL removes unused method parameters. One int is updated to a size_t to fixup a size comparison issue.


dj2:
@johnkslang PTAL.

johnkslang:
Most these changes should not be made, they are part of the design, throughout, and correctly communicate the right way to access that information.  Not-yet-used parameters should just have their names commented out to let the tools know this is the current state.

dj2:
PTAL.

PR #1904, opened 2019-09-17
Reflection will crash when the VS input symbol defines the same name with FS output symbol

ShchchowAMD:
[**PURPOSE**]:

1. The current process design for Uniform / Block / Pipe IO symbols reflection (during program linking) is as following :

- 1.1 using a global mapper called 'TNameToIndex' to store all the relationship of name (of symbols) to their indexes (in their own MapIndexToReflection vectors).

- 1.2 TNameToIndex mapper will be used during program linking and helps to check and merge duplicate symbols within each stage ( Uniform, Block and Pipe IO)

- 1.3 Different types of symbols will have their own index mapping storage. All those symbols will share TNameToIndex as a general searching mapper.

- 1.4 Only IN in first stage and OUT in last stage will be dealed within traversing functions.

2. Now, here we meet those problems:

- 2.1 In and Out variables for pipelines are mapping to different MapIndexToReflection vector (ioItems), but they may still have same names within the general symbol search mapper : TNameToIndex.

- 2.2 Then, when there are same symbols of IN in VS and OUT in FS, TNameToIndex could not tell the difference because it only stores one local index for one symbol (1:1) as a pair of KeyValue.

[**What fixed**]:

1. Seperate I/O from other symbols like Uniform and Block (it is wrong to keep them all in TNameToIndex), and save in new searching mappers called pipeInNameToIndex and pipeOutNameToIndex.

2. Expose new top-level functions defined as getReflectionPipeIOIndex and getPipeIOIndex for users who need to query Pipe I/O information (As they may reach those things through getUniformIndex and getReflectionIndex now, which is a confused way.)

3. As there are 2 mappers for above symbols, users needs to input second argument when they wanna reach those pipe I/O parameters, that's also why we need to modify GET functions either.

[**Test Case**]:

The shader is as following: 

######### VS ############

layout(location = 0) in vec4 g_position;
layout(location = 1) in vec4 g_color;
out StageData {
  vec4 color;
} g_vs_out;
void main() {
  gl_Position = g_position;
  g_vs_out.color = g_color;
}
 
########### FS #############
in StageData {
  vec4 color;
} g_fs_in;
layout(location = 0) out vec4 g_color;
void main() {
  g_color = g_fs_in.color;
}


johnkslang:
This is just reflection, right? Not glslang as a validator or translator to SPIR-V?

The first commit seems very odd (about markdown).  Can this PR just be one commit?

ShchchowAMD:
Yes, this is just for reflection part. I will merge them into one commit and re-submit it.

PR #1905, opened 2019-09-18
SPV_KHR_physical_storage_buffer/SPV: Add GL_EXT_buffer_reference_uvec2

johnkslang:
Adds uvec2 <-> reference constructor support.
Switches from EXT to KHR for physical_storage_buffer.

WIP, due to validation failures.  Need to sync with SPIRV-Tools on the KHR support.


johnkslang:
We have new tools (thanks @alan-baker) and headers now that remove the validation errors.

PR #1906, opened 2019-09-18
Reflection will crash when the VS input symbol defines the same name with FS output symbol

ShchchowAMD:
[**PURPOSE**]:
The current process design for Uniform / Block / Pipe IO symbols reflection (during program linking) is as following :

1.1 using a global mapper called 'TNameToIndex' to store all the relationship of name (of symbols) to their indexes (in their own MapIndexToReflection vectors).

1.2 TNameToIndex mapper will be used during program linking and helps to check and merge duplicate symbols within each stage ( Uniform, Block and Pipe IO)

1.3 Different types of symbols will have their own index mapping storage. All those symbols will share TNameToIndex as a general searching mapper.

1.4 Only IN in first stage and OUT in last stage will be dealed within traversing functions.

Now, here we meet those problems:

2.1 In and Out variables for pipelines are mapping to different MapIndexToReflection vector (ioItems), but they may still have same names within the general symbol search mapper : TNameToIndex.

2.2 Then, when there are same symbols of IN in VS and OUT in FS, TNameToIndex could not tell the difference because it only stores one local index for one symbol (1:1) as a pair of KeyValue.

[**What fixed**]:

Seperate I/O from other symbols like Uniform and Block (it is wrong to keep them all in TNameToIndex), and save in new searching mappers called pipeInNameToIndex and pipeOutNameToIndex.

Expose new top-level functions defined as getReflectionPipeIOIndex and getPipeIOIndex for users who need to query Pipe I/O information (As they may reach those things through getUniformIndex and getReflectionIndex now, which is a confused way.)

As there are 2 mappers for above symbols, users needs to input second argument when they wanna reach those pipe I/O parameters, that's also why we need to modify GET functions either.

[**Test Case**]:

The shader is as following:

######### VS ############

layout(location = 0) in vec4 g_position;
layout(location = 1) in vec4 g_color;
out StageData {
vec4 color;
} g_vs_out;
void main() {
gl_Position = g_position;
g_vs_out.color = g_color;
}

########### FS #############

in StageData {
vec4 color;
} g_fs_in;
layout(location = 0) out vec4 g_color;
void main() {
g_color = g_fs_in.color;
}

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -2048,6 +2048,9 @@ bool TProgram::buildReflection(int opts)
 
 unsigned TProgram::getLocalSize(int dim) const                        { return reflection->getLocalSize(dim); }
 int TProgram::getReflectionIndex(const char* name) const              { return reflection->getIndex(name); }
+int TProgram::getReflectionPipeIOIndex(const char* name, const bool inOrOut) const
+                                                                      { return reflection->getPipeIOIndex(name, inOrOut); }
+
 int TProgram::getNumUniformVariables() const                          { return reflection->getNumUniforms(); }
 const TObjectReflection& TProgram::getUniform(int index) const        { return reflection->getUniform(index); }
 int TProgram::getNumUniformBlocks() const                             { return reflection->getNumUniformBlocks(); }
--- a/glslang/MachineIndependent/reflection.cpp
+++ b/glslang/MachineIndependent/reflection.cpp
@@ -112,6 +112,10 @@ public:
             TReflection::TMapIndexToReflection &ioItems =
                 input ? reflection.indexToPipeInput : reflection.indexToPipeOutput;
 
+
+            TReflection::TNameToIndex &ioMapper =
+                input ? reflection.pipeInNameToIndex : reflection.pipeOutNameToIndex;
+
             if (reflection.options & EShReflectionUnwrapIOBlocks) {
                 bool anonymous = IsAnonymous(name);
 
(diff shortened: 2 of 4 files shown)


ShchchowAMD:
[https://github.com/KhronosGroup/glslang/pull/1904#issue-318139793](url)

ShchchowAMD:
[](https://github.com/KhronosGroup/glslang/pull/1904)

ShchchowAMD:
[Hi, sorry that I accidently closed the last PR, please check it here.](https://github.com/KhronosGroup/glslang/pull/1904)

ShchchowAMD:
I've merged them into one commit. FYI, this is just for reflection.

ShchchowAMD:
@johnkslang  FYI

ShchchowAMD:
@johnkslang  Hi, john. Is this commit OK to be merged?

johnkslang:
Can you add a test that goes from failing to succeeding with this change?

See the block of reflection tests in `Test/runtests`.

ShchchowAMD:
Hi, @johnkslang .

I checked those reflection test cases, but there is only one **single** shader source string within each case. 

This error happens when we link a program to initialize **pipeline** structure, which means we may need **two** shader source strings within one case for testing (Vertex Shader's input and Fragment Shader's output). Would you please give me some suggestions about where should I put the test case?

PS. Here is a simple example which I wanna to add for testing:

######### VS ############
layout(location = 0) in vec4 g_position;
layout(location = 1) in vec4 g_color;

out StageData {
    vec4 color;
} g_vs_out;

void main() {
    gl_Position = g_position;
    g_vs_out.color = g_color;
}

########### FS #############
in StageData {
vec4 color;
} g_fs_in;

layout(location = 0) out vec4 g_color;

void main() {
    g_color = g_fs_in.color;
}

PR #1907, opened 2019-09-18
Separate GLSLANG_WEB (min-size build) and Emscripten options

kainino0x:
Allows building non-min-size builds with Emscripten.
Adds ENABLE_GLSLANG_WEB_DEVEL.
Moves the glslang.js files to OSDependent/Web.
Small cleanups and docs update.

And use custom es6 modularization instead of MODULARIZE=1.


johnkslang:
@zoddicus can you review?

PR #1919, opened 2019-10-03
Move install directory for SPIRV/ folder.

dj2:
Currently the SPIRV/ folder will get installed into the include
directory. This folder is part of GLSLang, so it makes more sense under
glslang/SPIRV.

Currently, GLSLang will install a SPIRV/ folder while spirv-headers will
install a spirv/ folder. This is confusing and will cause issues on a
case sensitive filesystem if both are installed at the same time.


johnkslang:
Upper case is the way the projects are all named and is correct for the SPIR-V acronym.  I find it unfortunate that some projects did not follow this lead.

SPIRV/CMakeLists.txt
```diff
@@ -87,5 +87,6 @@ if(ENABLE_GLSLANG_INSTALL)
                 ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
     endif()
 
-    install(FILES ${HEADERS} ${SPVREMAP_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/SPIRV/)
+    install(FILES ${HEADERS} ${SPVREMAP_HEADERS}
+            DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/glslang/SPIRV/)
```
johnkslang: Oh, is this just a space change, not a capitalization change?

johnkslang: The line length is fine.  It is easier to see the real diff, with only the actual change in the diff.

johnkslang:
What all is at risk of breaking by relocating this?

SPIRV/CMakeLists.txt
dj2: done.

dj2:
I left the casing for SPIRV, I just moved it from SPIRV/ to glslang/SPIRV. The risk being, if someone downstream is depending on GLSLang being installed, and using the SPIRV/ folder from GLSLang then their build will break until they change their code to add the glslang/.

I don't know how many projects use the SPIRV/ from glslang instead of spirv-headers.

johnkslang:
Thanks.

Glslang is not supposed to be the provider of official SPIR-V headers, I think this was done by someone else who thought (?) otherwise?  There is some bootstrapping history here, as SPIR-V was coming online.

So, actually, maybe this is a good first step in getting consumers to use headers from the official path instead.

johnkslang:
> I don't know how many projects use the SPIRV/ from glslang instead of spirv-headers.

But, this moved headers that no one else provides, breaking downstream consumers.

dj2:
Yes, there is more stuff in the glslang SPIRV/ then in the spirv-headers spirv/ but there is also duplicated things.  It's moved not removed, so downstream would need to change their import locations.

PR #1932, opened 2019-10-15
Fix config issues and revert previous fixes for semi-colon issues

### LINKED ISSUE #1931: Chromium/ANGLE rollers broke because of semicolon warning
Examples:

https://chromium-review.googlesource.com/c/chromium/src/+/1860292
https://chromium-review.googlesource.com/c/angle/angle/+/1861874

Errors:

../../third_party/glslang/src/hlsl/hlslScanContext.cpp(86,2): error: extra ';' outside of a function is incompatible with C++98 [-Werror,-Wc++98-compat-extra-semi]
};
 ^

../../third_party/glslang/src/SPIRV/SpvTools.cpp(213,2): error: extra ';' outside of a function is incompatible with C++98 [-Werror,-Wc++98-compat-extra-semi]
}; // end namespace glslang
 ^

and a lot more files. Not sure if someone intentionally went and added extra semicolons at the end of every namespace. We need to either fix the semicolons or explicitly disable the warning in BUILD.gn. Mild preference to fix the semicolons as they aren't necessary.

zoddicus:
This reverts commit 41261d95a417bb3a121c162154d441433705a237.
This reverts commit 7c9accb6534d4cde050555b8091b434dc7e4d218.

Instead of trying to work around all of the potential semicolon issues
in glslang, making it conform to Chromium's style, mark the code
explicitly as non-chromium in the BUILD.gn, so chromium doesn't
attempt to enforce its style rules on glslang.

Fixes #1931

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -159,7 +159,7 @@ source_set("glslang_sources") {
   }
 
   if (is_clang) {
-    cflags_cc = [
+    cflags = [
       "-Wno-extra-semi",
       "-Wno-ignored-qualifiers",
       "-Wno-implicit-fallthrough",
@@ -168,7 +168,6 @@ source_set("glslang_sources") {
       "-Wno-unused-variable",
       "-Wno-missing-field-initializers",
       "-Wno-newline-eof",
-      "-Wextra-semi",
     ]
   }
   if (is_win && !is_clang) {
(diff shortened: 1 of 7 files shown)


johnkslang:
We can leave in place the semicolon improvements.  The ones around method definitions look much better fixed.  We just don't need to dogmatically push this to the extreme.  So, can we keep just the settings part of your PR, without erasing improvements?

zoddicus:
> We can leave in place the semicolon improvements. The ones around method definitions look much better fixed. We just don't need to dogmatically push this to the extreme. So, can we keep just the settings part of your PR, without erasing improvements?

Yup, PR updated.

PR #1933, opened 2019-10-15
Make runtests configurable.

dj2:
This CL allows setting the location of glslangValidator and spirv-remap
when calling the runtests script. A test target is added to CMake to
execute runtests and sets the location to the build folder for the two
applications.


dj2:
This target is getting created in shaderc currently, this allows removing the shaderc target, along with the code to copy Tests and rewrite runtest (https://github.com/google/shaderc/pull/838).

johnkslang:
Please follow local coding convention.

johnkslang:
Do you really want this as 3 different commits?

dj2:
Can you please let me know which coding convention I'm violating? I thought I was following local convention.

dj2:
No, I just assumed the squash and commit button would be used?

johnkslang:
Yes, I'll squash then.  The indentation does not match the surrounding lines.

dj2:
Fixed, I believe.

CMakeLists.txt
```diff
@@ -180,3 +180,33 @@ if(ENABLE_HLSL)
...
+    endif()
+
+    add_test(NAME glslang-testsuite
+        COMMAND bash ${IGNORE_CR_FLAG} runtests
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Test/)
+
+  
+    if (CMAKE_CONFIGURATION_TYPES)
+        set_property(TEST glslang-testsuite APPEND PROPERTY
+            ENVIRONMENT GLSLANG_PATH=${CMAKE_CURRENT_BINARY_DIR}/StandAlone/$<CONFIGURATION>/glslangValidator)
```
johnkslang: What is the lifetime of this environment setting?  It would seem odd if, say, leaving a shell and starting a new one got different behavior that staying in the original one.

dj2: As far as I understand, this is just for the invocation of glslang-testsuite. It should not change our shell outside of the cmake run. So, running on ctest and then running the runtests program from the shell will have different results because the settings aren't the same.

Issue #1939, opened 2019-10-18
Werror=deprecated-copy compilation issues with GCC 9.2.1


tsuoranta:
Including `intermediate.h` from glslang with GCC 9.2.1 (Ubuntu 19.10 comes with it) with `Werror=deprecated-copy` compilation option active results compilation errors. My understanding is that compiler now enforces (at least to some extent) rule of three.

Fix proposal in #1940

JohannesKauffmann:
It seems that this issue can be closed, as #1940 is merged.

PR #1940, opened 2019-10-18
Fix for GCC 9 / Werror=deprecated-copy

tsuoranta:
Changes to make compiler happier regarding `Werror=deprecated-copy` / rule of three related issues.

Fix proposal for #1939

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,9 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
                         -Wunused-parameter -Wunused-value  -Wunused-variable -Wunused-but-set-parameter -Wunused-but-set-variable -fno-exceptions)
     add_compile_options(-Wno-reorder)  # disable this from -Wall, since it happens all over.
     add_compile_options(-fno-rtti)
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "9.0.0")
+        add_compile_options(-Werror=deprecated-copy)
+    endif()
 elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options(-Wall -Wuninitialized -Wunused -Wunused-local-typedefs
                         -Wunused-parameter -Wunused-value  -Wunused-variable)
--- a/SPIRV/SpvBuilder.h
+++ b/SPIRV/SpvBuilder.h
@@ -512,7 +512,7 @@ public:
         Block &head, &body, &merge, &continue_target;
     private:
         LoopBlocks();
-        LoopBlocks& operator=(const LoopBlocks&);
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
                         -Wunused-parameter -Wunused-value  -Wunused-variable -Wunused-but-set-parameter -Wunused-but-set-variable -fno-exceptions)
     add_compile_options(-Wno-reorder)  # disable this from -Wall, since it happens all over.
     add_compile_options(-fno-rtti)
+    add_compile_options(-Werror=deprecated-copy)
```
tsuoranta: This is probably not a good idea to include in final merge at least as is, but is useful during testing.

tsuoranta: Fixed - now guarded with test to make sure compiler is new enough to support Werror=deprecated-copy

johnkslang:
So far, we haven't added syntax like `= delete` and `= default` to glslang, to keep it building on a wider range of platforms.

However, we have been preventing use of copy/assign/etc. as needed by declaring methods as private/protected. Are these unworkable that way? Or, are they sufficient motivation to make the switch over?

tsuoranta:
`= delete` and ` = default` are C++11 feature. In `README.md` the very first dependency for glslang is C++11 compiler, and seems like it has been there for couple of years at least.

It appears that declaring as private is not sufficient to make GCC9 happy.

CMakeLists.txt
turol: Instead of checking the compiler version it would be better to check if it supports `-Wdeprecated-copy`. This way it will also work seamlessly on clang if they start supporting it. I don't know how to check it using CMake though.

johnkslang:
Compilers rolled out C++11 over time; MSVS in particular drug it out over a long time.  The first version to support C++11 missed this feature.

Maybe it's time to move on though, in which case, I'd also like to do a more thorough job of fixing this issue.  Will still give this MR some more consideration.

CMakeLists.txt
tsuoranta: I find checking option-supported-or-not (cmake has `check_cxx_compiler_flag()`) slightly less secure, because in theory someone could use a compiler one day which implements the flag with different semantics. When we know specific compilers support the flag with matching semantics, we can add similar compiler and version identification.

I don't feel strongly about this, so if `check_cxx_compiler_flag()` is preferred, I can update this PR to use that instead.

tsuoranta:
Pinging this issue. There needs to be some solution for those using glslang with GCC9.

johnkslang:
Is GCC9 really rejecting valid C++ code, or are is this because of explicitly asking it to treat a deprecation warning as an error?

johnkslang:
I'm going to merge, and see if we still have issues with consumers not accepting `= delete`.

PR #1943, opened 2019-10-22
SPIR-V: prune unreachable merge block and continue target

dneto0:
Within a function, only emit a block if:
- it's reachable from the entry block by zero or more branches,
- it's an unreachable merge block, in which case it's just OpLabel and OpUnreachable
- it's an unreachable continue target, in which case it's OpLabel and an unconditional branch to the corresponding loop header.

Includes tests showing intersting cases.
This is WIP.  I haven't updated the existing tests that are affected by this change.


dneto0:
Existing tests affected:

[==========] 1372 tests from 54 test suites ran. (4038 ms total)
[  PASSED  ] 1354 tests.
[  FAILED  ] 18 tests, listed below:
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_constantbuffer_frag, where GetParam() = 16-byte object <09-B3 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_doLoop_frag, where GetParam() = 16-byte object <BB-B3 9D-00 01-00 00-00 08-AF 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_forLoop_frag, where GetParam() = 16-byte object <34-B5 9D-00 01-00 00-00 08-AF 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_if_frag, where GetParam() = 16-byte object <64-B8 9D-00 01-00 00-00 08-AF 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_structbuffer_frag, where GetParam() = 16-byte object <C3-C6 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_structbuffer_coherent_frag, where GetParam() = 16-byte object <54-C7 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_structbuffer_rw_frag, where GetParam() = 16-byte object <E7-C7 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00>
[  FAILED  ] ToSpirv/HlslVulkan1_1CompileTest.FromFile/hlsl_wavequery_frag, where GetParam() = 16-byte object <E8-CB 9D-00 01-00 00-00 08-AF 9D-00 01-00 00-00>
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_for_notest_vert, where GetParam() = "spv.for-notest.vert"
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_controlFlowAttributes_frag, where GetParam() = "spv.controlFlowAttributes.frag"
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_earlyReturnDiscard_frag, where GetParam() = "spv.earlyReturnDiscard.frag"
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_forwardFun_frag, where GetParam() = "spv.forwardFun.frag"
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_functionCall_frag, where GetParam() = "spv.functionCall.frag"
[  FAILED  ] Glsl/CompileVulkanToSpirvTest.FromFile/spv_merge_unreachable_frag, where GetParam() = "spv.merge-unreachable.frag"
[  FAILED  ] ToSpirv/RemapTest.FromFile/remap_similar_1a_none_frag, where GetParam() = 24-byte object <D8-F5 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00 00-00 00-00 00-00 00-00>
[  FAILED  ] ToSpirv/RemapTest.FromFile/remap_similar_1b_none_frag, where GetParam() = 24-byte object <F3-F5 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00 00-00 00-00 00-00 00-00>
[  FAILED  ] ToSpirv/RemapTest.FromFile/remap_similar_1a_everything_frag, where GetParam() = 24-byte object <0E-F6 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00 00-00 00-00 FF-00 00-00>
[  FAILED  ] ToSpirv/RemapTest.FromFile/remap_similar_1b_everything_frag, where GetParam() = 24-byte object <2F-F6 9D-00 01-00 00-00 F3-AE 9D-00 01-00 00-00 00-00 00-00 FF-00 00-00>

johnkslang:
I generated the test changes locally (easier to include them and have them here to see and comment on). Most look okay, but some are losing test cases now:

- hlsl.doLoop.frag: There is a `do { return; } while ()` in the middle.  I expected to see XXXX disappear in `do { return; XXXX; } while ()`, but not the entire rest of the shader
- hlsl.forLoop.frag, similar problem started with `for (;;) ;`
- spv.controlFlowAttributes.frag: also `for (;;) { }`
- hlsl.if.frag: probably need to just move the `if () return; else return;` to be the last thing.

Would like confirmation that this change should not cross the line into being an actual optimization. I think it's not, because at the physical CFG level there actually were no branches to the beginning of the code that disappeared.  Yes?

It will be required though to change the test sources a bit such that they don't have most their results eliminated by this change.  I think I identified the necessary changes above.

SPIRV/SpvPostProcess.cpp
```diff
@@ -328,24 +328,52 @@ void Builder::postProcessReachable(const Instruction&)
...
         }
     }
 
+    // Modify unreachable merge blocks and unreachable continue targets.
+    // Delete their contents.
+    for (auto merge_and_header : headerForUnreachableMerge) {
+        Block* merge = merge_and_header.first;
+        merge->forceDeadMerge();
+    }
+    for (auto continue_and_header : headerForUnreachableContinue) {
```
johnkslang: This `:` is one of those features not initially supported in MSVC with C++11.  There is another PR wanting to add `= default`, which is a similar situation.  It's pretty trivial to not do this, but then maybe it's time to try again and see if anyone still complains.

SPIRV/spvIR.h
```diff
@@ -268,10 +299,24 @@ class Block {
     bool unreachable;
 };
 
+// The different reasons for reaching a block in the inReadableOrder traversal.
+typedef enum ReachReason {
+  // Reachable from the entry block via transfers of control, i.e. branches.
+  ReachViaControlFlow = 0,
```
johnkslang: Four-space indent.

SPIRV/spvIR.h
```diff
@@ -226,6 +226,36 @@ class Block {
         return nullptr;
     }
 
+    // Change this block into a canonical dead merge block.  Delete instructions
+    // as necessary.  A canonical dead merge block has only an OpLabel and an
+    // OpUnreachable.
+    void forceDeadMerge() {
+       assert(localVariables.empty());
```
johnkslang: This all looks like 7-space indentation; probably you intended 8.

Test/spv.dead-complex-continue-after-return.vert
```diff
@@ -0,0 +1,19 @@
...
+void main() {
+  int i = 0;
+  o = 1;
+  // This has non-trivial continue target.
+  for (i=0; i < 5; ++i, o=99) {
+    o = 2;
+    return;
+    o = 3;
+  }
+  // This is considered reachable since we don't assume
```
johnkslang: Saying it stronger: we should be quite clear this PR is not an optimization and has no smarts, other than taking away something that was not hooked up.

Is that all true?

SPIRV/SpvPostProcess.cpp
```diff
@@ -328,24 +328,52 @@ void Builder::postProcessReachable(const Instruction&)
...
-            if (reachableBlocks.count(b) == 0) {
-                for (auto ii = b->getInstructions().cbegin(); ii != b->getInstructions().cend(); ii++)
+            if (headerForUnreachableMerge.count(b) != 0 || headerForUnreachableContinue.count(b) != 0) {
+                auto ii = b->getInstructions().cbegin();
+                ++ii; // Keep potential decorations on the label.
+                for (; ii != b->getInstructions().cend(); ++ii)
+                    unreachableDefinitions.insert(ii->get()->getResultId());
+            } else if (reachableBlocks.count(b) == 0) {
+                // The normal case for unreachable code.  All definitions are considered dead.
+                for (auto ii = b->getInstructions().cbegin(); ii != b->getInstructions().cend(); ++ii)
```
johnkslang: Because of GLSL scoping and load/store approach, do we never remove definitions that get consumed in non-deleted code?

Of course, I understand about undefined values/behavior/etc.  I'm only asking about syntactic level consumption of an SSA `<id>` that has no definition.

Test/spv.dead-complex-continue-after-return.vert
dneto0: Yes, that's the intent.

SPIRV/SpvPostProcess.cpp
dneto0: I believe this will never strand an SSA use without a def.  The argument is that:
- This code only removes definitions that are never reachable via the CFG
- Any *use* of such a definition must be (a) in a phi, or (b) dominated by the definition.
    - (a) glslang doesn't generate phis
    - (b) by dominance, all paths to the use must pass through the definition.  And since the definition is not reachable the use can't be reachable.

SPIRV/SpvPostProcess.cpp
dneto0: ack.  I'll fix in a subsequent patch

SPIRV/spvIR.h
dneto0: ack.

SPIRV/spvIR.h
dneto0: Fixed in the next commit.

SPIRV/spvIR.h
dneto0: fixed in the next commit.

SPIRV/SpvPostProcess.cpp
dneto0: Fixed in the next commit by using iterator iteration.

dneto0:
I added an option to get back to the old behaviour.
I added tests for both behaviours.

SPIRV/SpvPostProcess.cpp
dneto0: Thinking about this gain, maybe I missed the point of the question.
Is this the scenario you're thinking of?

       void main() {
            some code ...
            if (...)  {
                return;
                strandedCodeA;
            } 
            strandedCodeB;
       }

In the old behaviour, strandedCodeA would be removed already by the in-readable-order algorithm, but strandedCodeB would remain.
In the new behaviour both strandedCodeA and strandedCodeB are removed.

Is the concern that there is code in strandedCodeB that provokes a definition to be emitted (e.g. a function-scope variable)?  But in that case I think by the previous argument that variable will no longer be referenced anywhere. ?

dneto0:
I think this is ready for review now. 
I squashed the history to avoid ugliness.

SPIRV/SpvPostProcess.cpp
johnkslang: > glslang doesn't generate phis

Glslang does generate phis.  It is basically required for correct lowering of `&&` and `||` in the presence of short circuiting.  See `TGlslangToSpvTraverser::createShortCircuit(...)`.

Otherwise, for normal "variables", at least those visible in the source, glslang always loads/stores, thereby bypassing an use/def problem for that.  It would have to be intermediate definitions not explicitly visible in the source. 

So, the places to worry would be complex expressions, like those involving `?:`, `||`, and `&&` where control flow might be introduced.

I can believe it's all okay, just worth some semi-proof like thing, like the one you're laying out above.

johnkslang:
> I added an option to get back to the old behaviour.

I really hope this is non-modal.  There should be the one right way to do this, with one set of test results.

The amount of change seemed fine earlier (haven't caught up with the latest round of changes).

dneto0:
The Vulkan CTS tests has had targeted OpUnreachable tests since November 2015, or before Vulkan 1.0.  See this version of the SPIR-V assembly-based tests: https://github.com/KhronosGroup/VK-GL-CTS/blob/6632073b4f79df50f93b27c3d2ed005d21f2c367/external/vulkancts/modules/vulkan/spirv_assembly/vktSpvAsmInstructionTests.cpp#L315


This interesting different cases include:
 - **nonvoid fn**: OpUnreachable in a non-void function.   See https://github.com/KhronosGroup/VK-GL-CTS/blob/6632073b4f79df50f93b27c3d2ed005d21f2c367/external/vulkancts/modules/vulkan/spirv_assembly/vktSpvAsmInstructionTests.cpp#L453
 - **branch-to-unreach**:  OpUnreachable in place of an unconditional branch in a void function.  https://github.com/KhronosGroup/VK-GL-CTS/blob/6632073b4f79df50f93b27c3d2ed005d21f2c367/external/vulkancts/modules/vulkan/spirv_assembly/vktSpvAsmInstructionTests.cpp#L415  (See the "else" clause in the original GLSL main function, in comments at line 350)
- **void fn**: OpUnreachable at end of a void function.  https://github.com/KhronosGroup/VK-GL-CTS/blob/6632073b4f79df50f93b27c3d2ed005d21f2c367/external/vulkancts/modules/vulkan/spirv_assembly/vktSpvAsmInstructionTests.cpp#L424 

I'm attaching a .csv file that analyzes the test diffs in the current version of the patch.
They fall into the following cases:
- nonvoid fn: tested as above
- restructured:  Original GLSL rewritten to avoid wiping away most of the codegen output. Generally rewrite by calling out to functions, each with their own isolated case, e.g. infinite loop.
- branch-to-unreach: tested as above
- "nonvoid fn: id rename"  Like nonvoid vn, but one branch ID is changed by spirv-remap
- new test:  A test added for this MR.

[glslang-pruning-dead-code-cases.csv.txt](https://github.com/KhronosGroup/glslang/files/3785105/glslang-pruning-dead-code-cases.csv.txt)

I'm pretty happy with the existing test coverage.  So I'll update the PR to make the new behaviour the one and only behaviour. That will simplify the code and testing.

SPIRV/SpvPostProcess.cpp
dneto0: 1. For the createShortCircuit case, this triggers only for || and &&.  In that case the code generated is in a "triangle" shape of a simple if:    

              if-header
                  |       \
                  |        then-clause  implements the right-hand-side of the && or ||
                  |       /
               if-merge

The if-merge block and then-clause are CFG-reachable if and only if the if-header block is reachable.  

So the remaining thing to look at is whether the the last block generated for the then-clause is wiped away but still referenced by the phi in the if-merge block.  This would occur if there were some control flow in the RHS which can yield a final block which is not CFG-reachable.  Since we're in an expression context, I think this does not occur.   
For example, I can't use "discard" or "return" in the RHS of &&.
We'd be in trouble if the RHS had a function that was inlined.

2.  The ternary operator (?:),   is handled by TGlslangToSpvTraverser::visitSelection.  There are a few cases: 
- (a) Use OpSelect.  Nothing exotic here.
- (b) Execute one or both arms of the ternary value.  In this case an "if" construct is generated.   Again, the if-merge block is reachable if and only if the if-header is reachable. Also, no Phis are generated; instead a function-variable is created to hold the resulting value of the expression, and the code stores to it.

I scanned GlslangToSpv.cpp and the only two places where the spv::Builder::If helper is used is in the cases above: `||`, `&&`, `?:`

Thanks for pointing out these issues.  I understand the codegen better and I'm still convinced the new behaviour is good and safe and tested. 

I'll work on removing the old behaviour entirely, and get the MR closer to its original level of complexity.

dneto0:
I think this is ready review, and that questions have hopefully addressed.

SPIRV/SpvPostProcess.cpp
johnkslang: looks good

johnkslang:
We should bump up the `GetSpirvGeneratorVersion()` since code gen is changing.

Also, tag, etc.

I will do these, before merging.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1630,11 +1630,11 @@ void TGlslangToSpvTraverser::finishSpv()
     for (auto it = iOSet.cbegin(); it != iOSet.cend(); ++it)
         entryPoint->addIdOperand(*it);
 
-#ifndef GLSLANG_WEB
-    // Add capabilities, extensions, remove unneeded decorations, etc., 
+    // Add capabilities, extensions, remove unneeded decorations, etc.,
     // based on the resulting SPIR-V.
+    // Note: WebGPU code generation must have the opportunity to aggressively
```
johnkslang: This turned on lots of code having to do with types and capabilities that should not be relevant to WebGPU.

Don't we need to just turn on the CFG topology part?

I can look at that too.

johnkslang: Well, just reply with the specific subset needed by WebGPU, and I'll do the experiment/pruning of other things and verify how size is impacted.

(First, I'll merge this, and then do that as a second step to get space back for WebGPU.)

SPIRV/spvIR.h
```diff
@@ -268,10 +299,24 @@ class Block {
     bool unreachable;
 };
 
+// The different reasons for reaching a block in the inReadableOrder traversal.
+typedef enum ReachReason {
```
johnkslang: What's the purpose of `typedef` here without also declaring a name?  I'm getting warnings on this have no purpose.  Does it work just as well to remove the `typedef`?

johnkslang: Note, I removed the `typedef`.

johnkslang:
I'm going to push this now, with some modifications, including new test results due to bumping up the generator version number.

SPIRV/GlslangToSpv.cpp
dneto0: > Don't we need to just turn on the CFG topology part?

In principle, yes.

But when I looked more deeply at Builder::postProcess, I became worried that the WEB path was missing functionality, particularly around adding per-instruction capabilities and extensions.
Looking again, at the moment those additional things are only affecting:
-  8bit and 16bit storage
- physical storage
Neither of these are in the WEB variant for now, so that could be saved.
I can take on getting that code space back.

dneto0: I posted https://github.com/KhronosGroup/glslang/pull/1968 to reclaim the codespace wasted by this part of the change.

johnkslang: That looks like it solves it, thanks.

PR #1944, opened 2019-10-22
BUILD.gn: Fix fuchsia build

digit-google:
The Fuchsia platform build and the Chromium one use a completely
different set of configs to specify default warnings. The previous
version of BUILD.gn worked with Chromium, but failed to build
with Fuchsia.

This CL fixes the issue by identifying said configs and reusing
them consistently.

Change-Id: I7de526a57d2f14eb93c03e06401d2c3059d35e9a

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -33,6 +33,19 @@
 
 import("//build_overrides/glslang.gni")
 
+# Both Chromium and Fuchsia use by default a set of warning errors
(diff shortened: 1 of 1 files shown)


digit-google:
I have tested this with both a recent Fuchsia and Chromium Linux build. @zoddicus, can you please take a look at this pull request? Thanks.

PR #1972, opened 2019-11-11
Fix construction issue for 8 and 16 bit types.

andflo-Arm:
The intention of the previous code seems to be to translate the operator into the corresponding 32-bit version (e.g., `EOpConstructI8Vec2` -> `EOpConstructIVec2`). It works fine for the vector types, but the scalar types evaluates to the wrong enum variant. This is what the scalar types were translated into:
- float16_t: EOpConstructDouble
- int8_t: EOpConstructMat2x2
- uint8_t: EOpConstructMat2x3
- int16_t: EOpConstructDouble
- uint16_t: EOpConstructVec2

For the 8-bit types, this resulted in a crash downstream when asserting that a type was indeed a matrix. The others did compile, but I expect that is just by luck.

This was found in a shader like this:
```
#version 450
#extension GL_EXT_shader_8bit_storage : enable
#extension GL_EXT_shader_16bit_storage : enable

precision highp int;

void main()
{
    // Note: 32 bit int not relevant, it's just to avoid having to store the
    // int8_t to a storage buffer. Crash happens during construction of 8-bit
    // using non-32-bit parameter.
    int i16_to_i8_to_int = int(int8_t(int16_t(1)));
}
```


johnkslang:
@jeffbolznv any input on the scalar usage here?

jeffbolznv:
This looks reasonable, but it would be nice to add a test as part of this PR.

johnkslang:
Agreed the change should have a corresponding test-suite change.

andflo-Arm:
The test catches the crash, but it's not entirely correct. The integer cases seem fine; they truncate and extend between various integer widths as expected. However, the float case is incorrect. It first truncates a 32-bit integer into 8 (fine), but then uses that result in the `FConvert`, which is illegal because it takes another float type as its operand. I would expect to see an `OpConvertSToF` in-between.

This appears to happen for float16_t even before this patch. I think it's related to this intermediate aggregator thing. Before my patch the aggregate operator would be set with the wrong value during compilation, but it never ends up in the output.

johnkslang:
Probably doesn't matter for this bug fix, but you should sign the CLA.

johnkslang:
Will look at the fconvert problem separately.

johnkslang:
I opened https://github.com/KhronosGroup/GLSL/issues/100 for the remaining issue.  Glslang is neither giving an error nor using the right operator, but we should first be clear which way to go.

PR #1982, opened 2019-11-19
[GLSL][PP] extension pp check after non-pp tokens

ShchchowAMD:
**[Purpose]**

Currently we lack of checking whether "#extension" is used after any non-pp tokens/statements.

According to spec, this situation should be treated as an error.

Here is a fix for issues with following pattern:

**[Shader]**
...

void main{
...
}

GLSL should notify users that this would lead to a compilation error as "#extension" should not be used after non-pp tokens.

**[Concerns]**
I don't know whether it is OK to use a simple flag/mask attribute to represent current pp state.
Or, should I add this check following a different way?

I will add tests if this change is OK according to your concerns.

### DIFF
--- a/glslang/MachineIndependent/preprocessor/Pp.cpp
+++ b/glslang/MachineIndependent/preprocessor/Pp.cpp
@@ -873,6 +873,16 @@ int TPpContext::CPPextension(TPpToken* ppToken)
     int token = scanToken(ppToken);
     char extensionName[MaxTokenLength + 1];
 
+    // extension directives must occur before any non-preprocessor tokens
+    if (parseContext.getPpContext()->nonMacroStarted == true) {
+        if (parseContext.relaxedErrors())
+            parseContext.ppWarn(ppToken->loc, "extension directives must occur before any non-preprocessor tokens", "#extension", "");
+        else
+            parseContext.ppError(ppToken->loc, "extension directives must occur before any non-preprocessor tokens", "#extension", "");
+        return token;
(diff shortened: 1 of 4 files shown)


ShchchowAMD:
test fails due to ERROR: spv.400.frag:15: '#extension' : extension directives must occur before any non-preprocessor tokens

Will modify test cases if this PR modification is OK

johnkslang:
> According to spec, this situation should be treated as an error.

It was actually conditionally an error, not categorically an error. 

TL;DR: This is a confusing space, and I don't think we want to make this change.

Details

What the specification says is:

> Each extension can define its allowed granularity of scope. If nothing is said, the granularity is a shader (that is, a single compilation unit), and the extension directives must occur before any non-preprocessor tokens.

That means, it is up to each extension to define where it can be turned on or off.  This broke down when most extensions failed to do this, and for many it was actually okay to turn them on/off, and the current test suite takes full advantage of this.  I'm concerned that now this might be a backward incompatible change.

At minimum though, if we were to fix it, the fix would need a little database of all extensions stating at what granularity each if them can be enabled/disabled, so that the semantic check is implemented correctly per extension.

ShchchowAMD:
I fully understand and I think this should be considered to use a more formal solution as you mentioned(a little db would be better).

I will remove this PR and do some local changes currently (I will try to work on it as I find some refine commits already in initialize.cpp).

Thanks for your kind reply.

PR #1988, opened 2019-11-21
Check for ENABLE_SPVREMAPPER flag in CMakeList files.

dj2:
There is a flag to disable the SPVRemapper during the GLSLang build.
That flag is check in some, but not all spots so if you try to build
with SPVRemapper disabled you get CMake errors and compile errors.

This CL fixs up the build so building with -DENABLE_SPVREMAPPER=0 will
complete correclty.


johnkslang:
Yes, it seems to be broken.  I guess I just caused merge conflicts?  Can you look and fix?

johnkslang:
Okay, I fixed the conflicts locally and merged.

PR #1990, opened 2019-11-25
Implement INTEL_shader_integer_functions2

ianromanick:
The SPIR-V extension landed quite some time ago.  The GLSL extension is at https://github.com/KhronosGroup/OpenGL-Registry/pull/331.  This has been tested with the implementation of the VK extension in Intel's open-source Vulkan driver using the tests at https://gitlab.freedesktop.org/mesa/crucible/merge_requests/59.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2608,6 +2608,17 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
         atomic = true;
         break;
 
+    case glslang::EOpAbsDifference:
+    case glslang::EOpAddSaturate:
+    case glslang::EOpSubSaturate:
+    case glslang::EOpAverage:
+    case glslang::EOpAverageRounded:
+    case glslang::EOpMul32x16:
+        builder.addCapability(spv::CapabilityIntegerFunctions2INTEL);
+        builder.addExtension("SPV_INTEL_shader_integer_functions2");
+        binOp = node->getOp();
+        break;
+
     case glslang::EOpIgnoreIntersectionNV:
     case glslang::EOpTerminateRayNV:
     case glslang::EOpTraceNV:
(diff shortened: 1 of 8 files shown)


ianromanick:
There was a test failure in `Glsl/CompileVulkanToSpirvTest.FromFile/spv_specConstant_vert`.  It's not obvious to me that the failure is my fault, nor is it obvious what to do about it.  Suggestions?

johnkslang:
If it's just that index change, that sometimes happens, but in a portable way, meaning your local test run should also have seen that same change.

ianromanick:
With some difficulty, I managed to get the same results locally.  I updated the commit that regressed the test with a change to the expected test output.

SPIRV/GlslangToSpv.cpp
```diff
@@ -2608,6 +2608,15 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
         atomic = true;
         break;
 
+    case glslang::EOpAbsDifference:
+    case glslang::EOpAddSaturate:
+    case glslang::EOpSubSaturate:
+    case glslang::EOpAverage:
+    case glslang::EOpAverageRounded:
+    case glslang::EOpMul32x16:
+        binOp = node->getOp();
```
johnkslang: See note below about `addExtension` and `addCapability`.

SPIRV/GlslangToSpv.cpp
```diff
@@ -5231,6 +5240,42 @@ spv::Id TGlslangToSpvTraverser::createBinaryOperation(glslang::TOperator op, OpD
         binOp = spv::OpLogicalNotEqual;
         break;
 
+    case glslang::EOpAbsDifference:
+        builder.addCapability(spv::CapabilityIntegerFunctions2INTEL);
```
johnkslang: Consider doing these pair once above (previous note) instead of 6 times here?

```
        builder.addCapability(spv::CapabilityIntegerFunctions2INTEL);
        builder.addExtension("SPV_INTEL_shader_integer_functions2");
```

glslang/MachineIndependent/intermOut.cpp
```diff
@@ -213,6 +213,12 @@ bool TOutputTraverser::visitBinary(TVisit /* visit */, TIntermBinary* node)
     case EOpLogicalXor: out.debug << "logical-xor"; break;
     case EOpLogicalAnd: out.debug << "logical-and"; break;
 
+    case EOpAddSaturate:            out.debug << "addSaturate";           break;
+    case EOpSubSaturate:            out.debug << "subtractSaturate";      break;
+    case EOpAverage:                out.debug << "average";               break;
+    case EOpAverageRounded:         out.debug << "averageRounded";        break;
+    case EOpMul32x16:               out.debug << "multiply32x16";         break;
```
johnkslang: Is EOpAbsDifference missing here?

These show up when printing the AST (-i option from the command line).

SPIRV/GlslangToSpv.cpp
ianromanick: Ah, yes.

glslang/MachineIndependent/intermOut.cpp
ianromanick: It is missing.  Good catch.

ianromanick:
I'm not sure what to do about AppVeyor being grumpy, but I incorporated all of the feedback from @johnkslang.

PR #1995, opened 2019-11-26
Add support for ARB_sample_shading

ShchchowAMD:
Version : >= 130

Purpose:
Allow users to use features by enabling this extension, even in low versions. Related features have been already supported by SPIR-V and GLSL, just add extensions for users who want to enable this extension in low versions.

Extension Name: 
ARB_sample_shading

Builtin-variables:
"gl_SampleID"
"gl_SamplePosition"
"gl_SampleMask"
"gl_NumSamples"

Reference:
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_sample_shading.txt

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -5178,18 +5178,24 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
                 "flat in int gl_PrimitiveID;"
                 );
 
-        if (version >= 400) {
+        if (version >= 130) { // ARB_sample_shading
             stageBuiltins[EShLangFragment].append(
                 "flat in  int  gl_SampleID;"
                 "     in  vec2 gl_SamplePosition;"
-                "flat in  int  gl_SampleMaskIn[];"
                 "     out int  gl_SampleMask[];"
                 );
-            if (spvVersion.spv == 0)
+
+            if (spvVersion.spv == 0) {
                 stageBuiltins[EShLangFragment].append(
                     "uniform int gl_NumSamples;"
-                    );
+                );
+            }
         }
+
+        if (version >= 400)
+            stageBuiltins[EShLangFragment].append(
+                "flat in  int  gl_SampleMaskIn[];"
+            );
 
         if (version >= 430)
             stageBuiltins[EShLangFragment].append(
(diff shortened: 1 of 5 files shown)


glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8074,18 +8078,27 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
+            BuiltInVariable("gl_SampleID",           EbvSampleId,       symbolTable);
+            BuiltInVariable("gl_SamplePosition",     EbvSamplePosition, symbolTable);
+            BuiltInVariable("gl_SampleMask",         EbvSampleMask,     symbolTable);
+            if (profile != EEsProfile && version < 400) {
+                BuiltInVariable("gl_NumSamples",     EbvSampleMask,     symbolTable);
+                symbolTable.setVariableExtensions("gl_SampleMask",     1, &E_GL_ARB_sample_shading);
+                symbolTable.setVariableExtensions("gl_SampleID",       1, &E_GL_ARB_sample_shading);
+                symbolTable.setVariableExtensions("gl_SamplePosition", 1, &E_GL_ARB_sample_shading);
+                symbolTable.setVariableExtensions("gl_NumSamples",     1, &E_GL_ARB_sample_shading);
+            }
```
johnkslang: Put `}` and `else` on the same line to match the local style.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5861,19 +5861,23 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
                 "flat in int gl_PrimitiveID;"
                 );
 
-        if (version >= 400) {
```
johnkslang: These line numbers are off by about 800 from my version of master.  Curious if this is based on an old version of master?

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5861,19 +5861,23 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
...
             stageBuiltins[EShLangFragment].append(
                 "flat in  int  gl_SampleID;"
                 "     in  vec2 gl_SamplePosition;"
-                "flat in  int  gl_SampleMaskIn[];"
                 "     out int  gl_SampleMask[];"
                 );
             if (spvVersion.spv == 0)
                 stageBuiltins[EShLangFragment].append(
-                    "uniform int gl_NumSamples;"
-                    );
```
johnkslang: I think this was the style throughout; can we keep it here too?

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -5861,19 +5861,23 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
                 "flat in int gl_PrimitiveID;"
                 );
 
-        if (version >= 400) {
+        if (version >= 130) { // ARB_sample_shading
```
johnkslang: We should have a test that verifies the extension working in version 130.  Probably by just adding a section at the end of  130.frag that:

1. starts with attempts to use these names that fail
2. `#extension ARB_sample_shading : enable`
3. use of the names that succeed

This is the existing pattern, so hopefully not too difficult.

ShchchowAMD:
Thanks for your suggestions, I'm trying to fix those issues and adding test in 130.frag.

ShchchowAMD:
Hi John, I've fixed all the issues you mentioned before, including coding styles and adding tests as you suggested. Please have a review.

PR #1998, opened 2019-11-26
Add support for ARB_gpu_shader_fp64

ShchchowAMD:
GLSL Version : >= 150

Purpose:
Allow users to use features by enabling this extension, even in low versions.

Extension Name:
ARB_gpu_shader_fp64

Builtin-variables:
Nah

Builtin-functions:
functions overloaded for this extension, please check registry in reference.

Keywords:
Double

Features:
add support for type "double", also related functions/builtin functions
overloaded for DOUBLE. "double vector" input should be in
ARB_vertex_attrib_64bit

Reference:
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_gpu_shader_fp64.txt

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -690,7 +690,7 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
     //
     // double functions added to desktop 4.00, but not fma, frexp, ldexp, or pack/unpack
     //
-    if (profile != EEsProfile && version >= 400) {
+    if (profile != EEsProfile && version >= 150) {  // ARB_gpu_shader_fp64
         commonBuiltins.append(
 
             "double sqrt(double);"
@@ -1298,15 +1298,15 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "vec3   fma(vec3,   vec3,   vec3  );"
             "vec4   fma(vec4,   vec4,   vec4  );"
             "\n");
+    }
 
-        if (profile != EEsProfile) {
+    if (profile != EEsProfile && version >= 150) {  // ARB_gpu_shader_fp64
             commonBuiltins.append(
                 "double fma(double, double, double);"
                 "dvec2  fma(dvec2,  dvec2,  dvec2 );"
                 "dvec3  fma(dvec3,  dvec3,  dvec3 );"
                 "dvec4  fma(dvec4,  dvec4,  dvec4 );"
                 "\n");
-        }
     }
 
     if ((profile == EEsProfile && version >= 310) ||
@@ -1325,7 +1325,7 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "\n");
     }
 
-    if (profile != EEsProfile && version >= 400) {
+    if (profile != EEsProfile && version >= 150) { // ARB_gpu_shader_fp64
         commonBuiltins.append(
             "double frexp(double, out int);"
             "dvec2  frexp( dvec2, out ivec2);"
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -6091,7 +6091,7 @@ const TFunction* TParseContext::findFunction(const TSourceLoc& loc, const TFunct
     if (isEsProfile() || version < 120)
         function = findFunctionExact(loc, call, builtIn);
     else if (version < 400)
-        function = findFunction120(loc, call, builtIn);
+        function = extensionTurnedOn(E_GL_ARB_gpu_shader_fp64) ? findFunction400(loc, call, builtIn) : findFunction120(loc, call, builtIn);
     else if (explicitTypesEnabled)
         function = findFunctionExplicitTypes(loc, call, builtIn);
     else
--- a/glslang/MachineIndependent/glslang.m4
+++ b/glslang/MachineIndependent/glslang.m4
@@ -402,7 +402,9 @@ GLSLANG_WEB_EXCLUDE_ON
         $$ = parseContext.intermediate.addConstantUnion((unsigned short)$1.u, $1.loc, true);
     }
     | DOUBLECONSTANT {
-        parseContext.doubleCheck($1.loc, "double literal");
+        parseContext.requireProfile($1.loc, ECoreProfile | ECompatibilityProfile, "double literal");
+        if (! parseContext.symbolTable.atBuiltInLevel())
+            parseContext.doubleCheck($1.loc, "double literal");
         $$ = parseContext.intermediate.addConstantUnion($1.d, EbtDouble, $1.loc, true);
     }
     | FLOAT16CONSTANT {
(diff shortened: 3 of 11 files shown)


glslang/MachineIndependent/glslang.m4
```diff
@@ -1751,7 +1751,7 @@ type_specifier_nonarray
     }
 GLSLANG_WEB_EXCLUDE_ON
     | DOUBLE {
-        parseContext.doubleCheck($1.loc, "double");
+        parseContext.doubleCheck($1.loc, "double", parseContext.symbolTable.atBuiltInLevel());
```
johnkslang: Is adding an argument necessary?  `parseContext` already has `symbolTable`, so it can check itself.

That is, it seems there is not much use now for the form that does not check `atBuiltInLevel` so, just leave all the calls the same, and only update the function.  (Or, add a parseContext.doubleCheck() that checks the builtInLevel.)

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -1239,15 +1239,15 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "vec3   fma(vec3,   vec3,   vec3  );"
             "vec4   fma(vec4,   vec4,   vec4  );"
             "\n");
+    }
 
-        if (profile != EEsProfile) {
+    if (profile != EEsProfile && version >= 150) {  // ARB_gpu_shader_fp64
             commonBuiltins.append(
                 "double fma(double, double, double);"
```
johnkslang: What are the impacts to overloaded function resolution under implicit type conversion?  Version 400 has different rules than versions before 400.

glslang/MachineIndependent/glslang.m4
ShchchowAMD: Hi john, 
The reason for this modification is that: 
1. initialization (built-in functions and variables) happens before we update extensions status in state machine.
2. when I remove this builtin level check, glslang would report an error because it would treat "double" as an unavailable symbol in low versions during initialization when meets some overloaded functions with type "double".
3. if we add an builtin level check here:
    3.1 initialization process would be okay as "doublecheck()" now knows they are overloaded functions.
    3.2 users, still, can only use high core version or enabling extension if they need to use fp64.

For your suggestion, do you mean that I should add another doubleCheck() for builtinLevel symbols and keep calls same in .y file? Or maybe, try to do the builtinLevel check within doubleCheck()?

glslang/MachineIndependent/Initialize.cpp
ShchchowAMD: I'll test for this and record here soon.

glslang/MachineIndependent/glslang.m4
johnkslang: >  try to do the builtinLevel check within doubleCheck()?

Yes.  Does that work?

ShchchowAMD: Yes I tried before I committed in this version.

As if we put a breakpoint here, in doubleCheck(), we could know that: when a "double" symbol comes in, we have no extra information to know whether it is a built-in. We could also see that other check functions also use builtin flag to check types.

For built-in variables or functions, doubleCheck() is a part of initializing symbol table, we can't find that flag in context here as we have no context member variables. Otherwise, we still need to change doubleCheck's interface, or add context status, till here by other parameters.

I think for many types, like float16, their check functions already get this builtin flag with a default input parameter, may I know what suggestions for this part?

glslang/MachineIndependent/Initialize.cpp
ShchchowAMD: Thanks for reminding, implicit converting meets errors. I'll fix it.

ShchchowAMD: @johnkslang  Hi, John. I made some tests and find the difference is inside findFunction120 and findFunction400. Should "best match" rules be enabled in version 120-330 if I turned on type "double" with this commit?

ShchchowAMD: Currently, I allow glslang to use findFunction400 when we turn on this extension, as type 'double' could be introduced with this extension.

johnkslang: Yes, that might be best, especially if formalized.  Someone could change their shader behavior of unrelated code, just by enabling the extension, but if it's well specified that's how it works, it may be the best path.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -6091,7 +6091,7 @@ const TFunction* TParseContext::findFunction(const TSourceLoc& loc, const TFunct
     if (isEsProfile() || version < 120)
         function = findFunctionExact(loc, call, builtIn);
     else if (version < 400)
-        function = findFunction120(loc, call, builtIn);
+        function = extensionTurnedOn(E_GL_ARB_gpu_shader_fp64) ? findFunction400(loc, call, builtIn) : findFunction120(loc, call, builtIn);
```
ShchchowAMD: @johnkslang FYI, now when this extension turns on, we will use rules based on version 400.

I think the reason is we may still wanna remain the difference between old versions and versions larger than 400. And keep the change as small as possible (rules inside findFunction400 may be modified in the future).

When type "double" is introduced, a best match check would be enabled to allow extra implicit conversion, that's when those three rules work, to decide a best candidate.

When this extension is enabled, we have type "float", "int" and "double", same with version >= 400, which make no difference with what we do in latest version.

glslang/MachineIndependent/glslang.m4
johnkslang: It seems like this adds this extra Boolean parameter to almost all (all?) calls to `doubleCheck()`, but with a query to the same object the method is in, so the method could more directly do the query itself.

ShchchowAMD: FYI, this Boolean parameter has been removed. Built-in check has been moved when call this object. There's no difference in generated test results for latest commit.

glslang/MachineIndependent/glslang.m4
```diff
@@ -402,7 +402,9 @@ GLSLANG_WEB_EXCLUDE_ON
         $$ = parseContext.intermediate.addConstantUnion((unsigned short)$1.u, $1.loc, true);
     }
     | DOUBLECONSTANT {
-        parseContext.doubleCheck($1.loc, "double literal");
+        parseContext.requireProfile($1.loc, ECoreProfile | ECompatibilityProfile, "double literal");
+        if (! parseContext.symbolTable.atBuiltInLevel())
```
johnkslang: I don't understand why you don't do this test inside the method instead of outside.

glslang/MachineIndependent/glslang.m4
johnkslang: I was trying to get one copy of the test instead of N copies.

glslang/MachineIndependent/glslang.m4
ShchchowAMD: Inside the method, the current class is TParseVersions, instead of TParseContext or TParseContextBase.
We don't have a symbolTable at that time.

I think that's also why in other places, they also use a Boolean parameter to do the check inside.
Like, void int64Check(const TSourceLoc&, const char* op, bool builtIn = false) { }

ShchchowAMD: Or, move doubleCheck to TParseContext or TParseContextBase will be okay?

ShchchowAMD: It seems be to better, but should I also change other type check functions which use this Boolean parameter?

Roy-AMD: > It seems be to better, but should I also change other type check functions which use this Boolean parameter?

I think it ok to keep current change, you can create a new PR for move all check to TParseContext or TParseContextBase. @johnkslang @ShchchowAMD

johnkslang: Okay, thanks.

PR #2004, opened 2019-12-02
Update README.md to include other test requirements

ianromanick:
Knowing this in advance would have saved me some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ cd ../..
 ```
 
 If you wish to assure that SPIR-V generated from HLSL is legal for Vulkan,
-or wish to invoke -Os to reduce SPIR-V size from HLSL or GLSL, install
-spirv-tools with this:
+wish to invoke -Os to reduce SPIR-V size from HLSL or GLSL, or wish to run the
+integrated test suite, install spirv-tools with this:
 
 ```bash
 ./update_glslang_sources.py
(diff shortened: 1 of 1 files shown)


ianromanick:
It might be worth updating the "except" list so that we don't have to run the CI when just `README.md` is changed. :smile:

README.md
```diff
@@ -236,6 +236,11 @@ Running `runtests` script-backed tests:
 cd $SOURCE_DIR/Test && ./runtests
 ```
 
+If some tests fail with validation errors, there may be a mismatch between the
+version of `spirv-val` on the system and the version of glslang.  In this
+case, it is necessary to run `update_glslang_sources.py`.  See "Check-Out
+External Projects" above for more details.
+
```
johnkslang: One should always run `update_glslang_sources.py` for many reasons before doing any new development.  It is part of the code base for HLSL legalization, validation, and testing.

johnkslang: If we are not saying this, we need to fix that; it's not just about validation.

README.md
```diff
@@ -103,8 +103,8 @@ cd ../..
 ```
 
 If you wish to assure that SPIR-V generated from HLSL is legal for Vulkan,
-or wish to invoke -Os to reduce SPIR-V size from HLSL or GLSL, install
-spirv-tools with this:
+wish to invoke -Os to reduce SPIR-V size from HLSL or GLSL, or wish to run the
+integrated test suite, install spirv-tools with this:
```
johnkslang: Yes, okay.  These dependencies changed after the fact (comparatively recently).

johnkslang:
Thanks.

PR #2026, opened 2019-12-17
Add specialization constant support in the reflection API

### LINKED ISSUE #2011: Feature request: Specialization constants in reflection
I was wondering if it's possible to add specialization constant support in glslang's reflection API. I haven't seen something relevant in TProgram class. 

Thank you in advance.

godlikepanos:
Also added it when dumping using -q option from glslangValidator. Fixes #2011

### DIFF
--- a/glslang/MachineIndependent/reflection.cpp
+++ b/glslang/MachineIndependent/reflection.cpp
@@ -146,10 +146,11 @@ public:
         base.getType().contains([this](const TType* t) {
             bool containsSpecializationSize = t->isArray() && t->getArraySizes()->isOuterSpecialization();
             if (containsSpecializationSize) {
-                const int dimension = 0; // Spec consts can only be in the outermost dimension
-                TIntermTyped* typed = t->getArraySizes()->getDimNode(dimension);
-                assert(typed);
-                traverseToFindSpecConsants(typed);
+                for (int dim = 0; dim < t->getArraySizes()->getNumDims(); ++dim) {
+                    TIntermTyped* typed = t->getArraySizes()->getDimNode(dim);
+                    if (typed != nullptr)
+                        traverseToFindSpecConsants(typed);
+                }
             }
 
             return false;
@@ -1118,7 +1119,7 @@ int TObjectReflection::getBinding() const
 
 void TObjectReflection::dump() const
 {
-    if (type->getQualifier().storage == EvqConst) {
+    if (type->getQualifier().specConstant) {
         printf("%s: constantId %d, type %d\n", name.c_str(), type->getQualifier().layoutSpecConstantId, glDefineType);
     } else {
         printf("%s: offset %d, type %x, size %d, index %d, binding %d, stages %d", name.c_str(), offset, glDefineType,
@@ -1146,12 +1147,37 @@ void TObjectReflection::dump() const
 
 // Track any required attribute reflection, such as compute shader numthreads.
 //
-void TReflection::buildAttributeReflection(EShLanguage stage, const TIntermediate& intermediate)
+void TReflection::buildAttributeReflection(
+    EShLanguage stage, const TIntermediate& intermediate, TReflectionTraverser& reflectionTraverser)
 {
     if (stage == EShLangCompute) {
         // Remember thread dimensions
-        for (int dim=0; dim<3; ++dim)
-            localSize[dim] = intermediate.getLocalSize(dim);
+        for (int dim = 0; dim < 3; ++dim) {
+            if (intermediate.getLocalSizeSpecId(dim) >= 0) {
+                // Searches all symbols for a spec constant with a specific constant_id
+                class SpecConstTraverser : public TIntermTraverser
+                {
+                public:
+                    TReflectionTraverser* reflectionTraverser;
+                    int specId;
+
+                    void visitSymbol(TIntermSymbol* base)
+                    {
+                        if (base->getQualifier().specConstant && base->getQualifier().layoutSpecConstantId == specId)
+                            reflectionTraverser->addSpecConstant(*base);
+                    }
+                };
+
+                SpecConstTraverser traverser;
+                traverser.reflectionTraverser = &reflectionTraverser;
+                traverser.specId = intermediate.getLocalSizeSpecId(dim);
+                intermediate.getTreeRoot()->traverse(&traverser);
+
+                localSizeSpecId[dim] = intermediate.getLocalSizeSpecId(dim);
+            } else {
+                localSize[dim] = intermediate.getLocalSize(dim);
+            }
+        }
     }
 }
 
@@ -1196,9 +1222,9 @@ bool TReflection::addStage(EShLanguage stage, const TIntermediate& intermediate)
         intermediate.isRecursive())
         return false;
 
-    buildAttributeReflection(stage, intermediate);
-
     TReflectionTraverser it(intermediate, *this);
+
+    buildAttributeReflection(stage, intermediate, it);
 
     // put the entry point on the list of functions to process
     it.pushFunction(intermediate.getEntryPointMangledName().c_str());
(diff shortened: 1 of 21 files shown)


johnkslang:
Can we have one test result that shows the new information emitted?

godlikepanos:
@johnkslang thank you for looking into this. I'm writing a test at the moment.

Unfortunately the test revealed an issue where a spec constant should be "live" but it's not (in other words it's not visited by TReflectionTraverser::visitSymbol).

This is a weird case where the spec constant is used as a size to an array:

```
layout(constant_id = 4) const int constLiveArraySize = 1;

layout(binding = 0, std140) uniform uniformBlock {
	vec4 liveUniform[constLiveArraySize];
};

void main()
{
    gl_Position += liveUniform[gl_VertexIndex];
}
```

I'll try to figure out what's wrong. Any pointers are welcome :)

johnkslang:
I suspect that in `visitSymbol` you'll have to look to see if there is any internal array sizing with specialization constants:

```
    if (base->getType().containsSpecializationSize()) ...
```

godlikepanos:
Thanks for the help! The test revealed a few issues in the original code. I'm not sure how clean the solution is though. Happy for any feedback in this area.

Also, I tried to cover all the use cases I could think of in the test. Not sure if I covered everything though. Happy for feedback in the test as well.

Thanks

glslang/MachineIndependent/reflection.cpp
```diff
@@ -100,6 +99,63 @@ class TReflectionTraverser : public TLiveTraverser {
...
+        typed->traverse(&traverser);
+    }
+
+    // Search for spec constant subtypes associated with this symbol
+    void searchForSpecConstantsInArrayDimensions(const TIntermSymbol& base)
+    {
+        base.getType().contains([this](const TType* t) {
+            bool containsSpecializationSize = t->isArray() && t->getArraySizes()->isOuterSpecialization();
+            if (containsSpecializationSize) {
+                const int dimension = 0; // Spec consts can only be in the outermost dimension
```
johnkslang: Some interfaces add a layer of per-vertex arrayness, such that one side of an interface is a single vertex, and the other side is an array of the same type.

In these cases, after ignoring the outermost level of arrayness, then the next level of arrayness can be a specialization constant.

Really, fundamentally, any inner dimension could be specialized, and it's a semi-arbitrary rule preventing that right now.

So, safest thing (and required for vertex-arrayed interfaces) is to just check all dimensions, not just the outer dimension.

godlikepanos: Thank you for looking into this! I'm not sure I understand this use-case though. Maybe some GLSL code will clear things up for me. Want to understand it in order to update the test.

Thanks

godlikepanos: @johnkslang I think I've corrected the behavior. I still need some help with the test though.

glslang/MachineIndependent/reflection.cpp
```diff
@@ -100,6 +99,64 @@ class TReflectionTraverser : public TLiveTraverser {
...
+        // Simple traverser that searches for spec constants
+        class TArrayDimensionTraverser : public TIntermTraverser
+        {
+        public:
+            TReflectionTraverser* reflectionTraverser = nullptr;
+
+            void visitSymbol(TIntermSymbol* base)
+            {
+                if (base->getType().getQualifier().specConstant) {
+                    // Some symbols contain the specConstant qualifier but in reality they are not. Check that
```
johnkslang: Do you mean some are specialization constants, but don't have an ID?  I think that's true.  Not all specialization constants have IDs, but I think everything marked as a specialization constant is a specialization constant.

glslang/MachineIndependent/reflection.cpp
```diff
@@ -100,6 +99,64 @@ class TReflectionTraverser : public TLiveTraverser {
...
+                    if (base->getType().getQualifier().layoutSpecConstantId != TQualifier::layoutSpecConstantIdEnd) {
```
johnkslang: Is the point that you only want to reflect specialization constants that can be set through the API?

glslang/MachineIndependent/reflection.cpp
```diff
@@ -100,6 +99,64 @@ class TReflectionTraverser : public TLiveTraverser {
...
+    }
+
+    // Search for spec constant subtypes associated with this symbol
+    void searchForSpecConstantsInArrayDimensions(const TIntermSymbol& base)
+    {
+        base.getType().contains([this](const TType* t) {
+            bool containsSpecializationSize = t->isArray() && t->getArraySizes()->isOuterSpecialization();
+            if (containsSpecializationSize) {
+                for (int dim = 0; dim < t->getArraySizes()->getNumDims(); ++dim) {
+                    TIntermTyped* typed = t->getArraySizes()->getDimNode(dim);
```
johnkslang: This appears to search through all the dimensions, but you only did it if the outer dimension was a specialization constant.  I  think you want to do this without conditioning it on the outer dimension.

glslang/MachineIndependent/reflection.cpp
godlikepanos: That is a very good question. I added this comment and the following check because some symbols are marked as spec consts but I don't think they really are. In the reflection.spec_consts.vert for example, the `someConst1` and `someConst2` have the `TQualifier::specConstant` set to true. It didn't seem right to me to include these regular constants in the reflection.

I'm not sure what is the right solution for the likes of `someConst1` and `someConst2`. I can always rephrase the comment or replace the check with something more correct. Any advice welcome.

glslang/MachineIndependent/reflection.cpp
godlikepanos: Should I change this check to:
```bool containsSpecializationSize = t->isArray() && (t->getArraySizes()->isOuterSpecialization() || t->getArraySizes()->isInnerSpecialization());```

File: glslang/MachineIndependent/reflection.cpp (hunk 2 above)
Comment (johnkslang) [reply glslang/MachineIndependent/reflection.cpp]: The set of specialization constants is a super set of the set of symbols with a specialization-constant ID.

```
layout (constant_id = 2) const int a = 1;
const int b = a + 1;
```

Here, `b` is a specialization constant, but it does not have an ID.  That is by design and totally correct.  It is not some strange inconsistency.  The value of `b` changes based on what `a`  is specialized to, it does not stay fixed at 2.

If you only want to reflect things with a specialization constant ID, I think you could just check for that.

> some symbols are marked as spec consts but I don't think they really are.

I'm unaware of any issues here, but hopefully the above example clears up any misunderstandings.

godlikepanos: Thank you for your explanation. My understanding was wrong apparently. I'll change the comment since IMHO only spec constants with an ID should take part in the reflection.

glslang/MachineIndependent/reflection.cpp
johnkslang: Since you check all the dimensions yourself, I think you don't need to precheck.  Just run your loop and look directly.

godlikepanos:
@johnkslang I think I'll close the pull request. Your concerns have been voiced internally in Khronos and I agree with your view. If you don't mind please give me a confirmation so I can pull the plug. Thanks for all your help.

johnkslang:
I think it is okay to reflect GLSL-declared specialization constants, though it might make more sense as part of a SPIR-V reflector instead.  Is that what you are referring to as a reason to close?  That was discussed in another PR, not this one, but it could apply here as well.

If you don't want to follow through with this, I agree with you closing it.  Thanks.

godlikepanos:
The main reason for closing is seeing a presentation from the Barcelona F2F (See "SPIR-V tooling update" -> "Glslang: Other concerns"). I won't go into details since this document is confidential :)

The 2nd reason is that SPIRV-Cross covers my needs at the moment. @themaister was kind enough to fix some issues around reflection in SPIRV-Cross and I can use that instead.

PR #2030, opened 2019-12-24
"Switch" with no statements till end

ShchchowAMD:
From version 440, according to spec, GLSL

"Removed the rule about having no statement between a label and the end of the switch statement; existing compiler errors for this can turn into a warning.".

So this should be treated as a warning since version 440, but still error before version 440 for NoESProfile

### DIFF
--- a/Test/430.comp
+++ b/Test/430.comp
@@ -79,6 +79,12 @@ void fooaoeu() {
      double globalCoef = 1.0;
      int i = globalCoef;            // ERROR, can't convert from double to int
      double di = i;
+     int j = 0;
+     switch(j)                       // ERROR
+     {
+     case 1: break;
+     default:
+     }
 }
 
 in inb {     // ERROR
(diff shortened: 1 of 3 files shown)


johnkslang:
I think this was removed because it was not well defined, and the fix was (perhaps ambiguously) okay for multiple releases, not just the release it was published.  Will take some time/archaeology to be sure.

How important is this?

ShchchowAMD:
This fix is to follow specific GLSL documents. Many tests tools/frameworks will test difference among various versions of GLSL following specification documents.

Anyway, this is not a strictly defined check point as other issues. Before this fix, related error is relaxed as all GLSL on desktop will only get a warning rather than error.

johnkslang:
I think the current behavior is correct. Khronos just didn't go through all past versions of the specification to make a new revision just to fix this minor ill-defined issue.

PR #2031, opened 2019-12-24
Add flag to check whether offset is implicit or explicit

ShchchowAMD:
When we use unsized array in shader storage buffer, glslang calculate the offset during delcaring the block, it may lead to incorrect block offsets when its implicit array size changed.

So here is what we do:
1. For GLSL, we add flag explicitOffset in TQualifier, and set it when layout offset is specified explicitly
2. By using this flag we could tell difference as whether it is an explicit offset, and recalculate the block member offset conditionally in OpenGL.

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -553,6 +553,7 @@ public:
     // having a constant_id is not sufficient: expressions have no id, but are still specConstant
     bool specConstant : 1;
     bool nonUniform   : 1;
+    bool explicitOffset   : 1;
 
 #ifdef GLSLANG_WEB
     bool isWriteOnly() const { return false; }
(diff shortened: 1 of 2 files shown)


glslang/Include/Types.h
```diff
@@ -553,6 +553,7 @@ class TQualifier {
     // having a constant_id is not sufficient: expressions have no id, but are still specConstant
     bool specConstant : 1;
     bool nonUniform   : 1;
+    bool explicitOffset   : 1;
```
johnkslang: This is never consumed, only declared and set.

ShchchowAMD: Hi John,

We will use this flag in OpenGL to recalculate block offsets when we could know blocks' offsets have been implicitly set.

johnkslang: I was going to say something about generalized reflection information from the HLL.  But:

> in OpenGL

Do you mean in a driver, consuming the AST from glslang?  Or SPIR-V?

ShchchowAMD: Sorry to make you confused.

As you mentioned before, this flag will be used to :

> changing the number of elements in the run-time sized array, relocating where the run-time sized arrays' offset

Yes this will be consumed in a driver based on AST from glslang.

johnkslang: Okay, I think it is interesting.  It should maybe be part of your fork for your driver, but it is small enough, I think it is okay.

The issue is that it should keep working, but has no tests, because it is consumed elsewhere.  If you're willing to accept the burden of maintaining it, without regression tests, I think that is your choice to make.

PR #2033, opened 2019-12-24
Add Error check flag in io mapper

ShchchowAMD:
GLSL could detect contradictory uniform binding, but missing error reporting, even if we could now observe error message inside infosink.

Add error check when first time validate uniform binding. 
This error check flag will be used during linking time. To quickly report errors like contradictory uniform bindings.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -831,6 +831,7 @@ void TDefaultGlslIoResolver::reserverStorageSlot(TVarEntryInfo& ent, TInfoSink& 
                 if (iter->second != location) {
                     TString errorMsg = "Invalid location: " + name;
                     infoSink.info.message(EPrefixInternalError, errorMsg.c_str());
+                    hasError = true;
                 }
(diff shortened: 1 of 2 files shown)


glslang/MachineIndependent/iomapper.cpp
```diff
@@ -884,6 +884,7 @@ void TDefaultGlslIoResolver::reserverResourceSlot(TVarEntryInfo& ent, TInfoSink&
             if (iter->second != binding) {
                 TString errorMsg = "Invalid binding: " + name;
                 infoSink.info.message(EPrefixInternalError, errorMsg.c_str());
+                hasError = true;
```
johnkslang: I see three places using `EPrefixInternalError` that are not setting an error flag, all seem to be in the same class, `TDefaultGlslIoResolver`.  Should this be used in all three places?

ShchchowAMD: Thanks for reminding.
Other two checks for TDefaultGlslIoResolver has been added, please have a check.

PR #2034, opened 2019-12-24
Add implementation dependent limit for uniform buffer bindings

ShchchowAMD:
GLSL could correctly check uniform bindings for atomic counter and samplers. But failing to check binding limit for uniform block.

As other two cases use built-in MAX variables for checking and to provide to users to fetch and check, a new limit variable for uniform block could be used for limit checking and also shader coding usage.

GLSL doesn't have a built-in level variable for this, so we don't set it as a built-in, but an implementation dependent limit.

PS. uniform buffer bindings is an implementation dependent value, introduced in OpenGL Spec but not from glslang spec. We just want to use this macro to report limit error during compilation, otherwise any wired bindings (like 10k) could pass compilation and lead to unexpected errors.


ShchchowAMD:
Move to other CLS

Issue #1951, opened 2019-10-28
How to use glslang with C compiler?


sasmaster:
I read in the documentation that ShaderLang.h contains both C and C++ APIs. But I also see that C++ part of the header is not guarded against C compiler. So my C program spits tons of errors for all the .cpp API related syntax. Does it mean one can't use glslang with pure C compiler? Otherwise what do you suggest to do? Delete the CPP part from the header? 

Thanks.

johnkslang:
The "Sh" stuff was intended to be a C interface.  Is it okay to still use a C++ compiler to create the libraries?  If so, we should be able to expose just the C interface to the that library, maybe by changing the public header to exclude the C++ interface.

sasmaster:
Guys, IMHO it's totally unacceptable such a low level API doesn't provide clean C interface. There are companies out there which still write graphics software in C  ;)

johnkslang:
I'm happy to work with someone to re-establish a C interface.

STL has a data() method that should be useful for wrapping functions like `GlslangToSpv()`.

sasmaster:
> I'm happy to work with someone to re-establish a C interface.
> 
> STL has a data() method that should be useful for wrapping functions like `GlslangToSpv()`.

Then there should be another method that gives user the expected size of the buffer.  I am going to integrate glslang into commercial project so maybe I can help. C interface is of significant importance for me as I tend to use STL as seldom as possible.

johnkslang:
Given that size is not known until after the work is done, it seems the current method could be wrapped with the C interface (intentionally avoiding what might be the correct names for the moment):

`getSizeAndHandle()`:  get the size of the return and a handle to it if necessary.  Calls `GlslangToSpv()` to compile, get a handle to the result and the size, caching the result somewhere.

`getResult()`: copy out the result into user-provided memory

Maybe that results in new methods too, but I think it's fine to compute the result at the time of the first call, rather than trying to estimate size.

corporateshark:
@johnkslang 

We have bumped into the same problem - absence of any up-to-date C interface to `glslang` and `SPIRV`. Our use case is somewhat limited, we use these two to libraries at run time to convert GLSL shaders to SPIRV. Would there be any interest, if we publish our solution as a pull request here?

However, the only subset of features we use is the `TShader` interface and `glslang::GlslangToSpv()`.

sasmaster:
> @johnkslang
> 
[...]

PR #2038, opened 2019-12-24
Implemented C-language interface for glslang

corporateshark:
This PR implements a C-language interface for the glslang compiler (#1951).

Usage example (Vulkan 1.1):
```C++
	const char* shaderCodeVertex = ...;

	const glslang_input_t input =
	{
		.language = GLSLANG_SOURCE_GLSL,
		.stage = GLSLANG_STAGE_VERTEX,
		.client = GLSLANG_CLIENT_VULKAN,
		.client_version = GLSLANG_TARGET_VULKAN_1_1,
		.target_language = GLSLANG_TARGET_SPV,
		.target_language_version = GLSLANG_TARGET_SPV_1_3,
		.code = shaderCodeVertex,
		.default_version = 100,
		.default_profile = GLSLANG_NO_PROFILE,
		.force_default_version_and_profile = false,
		.forward_compatible = false,
		.messages = GLSLANG_MSG_DEFAULT_BIT,
	};

	glslang_initialize_process();

	glslang_shader_t* shader = glslang_shader_create( &input );

	if ( !glslang_shader_preprocess(shader, &input) )
	{
		// use glslang_shader_get_info_log() and glslang_shader_get_info_debug_log()
	}

	if ( !glslang_shader_parse(shader, &input) )
	{
		// use glslang_shader_get_info_log() and glslang_shader_get_info_debug_log()
	}

	glslang_program_t* program = glslang_program_create();
	glslang_program_add_shader( program, shader );

	if (!glslang_program_link(program, GLSLANG_MSG_SPV_RULES_BIT | GLSLANG_MSG_VULKAN_RULES_BIT))
	{
		// use glslang_program_get_info_log() and glslang_program_get_info_debug_log();
	}

	glslang_program_SPIRV_generate( program, input.stage );

	if ( glslang_program_SPIRV_get_messages(program) )
	{
		printf("%s", glslang_program_SPIRV_get_messages(program));
	}

	glslang_shader_delete( shader );

	const VkShaderModuleCreateInfo ci =
	{
		.sType    = VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO,
		.codeSize = glslang_program_SPIRV_get_size(program) * sizeof(unsigned int),
		.pCode    = glslang_program_SPIRV_get_ptr(program)
	};

	VkResult result = vkCreateShaderModule(device, &ci, nullptr, ...);

	glslang_program_delete( program );
```


johnkslang:
Just starting to review, but one early request: For things in the glslang C++ namespace, I'd like to keep the "glslang" namespace in C as well. I think SH was the original "namespace", but if we're trying this anew, it seems better to somehow use `glslang` or `GLSLANG`.

corporateshark:
@johnkslang Changed `SH_` to `GLSLANG_`.

glslang/Include/glslang_c_shader_types.h
```diff
@@ -0,0 +1,167 @@
...
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**/
+
+#ifndef C_SHADER_TYPES_H_INCLUDED
+#define C_SHADER_TYPES_H_INCLUDED
+
+/* EShLanguage counterpart */
```
johnkslang: I think we need both sides (counterpart) data structures to clearly say to edit both if you edit either one.

(Ideally, there'd be one place, rather than two to make changes.)

corporateshark: What do you think if we add `static_assert` inside `glslang_c_interface.cpp` to check equality of the integer values?

johnkslang: Maybe a static assert that the count changes, if that's possible.  That is, I doubt existing ones change value so much as something new getting added, but in only one place instead of two places.

corporateshark: Added `static_assert` for every enum used in the new C interface.

glslang/Public/ShaderLang.h
```diff
@@ -206,6 +212,7 @@ typedef enum {
     EShOptNone,
     EShOptSimple,       // Optimizations that can be done quickly
     EShOptFull,         // Optimizations that will take more time
+	 LAST_ELEMENT_MARKER(EshOptLevelCount),
```
johnkslang: Seem to be a few tab issues.  Should all be spaces.

Vironacorus:
How am I supposed to get glslang::DefaultTBuiltInResource?

corporateshark:
@Vironacorus The fast way is to get the default values from `glslang/StandAlone/ResourceLimits.cpp`. You can either copy the values or link this file to your project. The correct way is to construct the values yourself based on the caps of your target GPU.

Vironacorus:
I have a few questions:
-What's the difference between program and a shader? I'm confused by distinction between those.
-Should glslang_initialize_process be called every time I want to compile a shader or only once?

greg-lunarg:
Independent compilation units written in GLSL are called shaders. A program is a set of
shaders that are compiled and linked together, completely creating one or more of the
programmable stages of the API pipeline.

greg-lunarg:
glslang_initialize_process() should be called just once.

PR #2041, opened 2019-12-31
Reflection issue fix

Roy-AMD:
#### 1. Reserve unused std140 uniform block in reflection, and fix the uniform block struct member not affected by block matrix layout and memory layout.

>According to the spec OpenGL spec4.60.7:
>>7.6 Uniform Variables:
>>>All members of a named uniform block declared with a shared or std140 layout qualifier are considered active, even if they are not referenced in any shader in the program. The uniform block itself is also considered active, even if no member of the block is referenced.

We should reserve std140 block and shared block in reflection.

>According to the spec glsl4.60.7:
>>4.4.5. Uniform and Shader Storage Block Layout Qualifiers:
>>>The row_major and column_major qualifiers only affect the layout of matrices, including all
matrices contained in structures and arrays they are applied to, to all depths of nesting. These qualifiers can be applied to other types, but will have no effect.

We need ensure all matrix block member been effect, include the member of a struct in this block.

#### 2. Support EShMsgKeepUncalled in reflection. 

EShMsgKeepUncalled  is a link message for link program. We need only one option to control uncalled function optimization. If we set EShMsgKeepUncalled as false in link time, linker won't be keep the uncall function sequence in AST,  and if we set EShMsgKeepUncalled as true in link time, linker will keep all uncalled function sequence in AST.

#### 2. Recursively layout packing to "block member".
Layout packing isn't set recursively, it causes TReflection::getOffsets doesn't work correctly.

#### 3. Recursively layout matrix to "block member". 
Layout matrix isn't set recursively, it causes a matrix layout incorrectly issue.

So, in reflecte time, we just only travers all function sequence. It make EShMsgKeepUncalled  only work at linker, and can effect reflection.


Test/baseResults/300layout.vert.out
```diff
@@ -35,11 +35,11 @@ ERROR: node is still EOpNull!
...
+0:46                    'tblock' (layout( row_major std140) uniform block{layout( row_major std140 offset=0) uniform highp 4X4 matrix of float M1, layout( column_major std140 offset=64) uniform highp 4X4 matrix of float M2, layout( row_major std140 offset=128) uniform highp 3X3 matrix of float N1, layout( row_major std140 offset=176) centroid uniform highp float badf, layout( row_major std140 offset=180) uniform highp float badg, layout( row_major std140 offset=184) uniform highp float bad1, layout( row_major std140 offset=188) uniform highp float bad2, layout( row_major std140 offset=192) uniform highp float bad3})
```
johnkslang: Do you know why these lines are marked as different?

Test/baseResults/310.inheritMemory.frag.out
```diff
@@ -9,102 +9,102 @@ Shader version: 310
 0:19      'el' ( in mediump float)
 0:20  Function Definition: non_ro_funS(struct-S-f1[10]1; ( global void)
 0:20    Function Parameters: 
-0:20      's' ( in structure{ global 10-element array of mediump float buff})
+0:20      's' ( in structure{layout( std430) global 10-element array of mediump float buff})
```
johnkslang: This is an abstract structure, not a physical memory layout.  It was correct that it did not have layout information.

Test/baseResults/spv.16bitstorage-int.frag.out
```diff
@@ -127,7 +127,7 @@ spv.16bitstorage-int.frag
                               Decorate 49(B5) Block
                               Decorate 51(b5) DescriptorSet 0
                               Decorate 51(b5) Binding 0
-                              MemberDecorate 88(S2) 0 ColMajor
+                              MemberDecorate 88(S2) 0 RowMajor
```
johnkslang: There were two S2 here, one for column major (this one) and one for row major (see line 141).

You've made them both be RowMajor.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7858,6 +7860,41 @@ void TParseContext::fixBlockUniformOffsets(TQualifier& qualifier, TTypeList& typ
...
+                subQualifier = &(typeList[member].type->getQualifier());
+            }
+            fixBlockUniformLayoutMatrix(*subQualifier, *(typeList[member].type->getStruct()));
+        } else if (typeList[member].type->isMatrix() && typeList[member].type->getQualifier().layoutMatrix == ElmNone) {
+            typeList[member].type->getQualifier().layoutMatrix = qualifier.layoutMatrix;
+        }
+    }
+}
+
+// Spread LayoutPacking to block member, if a  block member is a struct, we need spread LayoutPacking to
```
johnkslang: This is incorrect.

The same struct can be used in multiple places with contradictory layouts, implying two lower-level structs with different layouts to back the one high-level/abstract struct.  

GLSL and the AST are higher level, representing all this with one struct.  SPIR-V is lower level, expanding this one struct to two structs.

I'm afraid you're trying to apply a low-level transform to a high-level representation.

Test/baseResults/300layout.vert.out
Roy-AMD: The block member' memory layout been sync to global declare.

Source code:
```
layout(std140) uniform Transform { // layout of this block is std140
    mat4 M1; // row_major
    layout(column_major) mat4 M2; // column major
    mat3 N1; // row_major
    centroid float badf;  // ERROR
    in float badg;        // ERROR
    layout(std140) float bad1;
    layout(shared) float bad2;
    layout(packed) float bad3;
} tblock;
```

old:

'tblock' (layout( row_major std140) uniform block{layout( row_major std140 offset=0) uniform highp 4X4 matrix of float M1, layout( column_major std140 offset=64) uniform highp 4X4 matrix of float M2, layout( row_major std140 offset=128) uniform highp 3X3 matrix of float N1, layout( row_major std140 offset=176) centroid uniform highp float badf, layout( row_major std140 offset=180) uniform highp float badg, layout( row_major std140 offset=184) uniform highp float bad1, layout( row_major **shared** offset=188) uniform highp float bad2, layout( row_major **packed** offset=192) uniform highp float bad3})


new :

'tblock' (layout( row_major std140) uniform block{layout( row_major std140 offset=0) uniform highp 4X4 matrix of float M1, layout( column_major std140 offset=64) uniform highp 4X4 matrix of float M2, layout( row_major std140 offset=128) uniform highp 3X3 matrix of float N1, layout( row_major std140 offset=176) centroid uniform highp float badf, layout( row_major std140 offset=180) uniform highp float badg, layout( row_major std140 offset=184) uniform highp float bad1, layout( row_major **std140** offset=188) uniform highp float bad2, layout( row_major **std140** offset=192) uniform highp float bad3})


In the past, bad2 and bad3 been set a layout share and packed, but now they been force set std140,
I think it should be an compiler error, std140, std430, shared, packed can't been apply to block member layout.
We can :
Report a compiler error  or  keep them layout in packed and shared, I prefer first one.

glslang/MachineIndependent/ParseHelper.cpp
Roy-AMD: I updated the code. Now when fix layoutMatrix and layoutPacking will clone a new struct, and will declare a new struct in Spirv code too. But I'm worry it will affect decompiling. because this structs in spirv have the same name, we make work out a way to give them a mangle name; like:
BlockName@structName .

Test/baseResults/310.inheritMemory.frag.out
Roy-AMD: Fixed

Test/baseResults/300layout.vert.out
Roy-AMD: Now I add a condition for it :
If the member' layoutPacking != ElpNone, we won't fix it.

Test/baseResults/reflection.linked.options.out
```diff
@@ -1,10 +1,10 @@
 reflection.linked.vert
 reflection.linked.frag
 Uniform reflection:
-ubo_block.unused_uniform: offset 0, type 1406, size 1, index 0, binding -1, stages 0
+ubo_block.unused_uniform: offset 0, type 1406, size 1, index 0, binding -1, stages 17
 ubo_block.shared_uniform: offset 4, type 1406, size 1, index 0, binding -1, stages 17
-ubo_block.vsonly_uniform: offset 8, type 1406, size 1, index 0, binding -1, stages 1
-ubo_block.fsonly_uniform: offset 12, type 1406, size 1, index 0, binding -1, stages 16
+ubo_block.vsonly_uniform: offset 8, type 1406, size 1, index 0, binding -1, stages 17
```
johnkslang: These seem incorrect, to change something that was active only in one stage to being active in many stages.  Or, is this in a test variant that shows the alternate way (correct way is with different options in a different run)?

On the other hand, if it is correct due to wanting interface consistency, then the test should be fixed to still have a way of testing the per-stage bits and using names that make sense.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: But, it is not correct at GLSL or AST level to have duplicate structures.  They really are the same type at the high level, but different types at the SPIR-V level.

SPIR-V was already doing the duplication, and I think the SPIR-V code was already correct, yes?

Test/baseResults/310.inheritMemory.frag.out
```diff
@@ -30,81 +30,81 @@ Shader version: 310
...
 0:40            Constant:
 0:40              0 (const int)
 0:40          Constant:
 0:40            3 (const int)
 0:41      Function Call: non_ro_funf(f1; ( global void)
-0:41        direct index ( readonly temp mediump float)
-0:41          buff: direct index for structure ( readonly global 10-element array of mediump float)
-0:41            s_ro: direct index for structure (layout( column_major std430 offset=40) readonly buffer structure{ global 10-element array of mediump float buff})
-0:41              'ro_buffer' (layout( binding=2 column_major std430) readonly buffer block{layout( column_major std430 offset=0) readonly buffer 10-element array of mediump float buff_ro, layout( column_major std430 offset=40) readonly buffer structure{ global 10-element array of mediump float buff} s_ro})
+0:41        direct index (layout( std430) readonly temp mediump float)
```
johnkslang: Where does this `std430` come from?  Is it part of the type of `S` now, or is it computed only at this point in the AST, due to being an operation on a buffer than was `std430`?

That is, if `S` had been used in both an `std430` and an `std140` buffer, would we see both `std430` and `std140` in these situations, within the same shader, depending on the local context?

Test/baseResults/spv.1.4.OpCopyLogical.comp.out
```diff
@@ -20,63 +20,81 @@ spv.1.4.OpCopyLogical.comp
...
+                              Name 51  "MyStruct"
+                              MemberName 51(MyStruct) 0  "foo"
+                              MemberName 51(MyStruct) 1  "sb"
+                              Name 53  "Nested"
+                              MemberName 53(Nested) 0  "f"
+                              MemberName 53(Nested) 1  "S"
+                              Name 54  "UBON"
+                              MemberName 54(UBON) 0  "N1"
+                              Name 56  "uBufN"
+                              Name 62  "MyStruct"
```
johnkslang: There are additional copies of types.  Was this a bug before, that the SPIR-V did not have enough?  Or, is this as unneeded expansion?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7864,6 +7866,48 @@ void TParseContext::fixBlockUniformOffsets(TQualifier& qualifier, TTypeList& typ
...
+            // block member can be declare a matrix style, so it should be update to the member's style
+            if (typeList[member].type->getQualifier().layoutMatrix == ElmNone) {
+                subQualifier = &qualifier;
+            } else {
+                subQualifier = &(typeList[member].type->getQualifier());
+            }
+            // Deep copy the type in pool.
+            // Because, struct use in different block may have different layout qualifier.
+            // We have to new a object to distinguish between them.
+            typeList[member].type = typeList[member].type->clone();
```
johnkslang: Do these type copies correctly get matched as the same type during semantic checks for same-type structures?

johnkslang:
There has been little discussion or answers about the biggest thing happening here.

- Currently, AST, GLSL, and HLSL share one high-level structure for multiply physical layouts.
- Currently, SPIR-V uses multiple structures for a single high-level structure, one for each used physical layout.

It's a pretty big change to instead have the AST in the SPIR-V bucket instead of the GLSL and HLSL bucket.  If this major architectural change is proposed, we should:

- be clear what the new architecture and high-level design is
- think deeply about what might break and add new tests for it
- get consensus this is how we want to change glslang operation

Maybe this is okay to do, but it should not just be a side effect of seeing a bug in code reading. It should be a proactively broadcast design change.

Test/baseResults/spv.1.4.OpCopyLogical.comp.out
Roy-AMD: Yes, SPIR-V is not have enough, the MyStruct in these buffers have different memory layout, so the struct member should been decorate a different offset.

In glslang, it's a big different to other program language, struct used in two block, it means we need base on the original type extend two struct, to bear different other own attribute. it some like the __attribute__((__aligned__((#)))) in cpp. Perhaps we can learn from this feature's implementation in C ++ in the future.


> There has been little discussion or answers about the biggest thing happening here.
> 
> * Currently, AST, GLSL, and HLSL share one high-level structure for multiply physical layouts.
> * Currently, SPIR-V uses multiple structures for a single high-level structure, one for each used physical layout.
> 
> It's a pretty big change to instead have the AST in the SPIR-V bucket instead of the GLSL and HLSL bucket. If this major architectural change is proposed, we should:
> 
> * be clear what the new architecture and high-level design is
> * think deeply about what might break and add new tests for it
> * get consensus this is how we want to change glslang operation
> 
> Maybe this is okay to do, but it should not just be a side effect of seeing a bug in code reading. It should be a proactively broadcast design change.

I agree that, and I think the most important is that we must be clear what the new architecture and high-level design is.

1. Currently, AST, GLSL, and HLSL share one high-level structure for multiply physical layouts.  ----- I agree.

2. Currently, SPIR-V uses multiple structures for a single high-level structure, one for each used physical layout.
---- I agree.

3. instead have the AST in the SPIR-V bucket instead of the GLSL and HLSL bucket.
---- I'm not agree.

Ideally, AST should be closer to the semantics of high-level languages, rather than to accommodate the structure of low-level languages. but, why do we encounter such problems?
I think it's a Typelist design issue. We use type list to contian the struct members details, but layoutPacking and layoutMartrix is not the struct member detail, it's a block attribute, even this attribute are declare on block member, but it still a block attribute. so I think a better representation is move qualifie form TType to TTypeLoc, just like loc:

rename :
TTypeLoc -> TTypeDetail

and :
typedef TVector<TTypeDetail> TTypeList;

```
struct TTypeDetail {
    TType* type;
    TQualifier qualifier
    TSourceLoc loc;
};
```

To make a separation of type and type qualifier and use TTypeDetail to instead the TType in symbol node. And it can help us solve problems from a theoretical level rather than an engineering level.

So I think the root cause is not share one high-level structure for multiply physical layouts, the root cause is we put two different things together in AST.

But it a huge change. This will be a huge contraction and bring about possible refactoring. We need to adjust in many places. It can be a long-term plan, but currently we need to solve this problem engineeringly, and cloning a new structure is the most convenient effective method. The biggest uncertainty brought by this solution is that the change of debugging information leads to more inaccurate decompiled results.

We are frequently developing and integrating on glslang. When new problems are exposed, we will follow up and fix them in a timely manner. Because this change caused a big change in the generated spirv, I had some negligence when I checked, and I am sorry for this. But the big changes also prove that our test cases cover a very wide range, and I'm glad to have such a comprehensive test and a careful reviewer like you. So don't worry too much, git gives us a truth: brave to modify, roll back if bad.

glslang/MachineIndependent/ParseHelper.cpp
Roy-AMD: I think it can get matched correctly during semantic checks. 

1. This two fix only work on block struct member.  we cloned a new structure and modify the matrixlayout and layoutpacking, this all happened after we index the struct details from symbol table.

2. Before we build block type, we only checked layoutType in layoutTypeCheck, and resize mesh view dim in checkAndResizeMeshViewDim. we don't make any same type check in those two function.

3. During semantic checks, in  https://github.com/KhronosGroup/glslang/pull/2046 
I add glslangTypeToIdMap to map struct type and symbol Id, I'm try to avoid use structure reference to do any check.

Roy-AMD:
> There has been little discussion or answers about the biggest thing happening here.
> 
> * Currently, AST, GLSL, and HLSL share one high-level structure for multiply physical layouts.
> * Currently, SPIR-V uses multiple structures for a single high-level structure, one for each used physical layout.
> 
> It's a pretty big change to instead have the AST in the SPIR-V bucket instead of the GLSL and HLSL bucket. If this major architectural change is proposed, we should:
> 
> * be clear what the new architecture and high-level design is
> * think deeply about what might break and add new tests for it
> * get consensus this is how we want to change glslang operation
> 
> Maybe this is okay to do, but it should not just be a side effect of seeing a bug in code reading. It should be a proactively broadcast design change.

I agree that, and I think the most important is that we must be clear what the new architecture and high-level design is.

1. Currently, AST, GLSL, and HLSL share one high-level structure for multiply physical layouts.  ----- I agree.

2. Currently, SPIR-V uses multiple structures for a single high-level structure, one for each used physical layout.
---- I agree.

3. instead have the AST in the SPIR-V bucket instead of the GLSL and HLSL bucket.
---- I'm not agree.

Ideally, AST should be closer to the semantics of high-level languages, rather than to accommodate the structure of low-level languages. but, why do we encounter such problems?
I think it's a Typelist design issue. We use type list to contian the struct members details, but layoutPacking and layoutMartrix is not the struct member detail, it's a block attribute, even this attribute are declare on block member, but it still a block attribute. so I think a better representation is move qualifie form TType to TTypeLoc, just like loc:

rename :
TTypeLoc -> TTypeDetail

and :
typedef TVector<TTypeDetail> TTypeList;

```
struct TTypeDetail {
    TType* type;
    TQualifier qualifier
    TSourceLoc loc;
};
```

To make a separation of type and type qualifier and use TTypeDetail to instead the TType in symbol node. And it can help us solve problems from a theoretical level rather than an engineering level.

So I think the root cause is not share one high-level structure for multiply physical layouts, the root cause is we put two different things together in AST.

But it a huge change. This will be a huge contraction and bring about possible refactoring. We need to adjust in many places. It can be a long-term plan, but currently we need to solve this problem engineeringly, and cloning a new structure is the most convenient effective method. The biggest uncertainty brought by this solution is that the change of debugging information leads to more inaccurate decompiled results.

We are frequently developing and integrating on glslang. When new problems are exposed, we will follow up and fix them in a timely manner. Because this change caused a big change in the generated spirv, I had some negligence when I checked, and I am sorry for this. But the big changes also prove that our test cases cover a very wide range, and I'm glad to have such a comprehensive test and a careful reviewer like you. So don't worry too much, git gives us a truth: brave to modify, roll back if bad.

Test/baseResults/spv.1.4.OpCopyLogical.comp.out
johnkslang: Glslang did already expand these out.  The source for this test has 3 needs for MyStruct:
- in a std430 SSBO
- in a std140 UBO
- as function local variable

The original test result has 3 instances of this type; at ids 12, 16, and 32.

The new test results have 6 instances of the type.  See ids 12, 16, 26, 35, 51, and 62.

This appears to take glslang from doing the optimal thing with this type to doubling the number of types.  Can you point out the bug in the existing test result?

Roy-AMD:
Hi John,

We have other development and refinement plans, so I hope this pr can be merged as soon as possible, if nothing else.

Thanks
Ruoyu

Roy-AMD:
Hi @johnkslang ,

Best wishes for you. beside that, could this change been merge to master now?
If the code doesn't have other issue, I hope it can be merged as soon as possible, which will make our next development plan, and code management more convenient and fluid.

Thanks
Ruoyu

Roy-AMD:
Hi John,
@johnkslang  we have other Pr depend on this change, can we just merge it? if it has any issue, we can fixed it immediately.

Thanks
Ruoyu

johnkslang:
The currently open "Conversation" issue should be addressed somehow.  I view this as an example issue that it already has that needs a fix.  Repeating here, but hopefully we can discuss in the issue: 

> Glslang did already expand these out.  The source for this test has 3 needs for MyStruct:
> - in a std430 SSBO
> - in a std140 UBO
> - as function local variable

> The original test result has 3 instances of this type; at ids 12, 16, and 32.

> The new test results have 6 instances of the type.  See ids 12, 16, 26, 35, 51, and 62.

> This appears to take glslang from doing the optimal thing with this type to doubling the number of types.  Can you point out the bug in the existing test result?

Test/baseResults/spv.1.4.OpCopyLogical.comp.out
Roy-AMD: Sorry, I ignored your additional comments.

Roy-AMD:
> The currently open "Conversation" issue should be addressed somehow.  I view this as an example issue that it already has that needs a fix.  Repeating here, but hopefully we can discuss in the issue:
> 
> Glslang did already expand these out.  The source for this test has 3 needs for MyStruct:
> 
> in a std430 SSBO
> in a std140 UBO
> as function local variable
> 
> 
> 
> The original test result has 3 instances of this type; at ids 12, 16, and 32.
> 
> 
> The new test results have 6 instances of the type.  See ids 12, 16, 26, 35, 51, and 62.
> 
> 
> This appears to take glslang from doing the optimal thing with this type to doubling the number of types.  Can you point out the bug in the existing test result?

This issue have been fixed in those commit, and the solution is create a map to record the structure copy, at first time, we create a structure copy as a temp. And fix it's layoutPacking or layoutMatrix on the temp structure, after that, we will check this 2 qualifier whether been changed, if it been changed, we will use the copied structure replace the original structure.

Based on this design, in order to ensure that our records can be properly recorded and recursive, layoutMatrix will also be recorded in the TType of the structure, which is why some case's AST dump has changed, in which more layout declarations came out. But I think it's expected and it's correct, It can also make sure TQualifier::isMatrix can return a correct result.

Thanks
Ruoyu

Test/baseResults/310.inheritMemory.frag.out
Roy-AMD: Now, we will create a struct copy to save it, and it will be reflected in spirv where struct is defined multiple times.

Roy-AMD:
What is the progress of the review, do you have any other suggestions? I hope this Pr can be merged into the master as soon as possible. It's a huge change, and our work been blocked on it.
If there are any other problems after merging the code, we will fix them as soon as possible.
 @johnkslang

Roy-AMD:
Hi John,
@johnkslang . We are still waiting for this change to merge to ensure that our other development based on this change can be completed as soon as possible. Now that this pr cannot be merged has affected our project. I hope you can speed up the review. If there are other problems, I will fix them as soon as possible. If there are still problems after the code is merged, you can roll back or notify me to fix them as soon as possible. Putting this Pr on hold is not a wise decision. In fact, it fixes a glslang hidden problem.

Best Regards,
Ruoyu

johnkslang:
Yes, I am actually reviewing now.  Hope to finish today.

johnkslang:
Sorry this is taking so long, part of that is my fault.  For other parts, it will be faster to handle PRs when:

- they are smaller; I think this one got better
- if a reviewer raises an issue, let the reviewer close the conversation after reading and agreeing with the resolution (what happened here a few times is you answered and then closed the conversation, so I did not see the answer, or sometimes also did not agree to it, but it was somewhat hidden)

Test/baseResults/reflection.linked.out
```diff
@@ -3,7 +3,8 @@ reflection.linked.frag
 Uniform reflection:
 ubo_block.shared_uniform: offset 4, type 1406, size 1, index 0, binding -1, stages 17
 ubo_block.vsonly_uniform: offset 8, type 1406, size 1, index 0, binding -1, stages 17
-ubo_block.fsonly_uniform: offset 12, type 1406, size 1, index 0, binding -1, stages 16
+ubo_block.unused_uniform: offset 0, type 1406, size 1, index 0, binding -1, stages 17
```
johnkslang, line 6: The point of the test was to ensure that `unused_uniform` was not reflected, and that `fsonly_uniform` was only reflected for the fragment stage.

Is there a reason to be breaking this?

I'm pretty sure some users depend on reflection not reporting unused items.  There is a separate options to turn on to get more information.

glslang/MachineIndependent/ParseHelper.cpp
johnkslang: Was there already an existing problem with GLSL/AST level starting to push information into nested structures, but incorrectly? It seems there are two paths here:

A. fix it from doing that, at the high level, there should be only one
B. do it more, completely, requiring structure copying, which is a pretty major change

So, I guess three questions here:

1. is it true this issue was present?  (I can believe it from the memory qualifier test)
2. is it the case you are choosing path B instead of fixing the code to match the design with path A?
3. What are all the implications of changing GLSL/AST from single copy to multiple copy of the same struct?

It is possible you are fixing real problems here, but I worry it is by changing design and interface, and wonder about path A as well.

Test/baseResults/reflection.linked.out
Roy-AMD: > The point of the test was to ensure that `unused_uniform` was not reflected, and that `fsonly_uniform` was only reflected for the fragment stage.
> 
> Is there a reason to be breaking this?
> 
> I'm pretty sure some users depend on reflection not reporting unused items. There are separate options to turn on to get more information.

>According to the spec OpenGL spec4.60.7:
>>7.6 Uniform Variables:
>>>All members of a named uniform block declared with a shared or std140 layout qualifier are considered active, even if they are not referenced in any shader in the program. The uniform block itself is also considered active, even if no member of the block is referenced.

```glsl
layout(binding = 0, std140) uniform ubo_block {
	float unused_uniform;
	float shared_uniform;
	float vsonly_uniform;
	float fsonly_uniform;
} ubo;
```

If app user don't want compiler reporting unused uniform block, they should use std430 layout.
If gpu driver user don't want compiler reporting unused uniform block, they can implement a  
reflection out of glslang, and travers again. In fact, when we encounter similar problems, we do just that. Glslang should correctly based on the spec, and venders needs to find a way to meet its own needs.

glslang/MachineIndependent/ParseHelper.cpp
Roy-AMD: > Was there already an existing problem with GLSL/AST level starting to push information into nested structures, but incorrectly? It seems there are two paths here:
> 
> A. fix it from doing that, at the high level, there should be only one
> B. do it more, completely, requiring structure copying, which is a pretty major change
> 
> So, I guess three questions here:
> 
> 1. is it true this issue was present?  (I can believe it from the memory qualifier test)
> 2. is it the case you are choosing path B instead of fixing the code to match the design with path A?
> 3. What are all the implications of changing GLSL/AST from single copy to multiple copy of the same struct?
> 
> It is possible you are fixing real problems here, but I worry it is by changing design and interface, and wonder about path A as well.

Yes, there is a existing problem with GLSL/AST level.
1. is it true this issue was present?
    * This issue was present.
2. is it the case you are choosing path B instead of fixing the code to match the design with path A?
    * I chooise the path B to fix it.
3. What are all the implications of changing GLSL/AST from single copy to multiple copy of the same struct?
    * This will result in multiple copies of the structure in memory in addition to the original structure, but these copies are unique if they have the same memory layout or the same matrix layout. That's vary important, it can help we not generate duplicate struct in spirv code.  So, it won't implicate spirv generate. One practical effect he may bring is that when we want to modify only the origin structure to affect all struct used in block, we may not get the results we want. If we want to affect all struct used in block, we need to process all the structural branches in the record.

In fact, I chose path B to reduce the impact on the design and interface. If we want fix it in high level, we have to modify the TTypeLoc (path A), and save layout information, matrix information in it, we have discussed internally, I think this is more fit, but the risk will be higher. path B won't impact interface, we only make a struct copy in AST, When we choose pathB, this is the only way, If the behavior of some interfaces is affected, I think this is also easy to fix, just apply the behavior of the interface to all records.

Roy-AMD:
> Sorry this is taking so long, part of that is my fault. For other parts, it will be faster to handle PRs when:
> 
> * they are smaller; I think this one got better
> * if a reviewer raises an issue, let the reviewer close the conversation after reading and agreeing with the resolution (what happened here a few times is you answered and then closed the conversation, so I did not see the answer, or sometimes also did not agree to it, but it was somewhat hidden)

The code of this change itself is not very large, but it affects a lot of cases, I did not do enough checks, so a lot of errors occurred at the beginning (Self-confidence is a good thing , but I should not check it when it affects so many cases :)).
In AMD, We use review board, when I fixed a issue, I will click button [Fixed] to mark it, I thought resolve button same as it, but it's OK, I will keep the state in unsolved in future.
You may be able to invite more contributors to help you view the code, such as Rexu.

I can't blame you. Actually I am very worried about your life and hope you pay attention to safety and stay healthy during COV-19, no one would like to lose a great engineer / scientist. If you need any help from china, you may contect me.
Best wish to you.

Test/baseResults/reflection.linked.out
johnkslang: Reflection isn't a normative specification thing, it as a tooling thing added on by users that needed that specific tooling.

Khronos does not have a specification for this tooling feature.

Maybe you need something similar, but need to add what it is you need, not re-specify and break something that people are already using.

johnkslang:
First, despite my strict review, I want to thank you for doing much hard work here and figuring all this out.  I am sure you have found important improvements, and thank you for that.

I think there are two things going on here:

1.  there are some bugs with nested structures that you found fixes for
1. you want to change definition of how reflection works

The first one is great, while the second one is worrisome.

A few paths forward:

A) separate this into multiple PRs that are clearly fixes versus behavior changes independent of the fixes,
B) or, champion getting approval of downstream users for the semantic changes you want to make,
C) or, put your semantic changes behind a new option, such that default operation remains the same.

Test/baseResults/reflection.linked.out
Roy-AMD: > Reflection isn't a normative specification thing, it as a tooling thing added on by users that needed that specific tooling.
> 
> Khronos does not have a specification for this tooling feature.
> 
> Maybe you need something similar, but need to add what it is you need, not re-specify and break something that people are already using.

You can't use "Reflection isn't a normative specification thing" to deny that reflection needs to abide by glsl norms. The compiler is just a tooling thing too, but since this compiler is a glsl compiler, it should be subject to the glsl norms, the same Since reflection is glsl's reflection, it should also be subject to glsl's specifications, otherwise it is not glsl's reflection, or you should rename it to someotherReflection. So, I think glsl should honor to the spec, not to mention that I have not seen negative comments or questions from users under this Pr. Bringing people more standard tools is not a break.
The changes we have made are based on the test results from CTS. There are sufficient and legitimate reasons for glslang to respect and restore the provisions of the spec from all angles.

Roy-AMD:
> First, despite my strict review, I want to thank you for doing much hard work here and figuring all this out. I am sure you have found important improvements, and thank you for that.
> 
> I think there are two things going on here:
> 
> 1. there are some bugs with nested structures that you found fixes for
> 2. you want to change definition of how reflection works
> 
> The first one is great, while the second one is worrisome.
> 
> A few paths forward:
> 
> A) separate this into multiple PRs that are clearly fixes versus behavior changes independent of the fixes,
> B) or, champion getting approval of downstream users for the semantic changes you want to make,
> C) or, put your semantic changes behind a new option, such that default operation remains the same.

About A:
The code review cycle is too long, I hope that these codes will be merged as soon as possible, so that we can mention another Pr to achieve C.

About B:
I think CTS is the best interpretation of B.

About C:
Actually, We have added options for them, which have been implemented in our local changes, but because this change has not been merged into, we have not submitted the relevant Pr for these codes, this is some very huge changes, when this change After being merged , we will immediately organize the local code and submit it. These codes are scattered in multiple changes, in order to facilitate our local better management of them.

johnkslang:
It is partly my fault, but also the code review cycle is long partly because of the size of the original changes, and because each review finds non-backward-compatible changes.

Smaller PRs without backward incompatible changes can progress much faster.

I'm not understanding why we can't incrementally make the fixes without breaking existing users.

I did not know CTS was measuring this.  CTS tests drivers, not glslang reflection, yes?

Glslang reflection was a side-band tool, for a particular GLSL purpose, not a legally normative implementation of any specification.

Can you help us understand better?

Test/baseResults/hlsl.reflection.vert.out
```diff
@@ -1,5 +1,9 @@
 hlsl.reflection.vert
 Uniform reflection:
+anonDeadMember2: offset 64, type 8b52, size 1, index 0, binding -1, stages 1
+ufDead4: offset 28, type 1406, size 1, index 1, binding -1, stages 1
```
Roy-AMD, line 4: Those case are under the influence of EShMsgKeepUncalled option, this option did not work in old implementation, but now it can work normally. Therefore, some changes have taken place when traversing the grammar tree, which will result in the change of some variables and index used in EShMsgKeepUncalled.

Roy-AMD:
> It is partly my fault, but also the code review cycle is long partly because of the size of the original changes, and because each review finds non-backward-compatible changes.
> 
> Smaller PRs without backward incompatible changes can progress much faster.
> 
> I'm not understanding why we can't incrementally make the fixes without breaking existing users.
> 
> I did not know CTS was measuring this. CTS tests drivers, not glslang reflection, yes?
> 
> Glslang reflection was a side-band tool, for a particular GLSL purpose, not a legally normative implementation of any specification.
> 
> Can you help us understand better?

I add a option for disable this reflection, please review again, if no problem, after test passed, you can merge it now.

Test/baseResults/hlsl.reflection.vert.out
Roy-AMD: @johnkslang ,I rebased all commit for you easily review code.

PR #2044, opened 2020-01-03
Fix binding for arrays.

ShchchowAMD:


ShchchowAMD:
Close as optional ssbo layout needs further modification, would use another CL.

PR #2046, opened 2020-01-03
Fix glslang can't link multiple AST in a single stage

Roy-AMD:
Root cause:
GlslangToSpv use symbol structure's ptr as a map key, but multiple shader object can build a new AST.
In the AST the the same symbol has different ptr point to their structure, so indext map faild.

solution:
Add a new map glslangTypeToIdMap to map ptr to symbol id, and use symbol id to index memberRemapper.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -245,7 +245,9 @@ protected:
     std::unordered_map<std::string, spv::Function*> functionMap;
     std::unordered_map<const glslang::TTypeList*, spv::Id> structMap[glslang::ElpCount][glslang::ElmCount];
     // for mapping glslang block indices to spv indices (e.g., due to hidden members):
-    std::unordered_map<const glslang::TTypeList*, std::vector<int> > memberRemapper;
+    std::unordered_map<int, std::vector<int>> memberRemapper;
+    // for mapping glslang symbol struct to symbol Id
+    std::unordered_map<const glslang::TTypeList*, int> glslangTypeToIdMap;
     std::stack<bool> breakForLoop;  // false means break for switch
     std::unordered_map<std::string, const glslang::TIntermSymbol*> counterOriginator;
     // Map pointee types for EbtReference to their forward pointers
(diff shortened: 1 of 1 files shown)


johnkslang:
So, the AST merging in linkValidate.cpp got the Id's correct, but with aliased pointers?

Is it correct WRT to duplicate-appearing declarations (same members, same order, ...) that are still not the same?  And, with one compilation unit, is it 1:1 with pointers and types, but with multiple compilation units it is many:1?

SPIRV/GlslangToSpv.cpp
```diff
@@ -1753,6 +1758,12 @@ void TGlslangToSpvTraverser::visitSymbol(glslang::TIntermSymbol* symbol)
 bool TGlslangToSpvTraverser::visitBinary(glslang::TVisit /* visit */, glslang::TIntermBinary* node)
 {
     builder.setLine(node->getLoc().line, node->getLoc().getFilename());
+    if (node->getLeft()->getAsSymbolNode() != nullptr && node->getLeft()->getType().isStruct()) {
```
johnkslang: We need both here and in `visitSymbol`?  (Something seems odd about this pair being the right subset of places to check.)

Roy-AMD: Yes, it's a pre-Visit travers, if a symbol use as a[0][1], 'a' still not in map, glslangTypeToIdMap will create a new key-value  - > {id_a: 0} and return '0', use '0' to index memberRemapper will get a unsized array. and use 'i' index this array can cause a crash.

Roy-AMD: Do you have a better solution? I think this code weird too. but I don't have a better way to solve it.

Roy-AMD:
> So, the AST merging in linkValidate.cpp got the Id's correct, but with aliased pointers?
> Is it correct WRT to duplicate-appearing declarations (same members, same order, ...) that are still not the same?  And, with one compilation unit, is it 1:1 with pointers and types, but with multiple compilation units it is many:1?

Basically yes.
So, it's not  reliable to use aliased pointers index map.

PR #2048, opened 2020-01-06
Add missing extension defination

Roy-AMD:
This micro should been defined 1, if opengl 2.0 is required.

See extension spec https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_blend_equation_advanced.txt


johnkslang:
There is a line doing that at line 341, how does this compare to that?

Roy-AMD:
> There is a line doing that at line 341, how does this compare to that?

Sorry, I forgot to remove this line when I port local change to git. this micro should be define in OpenGL and OES both. Now, I fixed it.

Issue #2053, opened 2020-01-07
Buffer references can generate invalid forward references


alan-baker:
The test spv_bufferhandle6_frag generates a forward reference to an array. This is now caught by the validator.

johnkslang:
I only see forward references to pointer to structures, but they then get consumed by an array, which is required for a struct containing an array of pointers to its own type.

Either the spec needs to be relaxed to allow that, or that functionality needs to be disallowed, but I don't think it's right to have the GLSL support a struct with an array of pointers to the structs type (seems perfectly valid) and say the natural SPIR-V for that is illegal.

johnkslang:
Going back to basics, what's the spec language preventing what glslang is producing?

alan-baker:
The heart of it is that the arrays do not allow a forward reference to the element type. So when glslang uses the forward pointer in the array declaration, that is invalid as the spec stands.

johnkslang:
Yes, I think the spec. meant to allow a struct pointing to itself and says so very directly. We should discuss within Khronos if it's meant to disallow a struct pointing to itself through an array (allowing a tree rather than just a linear list).

johnkslang:
Khronos has agreed that other type declarations instructions can consume an OpTypeForwardPointer along the way toward its' OpTypePointer declaration. So, glslang is okay.  This will appear in the next revision of the SPIR-V specification.

PR #2054, opened 2020-01-07
Update spirv tools

alan-baker:
Updates the known good SPIRV-Tools commit and known test failures (see #2053).


johnkslang:
@jeffbolznv, how is recursiveness supposed to be handled?  The GLSL is:

```
layout(buffer_reference) buffer T1;
layout(set = 3, binding = 1, buffer_reference) buffer T1 {
   layout(offset = 0) int a[2]; // stride = 4 for std430, 16 for std140
   layout(offset = 32) int b;
   layout(offset = 48) T1  c[2]; // stride = 8 for std430, 16 for std140
   layout(offset = 80) T1  d;
} x;
```

And the SPIR-V makes a forward reference to an array (SPIR-V only allows to a struct), as well as making two copies of the struct:

```
               6:             TypeInt 32 1
               9:      6(int) Constant 0
              10:             TypeInt 32 0
              11:     10(int) Constant 2
              12:             TypeArray 6(int) 11
                              TypeForwardPointer 13 PhysicalStorageBufferEXT
              14:             TypeArray 13 11
          15(T1):             TypeStruct 12 6(int) 14 13
              16:             TypeArray 6(int) 11
              17:             TypeArray 13 11
          18(T1):             TypeStruct 16 6(int) 17 13
              13:             TypePointer PhysicalStorageBufferEXT 18(T1)
              19:             TypeArray 13(ptr) 11
              20:             TypePointer StorageBuffer 15(T1)
```

johnkslang:
Well, I no longer see a forward reference to an array, but that's what the change says it addresses, so I'm not sure now.

johnkslang:
I think the issue is that the specification was for a struct pointing to its own type, but that needs to be generalized to allowing a struct containing an array of pointers to the struct's type, so maybe this should be a specification change, not a validator change.

jeffbolznv:
This seems related to https://gitlab.khronos.org/spirv/SPIR-V/issues/394 which is still open. At first glance I don't see what's wrong with the SPIR-V quoted above (the duplicate structure is unexpected, but not invalid). Is there more info somewhere on why it's considered invalid?

johnkslang:
SPIR-V says 

>  Subsequent OpTypeStruct instructions can use Pointer Type as an operand.

Maybe that gets interpreted as "only a structure can consume the forward-declared pointer", but it doesn't really restrict to that because it says "can" not "must/only".  But, in the GLSL, the structure contains an array of the structure, making the forward-declared pointer consumed by an OpTypeArray, which I now suspect the spec should allow.

jeffbolznv:
That's part of what's covered in https://gitlab.khronos.org/spirv/SPIR-V/issues/394. I think this rule probably ought to be disabled until that spec issue is resolved.

alan-baker:
https://github.com/KhronosGroup/SPIRV-Tools/pull/3131

johnkslang:
Khronos has decided to allow extra levels of indirection between the forward-declared pointer and the consumption in the struct, which allows glslang's current behavior as being valid.

PR #2061, opened 2020-01-24
For glslang.js, lower the target to SPV1.0 on Vk1.0 for WebGPU

kainino0x:
Higher values don't make sense since WebGPU will probably target Vulkan 1.0 devices.


johnkslang:
Unless, WebGPU stays within a functionality envelop that allows easy transform from higher version to a lower version.

Even then, you could choose to target just 1.0, per the PR, so no strong opinion either way.  I'll wait for a response before merging.

kainino0x:
WebGPU certainly will stay in that envelope (plus any optional extensions). Based on David's response it seems like we shouldn't make this change. Depending on whether wgpu's tooling can change to support 1.3(/1.5?), maybe I'll open another PR to make this configurable.

johnkslang:
Okay. Marking this as WIP until you decide..

kainino0x:
I decided it should just be configurable.

PR #2064, opened 2020-01-27
Use correct enum type in case statement

zoddicus:
Two similarly named enums led to the wrong one being used here, this
is getting caught in the Chromium and Dawn DEPS rolls.


zoddicus:
ugh, sorry signed this with my chromium account, not my github, I will recreate

PR #2067, opened 2020-01-28
Use newer gtest API, replacing deprecated one

dneto0:
Use INSTANTIATE_TEST_SUITE_P instead of
deprecated INSTANTIATE_TEST_CASE_P.

This transition has been going on for about a year,
and this is a good time to change.

### DIFF
--- a/gtests/AST.FromFile.cpp
+++ b/gtests/AST.FromFile.cpp
@@ -59,7 +59,7 @@ TEST_P(CompileToAstTestNV, FromFile)
 }
 
 // clang-format off
-INSTANTIATE_TEST_CASE_P(
+INSTANTIATE_TEST_SUITE_P(
     Glsl, CompileToAstTest,
     ::testing::ValuesIn(std::vector<std::string>({
         "sample.frag",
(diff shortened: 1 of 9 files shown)


dneto0:
The deprecation is noted here: https://github.com/google/googletest/blob/master/googletest/include/gtest/gtest-param-test.h#L503

Git blame shows the new API has been available for 12 months. https://github.com/google/googletest/blame/master/googletest/include/gtest/gtest-param-test.h#L506

johnkslang:
Agreed this fix is overdue.  The bot failure looks genuine though.

johnkslang:
Maybe things are getting done out of order.

We still advise to use the Sept. 2018 gtests due to compatibility issues.

I think the steps needed are:

1. get everyone off 2013 -> 2015 (includes one of my systems, not all, as the test of still supporting 2013)
1. decide what new known-good of the gtests to use
1. update advice scripts with result of step 2
1. make the changes in this PR

johnkslang:
> The deprecation is noted here:...

But that is added in a later release that the one we've locked onto for glslang.  (In the README and in appveyor.yml.)

If we can now drop 2013 finally, I'm agree adapting to the later versions.

This is just an explanation for why we don't see the deprecated assert.

johnkslang:
I checked the past issues about this.  The last I see, following discussions during 2018 is confirmation (in Feb. and in Nov.) that visual studio 2013 is still supported.

There was some talk of 5 years availability of tools before deprecation.  For visual studio 2015, that will be July this year.  Should we propose (in a separate issue) dropping 2013 support in July 2020?

dneto0:
Ah.  That explains my sense of deja vu.  Sorry to make you go thorugh that analysis again.

dneto0:
I believe this is not urgent from end.

johnkslang:
Closing as superseded by #2365, which is also a bit more comprehensive.

PR #2078, opened 2020-02-06
GLSL: Fix semantic error for GL_KHR_memory_scope_semantics

amdrexu:
When 2DMS or 2DMSArray is used, IMAGE_PARAMS will take an additional
paramter to represent sample number. The semantics check should get the
correct semantics values, taking this case into account.


amdrexu:
A shader of mine fails in compilation. Mistaken errors are reported. 

```
#version 450 core

#extension GL_KHR_memory_scope_semantics: enable
 
layout(binding = 0, r32i)  uniform iimage2DMS i2DMS;

layout(binding = 1) buffer Buf
{
    int i32;
    ivec4 i32v4;
};

void main()
{
    i32 = imageAtomicAdd(i2DMS, i32v4.xy, i32v4.z, i32, gl_ScopeDevice, gl_StorageSemanticsImage, gl_SemanticsRelaxed);
}
```

PR #2081, opened 2020-02-07
CMake: Split "is emscripten" config from "enable glslang.js"

### LINKED ISSUE #2072: Full-featured build on Web
Hello everyone!

I'm working on C++ framework that builds on all major platforms (Desktop, Mobile, Web) and I want to use glslang as 3rd-party library there.

Therefore, I need to build `glslang` on Web as if it is Desktop platform, i.e. I want to have 100% same behavior and functionality.
I don't need any grammar featurecut or minimalistic build or whatever.

Can you please help me properly configure `glslang` with this goal in mind?

I'm not sure that I have to follow instructions from "Building to WASM for the Web and Node". It seems that these instructions are not for my specific case.

This folder is also quite suspicious:
https://github.com/KhronosGroup/glslang/tree/master/glslang/OSDependent/Web
I'm not sure that I need this "zeroCopy" thing in my build but I cannot disable it.
I also don't understand how TLS thing is handled in Web build since there's no implementation for `OS_AllocTLSIndex` and so on in `OSDependent/Web`.

Thanks,
Eugene.

kainino0x:
This should enable glslang to be used as a dependency of other projects
that target WebAssembly, instead of assuming "EMSCRIPTEN" => "Want
glslang.js".

Should fix #2072.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,12 +32,29 @@ option(ENABLE_SPVREMAPPER "Enables building of SPVRemapper" ON)
 
 option(ENABLE_GLSLANG_BINARIES "Builds glslangValidator and spirv-remap" ON)
 
-option(ENABLE_GLSLANG_WEB "Reduces glslang to minimum needed for web use" OFF)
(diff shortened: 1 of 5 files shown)


eugeneko:
Thank you!

PR #2083, opened 2020-02-14
Add pkg-config files

DanySpin97:
This PR is based on Fedora patch: https://src.fedoraproject.org/rpms/glslang/blob/master/f/0001-pkg-config-compatibility.patch


CMakeLists.txt
```diff
@@ -168,18 +173,10 @@ if(BUILD_EXTERNAL AND IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/External)
 	add_subdirectory(External)
 endif()
 
-if(NOT TARGET SPIRV-Tools-opt)
-    set(ENABLE_OPT OFF)
-endif()
-
-if(ENABLE_OPT)
+# This is always true
+if(${SPIRV_TOOLS_FOUND} EQUAL 1)
```
johnkslang: This, or something like it, seems to have broken CMake, at least for AppVeyor.

DanySpin97: AppVeyor build uses Windows and does not ship pkg-config tool. I'm looking for a way to fix this. If Windows cannot be fixed, I'm going to add a check and use the provided SPIRV-Tools on it.

DanySpin97:
Moreover, when -DBUILD_SHARED_LIBS=OFF (which is the defaults), [this fix](https://github.com/KhronosGroup/SPIRV-Tools/pull/3189) for SPIRV-Tools is required.

CMakeLists.txt
DanySpin97: I'll open a separate PR to build using system SPIRV-Tools.

SPIRV/CMakeLists.txt
```diff
@@ -40,6 +40,9 @@ target_include_directories(SPIRV PUBLIC
 	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
 	$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
+set(SPIRV_NAME spirv)
+set(SPIRV_VERSION 1.5)
```
johnkslang: Why is this needed?  We seem to be increasing the number of random places that will need to be changed when SPIR-V is updated.

glslang/CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ set(HEADERS
     MachineIndependent/preprocessor/PpContext.h
     MachineIndependent/preprocessor/PpTokens.h)
 
+set(VERSION 8.13.3559)
```
johnkslang: This number is already out of date...

SPIRV/CMakeLists.txt
DanySpin97: It is needed to set a version in the `spirv.pc` file. What could be done is to pass SPIRV version from cmake to `spirv.hpp`.

glslang/CMakeLists.txt
DanySpin97: Same thing here.

johnkslang:
There is enough feedback here that a better alternate way is needed, so closing this one.  Thanks.

PR #2086, opened 2020-02-18
SPV_AMD_shader_image_load_store_lod is now validated by spirv-tools

samuelig:
SPIRV-Tools Pull Request [#3186](https://github.com/KhronosGroup/SPIRV-Tools/pull/3186) is blocked because spv.imageLoadStoreLod.frag.out is expecting to return a validation failed result.


samuelig:
And now the CI is failing here because https://github.com/KhronosGroup/SPIRV-Tools/pull/3186 is not merged.

As we have a circular dependency between glslang and spirv-tools for this case, I don't know which is the procedure: land first this pull request and then spirv-tools one, or the other way around.

johnkslang:
You can land here first, just need a consistent set of test results and known-good, etc.

It is okay if the passing consistent set says there is a validation failure, if that is the known intention.

samuelig:
https://github.com/KhronosGroup/SPIRV-Tools/pull/3186 is merged.

johnkslang:
Is this just waiting a test update?

Or, do you want to update update_glslang_sources.py with the new SPIRV-Tools commit and repush?

johnkslang:
I see now the only thing this PR did was change a  test result.  That never makes sense as it will break all the regression test runs.  Something has to change to cause a test result to change.  Those need to be done together.

Changes should be driven by an actual functional change, and then the tests updated with the new results.  So, this PR never made sense.  In this case the functional change is a new SPIRV-Tools to point to in update_glslang_sources.py.  Making that change will naturally result in the test-result change, and the pair together keep glslang consistent.

samuelig:
> Is this just waiting a test update?
> 
> Or, do you want to update update_glslang_sources.py with the new SPIRV-Tools commit and repush?

Yes, I know. Yesterday I was unable to find the time to update the branch after the merge and test it. It is now updated.

Thanks!

johnkslang:
No problem.  I was getting asked about it, but it all looks good now.  Thanks.

PR #2088, opened 2020-02-19
Fix xfb stride limit issue

Roy-AMD:
The key word xfb_stride are introduced by extension [ARB_enhanced_layouts.txt](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_enhanced_layouts.txt) start from glsl 4.4, and been introduced to core spec and glsl 4.6


johnkslang:
Thanks.  Can we also add tests to show it working and for future regression?

Roy-AMD:
> Thanks. Can we also add tests to show it working and for future regression?

Hi John,

I add some case for xfb limit negative test. I'm sorry, but I've been busy with other things lately.

Thanks
Ruoyu

Roy-AMD:
Hi @johnkslang ,

Hope you well.
can we merge this commit now?

Thanks
Ruoyu

Test/xfbUnsizedArray.error.vert
```diff
@@ -0,0 +1,16 @@
+#version 430 core
+#extension GL_ARB_enhanced_layouts : require
+
+layout (xfb_offset = 0) out vec4 xfb_buffer[]; // unsized array can been use on xfb buffer
```
johnkslang, line 4: It means to say this?

```
// ERROR: unsized array cannot be used for an xfb buffer?
```

Test/baseResults/xfbUnsizedArray.error.vert.err
```diff
@@ -0,0 +1,5 @@
+xfbUnsizedArray.error.vert
+ERROR: 0:4: 'xfb_offset' : unsized array at in buffer 0
```
johnkslang, line 2: I think will want this to say:

```
unsized array in buffer 0
```

I suggest a code change below.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5684,6 +5684,8 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
         int repeated = intermediate.addXfbBufferOffset(type);
         if (repeated >= 0)
             error(loc, "overlapping offsets at", "xfb_offset", "offset %d in buffer %d", repeated, qualifier.layoutXfbBuffer);
+        if (type.isUnsizedArray())
+            error(loc, "unsized array at", "xfb_offset", "in buffer %d", qualifier.layoutXfbBuffer);
```
johnkslang, line 5688: Delete `at`, just `unsized array`.

Test/xfbUnsizedArray.error.vert
Roy-AMD: Yes, it's a typo.
It seems that one of the main tasks in the future is improve my English : )

Roy-AMD:
Hi John,

@johnkslang This Pr are in review now?

Thanks
Rouyu

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5776,6 +5778,12 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
 
     // "The offset qualifier can only be used on block members of blocks..."
     if (qualifier.hasOffset()) {
+        if (isEsProfile())
+            profileRequires(loc, EEsProfile, 300, E_GL_ARB_enhanced_layouts, "\"offset\" can only use to atomic counter layout qualifiers");
+        else {
+            if (version <= 430)
+                profileRequires(loc, ~EEsProfile, 140, E_GL_ARB_enhanced_layouts, "\"offset\" can only use to atomic counter layout qualifiers");
```
johnkslang, line 5785: The error should not come out if the version is greater than the 140 in the 3rd parameter, so this test seems extra too.

These calls were set up to be declarative, rarely needing extra control flow around them.

Roy-AMD: How about change the code like blow:
```
    // "The offset qualifier can only be used on block members of blocks..."
    if (qualifier.hasOffset()) {
        if (version <= 430) {
            profileRequires(loc, ~EEsProfile, 140, E_GL_ARB_enhanced_layouts, "\"offset\" can only use to atomic counter layout qualifiers");
            profileRequires(loc, EEsProfile, 300, E_GL_ARB_enhanced_layouts, "\"offset\" can only use to atomic counter layout qualifiers");
        }
        if (type.getBasicType() == EbtBlock)
            error(loc, "only applies to block members, not blocks", "offset", "");
    }
```
profileRequires not provided a param to specified the maxVersion, but enhanced_layout extension are merge to core spec after version 430, so when version number is higher then 430, the extension are not force require.

Adding new parameters to profileRequires would be a huge task,
I think we can add a TODO for profileRequires: in the future we will add new parameters to it for specified a maxVersion, help us limite the version to a range. Which will help us keep the code style of the declarative, but for now keep this condition will help us to standard and closer to spec

johnkslang: > specified the maxVersion

Generally, features don't go away, so we shouldn't need that.

> Adding new parameters to profileRequires would be a huge task,

I agree we don't want to do that.

The version number to `profileRequires` means if the current version is at or above the specified version, the checks for the extension are not done.

Do you actually get different test results if you delete the if test?

johnkslang: Maybe it's a side subject, but are you talking about XFB, or about `offset`?  These are not the same features.

johnkslang: What are you applying `offset` to?  That might be what's important.

Roy-AMD: Sorry, It's a typo, I updatedd the comment now. we are talking about enhanced layout and offset, offset been introduce by enhance layout, and it can be use when 130 <= version number <= 430 and extension GL_ARB_enhanced_layouts : require, or only when version number >= 440. So `if (version <= 430)` is for some corner case.

Roy-AMD: @johnkslang Let's conitnue this code review. I think the `if (version <= 430` should been keep.

johnkslang: So, we have this

```
    profileRequires(loc, ~EEsProfile, 140, ...);
...
...profileRequires(const TSourceLoc& loc, int profileMask, int minVersion...
{
    if (profile & profileMask) {
        bool okay = minVersion > 0 && version >= minVersion;
        // minVersion is 140, so 'okay' will be true for any 'version' bigger than 140, no matter what number it is; 430 and 440 will be treated the same way, which is to do nothing
        ....
        // nothing sets okay to false
       if (! okay)
            error(...
```

Also, it doesn't make sense to do an ES check under a desktop version number check.

Roy-AMD: > So, we have this
> 
> ```
>     profileRequires(loc, ~EEsProfile, 140, ...);
> ...
> ...profileRequires(const TSourceLoc& loc, int profileMask, int minVersion...
> {
>     if (profile & profileMask) {
>         bool okay = minVersion > 0 && version >= minVersion;
>         // minVersion is 140,
>         // so 'okay' will be true for any 'version' bigger than 140,
>         // no matter what number it is; 430 and 440 will be treated the same way,
>         // which is to do nothing
>         ....
>         // 'okay' is never set to false
>         if (! okay)
>             error(...
> ```
> 
> Also, it doesn't make sense to do an ES check under a desktop version number check.

I don't think so. es max version is under 330, so it always run into `if (version <= 430) {` ,
if it's es profile, `profileRequires(loc, ~EEsProfile` will be ignore, if it's core profile `profileRequires(loc, EEsProfile` will be ignore, it will always make a  version check.
The only problem is that after the version number of es glsl increases, there will be a potential error, but this way to write statements that can reduce the if condition useage, maybe after certain days the profile needs to support the maximum version number limit This line of code will be deleted.
Based on your suggestion, maybe I should put `profileRequires (loc, EEsProfile` outside

Roy-AMD: > @johnkslang Let's conitnue this code review. I think the `if (version <= 430` should been keep.

If there are no other problem, I hope that this change can be merged as soon as possible, and I have other negative test case and fix based on this Pr.

johnkslang: Okay, I'd like to finish quickly too.

I'm sorry, however, I'm still concerned why `profileRequires` might not be working correctly.  The pattern is

```
    if (functionality is getting used)
        profileRequires(minimum version, extension)
```

Where, it should be "either or", not that both are needed.

Since the purpose of `profileRequires` is to avoid logic testing version numbers everywhere, basically to be declarative rather than procedural, I don't want to be setting the precedent that this is not true.

I took the time to track down and post the relevant code.  If I'm still mistaken, could you respond with what's not working correctly here?

johnkslang: I think I realized what's wrong.

The minimum version is the minimum core version that supports it without an extension, not the oldest version the extension document says the extension is allowed to operate in.

johnkslang: GL_ARB_enhanced_layouts says

> OpenGL 3.1 and GLSL 1.40 or later are required.

But, we don't really have a way of enforcing that per-semantic right now, basically, it's done by not adding the built-in functions into the symbol table unless the version is high enough.  That's not quite perfect, but mostly works.

The version passed to `profileRequires` is the first version that the feature exists in core.

So, you should not be passing 140 in, you should be passing 430 or 440 or whatever version pulled it into core.

johnkslang: Looks like the code you need is, without any if-testing:

```
    profileRequires(loc, ~EEsProfile, 440, E_GL_ARB_enhanced_layouts, "\"offset\" can only used for atomic counter layout qualifiers");
```

I also fixed the English grammar.

Roy-AMD: > I think I realized what's wrong.
> 
> The minimum version is the minimum core version that supports it without an extension, not the oldest version the extension document says the extension is allowed to operate in.

I'm not very agree that, the 4th param of `profileRequires` is extension require, it's means the `profileRequires` is use check the oldest version the extension document says the extension is allowed to operate in, or the 4th param will been useless.
if we use  `profileRequires(loc, ~EEsProfile, 440, E_GL_ARB_enhanced_layouts, "\"offset\" can only used for atomic counter layout qualifiers");` that will cause even in version 440, shader user still need a entension require, and they can't use it in older version.
The most important thing about this problem is that the extension requirement is in a range, it cannot be use below the range, and it not needed be require when above this range, so the essence of the problem is that it does not provide enough powerful functions to help us implement purely declarative code. If we want it to be beautiful, we need to modify profileRequires, and we cannot change the version number defined in the spec. I would like to solve this problem in the future Pr, rather than making the current Pr more bloated and difficult to control.

johnkslang: For something like this, there are two choices for what it means to have a version and an extension:

1. you need to meet *both* the version *and* the extension requirement
2. you need to meet *either* the version *or* the extension requirement

The current system is 2, not 1.   Do you think it should be 1?  I agree this PR should not be held up for a redesign of this 1 versus 2.  But, not redesigning means using it as it currently is.

I don't think there is a right or wrong choice for design on 1 versus 2, either design can work.  But, it has been 2 and is implemented as 2, and the problem here I think is trying to treat it as if it were 1.

There are around 100 of these that need to be fixed to do it the way you are suggesting, instead of just correcting this PR.

johnkslang: Here is a subset of the comment from the top of Versions.cpp:

```
//    The mostly likely scenario is that Feature F can only be used with a
//    particular profile if XXX_extension_X is present or the version is
//    high enough that the core specification already incorporated it.
//
//        // following the requireProfile() call...
//        profileRequires(loc,
//                        ECoreProfile | ECompatibilityProfile,
//                        420,             // 0 if no version incorporated the feature into the core spec.
//                        XXX_extension_X, // can be a list of extensions that all add the feature
//                        "Feature F Description");
//
//    This allows the feature if either A) one of the extensions is enabled or
//    B) the version is high enough.  If no version yet incorporates the feature
//    into core, pass in 0.
```

Roy-AMD: > For something like this, there are two choices for what it means to have a version and an extension:
> 
> 1. you need to meet _both_ the version _and_ the extension requirement
> 2. you need to meet _either_ the version _or_ the extension requirement
> 
> The current system is 2, not 1. Do you think it should be 1? I agree this PR should not be held up for a redesign of this 1 versus 2. But, not redesigning means using it as it currently is.
> 
> I don't think there is a right or wrong choice for design on 1 versus 2, either design can work. But, it has been 2 and is implemented as 2, and the problem here I think is trying to treat it as if it were 1.
> 
> There are around 100 of these that need to be fixed to do it the way you are suggesting, instead of just correcting this PR.

I think we found a point of disagreement. I didn't read the code seriously, which led me think that the behavior of the function is 2. There are already related checks in glslang. I found a more suitable place to put these codes.

Roy-AMD: > Here is a subset of the comment from the top of Versions.cpp:
> 
> ```
> //    The mostly likely scenario is that Feature F can only be used with a
> //    particular profile if XXX_extension_X is present or the version is
> //    high enough that the core specification already incorporated it.
> //
> //        // following the requireProfile() call...
> //        profileRequires(loc,
> //                        ECoreProfile | ECompatibilityProfile,
> //                        420,             // 0 if no version incorporated the feature into the core spec.
> //                        XXX_extension_X, // can be a list of extensions that all add the feature
> //                        "Feature F Description");
> //
> //    This allows the feature if either A) one of the extensions is enabled or
> //    B) the version is high enough.  If no version yet incorporates the feature
> //    into core, pass in 0.
> ```
> 
> Note that 0 means "not in core", it does not mean "extension supported in all versions".
> 
> We are simply missing the way to gracefully enforce "minimum version that allows the extension". That's the existing hole that is okay to continue having while accepting this PR.

I decided to give up checking the lowest version in this Pr, it can be part of the TODO list.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7452,10 +7454,8 @@ void TParseContext::declareBlock(const TSourceLoc& loc, TTypeList& typeList, con
         if (memberType.isArray())
             arraySizesCheck(memberLoc, currentBlockQualifier, memberType.getArraySizes(), nullptr, member == typeList.size() - 1);
         if (memberQualifier.hasOffset()) {
-            if (spvVersion.spv == 0) {
```
Roy-AMD, line 7455: I think the version and extension check can't be effect by spv version, so I removed this condition.

johnkslang: It's likely that Vulkan allowed putting offset everywhere, as it's required, and it's only when not compiling for Vulkan that the normal OpenGL(ES) rules must be followed for offset.

gtests/AST.FromFile.cpp
```diff
@@ -285,6 +294,18 @@ INSTANTIATE_TEST_CASE_P(
     })),
     FileNameAsCustomTestSuffix
 );
+
+INSTANTIATE_TEST_CASE_P(
+    Glsl, CompileToAstTestError,
+    ::testing::ValuesIn(std::vector<std::string>({
+        "glsl.es300.layoutOffset.error.vert",
```
johnkslang, line 301: If this section is just for checking that errors come out, we already have test lists for that, and using them can probably avoid copying/pasting code.

glslang/MachineIndependent/ParseHelper.cpp
Roy-AMD: > It's likely that Vulkan allowed putting offset everywhere, as it's required, and it's only when not compiling for Vulkan that the normal OpenGL(ES) rules must be followed for offset.

Why we not use `spvVersion.vulkanGlsl` or `spvVersion.openGl` to get the environment? `spvVersion.spv` seem does not reliable.

johnkslang: `spvVersion` predates the others, and within core glslang is what is used to reliably know whether SPIR-V is targeted, etc.

The others were added to support a rich interface, and are translated to the existing internal ones before doing translation.

We could switch over to the newer interface, but we would need a PR updating the >200 places the original ones are used.

To reiterate, the original ones are indeed reliable and sufficiently robust and communicative for translation.

Roy-AMD: > `spvVersion` predates the others, and within core glslang is what is used to reliably know whether SPIR-V is targeted, etc.
> 
> The others were added to support a rich interface, and are translated to the existing internal ones before doing translation.
> 
> We could switch over to the newer interface, but we would need a PR updating the >200 places the original ones are used.
> 
> To reiterate, the original ones are indeed reliable and sufficiently robust and communicative for translation.

Thanks for your explanation, I restored the code and deleted the redundant code in the test. I think it would be better to respect the tradition.

gtests/AST.FromFile.cpp
Roy-AMD: @johnkslang  could we merge it?

PR #2098, opened 2020-03-04
GL_EXT_debug_printf implementation

jeffbolznv:


jeffbolznv:
@johnkslang I can't tell what's going on with the failing CI jobs. They're failing to include the new header file, but the file is there in the same directory in the pull request.

glslang/MachineIndependent/preprocessor/PpScanner.cpp
```diff
@@ -1032,6 +1032,70 @@ int TPpContext::tStringInput::scan(TPpToken* ppToken)
             ch = getch();
             while (ch != '"' && ch != '\n' && ch != EndOfInput) {
                 if (len < MaxTokenLength) {
+                    if (ch == '\\' && pp->parseContext.extensionTurnedOn(E_GL_EXT_debug_printf)) {
```
johnkslang: Parsing this way has historically caused both performance and semantics issues, as opposed to the normal way of accept everything, and then require the extension to be present after seeing what was accepted.

I'll have to think through it still.  Just doing a quick review before looking into the missing header-file problem.

johnkslang:
Is it because the new header does not appear in SPIRV/CMakeLists.txt?

glslang/MachineIndependent/preprocessor/PpScanner.cpp
johnkslang: We also need to address the comment above:

```
            // TODO: If this gets enhanced to handle escape sequences, or
            // anything that is different than what #include needs, then
            // #include needs to use scanHeaderName() for this.
```

johnkslang: So, the semantic problem is that ideally, and I believe we're virtually 100% successful, extensions add features, but don't switch semantics.  At the keyword level we do this, but not normally while building a token.

So, I think the big questions here are:

- Do #include strings have different semantics than debug strings?
- Do strings really change their different semantics just because an extension is present?

Not sure yet if it helps, but, for example:

1. we could steer off (if not already) #include strings to a different algorithm

2. for what's left here, we could note whether or not escape characters were used and then later check for extensions that enable such usage

What works here will also depend on differences, if any, with HLSL strings.

If we add strings to GLSL, I claim we should add one kind of string, not multiple kinds (or one for #include and one for everything else), and hence not need micro-testing inside the token to see how to parse it.

It could be we just want high-level switches.  GLSL vs. HLSL, #include vs. anything else, and no extension from inside the middle of a token.

jeffbolznv: The motivation for this check is that HLSL include strings apparently don't use escape sequences. For example, hlsl.include.vert does `#include "inc2\bar.h"`, and using escape sequences treats \b as a backspace, so it turns into `"incar.h"`.

So yes, HLSL include strings are different than how strings are defined in this GLSL extension. The GLSL extension specifies how escape sequences are handled, so basing it off the extension enable seems pretty natural.

I wasn't aware of scanHeaderName, I'll see if it's possible to use it.

johnkslang: For headers, we just need to make sure they don't start using the escape sequences.

If it's HLSL vs. GLSL, there is a switch for that we can use, far cheaper than querying the extension.

johnkslang: Probably, headers go to a different path, and then we just want to query

```
parseContext.intermediate.getSource() != EShSourceHlsl
```

for the escape sequences.  This is assuming we want GLSL-level strings (not pre-processor, not HLSL) to all have the same semantics, which I think is a good starting point.

johnkslang: Note the GLSL extension did not define the strings as being specific to the argument to printf, but seems to independently add strings and printf, even though it's only the initial use.

jeffbolznv: I've changed CPPInclude to use scanHeaderName for both types of includes. For now, I've left this parsing to use escape sequences for GLSL and HLSL, but I'm happy to change it if you want.

johnkslang: That's probably fine; the distinction was header vs. non-header, rather than HLSL vs. GLSL or extension vs. non-extension.

dneto0: It looks like the filename string for #line can be corrupted by this change.  See #2108

PR #2105, opened 2020-03-09
Allow nonuniformEXT() on sampler types

jeffbolznv:
Also add missing NonUniformEXT decoration on OpImageTexelPointer.


jeffbolznv:
Picked up the spirv-tools fix, I think this is ready now.

SPIRV/SpvBuilder.cpp
```diff
@@ -2731,7 +2731,13 @@ Id Builder::accessChainLoad(Decoration precision, Decoration nonUniform, Id resu
         }
 
         // load through the access chain
-        id = createLoad(collapseAccessChain(), memoryAccess, scope, alignment);
+        id = collapseAccessChain();
+        // Apply nonuniform both to the access chain and the loaded value.
+        // Buffer accesses need the access chain decorated, and this is where
+        // loaded image types get decorated. TODO: This should maybe move to
+        // createImageTextureFunctionCall.
+        addDecoration(id, nonUniform);
```
jeffbolznv: This change is conservative, but I think it's OK. I wonder if the nonuniform stuff should be implemented more like the coherentFlags, but that's a bigger change than I want to make right now.

PR #2107, opened 2020-03-09
Use strcmp for all of the extended instruction set checks in the disassembler

### LINKED ISSUE #2106: Potential read beyond end of buffer in SPIRV/disassemble.cpp
This code specifically isn't safe:
```
const char* name = idDescriptor[stream[word - 2]].c_str();
if (0 == memcmp("OpenCL", name, 6)) {
    extInstSet = OpenCLExtInst;
} else if (0 == memcmp("NonSemantic.DebugPrintf", name, 23)) {
```
Because even though `name` should be null terminated, `memcmp` doesn't stop at the null, and there is no guarantee that name will be 6 or 23 characters long.

ASAN bots for shaderc picked up this issue when trying to integrate the newest version of glslang.

I will send up a PR with a patch for this.

zoddicus:
Fixes #2106

### DIFF
--- a/SPIRV/disassemble.cpp
+++ b/SPIRV/disassemble.cpp
@@ -481,9 +481,11 @@ void SpirvStream::disassembleInstruction(Id resultId, Id /*typeId*/, Op opCode, 
             if (opCode == OpExtInst) {
                 ExtInstSet extInstSet = GLSL450Inst;
                 const char* name = idDescriptor[stream[word - 2]].c_str();
-                if (strlen(name) >= 6 && 0 == memcmp("OpenCL", name, 6)) {
+                if (strcmp("OpenCL.std", name) == 0) {
                     extInstSet = OpenCLExtInst;
-                } else if (strlen(name) >= 23 && 0 == memcmp("NonSemantic.DebugPrintf", name, 23)) {
+                } else if (strcmp("OpenCL.DebugInfo.100", name) == 0) {
+                    extInstSet = OpenCLExtInst;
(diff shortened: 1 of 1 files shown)


johnkslang:
Wouldn't it be better to be using a real string compare instead of memory compare?

I'm sure it is full string matching you want, not just the leading characters.

zoddicus:
I agree that using string compare would be better, if it is correct, but I don't have enough context/knowledge to say if it is correct.

@jeffbolznv & @amdrexu  Can you comment on whether or not it would be correct to replace checking the leading characters with using a full string comparison?

jeffbolznv:
I don't know why it's not a string comparison, I just copied the existing style. IMO a string comparison would probably be better.

zoddicus:
Thanks for the feedback. I will look at updating the PR later today to use strcmp.

zoddicus:
Looking at the various extension specs, it appears there are two potential strings that begin with OpenCL, "OpenCL.std" and "OpenCL.DebugInfo.100". I talked with David about this, and I think I understand what is going on there. There are two related OpenCL extension sets.

I have converted the existing code to all use strcmp, and both of the OpenCL strings go to the same instruction set, which is the existing behaviour, but potentially semantically incorrect.

I have filed #2114 for determining if other changes are needed to be semantically correct.

zoddicus:
Apparently the mac bot couldn't find gmock, which I think is unrelated to my patch

johnkslang:
Merging; will watch and see if bots do okay in master.

johnkslang:
It passed in master.

PR #2109, opened 2020-03-10
disable escape sequences for #line and #error

### LINKED ISSUE #2108: #line  file pathnames are being processed for escapes
The problem is caused by https://github.com/KhronosGroup/glslang/pull/2098

Here's a reproducer.  There are two files in directory c:\build\no:  a.frag and b.glsl
The gv-before.exe is glslangValidator at the commit before #2098 was merged.
The gv-jb.exe is glslangValidator at #2098 

```
C:\build>type no\a.frag
#version 450
#extension GL_GOOGLE_include_directive : enable

content a before

#include "b.glsl"

content a after

C:\build>type no\b.glsl
void main() {}

C:\build>gv-before.exe -Ic:\build\no c:\build\no\a.frag -E
#version 450
#extension GL_GOOGLE_include_directive : enable

content a before

#line 1 "c:/build/no/b.glsl"
 void main(){ }
#line 7 "c:\build\no\a.frag"

content a after


C:\build>gv-jb.exe -Ic:\build\no c:\build\no\a.frag -E
#version 450
#extension GL_GOOGLE_include_directive : enable

content a before

#line 1 "c:/build/no/b.glsl"
 void main(){ }
#line 7 "cuild
o.frag"

content a after

```

The processing of the path string in the last #line directive is corrupted.  The \b and \n have been interpreted as backspace and newline.

The original PR #2098 ensured the header name for #include is not corrupted by escape processing, but there are other cases that need to be handled too:  #line and probably #error.  I don't know if that's the full list.

This was caught by our tests in Shaderc https://github.com/google/shaderc/issues/996 and VKB: https://github.com/google/vkb/pull/82

cc: @jeffbolznv

jeffbolznv:
Fixes #2108

### DIFF
--- a/glslang/MachineIndependent/preprocessor/PpScanner.cpp
+++ b/glslang/MachineIndependent/preprocessor/PpScanner.cpp
@@ -1030,7 +1030,7 @@ int TPpContext::tStringInput::scan(TPpToken* ppToken)
             ch = getch();
             while (ch != '"' && ch != '\n' && ch != EndOfInput) {
                 if (len < MaxTokenLength) {
-                    if (ch == '\\') {
+                    if (ch == '\\' && !pp->disableEscapeSequences) {
                         int nextCh = getch();
                         switch (nextCh) {
                         case '\'': ch = 0x27; break;
--- a/Test/baseResults/hlsl.pp.line4.frag.out
+++ b/Test/baseResults/hlsl.pp.line4.frag.out
@@ -9,8 +9,8 @@ hlsl.pp.line4.frag
                               EntryPoint Fragment 5  "MainPs" 70 74
                               ExecutionMode 5 OriginUpperLeft
                1:             String  "hlsl.pp.line4.frag"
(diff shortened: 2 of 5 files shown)


glslang/MachineIndependent/preprocessor/PpScanner.cpp
```diff
@@ -1030,7 +1030,7 @@ int TPpContext::tStringInput::scan(TPpToken* ppToken)
             ch = getch();
             while (ch != '"' && ch != '\n' && ch != EndOfInput) {
                 if (len < MaxTokenLength) {
-                    if (ch == '\\') {
+                    if (ch == '\\' && !pp->disableEscapeSequences) {
```
johnkslang: I thought out decision and result was to use the scanHeaderName() for includes, not the full string parser.

Will try to look deeper and understand better.

johnkslang:
Will merge to fix the current issue, but:

1. we should have tests for backslash in more places
1. it would be better to have a single way of avoiding escape sequences

Issue #2112, opened 2020-03-10
Runtime error on tests with Address and Undefined behavior sanitizer


durswd:
I added ``` set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined,address") ``` in Cmake and run tests.

These runtime errors are caused. Overflow and accessing uninitialized variables are caused.

```
glslang/glslang/MachineIndependent/linkValidate.cpp:683:53: runtime error: signed integer overflow: -1094795586 * 4 cannot be represented in type 'int'
glslang/hlsl/hlslParseHelper.cpp:5881:28: runtime error: reference binding to null pointer of type 'const struct TIntermSequence'
glslang/glslang/MachineIndependent/Constant.cpp:532:54: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
glslang/glslang/MachineIndependent/Constant.cpp:539:56: runtime error: negation of -9223372036854775808 cannot be represented in type 'long long int'; cast to an unsigned type to negate this value to itself
glslang/glslang/MachineIndependent/preprocessor/Pp.cpp:377:33: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 240, which is not a valid value for type 'bool'
glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 58, which is not a valid value for type 'bool'
glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 18, which is not a valid value for type 'bool'
```

johnkslang:
For linkValidate, it's saying a garbage value for `maxTransformFeedbackInterleavedComponents`, which should default to 64.  But, it depends on you to set up the resources.  How did you set the resources to do this run?

I think for the Constant.cpp, we are relying on the host being 2's complement and doing (as tested) the right thing here on behalf of what the target would do in the same situation.  Maybe similar in others as well.

For localintermediate.h, where is the load happening?  Line 100 is just the data structure. I said more in #2051.

durswd:
I'm sorry for the slow reply.
For linkValidate, it is 64 bit in almost tests.
But  it is wrong in some tests such as 

```
Glsl/LinkTest.FromFile/0
Glsl/LinkTestVulkan.FromFile/0
[...]

PR #2113, opened 2020-03-10
Fix a bug that uninitialized valuables are accessed

durswd:
I fix bugs about

glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 240, which is not a valid value for type 'bool'
glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 58, which is not a valid value for type 'bool'
glslang/glslang/MachineIndependent/localintermediate.h:100:8: runtime error: load of value 18, which is not a valid value for type 'bool'

in https://github.com/KhronosGroup/glslang/issues/2112

### DIFF
--- a/glslang/MachineIndependent/localintermediate.h
+++ b/glslang/MachineIndependent/localintermediate.h
@@ -98,7 +98,10 @@ private:
 // A "call" is a pair: <caller, callee>.
 // There can be duplicates. General assumption is the list is small.
 struct TCall {
-    TCall(const TString& pCaller, const TString& pCallee) : caller(pCaller), callee(pCallee) { }
+    TCall(const TString& pCaller, const TString& pCallee)
+        : caller(pCaller), callee(pCallee), visited(true), currentPath(true), errorGiven(true), calleeBodyPosition(0)
+    {
+    }
     TString caller;
     TString callee;
     bool visited;


glslang/MachineIndependent/localintermediate.h
```diff
@@ -98,7 +98,10 @@ class TSwizzleSelectors {
 // A "call" is a pair: <caller, callee>.
 // There can be duplicates. General assumption is the list is small.
 struct TCall {
-    TCall(const TString& pCaller, const TString& pCallee) : caller(pCaller), callee(pCallee) { }
+    TCall(const TString& pCaller, const TString& pCallee)
+        : caller(pCaller), callee(pCallee), visited(true), currentPath(true), errorGiven(true), calleeBodyPosition(0)
```
johnkslang: These are probably not good defaults, but that opens the discussion as to what are good defaults, and I don't think there are any.

The model of use is a higher-level traverser will modify these at the right times.  E.g., see `checkCallGraphCycles()`.

It is in fact misleading to move those here though, implying that the user no longer has to worry about it.  It is more communicative to make it more clear these are not yet set.  Unless, that's really not the model.

Did you actually see code that cares about one of these values, before the value is set?

durswd:
It seem that there are bugs because a sanitizer notify that uninitialized values are used. But this PR is not suitable for this bug.

johnkslang:
I think the sanitizer might be notifying the unitialized values got read, but if all they did was initialize another variable that also does not need to be initialized, where all such variables still get correctly set before any actual decision is based on their value, then everything is still okay.

Issue #1816, opened 2019-06-30
Namespace CMake build options


DiligentGraphics:
Glslang now uses very generic names for CMake options such as `BUILD_SHARED_LIBS`. This may cause conflicts/confusion when integrating glslang in larger projects (https://github.com/DiligentGraphics/DiligentEngine/issues/45).
It is preferred to namespace global settings such as `GLSLANG_BUILD_SHARED_LIBS` (similar to for instance how [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/CMakeLists.txt#L66) does that) to clearly separate them.

DiligentGraphics:
`ENABLE_CTEST` is now another conflicting option

jeremyong:
This should be higher priority. The project is extremely unfriendly to other projects that want to transitively include glslang. There's also `BUILD_TESTING` in the root cmake file.

juan-lunarg:
For context this is mainly useful for `add_subdirectory` or `FetchContent` users (Which by default uses add_subdirectory).

This really doesn't impact find_package users. Not saying this shouldn't be addressed. Just giving context on who this primarily impacts.

Issue #1964, opened 2019-11-06
Remove deprecated install path for SPIRV/ folder


dj2:
With #1963 we are installing the SPIRV/ folder into both `CMAKE_INSTALL_PREFIX/SPIRV` and `CMAKE_INSTALL_PREFIX/glslang/SPIRV`. This was to fix #1959  which broken when 7b0e236 was committed which moved the `SPIRV` folder into the `glslang` subfolder.

This is a tracking bug to remove the deprecated `CMAKE_INSTALL_PREFIX/SPIRV` installation. The plan of record is to land #1963, then update the `README.md` with a deprecation note and a 6 month lead time.  This bug will stay open during that time if folks have issues with the deprecation, please comment here. We will be installing to both `SPIRV` and `glslang/SPIRV` during the deprecation period to allow applications to migrate.

At that point, we will stop installing headers to the `CMAKE_INSTALL_PREFIX/SPIRV` folder.

dj2:
One part we're going to have to work out is, the `SPIRV` folder exists at the top level of the glslang directory. So, currently things using CMake would do `SPIRV/GlslangToSpv.h` and it will work, that doesn't depend on the include directory.

But, if users expect to build with an installed GLSLang they will need to include `glslang/SPIRV/GlslangToSpv.h` which means your include path is different between cmake include and system include.

To fix we'd need to move the `SPIRV` folder under `glslang` in the repo, or do ... something? in cmake.

dj2:
The deprecation date is set to: May 1, 2020

johnkslang:
If the problem you're solving is the collision of SPIR-V headers proper (not glslang headers), and it's arguably better to use proper SPIR-V headers anyway, rather than glslang's snapshot, should we just be focusing on having glslang not overwrite those files, rather than on moving files unique to glslang?

johnkslang:
@dj2 what's the latest best plan for this now?

It is okay to execute on it, if you'd like to submit a PR with the latest wisdom we have regarding this.

See #2216, thanks.

PR #2115, opened 2020-03-11
Prefix cmake options and compiler definitions with GLSLANG_

jeremyong:
Addresses #1816 

Most changes are direct find-replaces with the standard `GLSLANG_` prefix. Note that there is some pre-existing inconsistency where some options/definitions were `GLSLANG_ENABLE` and others were `ENABLE_GLSLANG`. I opted not to change the latter options to keep what is a sizable commit slimmer and reduce downstream disruption.


johnkslang:
Agreed in principle, and that it's the right direction to go.

My concern is breakage with existing usages that have coded the current names.  I'd like some idea of how many consumers will break, and if there are any, how we manage the transition to the better way.

jeremyong:
That sentiment is understandable. My thought process is that there is already a planned deprecating change in #1964 and perhaps this would be a good opportunity to clean up other aspects of the project. Users that rely on the installed binary would be unaffected.

jeremyong:
Another consideration is to not even have the GLSL dedicated option for building shared vs static libraries. Cmake already provides `BUILD_SHARED_LIBS` and generally it's not recommended to mix static and shared linkage. The breakage I'm envisioning is libraries that depend on glslang but do not know to forward the `BUILD_SHARED_LIBS` option to `glslang`, so I'm of the opinion that having `GLSLANG_BUILD_SHARED_LIBS` is itself an anti-pattern.

dj2:
I'd recommend accepting both for now. Add an entry to the README that the old versions are deprecated and will be removed in ~6 months. All the glslang internal code should be updated to the new flags and have the CMake files detect the old ones and set the new ones. We then remove the cmake bits when we hit the deprecation date.

This will make it easier to roll out to the ecosystem without causing a lot of breakage.

dneto0:
> GLSLANG_BUILD_SHARED_LIBS is itself an anti-pattern.

I'm coming to that belief too.   People have said that of Shaderc before, and they've won me over.

Related:  

- supporting shared libs "properly" in SPIRV-Tools  https://github.com/KhronosGroup/SPIRV-Tools/issues/3214 
- supporting shared libs "properly" in Shaderc https://github.com/google/shaderc/issues/994
- prior relevant discussion on this abandoned Shaderc PR  https://github.com/google/shaderc/pull/498

I'd really want to solve this problem one way across all these projects (and others I influence).   Thanks everybody for your patience.

jeremyong:
Just for clarity, should I leave this PR in a holding pattern then until it's decided how to handle the deprecation across the ecosystem? I'm happy to update the PR but would prefer to wait for consensus from the relevant stakeholders

ben-clayton:
Somewhat unrelated to the discussion above, but there's a pattern I've been using to try and reduce the number of settings that cannot be modified appearing in CMake projects that include other CMake projects.

Consider CMake project `Outer` that  depends on `Inner`. `Inner` has a number of `option()`s that `Outer` **requires** to be set to particular values.

Typically  `Outer` will use `FORCE CACHE` to set the `option()` to a particular value before including `Inner` with `add_subdirectory()`. This option still appears in the CMake settings list for no good reason, and worse still, changing this value will do nothing.

Instead, I've been using [this pattern](https://github.com/google/marl/blob/62f209bbb1d944435c5572151aff99464c792fbf/CMakeLists.txt#L21-L37) in the `Inner` project:
```cmake
function (option_if_not_defined name description default)
    if(NOT DEFINED ${name})
        option(${name} ${description} ${default})
    endif()
endfunction()

option_if_not_defined(MARL_WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
option_if_not_defined(MARL_BUILD_EXAMPLES "Build example applications" OFF)
option_if_not_defined(MARL_BUILD_TESTS "Build tests" OFF)
option_if_not_defined(MARL_BUILD_BENCHMARKS "Build benchmarks" OFF)
```

This acts as a regular `option()` when `Inner` is used as a standalone project, or if the setting is not set by `Outer`.
However, if `Outer` does `set()` any of these settings before calling `add_subdirectory()`, then they do not act as options, nor show up in the CMake settings / `CMakeCache.txt`.

johnkslang:
Offline, Ben also pointed to https://stackoverflow.com/questions/48524359/naming-convention-for-components-and-namespaces-in-cmake for naming possibilities; just capturing that here as well.

jeremyong:
Strictly speaking, exported targets should following `::` convention (e.g. `glslang::glslang`) because CMake knows that linking against a name with a double-colon is strictly a target (and not, for example, a library/file on the filesystem somewhere).

dnovillo:
@jeremyong This is a rather old PR. Is it still relevant?

dnovillo:
@jeremyong no response in more than a year. Closing for now. Please re-open if you are still interested in this.

PR #2122, opened 2020-03-12
Fix calculation of log2 and exp2.

ShchchowAMD:
Purpose:
Fix calculation of log2 and exp2.

Modification:
During test, the result of 'const float n = log2(14.0 + float(int(-3.0) - int(3.0)));' is incorrect, the expression is fold to constant at compile-time, but glslang fold the log2() to 1.44 * log(), which made the result incorrect, in fact the system has provided such built-in functions, Also exp2 has the same issues, so the issue about exp2 is also fixed in this change.

### DIFF
--- a/glslang/MachineIndependent/Constant.cpp
+++ b/glslang/MachineIndependent/Constant.cpp
@@ -599,17 +599,11 @@ TIntermTyped* TIntermConstantUnion::fold(TOperator op, const TType& returnType) 
             newConstArray[i].setDConst(log(unionArray[i].getDConst()));
             break;
         case EOpExp2:
-            {
-                const double inv_log2_e = 0.69314718055994530941723212145818;
-                newConstArray[i].setDConst(exp(unionArray[i].getDConst() * inv_log2_e));
-                break;
-            }
+            newConstArray[i].setDConst(exp2(unionArray[i].getDConst()));
+            break;
         case EOpLog2:
-            {
-                const double log2_e = 1.4426950408889634073599246810019;
-                newConstArray[i].setDConst(log2_e * log(unionArray[i].getDConst()));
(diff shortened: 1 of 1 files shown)


johnkslang:
I just want to note that this is one of many "draft" PRs now in glslang.

Can you clarify your intent for these, regarding process?

I'm currently not doing much with them, treating them like WIP.  But with so many, we should probably be operating under the same assumptions.

ShchchowAMD:
Hi John, I'm currently working to sync code from this master branch to my local work space. Those are some commits when we find bugs during development or some missing features during testing.

I'll try to create PR next time ASAP.

For those PRs, each one map to an issue found recently. I am appending tests now. Maybe I could keep them in one PR with several commits?

ShchchowAMD:
I don't make sure whether there would be any conflicts with other guys' PRs. So I just make those PRs as drafts now.

After some testings this week and making sure they are okay, I'll try to publish them for reviews. Local testing seems good for them currently.

johnkslang:
I think it is okay to do it how is best for you, I only want clarity on what process is expected.

For example, is it okay to largely ignore draft PRs (no problem having them present, just not getting attention)?  We would just look seriously at non-draft PRs.

ShchchowAMD:
Yes, I think we can assume that.

Only when a PR is ready, it will be published for reviewing. Please ignore those drafts.

And also please send me a message if you think a draft PR remains too long on the list, which should be closed
(Sometimes, those PRs are waiting to be re-implemented or still have bugs).

johnkslang:
Let's deal with it that way for now.  The ongoing issue is the contexts where the only way to see it's a draft or not is the slight color difference between grey and green. :)

PR #2123, opened 2020-03-12
Fix #2005. Allow multiple compilation units to declare identical push_constant blocks

### LINKED ISSUE #2005: Only one push_constant block is allowed per stage - identical push_constant block
Compiling for Vulkan.
I have a TProgram that is built up using multiple TShader objects for the same stage. Some contain utility functions, and one contains the main() for that stage. They all declare the same push_constant block. Each shader may use only a few variables of that block, but they all declare the exact same struct since a differing struct for the same stage would make no sense.
GLSLang fails with the error "Only one push_constant block is allowed per stage".
It would seem that GLSLang should compare the push_constant blocks and only error if any of the shaders have block that differs from another. Or, is there another way to share the block between the shaders that I'm missing?

Is this just a missing feature of GLSLang, or is what I'm doing somehow against a spec?
Thanks

mbechard:
I didn't add any unit tests for this since it seems like the framework isn't setup to do anything with multiple compilation units currently. Hope that's ok.


johnkslang:
There are already some tests, in the normal framework.  See:

- `glslang/gtests/Link.FromFile.cpp`
- `glslang/gtests/Link.FromFile.Vk.cpp`

I believe they cover validation level testing, negative vulkan testing, and positive vulkan testing, though the latter is pretty slim.

johnkslang:
Marking WIP until a test for push constants is added.

mbechard:
Thanks for the pointer, didn't look at that file. Updated with unit tests.

mbechard:
Also fixed uninitialized variable in Link.FromFile.Vk.cpp which caused the test results to have inconsistent output when compilation failure occurs.

Issue #2134, opened 2020-03-18
Recent EXT_ray_tracing change breaks VK-GL-CTS compilation


y-novikov:
After https://github.com/KhronosGroup/glslang/pull/2130 we are getting this compilation error:
https://luci-milo.appspot.com/p/chromium/builders/ci/GPU%20FYI%20Win%20x64%20dEQP%20Builder/52743

 ../../third_party/angle/third_party/VK-GL-CTS/src/external/openglcts/modules/common/glcSpirvUtils.cpp(62,3): error: use of undeclared identifier 'EShLangRayGenNV'; did you mean 'EShLangRayGen'?
                 EShLangRayGenNV, EShLangAnyHitNV, EShLangClosestHitNV, EShLangMissNV, EShLangIntersectNV, EShLangCallableNV

@johnkslang, @dgkoch

johnkslang:
That error seems to be outside glslang.

It does seem that the recent ray-tracing changed the interface to glslang, and maybe that should be done with aliases instead of replacements?

@dgkoch can we keep both the old and the new names in the interface, to not break existing consumers?

dgkoch:
Sounds like keeping both might be safer cc @alelenv

PR #2135, opened 2020-03-18
Re-add NV enums for raytracing to prevent build breaks.

alelenv:
Fixes glslang issue #2134


johnkslang:
This indeed fixed the problem, thanks.

PR #2138, opened 2020-03-19
for link validation, match blocks by block-name, not instance-name

### LINKED ISSUE #2136: Shouldn't TIntermediate::mergeLinkerObjects be matching blocks by block-name, not instance-name
Currently if you have two blocks with different layouts in two compilation units, with the same block-name, but different instance-name, the validator does not fail.
[SameBlockDiffInstance2.txt](https://github.com/KhronosGroup/glslang/files/4351610/SameBlockDiffInstance2.txt)
[SameBlockDiffInstance1.txt](https://github.com/KhronosGroup/glslang/files/4351611/SameBlockDiffInstance1.txt)
I think this should fail, no?

Conversely, if you have 2+ anonymous blocks (legal ones) it will fail if they are declared in different order, since it will be trying to match up by anon@0, naming, not the block names.

[MultiAnonBlocks1.txt](https://github.com/KhronosGroup/glslang/files/4351633/MultiAnonBlocks1.txt)
[MultiAnonBlocks2.txt](https://github.com/KhronosGroup/glslang/files/4351634/MultiAnonBlocks2.txt)

Seems like the root issue is that TIntermediate::mergeLinkerObjects is matching up the blocks by instance name and not block name.

Working on a patch, but let me know if I'm off base. Thanks

mbechard:
Fixes #2136

Also the output from 150.tesc.out test case changed due to
mismatch from gl_PerVertex declaration that previously wasn't
picked up, but is now.

### DIFF
--- a/Test/baseResults/150.tesc.out
+++ b/Test/baseResults/150.tesc.out
@@ -960,6 +960,9 @@ Linked tessellation evaluation stage:
 ERROR: Linking tessellation evaluation stage: can't handle multiple entry points per stage
 ERROR: Linking tessellation evaluation stage: Multiple function bodies in multiple compilation units for the same signature in the same stage:
     main(
+ERROR: Linking tessellation evaluation stage: Types must match:
+ERROR: Linking tessellation evaluation stage: Storage qualifiers must match:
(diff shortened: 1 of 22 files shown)


johnkslang:
It is not clear this is the right direction. Discussion in #2136.

mbechard:
Will create a new one when ready.

PR #2130, opened 2020-03-17
Add support for GLSL_EXT_ray_tracing

dgkoch:
and SPV_KHR_ray_tracing

PR #2139, opened 2020-03-19
GL_EXT_ray_query updates

neslimsah:
rayQuery extension updates.


johnkslang:
@neslimsah can you make an account and sign the CLA?  Or, possibly have this come from someone who has done that?

Thanks.

SPIRV/spirv.hpp
```diff
@@ -1939,6 +1940,31 @@ inline void HasResultAndType(Op opcode, bool *hasResult, bool *hasResultType) {
     case OpRayQueryGetIntersectionObjectToWorldKHR: *hasResult = true; *hasResultType = true; break;
     case OpRayQueryGetIntersectionWorldToObjectKHR: *hasResult = true; *hasResultType = true; break;
     case OpExecuteCallableNV: *hasResult = false; *hasResultType = false; break;
+    case OpTypeAccelerationStructureKHR: *hasResult = true; *hasResultType = false; break;
```
johnkslang: This should be a machine generated file, from the SPIRV-Headers repository.

Has SPIRV-Headers been updated with a new grammar that creates this?

Once so, I will copy the resulting spirv.hpp here, and push that before this PR goes in.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -4322,8 +4323,33 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "\n");
         }
 
-    // Builtins for GL_NV_ray_tracing/GL_EXT_ray_tracing
+    // Builtins for GL_NV_ray_tracing/GL_EXT_ray_tracing//GL_EXT_ray_query
```
johnkslang: Nit question: Was only one `/` intended to be added?

SPIRV/spirv.hpp
neslimsah: We have updated SPIRV-Headers also. However, I do not know if they are live or not yet.

glslang/MachineIndependent/Initialize.cpp
neslimsah: Intention was single /. Updated.

SPIRV/spirv.hpp
johnkslang: Where?  I didn't see a PR for it.

Tobski: Was this not done as part of #2130?

Tobski: Indeed it was.

johnkslang: Not quite.  #2130 was preceded by f368dcbb7d8af23f0cba3015d0f4dda9dc3aa66d, which is the step where I bring headers from SPIR-V headers into glslang.  #2130 then showed no diff with spirv.hpp, because they then matched, which is the goal.  My concern is in not seeing that happening here.

johnkslang: To be specific, the folllowing is not in SPIR-V headers.  It is also not in glslang, because those match.

```
case OpTypeAccelerationStructureKHR
```

jiaolu:
the test case are raygeneration shader stage. should add graphics and compute shader stages test cases

neslimsah:
> the test case are raygeneration shader stage. should add graphics and compute shader stages test cases

You mean graphics and compute shader stage test cases for rayQuery extention? Thanks.

johnkslang:
I see bots are reporting:

```
All checks have passed
2 successful checks
```

We will need to know why there are only 2 checks.  Travis CI, for example, did not run.  I see a change to allow 30 minutes, but hopefully that's actually not the point.  Some bot runs occasionally fail, due to timeouts, where it appears some part of the process hung.  But, I'm not sure now what the current situation is.

.travis.yml
```diff
@@ -43,7 +43,7 @@ addons:
     packages:
       - clang-3.6
 
-install:
+install: travis_wait 30 mvn install
```
johnkslang: I would like to remove this again, as I'd rather consider it an error if things did start taking this long.

SPIRV/spirv.hpp
johnkslang: Just a note that this still needs resolution.

The SPIRV-Headers project should auto generate this file.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3241,6 +3259,11 @@ void TParseContext::accStructCheck(const TSourceLoc& loc, const TType& type, con
 
 }
 
+void TParseContext::rayQueryCheck(const TSourceLoc& loc, const TType& type, const TString& identifier)
```
johnkslang: Is this something still desired to be present?

neslimsah: I removed this function.

.travis.yml
neslimsah: Updated to origin.

SPIRV/spirv.hpp
neslimsah: For the SPIRV-Headers, it is up-to-date with rayQuery extension updates. How can I auto-generate it to be used in this update?

johnkslang: We are missing a PR to the the SPIRV-Headers project that reserves the enumerants in the JSON grammar file.

neslimsah: I updated https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.core.grammar.json and https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.h at the SPIRV-Headers. The master branch contains the updates. I think I am missing a point. Are these the files that needs to be updated?

dgkoch: @johnkslang the SPIRV-Headers project has the enumerants reserved already, but it doesn't seem to include these changes. 

@neslimsah did you add these changes manually here or if not, how did you generate them?

neslimsah: I added them manually here, I do not know about the process of this auto-generation.

dgkoch: Actually these are all present in this file already. Lines 1917-1941.  
Only OpTypeAccelerationStructureKHR isn't added because OpTypeAccelerationStructureNV is already present and they have the same value. 
You need to just drop these changes from the PR.

neslimsah: Remove all OpTypeAccelerationStructureKHR from this PR?

dgkoch: No, just remove all changes to SPIRV/spirv.hpp in this CL. They should be extraneous, and seem like they are more a result of a bad merge (and would actually cause build failures).

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -3,6 +3,7 @@
 // Copyright (C) 2012-2015 LunarG, Inc.
 // Copyright (C) 2015-2020 Google, Inc.
 // Copyright (C) 2017 ARM Limited.
+// Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
```
dgkoch: Remove copyright notice changes when there are no other changes to the file.

glslang/MachineIndependent/ParseContextBase.cpp
```diff
@@ -1,6 +1,7 @@
 //
 // Copyright (C) 2002-2005  3Dlabs Inc. Ltd.
 // Copyright (C) 2016 Google, Inc.
+// Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
```
dgkoch: Remove copyright notice changes when there are no other changes to the file.

glslang/MachineIndependent/ParseHelper.h
```diff
@@ -2,6 +2,7 @@
 // Copyright (C) 2002-2005  3Dlabs Inc. Ltd.
 // Copyright (C) 2012-2013 LunarG, Inc.
 // Copyright (C) 2015-2018 Google, Inc.
+// Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
```
dgkoch: Remove copyright notice changes when there are no other changes to the file.

SPIRV/spirv.hpp
neslimsah: Reverted spirv.hpp in this CL to commit (f368dcbb7d8af23f0cba3015d0f4dda9dc3aa66d) - commit before my updates.

dgkoch: Still a spacing change at the end of the file, but otherwise that looks better now.

alelenv:
@Tobski @neslimsah  : rayQuery should be passed by reference even to user functions similar to samplers right?
Current code seems to do pass by value.
Also rayQuery is modifiable via intrinsics but can't be assigned to so I think the lvalue check should still be there

Tobski:
> @Tobski @neslimsah : rayQuery should be passed by reference even to user functions similar to samplers right?
> Current code seems to do pass by value.

I think this is correct? The "rayQuery" itself is a reference. It's a bit weird but I think it's consistent with how we've defined SPIR-V (which is also weird because reasons...). I could be missing something though?

glslang/MachineIndependent/Intermediate.cpp
dgkoch: This file still doesn't have any changes besides the copyright change.

SPIRV/spirv.hpp
johnkslang: We can live with that for the moment.

SPIRV/GlslangToSpv.cpp
```diff
@@ -7615,6 +7721,98 @@ spv::Id TGlslangToSpvTraverser::createMiscOperation(glslang::TOperator op, spv::
...
+        opCode = spv::OpRayQueryGetWorldRayOriginKHR;
+    } break;
+    case glslang::EOpRayQueryGetIntersectionObjectToWorld: {
+        typeId = builder.makeMatrixType(builder.makeFloatType(32), 4, 3);
+        opCode = spv::OpRayQueryGetIntersectionObjectToWorldKHR;
+    } break;
+    case glslang::EOpRayQueryGetIntersectionWorldToObject: {
+        typeId = builder.makeMatrixType(builder.makeFloatType(32), 4, 3);
+        opCode = spv::OpRayQueryGetIntersectionWorldToObjectKHR;
+    } break;
```
johnkslang: Nit: All these curly braces seem superfluous.  Yes/no?  It is also a novel way of layout for glslang curly braces, though I think acceptable anyway.

I find it stranger though to have both returns and breaks for some cases.  The `return` makes the `break` dead code.

Couldn't the code by done like the following?

```
    case glslang::EOpRayQueryInitialize:
        builder.createNoResultOp(spv::OpRayQueryInitializeKHR, operands);
        return 0;
    case glslang::EOpRayQueryTerminate:
        builder.createNoResultOp(spv::OpRayQueryTerminateKHR, operands);
        return 0;
    case glslang::EOpRayQueryGenerateIntersection:
        builder.createNoResultOp(spv::OpRayQueryGenerateIntersectionKHR, operands);
        return 0;
    case glslang::EOpRayQueryConfirmIntersection:
        builder.createNoResultOp(spv::OpRayQueryConfirmIntersectionKHR, operands);
        return 0;
    case glslang::EOpRayQueryProceed:
        typeId = builder.makeBoolType();
        opCode = spv::OpRayQueryProceedKHR;
        break;
    ...
```

johnkslang:
I re-reviewed the whole thing.

I only had one nit/question/desired-improvement about the odd `switch` statement layout above.

Modulo that, I'm okay with this change, but see there might be some open questions you still want to resolve.

After answering my `switch` statement question, just let me know when your satisfied with it, and I can merge it.

SPIRV/GlslangToSpv.cpp
neslimsah: Agree, it looks cleaner this way with no dead code and updated as suggested.

dgkoch:
@johnkslang looks like all issues have been addressed and is passing now, can we get it merged?

Thanks.

alelenv:
My question was not resolved and i have filed issue #2150 to discuss/resolve there

neslimsah:
pass-by-reference issues #2150 and #2153 are fixed by #2157.

Issue #2140, opened 2020-03-19
Build error using MSVC


courtney-g:
FAILED: obj/third_party/glslang/src/glslang_sources/parseConst.obj 

../../third_party/glslang/src/glslang/MachineIndependent/parseConst.cpp(1): error C2220: the following warning is treated as an error
../../third_party/glslang/src/glslang/MachineIndependent/parseConst.cpp(1): warning C4828: The file contains a character starting at offset 0x19ce that is illegal in the current source character set (codepage 65001).

Looking at the file on Windows I see:
                        // If there is a single scalar parameter to a matrix
                        // constructor, it is used to initialize all the
                        // components on the matrix<92>s diagonal, with the
                        // remaining components initialized to 0.0.

I left a comment on the CL for the suspected line that's at issue.

PR #2141, opened 2020-03-19
Fix MSVC build issue - remove invalid character

courtney-g:
On Windows the single quote used in a comment was showing up as some
special character that the Microsoft compiler didn't like.

Bug #2140


johnkslang:
Note this has had no problem on multiple MSVC compilers, so not sure what the correct characterization is, but it's still a good change to make, merging.

courtney-g:
> Note this has had no problem on multiple MSVC compilers, so not sure what the correct characterization is, but it's still a good change to make, merging.

Maybe MSVC 2019 is pickier.

PR #2146, opened 2020-03-21
Make file formatting comply with POSIX and Unix standards

aaronfranke:
UTF-8, LF, and no BOM are all recommended for file formatting. These settings are the most compatible with various tools and also take up the least amount of space (CRLF -> LF in a file means saving one byte per line, and no BOM saves two bytes per file). If the whole file has diffs, that's CRLF -> LF.

* For some reason the BUILD files had rogue carriage return characters in the middle of them... how does that even happen?

The POSIX standard for text files says that each text file should end with a newline, otherwise it is by definition not a text file. GitHub shows a warning when a file is committed without a terminating newline (see the "before" diffs of this PR), and GCC shows warnings when compiling files that don't end with newlines. See [this SO question](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/25322168) and [this Unix SE question](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) for more information. 

I didn't touch any of the shader test files such as `.vert` and `.frag` files, but it's possible that this would also be a beneficial change for those files.


johnkslang:
Some window's developers might accidentally add CR. My environments should not, and I thought we had .gitattributes set up to fix most.  Maybe a file type is missing.

johnkslang:
@jeffbolznv is your .gitattributes somehow not working correctly?  I think you can use no CR everywhere.

jeffbolznv:
My .gitattributes is unmodified from what's in the repo.

PR #2149, opened 2020-03-23
Fix build on CMake 2.8, and fix Web build

### LINKED ISSUE #2145: Cmake minimum version
Should the cmake set a minimum version of 3.13.5?

I'm not too familiar with the details of cmake, but when trying to compile with versions less than 3.13.5, I run into an error for `add_link_options`, which I see on the cmake page appears to have been added in that version.

kainino0x:
- Replace `target_link_options` usages with `target_link_libraries`.
- Fix some compile breakage from recent glslang changes.
- Avoid using `-std=c++11` on C files, needed for a sentinel C file that seems to be generated by the Emscripten build.
- Suppress some Clang warnings that are too verbose in Web builds.
- Brought README up to date.

Fixes #2145


kainino0x:
Rendered version of the updated readme:
https://github.com/KhronosGroup/glslang/blob/817d0f63d354c968e94767c03d82161756971506/README.md#building-to-wasm-for-the-web-and-node

johnkslang:
Awesome, thanks.

PR #2154, opened 2020-03-26
Add missing braces to if condition

GMNGeoffrey:
The indentation implies this was the intention. Noticed the issue while trying to compile our code with -Werror -Wall


johnkslang:
Note this is a performance only change (i.e., no bug was present), but certainly worth doing.

PR #2156, opened 2020-03-26
Shader interface matching rework to fix #2136

### LINKED ISSUE #2136: Shouldn't TIntermediate::mergeLinkerObjects be matching blocks by block-name, not instance-name

mbechard:


glslang/Include/Types.h
```diff
@@ -1616,6 +1628,24 @@ class TType {
...
+            return EsiNone;
+        case EvqVaryingIn:
+            return EsiInput;
+		case EvqVaryingOut:
+            return EsiOutput;
+        case EvqUniform:
+            return EsiUniform;
+        case EvqBuffer:
+            return EsiBuffer;
+		}
```
johnkslang: Something happening here with tabs.  Project is all spaces.

glslang/Include/Types.h
```diff
@@ -473,6 +473,18 @@ enum TInterlockOrdering {
     EioCount,
 };
 
+enum TShaderInterface
+{
+    EsiUniform = 0,
```
johnkslang: Shouldn't we have just 3 interfaces?  In, out, and uniform?

mbechard: Good question. I just checked and on Windows neither the Nvidia nor the Intel compiler allow for duplicated block names between uniform blocks and buffer blocks, however the spec states:

> A shader interface is defined to be one of these:
> • All the uniform variables and uniform blocks declared in a program. This spans all compilation
> units linked together within one program.
> • All the buffer blocks declared in a program.
> • The boundary between adjacent programmable pipeline stages: This spans all the outputs
> declared in all compilation units of the first stage and all the inputs declared in all compilation
> units of the second stage. Note that for the purposes of this definition, the fragment shader and
> the preceding shader are considered to have a shared boundary even though in practice, all
> values passed to the fragment shader first pass through the rasterizer and interpolator.

Which to me means buffers are a different interface than uniform blocks.

glslang/Include/Types.h
mbechard: Sorry, the clang-format is catching most of these but I think one action in vim I'm doing is still inserting tabs. I'll push a fix

glslang/Include/Types.h
johnkslang: I think experts in this space commonly consider buffer and uniform to be in the same space, which agrees with your driver experience, so let's go with that.  It will certainly be more portable.

mbechard: Ok sounds good. I'll make an issue report over on the GLSL spec repo since I feel this can be more clearly stated then.

mbechard: Pushed

PR #2170, opened 2020-04-02
Remove extra semicolons

zoddicus:
This is causing downstream users compiler errors if they have Werror or other
particularly restrictive flags turned on.


zoddicus:
Apparently emacs felt the need to trim some excess whitespace also.

Issue #2169, opened 2020-04-01
missing support for gl_RayFlagsSkipAABBEXT and gl_RayFlagsSkipTrianglesEXT


dgkoch:
These should be enabled by both GL_EXT_ray_query and GL_EXT_ray_tracing extensions. 

When either of these flags is used (and only when they are used) the SPIR-V emitted should include the RayTraversalPrimitiveCullingProvisionalKHR capability. 

cc @alelenv

dgkoch:
implemented a while ago.

PR #2173, opened 2020-04-06
Add support for EXT_ray_flags_primitive_culling.

alelenv:
Fixes issue #2169.


SPIRV/GlslangToSpv.cpp
```diff
@@ -1448,6 +1448,10 @@ TGlslangToSpvTraverser::TGlslangToSpvTraverser(unsigned int spvVersion,
         builder.addExecutionMode(shaderEntry, spv::ExecutionModeXfb);
     }
 
+    if (sourceExtensions.find("GL_EXT_ray_flags_primitive_culling") != sourceExtensions.end()) {
```
johnkslang, line 1451: Are semantics actually changed by declaring the extension, or is there some feature that must get used to truly need the capability?

johnkslang: Generally, we are triggering capability need by use of the ability.

johnkslang: See for example `TGlslangToSpvTraverser::TranslateBuiltInDecoration`

dgkoch: RayFlag values 256 and 512 become valid when this capability is enabled, but there are more ways that you could get the values 256 and 512 than by just using the gl_RayFlagsSkipAABBEXT and gl_RayFlagsSkipTrianglesEXT constants.

johnkslang: So, on the one hand it sounds like we should be testing for emitting 256 and 512 as the reason to require the capability, but on the other hand I'm not clear if that's your position.

Speculating: If this can't be detected 100% by a feature use on translation, we also have `Builder::postProcessFeatures()` to detect what was needed after translation completed.

dgkoch: Ultimately, there is no guaranteed way to know at compile time if this capability is needed. 
The "Ray Flags" argument to OpTraceRayKHR is any Id. No requirement for it to be a constant. It could be loaded from memory, or passed in from another stage (as is done with the IncomingRayFlagsKHR builtin). This ultimately needs run-time validation, so we believe we need a separate extension to enable the shader author to specifically opt-in to using the capability.

johnkslang: I see.  It's a feature that the compiler cannot tell is being used, so the extension does nothing except signal a run-time value might appear.  Have we done this much?

The way to semantically guard such a feature would be, for example, to have the #extension declare a keyword/type/declaration/etc. that syntactically declares what is in use.  These could turn into modes in SPIR-V rather than capabilities.  There are choices in all those steps, but the point is that it is quite possible and common to still have a syntactic or semantically statically visible sign of what feature is in use.

Saying it more shortly, the #extension line by itself should not be what changes value semantics, it should be what enables a language feature that says what happens to semantics.

johnkslang: On re-reading the GLSL extension, the dynamic part really does not come through.  It is focused on providing the two built-ins, not on adding a new mode that becomes present whether or not the new built-ins are consumed.

I think this deserves better documentation, if not actual language semantics.

dgkoch: We can improve the documentation in the GLSL extension.  I don't know that we have precedent for anything like this, and I'm not sure how any modes or other declarations would really be an improvement, but can you add further suggestions in https://gitlab.khronos.org/vulkan/vulkan/issues/2073 and we can consider changing that for final? 
Right now we are blocked on actually being able to test these features until something is done here.

PR #2175, opened 2020-04-08
Support multiple swizzled out operands

cnorthrop:
Swizzled out operands were added in bbbd9a2a. This was sufficient
for most tests, but we ran into problems with umulExtended and
imulExtended, which have two.

This CL converts the tracking values to vectors so multiple operands
can be supported.

Test: KHR-GLES31.core.shader_bitfield_operation.*


johnkslang:
We can't see what effect this has without test cases added.  These also prevent regression when future changes are made.

SPIRV/GlslangToSpv.cpp
```diff
@@ -3072,9 +3072,11 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
         }
         if (invertedType != spv::NoResult)
             result = createInvertedSwizzle(precision, *glslangOperands[0]->getAsBinaryNode(), result);
-        else if (temporaryLvalue != spv::NoResult) {
-            builder.setAccessChain(complexLvalue);
-            builder.accessChainStore(builder.createLoad(temporaryLvalue));
+        else if (!temporaryLvalues.empty()) {
+            for (unsigned int i = 0; i < temporaryLvalues.size(); ++i) {
```
johnkslang, line 3076: I see there was some existing incorrect assumption about what was 1:1, because of the scalar nature of `invertedType`, which does itself assume there is only one such argument.

This makes it a bit worse though, in generalizing to handle multiple non-invertedType but swizzled types, yet still making it keyed of the scalar inverted type.  It is the same true-but-coincidental pairing, but now reads more strangely.

I think it works better to decouple these.  It is inherent in `invertedType` that you can't have more than one.

So, instead of 

```
    else /* of a non-sensical complement of invertedType */ if (!temporaryLvalues.empty() /* redundant test */) {
        for (unsigned int i = 0; i < temporaryLvalues.size(); ++i) {
```

Just say

```
    // no else
    for (unsigned int i = 0; i < temporaryLvalues.size(); ++i) {
```

SPIRV/GlslangToSpv.cpp
```diff
@@ -2972,10 +2972,10 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
                 // reduce to a simple access chain.  So, we need a temporary vector to
                 // receive the result, and must later swizzle that into the original
                 // l-value.
-                complexLvalue = builder.getAccessChain();
-                temporaryLvalue = builder.createVariable(spv::StorageClassFunction,
-                    builder.accessChainGetInferredType(), "swizzleTemp");
-                operands.push_back(temporaryLvalue);
+                complexLvalues.push_back(builder.getAccessChain());
```
johnkslang, line 2975: Note, in contrast to the note below, here, lvalues are either inverted or needing this swizzle temporary, and `invertedType` (especially if generalized) is set each time through the loop.  Below, it is just remembering the one answer from here, regardless of how many arguments of each type exist.

SPIRV/GlslangToSpv.cpp
cnorthrop: Okay, made the recommended change.

cnorthrop:
Okay, added a test for umulExtended and updated the golden.  Hard to tell by inspection if the spir-v is still good, but the dEQP test still passes.  Let me know if you wanted more cases than that.

johnkslang:
I hand verified the SPIR-V is correct.

Easier to review if:

- first add the test in one commit
- second update the code in another commit

Then, the only difference in the test result is the difference caused by the code change.

(Also, swizzle tests should probably swizzle; the right shuffle was there, but 'xyz' is a close to a degenerate swizzle.)

PR #2182, opened 2020-04-16
Add googletest as submodule instead of manually cloning

D3PSI:
Title pretty much says it. You can now clone googletest by cloning the repository recursively. If you do not intend to use googletest you can simply omit the `--recursive`-flag in the `git clone` command

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -47,7 +47,7 @@ matrix:
 install:
   - C:/Python27/python.exe update_glslang_sources.py
   - set PATH=C:\ninja;C:\Python36;%PATH%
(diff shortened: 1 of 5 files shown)


johnkslang:
@dneto0 do you know of any issues this would create?

johnkslang:
I am getting conflicting input on whether on not to switch to submodules.  It seems reasonably balanced WRT advantages and disadvantages.  Given that state of the moment, I think making people change how this works might not be worthwhile, but am open to deeper understanding of what this buys and costs.

johnkslang:
Closing, until someone can make a clear case for cost versus benefit of making this change, which those that oppose it can respond to.

PR #2184, opened 2020-04-16
Add support for extension GL_ARB_shader_storage_buffer_object.

pmistryNV:
Enable below features for GL Core version 420:
* layout qualifier "std430"
* storage qualifier "buffer"
* atomic memory functions

### DIFF
--- a/glslang/MachineIndependent/Scan.cpp
+++ b/glslang/MachineIndependent/Scan.cpp
@@ -917,7 +917,8 @@ int TScanContext::tokenizeIdentifier()
     case BUFFER:
         afterBuffer = true;
         if ((parseContext.isEsProfile() && parseContext.version < 310) ||
-            (!parseContext.isEsProfile() && parseContext.version < 430))
+            (!parseContext.isEsProfile() && (parseContext.version < 430 &&
+            !parseContext.extensionTurnedOn(E_GL_ARB_shader_storage_buffer_object))))
             return identifierOrType();
         return keyword;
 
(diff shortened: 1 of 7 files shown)


pmistryNV:
@johnkslang kindly review the changes.

glslang/MachineIndependent/Scan.cpp
```diff
@@ -917,7 +917,7 @@ int TScanContext::tokenizeIdentifier()
     case BUFFER:
         afterBuffer = true;
         if ((parseContext.isEsProfile() && parseContext.version < 310) ||
-            (!parseContext.isEsProfile() && parseContext.version < 430))
+            (!parseContext.isEsProfile() && parseContext.version < 420))
```
johnkslang, line 920: Is this incorrectly classifying something as a keyword even if the extension is not enabled?

pmistryNV: This extensions makes the BUFFER token legal for version 420. However at this point we cannot determine if the extension is enabled. This is determined later in TParseContext::setLayoutQualifier.

johnkslang: You can test that the extension is enabled, right here, before deciding it is a keyword.  You can see some other examples of this in that neighborhood.  Look for `parseContext.extensionTurnedOn(...`.  Best to do it after the simple testing, as it is a bit expensive.

pmistryNV: Sure I can add it, but this would also mean compiler will not proceed further and flag errors beyond it.

johnkslang: I don't understand.  Should buffer be a keyword or not depending on whether the extension is enabled?  Why would that give the wrong result?

pmistryNV: Yeah I understood what you meant. I have updated the patch.

PR #2185, opened 2020-04-16
Add support for extension GL_ARB_shader_image_size

pmistryNV:
Based on #2183 which should be merged first. And then the file 150.frag.out needs to be updated, before merging the change.

### DIFF
--- a/Test/420.frag
+++ b/Test/420.frag
@@ -12,3 +12,8 @@ layout(depth_less) in float depth; // ERROR: depth_less only applies to gl_FragD
 layout(depth_any) out float gl_FragDepth;  // ERROR, done after use
 
 layout(binding=0) uniform atomic_uint a[];
+
+uniform writeonly image2D      i2D;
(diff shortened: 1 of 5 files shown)


Test/420.frag
```diff
@@ -12,3 +12,7 @@ layout(depth_less) in float depth; // ERROR: depth_less only applies to gl_FragD
 layout(depth_any) out float gl_FragDepth;  // ERROR, done after use
 
 layout(binding=0) uniform atomic_uint a[];
+
+#extension GL_ARB_shader_image_size : enable
+uniform writeonly image2D      i2D;
+ivec2 t = imageSize(i2D);
```
johnkslang, line 18: Can we see this both before and after the extension is enabled, to ensure we still get the correct error message for version 420?

pmistryNV: Updated the test by calling imageSize before and after enabling GL_ARB_image_size extenson

PR #2187, opened 2020-04-20
Force SPIRV and HLSL projects to be static libs

Joshua-Ashton:
These projects are dependent on glslang which is dependent on these projects.
One of these needs to be static to avoid external linker errors when BUILD_SHARED_LIBS = ON

This PR reverses the dependency order for the SPIRV and glslang libraries, and makes the SPIRV and HLSL projects static.

This allows glslang to compile as a shared library and fixes compiling with BUILD_SHARED_LIBS = ON if pulled in as an external subdir in CMake.


johnkslang:
Why isn't this working?

> set(LIB_TYPE STATIC)

Joshua-Ashton:
Because:

```
set(LIB_TYPE STATIC)

if(BUILD_SHARED_LIBS)
    set(LIB_TYPE SHARED)
endif()```

and `BUILD_SHARED_LIBS` = ON

**Comment by johnkslang** [comment]
Right, okay.

Is it intentional to leave out this one?

https://github.com/KhronosGroup/glslang/blob/c9b28b9f3388c607ea5506f5e6197b7150238ad3/SPIRV/CMakeLists.txt#L45

**Comment by Joshua-Ashton** [comment]
Yes, because that's SPVRemap, that's not in this funny dependency mess and can be built however the user wants.

File: SPIRV/CMakeLists.txt [hunk 1]
```diff
@@ -59,12 +58,12 @@ if(ENABLE_OPT)
...
         PRIVATE ${spirv-tools_SOURCE_DIR}/source
     )
-    target_link_libraries(SPIRV glslang SPIRV-Tools-opt)
+    target_link_libraries(SPIRV SPIRV-Tools-opt)
     target_include_directories(SPIRV PUBLIC
 		$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External>
 		$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External>)
 else()
-    target_link_libraries(SPIRV glslang)
+    target_link_libraries(SPIRV)
```
SpaceIm, line 66: But SPIRV depends on glslang, this modification hides dependencies which actually exist.
glslang also depends on SPIRV since 8.13.3743 because compilation unit of C implementation was added in glslsang lib, but I think it's a mistake.

Joshua-Ashton: They are co-dependent, and SPIRV is statically linked here, therefore it should be the one listed without the dependency.

(actually I think this line should go all together)

Joshua-Ashton:
Bumping this.

johnkslang:
@ben-clayton has recently fixed many things around the library space.  Does that interact with the need to make this change?

Note it contains conflicts, so can't be merged right now, and those conflicts might be clues to the question above.

ben-clayton:
Shared objects should be buildable now, but there is a known ODR violation that may prevent the shared objects from being usable. I'm on vacation at the moment, but will look into fixing this when I return.

arcady-lunarg:
Closing this due to staleness, but we do need to do something like this in the long run.

PR #2199, opened 2020-04-27
Add SPIR-V capabilities needed for spec constants

### LINKED ISSUE #2198: glslang fails to include needed SPIR-V capabilities for spec constants
When creating specialization constants using non-basic types, glslang fails to include the needed capabilities in the generated SPIR-V output, resulting in the SPIR-V validator complaining about the missing capabilities if the program is verified with spirv-val, as used in VK-GL-CTS.

Example:
```
#version 450
#extension GL_EXT_shader_explicit_arithmetic_types_int16 : require

layout(constant_id = 1) const int16_t sc0 = 20000s;
layout(constant_id = 2) const int16_t sc1 = -20000s;

layout (set = 0, binding = 0, std430) writeonly buffer Output {
    int16_t r0;
    int16_t r1;
} sb_out;


void main (void)
{
    sb_out.r0 = sc0;
    sb_out.r1 = sc1;
}
```

Results in the following output:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 8
; Bound: 18
; Schema: 0
               OpCapability Shader
               OpCapability StorageBuffer16BitAccess
               OpExtension "SPV_KHR_16bit_storage"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 1
               OpSource GLSL 450
               OpSourceExtension "GL_EXT_shader_explicit_arithmetic_types_int16"
               OpName %main "main"
               OpName %Output "Output"
               OpMemberName %Output 0 "r0"
               OpMemberName %Output 1 "r1"
               OpName %sb_out "sb_out"
               OpName %sc0 "sc0"
               OpName %sc1 "sc1"
               OpMemberDecorate %Output 0 NonReadable
               OpMemberDecorate %Output 0 Offset 0
               OpMemberDecorate %Output 1 NonReadable
               OpMemberDecorate %Output 1 Offset 2
               OpDecorate %Output BufferBlock
               OpDecorate %sb_out DescriptorSet 0
               OpDecorate %sb_out Binding 0
               OpDecorate %sc0 SpecId 1
               OpDecorate %sc1 SpecId 2
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %short = OpTypeInt 16 1
     %Output = OpTypeStruct %short %short
%_ptr_Uniform_Output = OpTypePointer Uniform %Output
     %sb_out = OpVariable %_ptr_Uniform_Output Uniform
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
        %sc0 = OpSpecConstant %short 20000
%_ptr_Uniform_short = OpTypePointer Uniform %short
      %int_1 = OpConstant %int 1
        %sc1 = OpSpecConstant %short -20000
          %5 = OpLabel
         %14 = OpAccessChain %_ptr_Uniform_short %sb_out %int_0
               OpStore %14 %sc0
         %17 = OpAccessChain %_ptr_Uniform_short %sb_out %int_1
               OpStore %17 %sc1
               OpReturn
               OpFunctionEnd
```

Which spirv-val complains about:

```
error: line 34: Cannot form constants of 8- or 16-bit types
  %sc0 = OpSpecConstant %short 20000
```

rg3igalia:
Fixes #2198.


johnkslang:
Could you also include a test result change?  Likely a new test added. Maybe adding to or adding another `spv.specConst....` file in `glslang/Test`.

johnkslang:
Also, I'm not clear yet on whether this is better done this way and/or why  SpvPostProcess.cpp wasn't picking it up, or couldn't easily catch it.

rg3igalia:
> Could you also include a test result change? Likely a new test added. Maybe adding to or adding another `spv.specConst....` file in `glslang/Test`.

Will do.

> Also, I'm not clear yet on whether this is better done this way and/or why SpvPostProcess.cpp wasn't picking it up, or couldn't easily catch it.

First of all, this is my first real contact with the glslang sources, so take anything I say with a grain of salt. I believe the checks I added need to stay there in some form in the sense that the SPIR-V specification for Int16, for example, says the capability is needed to use OpTypeInt to declare 16-bit integers, in general, so by merely using `%short = OpTypeInt 16 1` glslang should declare the capability. The same goes for other type capabilities.

The question is "why does glslang not add the capabilities when it generates the type instruction?" Indeed, in SpvBuilder.cpp, makeIntegerType adds the Int64 capability right there, whereas for the 8-bit and 16-bit cases it says "these are currently handled by storage-type declarations and post processing". The same goes for makeFloatType. Indeed, those files were modified by 31aa3d6019945f9efdddcec05a5cbda22c983b6a, but I lack context for those changes. I don't know if they were trying to fix some other issue. In my opinion, the original lines at least in those two methods looked good and followed the SPIR-V spec.

The reason why SpvPostProcess is not catching those, for the provided example, is because it's going through postProcessType, case OpStore, reaches width == 16 and then the storage class is StorageClassUniform and it hits the break without generating the basic capabilities.

glslang correctly declares the StorageBuffer16BitAccess capability but I believe storage capabilities and type capabilities to be somewhat independent. The storage capabilities cannot replace the type capabilities. You may need 16-bit access to a storage buffer because probably you're using 16-bit integers or floats, but you need to know which and declare the specific type capability in advance, IMHO.

Restoring the lines in makeIntegerType and makeFloatType also fixes the issue for me. Do you prefer that solution?

rg3igalia:
I stand corrected: the capabilities StorageBuffer16BitAccess, UniformAndStorageBuffer16BitAccess, StorageBuffer8BitAccess and UniformAndStorageBuffer8BitAccess allow using 16-bit and 8-bit types by themselves.

The reason why the postprocessing isn't catching these cases, then, is because the OpSpecConstant operation does not reside in a function block, as per here:

https://github.com/KhronosGroup/glslang/blob/e43e8ba9f385edc4a7f3255635f8fdb69a06de32/SPIRV/SpvPostProcess.cpp#L407

I will be updating this PR with new tests.

PR #2210, opened 2020-05-08
Explicitly mark EShLanguageMask as unsigned

Flakebi:
This allows casting from and to any unsigned value, previously this was
undefined behavior.

This fixes ubsan complaining in `TParseContext::layoutQualifierCheck`,
where `~EShLangComputeMask` is used.

---
This change is [Reviewable](https://reviewable.io/reviews/khronosgroup/glslang/2210)

### DIFF
--- a/glslang/MachineIndependent/Versions.h
+++ b/glslang/MachineIndependent/Versions.h
@@ -52,7 +52,7 @@
 // Don't maintain an ordinal set of enums (0,1,2,3...) to avoid all possible
 // defects from mixing the two different forms.
 //
-typedef enum {
+typedef enum : unsigned {
     EBadProfile           = 0,
     ENoProfile            = (1 << 0), // only for desktop, before profiles showed up
     ECoreProfile          = (1 << 1),
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -109,7 +109,7 @@ typedef enum {
     LAST_ELEMENT_MARKER(EShLangCount),
 } EShLanguage;         // would be better as stage, but this is ancient now
 
-typedef enum {
+typedef enum : unsigned {
     EShLangVertexMask         = (1 << EShLangVertex),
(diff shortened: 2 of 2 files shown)


johnkslang:
Some platforms were previously not support that syntax.  Maybe that's fixed now.

Currently, no `typedef enum` in glslang or in the SPIRV-Tools it includes have the `typedef enum : <type>` syntax, so it seems a bit strange to add it to just one enum.

Is this really the only one in this bucket?

johnkslang:
Also, I think this is not about signed/unsigned, just about knowing the width, yes?

Flakebi:
Thanks for taking a look @johnkslang!

> Also, I think this is not about signed/unsigned, just about knowing the width, yes?

Yes, this is only about the width, the current type should be deduce to `int` if I understand the logic behind enums right, so `: unsigned` will change that. I thought `unsigned` makes more sense for this enum but it shouldn’t matter.

> Some platforms were previously not support that syntax. Maybe that's fixed now.

As far as I read it [here](https://en.cppreference.com/w/cpp/language/enum), specifying a base should be supported by C++11?

> Is this really the only one in this bucket?

This one popped up in some tests, that’s why I changed this specific enum, I’ll have a look at others.

Flakebi:
> > Is this really the only one in this bucket?
>
> This one popped up in some tests, that’s why I changed this specific enum, I’ll have a look at others.

I coarsely looked at the other enums in `ShaderLang.h` and they don’t seem to have this problem. The part that makes `EShLanguageMask` special is that it is a mask and is negated sometimes, which converts it to an int that is not inside the enum range.

johnkslang:
Off the top of my head, I wonder what's going on differently with all the `~EEsProfile`.

Flakebi:
I missed that one, I’ll add unsigned there too.

Flakebi:
Done, I also added `unsigned` to `EShMessages` because `~EShMsgReadHlsl` is used in one place.

Flakebi:
Thanks!

PR #2215, opened 2020-05-13
Update tests according to spirv-opt update

jaebaek:
We refactored function inlining pass of spirv-opt and it results in
different numbering of result ids in SPIR-V code. This commit updates
test cases to avoid a test failure according to the spirv-opt update.


jaebaek:
This change is related to https://github.com/KhronosGroup/SPIRV-Tools/pull/3328.
Since we did not submit the PR, HLSL legalization tests of glslang will return in failures.

I prepared this PR to show why shaderc test of https://github.com/KhronosGroup/SPIRV-Tools/pull/3328 fails.
Once the spirv-tools PR will be submitted, I will update the SPIRV-Tools submodule of glslang in this PR.

johnkslang:
Glslang gets the correct results relative to its own declaration of known good SPIRV-Tools, as per the script `./update_glslang_sources.py`.  What matters is that these are consistent.  If you use a different version, you might get different test results, but that doesn't mean anything is incorrect.

If it is important to change glslang's known good, then that should be done at the same time as updating the test results, so that they stay consistent.

Anticipating a bit here, the generated code seems somewhat different, which might affect things downstream, and potentially means we need to bump up glslang's code generation version number.

Can you briefly say what the nature of the code-gen changes are?

PR #2220, opened 2020-05-15
Add support for primitive culling layout qualifier.

alelenv:
Fixes https://gitlab.khronos.org/vulkan/vulkan/issues/2073

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -5173,6 +5173,12 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
             }
         }
     }
+
+    if (id == "primitive_culling") {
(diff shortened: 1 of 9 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -8368,6 +8374,12 @@ void TParseContext::updateStandaloneQualifierDefaults(const TSourceLoc& loc, con
     {
         checkIoArraysConsistency(loc);
     }
+
+    if (publicType.shaderQualifiers.layoutPrimitiveCulling) {
+        // Exit early as this qualifier has no default storage class
```
johnkslang, line 8379: Did you do the semantic checking for this rule?

What happens if a storage classes is included, etc.

alelenv: Good point. I missed that
Updated with error checks and a new negative test.

Issue #2221, opened 2020-05-16
HLSL Compute shader: Vulkan Validation Layers and SPIRV-Tools show Validation Error


Andreyogld3d:
Hi! 
I have error message from Vulkan Validation Layers:

```
[UNASSIGNED-CoreValidation-Shader-InconsistentSpirv] Code 7060244 : Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0xfd1cb80, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: OpFunctionCall Argument <id> '299[%299]'s type does not match Function <id> '55[%Frustum]'s parameter type.
  %302 = OpFunctionCall %bool %SphereInFrustum_struct_Frustum_vf4_6__i1_8_1_vf4_ %299 %301
```

also the glslang tool create invalid SPIR-V module. 
```
glslangValidator -S comp -e CullObjects -V -D GPUInstanceCulling.hlsl.comp --spirv-val
GPUInstanceCulling.hlsl.comp
error: SPIRV-Tools Validation Errors
error: Result type (OpTypeArray) does not match the type that results from indexing into the composite (OpTypeArray).
  %756 = OpCompositeExtract %_arr_v4float_uint_6 %476 0
```

Full shader code:

```
struct InstanceGPUData {
	float4 InstanceData[3];
};
typedef float4 BoundingSphere;
struct BoundingBox {
	float4 minPoint; 
	float4 maxPoint;
};
struct CullingObjInfo {
	BoundingBox boundingBox;
	float4 boundingSphere; // radius(float) + position(float3)
	InstanceGPUData instanceData;
	float4 lodDistance;
};
struct DrawIndexedIndirectCommand {
	uint indexCount;
	uint instanceCount;
	uint firstIndex;
	int  vertexOffset;
	uint firstInstance;
};

typedef float4 Plane;

static const uint	numPlanes = 6;

struct Frustum {
	Plane planes[numPlanes];
	int nearPointMasks[numPlanes + 2];
};

struct CameraInfo {
	Frustum frustum;
	float4 cameraPosition;
};

layout(binding = 0) RWByteAddressBuffer instanceVB : register(u0);
layout(binding = 1) RWByteAddressBuffer numObjects : register(u1);
layout(binding = 2) StructuredBuffer<CullingObjInfo> cullingObjInfo : register(t0);
[...]

Issue #2218, opened 2020-05-15
Provide option to skip default binding decoration, or a mechanism for tracking where the defaults have been set.


nitronoid:
Quoting from my previous comment in an older issue:

> "The tool I'm working on is feeding ESSL through and producing SPIRV, with the intention to patch any blocks which haven't been given explicit bindings by generating them ourselves. This becomes difficult with the changes in the PR you referenced as a binding of 0 is perfectly valid and offers no way of knowing that this uniform wasn't given an explicit binding.
> 
> I'm wondering if you have any suggestions on how I could work around this problem? Currently I have subclassed `TGlslIoMapper` to modify the binding values when not explicitly provided to a sentinel value which we check for later, i.e.
> 
> `if (!qualifier.hasBinding()) qualifier.layoutBinding = sentinel;`
> But this feels like a hack and may be fragile.
> 
> Perhaps instead of just setting the binding to zero, these uniforms could also be tracked by the shader so we can at least tell where default values have been assigned?"

It was suggested that the option `--shift-ssbo-binding` could be of use. Does this option apply to all ssbos or just the ones that have been given the default binding of `0`. In any case, this is a possible way to recreate the work around I am already using, when the ideal case would be to avoid decorating them with a binding.

Roy-AMD:
See StandAlone.cpp line: 1115 `if (! Program.mapIO ())` If you want use TGlslIoMapper you should use TGlslIoMapper and TDefaultGlslIoResolver as the function (mapIO) parameter, it's beta code, we only use it in Opengl driver. so glslangValidator won't be effect by this class and this rule, If you want try, you can modify the code in local. When we finished develop plan and the test case is done, we will add it to glslangValidator and use new option to control it.

nitronoid:
To be clear, what I really want is for **_no_** IO mapping to be performed.
Simply compiling the glsl shader gives me very close to what I want, except for the `OpDecoration Binding 0` being applied to everything.
The only reason I am using an `IoMapper` at all is to set the binding value for anything which doesn't have one in advance to prevent it from being defaulted to zero.

This is the code I am using:
```cpp
class DummyMapper : public glslang::TGlslIoMapper
{
public:
	DummyMapper (EShLanguage stage) : glslang::TGlslIoMapper{}, stage{ stage } {}
	EShLanguage stage;

	virtual bool doMap(glslang::TIoMapResolver* resolver, TInfoSink& info) override
	{
[...]

PR #2224, opened 2020-05-18
Add an spirv option to control default binding

Roy-AMD:
Turn off this option can help use generate a no-default binding spirv.
default binding should be generate in iomapper. so if we not do io mapper auto binding, we should get a spirv code without auto binding.
related issue https://github.com/KhronosGroup/glslang/issues/2221

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -8121,7 +8121,11 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
         builder.addDecoration(id, spv::DecorationBinding, symbol->getQualifier().layoutBinding);
     else if (IsDescriptorResource(symbol->getType())) {
         // default to 0
-        builder.addDecoration(id, spv::DecorationBinding, 0);
+        // If default is disable, binding to -1
+        if (options.defaultBinding)
+            builder.addDecoration(id, spv::DecorationBinding, 0);
+        else
+            builder.addDecoration(id, spv::DecorationBinding, -1, true);
     }
     if (symbol->getQualifier().hasAttachment())
         builder.addDecoration(id, spv::DecorationInputAttachmentIndex, symbol->getQualifier().layoutAttachment);
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -1202,7 +1202,7 @@ void Builder::addMemberName(Id id, int memberNumber, const char* string)
     names.push_back(std::unique_ptr<Instruction>(name));
 }
 
-void Builder::addDecoration(Id id, Decoration decoration, int num)
+void Builder::addDecoration(Id id, Decoration decoration, int num, bool allowIllegal)
 {
     if (decoration == spv::DecorationMax)
         return;
@@ -1210,7 +1210,7 @@ void Builder::addDecoration(Id id, Decoration decoration, int num)
     Instruction* dec = new Instruction(OpDecorate);
     dec->addIdOperand(id);
     dec->addImmediateOperand(decoration);
-    if (num >= 0)
+    if (allowIllegal || num >= 0)
         dec->addImmediateOperand(num);
 
     decorations.push_back(std::unique_ptr<Instruction>(dec));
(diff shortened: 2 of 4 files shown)


SPIRV/GlslangToSpv.cpp
```diff
@@ -8121,7 +8121,11 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
         builder.addDecoration(id, spv::DecorationBinding, symbol->getQualifier().layoutBinding);
     else if (IsDescriptorResource(symbol->getType())) {
         // default to 0
-        builder.addDecoration(id, spv::DecorationBinding, 0);
+        // If default is disable, binding to -1
+        if (options.defaultBinding)
+            builder.addDecoration(id, spv::DecorationBinding, 0);
+        else
+            builder.addDecoration(id, spv::DecorationBinding, -1);
```
johnkslang, line 8128: So, `addDecoration()` will not add the number (it skips negative values), leaving illegal SPIR-V.  Is that really the design we want?  It is counting on a future iomapping to correct it?

Roy-AMD: > So, `addDecoration()` will not add the number (it skips negative values), leaving illegal SPIR-V. Is that really the design we want? It is counting on a future iomapping to correct it?

I guess they may want to patch spirv binary, But instead of adding a default binding in the iomapper, it is battle to adding a glslangToSpirv option.

https://github.com/KhronosGroup/glslang/issues/2218 
We can ask @nitronoid and work out a batter solution.

nitronoid: Hi @johnkslang, not sure if you saw the [issue I created for this topic](https://github.com/KhronosGroup/glslang/issues/2218).
@Roy-AMD  is correct, I want to patch the SPIR-V with the correct binding after the call to `GlslangToSpv`. By setting any omitted bindings to zero, it becomes difficult to identify the decorations which have been added by `GlslangToSpv`, especially as a binding of 0 is very common.

I am using a work around to mark any unspecified bindings, as detailed in the issue, however this PR would simplify our process and be more robust.

Perhaps the option can be renamed to highlight the fact it will produce illegal SPIR-V?

Roy-AMD: > So, `addDecoration()` will not add the number (it skips negative values), leaving illegal SPIR-V. Is that really the design we want? It is counting on a future iomapping to correct it?

I'm not sure is it suitable to remove condition `if (num >= 0)` direct, or we need add a build option for it.

johnkslang: It seems the above is trying to transfer AST information into illegal SPIR-V, but note that `mapIO()` allows custom operations and runs before making SPIR-V, is it possible to take advantage of that (which would actually be using the system as intended)?

nitronoid: Unfortunately, the problem is that the binding is not known at the time of compilation. We perform our own linking phase, where the SPIR-V gets patched and the missing decorations required for linking get generated.

This is why I am using `mapIO` to mark these bindings so they may be identified later.

To perform the assignment of bindings through `mapIO()` would be somewhat limiting in our case. We would need to shift to using your linking mechanisms, and we would only receive a single SPIR-V shader representing the whole program which would misrepresent the input which we actually received, as well as adding complications when trying to reuse shaders.

Additionally, I think it's reasonable for incomplete SPIR-V to be generated from incomplete GLSL, rather than making up information.

johnkslang: Would it help if these were specialization constants that could be modified later?

nitronoid: Hi John,
Please could you show an example of the resulting SPIR-V? It's not clear to me how using specialization constants would help things in this case?

nitronoid: I think I see what you're getting at with specialization constants. The main issue isn't the modification of these binding values (which can be done by patching the shader itself). What we're having issues with is being able to identify which ones we will need to modify, and we need to know this post compilation of the shader.

Since using a spec constant for the binding is legal, I don't think we can use that fact alone to select the auto-generated bindings.

As an alternative to omitting the default binding decoration, is there any way you could mark these bindings as being auto-generated with legal SPIR-V?

PR #2225, opened 2020-05-18
Binding checks following specs

ShchchowAMD:
1.Add binding error checks for
    1.1.atomic_uint,
     1.2.uniform buffer block
     1.3.storage buffer block.

2.Add warn for unsized array binding check.

3. add error check "It is a compile-time error to apply location layout qualifier to atomic counter."

4. fix last binding calculation, add binding checks for gl_MaxImageUnits and gl_MaxCombinedTextureImageUnits


ShchchowAMD:
@johnkslang  Hi John, is this Okay to merge?

Test/baseResults/310.comp.out
```diff
@@ -86,7 +87,9 @@ WARNING: 0:238: '#define' : names containing consecutive underscores are reserve
 ERROR: 0:244: 'gl_DeviceIndex' : required extension not requested: GL_EXT_device_group
 ERROR: 0:245: 'gl_ViewIndex' : undeclared identifier 
 ERROR: 0:255: 'gl_ViewIndex' : undeclared identifier 
-ERROR: 85 compilation errors.  No code generated.
+ERROR: 0:258: 'binding' : buffer binding is too large; no more than  16384
```
johnkslang, line 90: 0 is a legal binding, is there an off-by-1 possibility here?

johnkslang: Should this refer to the symbol instead (or in addition to)?

See

```
void TParseContext::arrayLimitCheck(const TSourceLoc& loc, const TString& identifier, int size)
void TParseContext::limitCheck(const TSourceLoc& loc, int value, const char* limit, const char* feature)
```

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5741,6 +5741,10 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
         case EvqBuffer:
             if (type.getBasicType() == EbtBlock)
                 error(loc, "cannot apply to uniform or buffer block", "location", "");
+
+            // "It is a compile-time error to apply location layout qualifier to atomic counter."
```
johnkslang, line 5745: Should be "to an atomic counter".

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5741,6 +5741,10 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
...
+            else if (type.getBasicType() == EbtAtomicUint)
+                error(loc, "cannot apply to atomic counter", "location", "");
```
johnkslang, line 5747: "...to an atomic..."

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5801,55 +5805,74 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
...
-        if (type.getBasicType() == EbtSampler) {
-            int lastBinding = qualifier.layoutBinding;
-            if (type.isArray()) {
-                if (spvVersion.vulkan > 0)
-                    lastBinding += 1;
+        unsigned int lastBinding = qualifier.layoutBinding;
+        if (type.isArray()) {
+            if (spvVersion.vulkan == 0 && !type.isAtomic()) {
+                if (type.isSizedArray())
+                    lastBinding += type.getCumulativeArraySize() - 1;
```
johnkslang, line 5812: Why ` - 1`?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5801,55 +5805,74 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
...
+            else {
+                if (spvVersion.vulkan == 0 && lastBinding >= (unsigned int)resources.maxCombinedTextureImageUnits)
+                    error(loc, "sampler binding not less than gl_MaxCombinedTextureImageUnits", "binding", type.isArray() ? "(using array)" : "");
+            }
 #endif
         }
         if (type.isAtomic()) {
-            if (qualifier.layoutBinding >= (unsigned int)resources.maxAtomicCounterBindings) {
+            if (lastBinding >= (unsigned int)resources.maxAtomicCounterBindings) {
                 error(loc, "atomic_uint binding is too large; see gl_MaxAtomicCounterBindings", "binding", "");
```
johnkslang, line 5837: Maybe all of these should use the limit functions mentioned above.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -5801,55 +5805,74 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
...
                 return;
             }
         }
-    } else if (!intermediate.getAutoMapBindings()) {
-        // some types require bindings
 
+        if (type.getBasicType() == EbtBlock) {
+            if (qualifier.storage == EvqBuffer) {
+                if (lastBinding >= (unsigned int)resources.maxStorageBufferBindings) {
+                    error(loc, "buffer binding is too large; no more than ", "binding", "%d", resources.maxStorageBufferBindings);
```
johnkslang, line 5845: A lot of this seems like just rearranging, but without knowing, this will take some care reviewing.

Maybe I need a smarter differ.

glslang/MachineIndependent/ParseHelper.cpp
ShchchowAMD: lastBinding means an index starting from 0.
getCumulativeArraySize() means size value starting from 1.

glslang/MachineIndependent/ParseHelper.cpp
ShchchowAMD: Those kinds of checks are found in failure-expected CTS when checking binding with a super huge value.

ShchchowAMD:
1.  Fix error message for max UBO and SSBO bindings check.
There's no built-in constant for gl_MaxShaderStorageBufferBindings and gl_MaxUniformBufferBindings (only exists in core spec for api check).

Can we add both of them? Otherwise I could only use resource limits to check here.

2. Fix error in message and comments for atomic counter.

3.  Use limit check for other resources limits.
According to spec, some limitation indicate as "no more than or equal to", which for others are "no more than",
so add a flag to control different type of limit checks and related error msg.

Test/310.comp
```diff
@@ -254,3 +254,10 @@ void devie()
     gl_DeviceIndex;
     gl_ViewIndex;   // ERROR, never this stage
 }
+
+layout(binding=16384, std430) buffer buffer_block_bindingMax_block_bindingMax {
+    float bufferMax_a[2];
+} ; // ERROR
+layout(binding=16384, std140) uniform uniform_block_bindingMax {
+    float uniformMax_a[2];
+} ; // ERROR
```
johnkslang, line 263: Missing final newline.

johnkslang:
I think this looks okay, except for the interface change, which needs the version, though we've not always been doing that.

Even with a version change, consumers are not always realizing their resource structure is out of date, and I'm wondering how to address that.  Possibly putting new members at the end helps; I'm not sure.

Note interaction between changing the version and #2277.

The currently method gets the top/first version number from here:

```
{
    // return 1; // start
}
```

ShchchowAMD:
> I think this looks okay, except for the interface change, which needs the version, though we've not always been doing that.
> Even with a version change, consumers are not always realizing their resource structure is out of date, and I'm wondering how to address that.  Possibly putting new members at the end helps; I'm not sure.
> Note interaction between changing the version and #2277.
> The currently method gets the top/first version number from here:
> // For low-order part of the generator's magic number. Bump up
> // when there is a change in the style (e.g., if SSA form changes,
> // or a different instruction sequence to do something gets used).
> int GetSpirvGeneratorVersion()
> {
>     // return 1; // start
>     // return 2; // EOpAtomicCounterDecrement gets a post decrement, to map between GLSL -> SPIR-V
>     // return 3; // change/correct barrier-instruction operands, to match memory model group decisions
>     // return 4; // some deeper access chains: for dynamic vector component, and local Boolean component
>     // return 5; // make OpArrayLength result type be an int with signedness of 0
>     // return 6; // revert version 5 change, which makes a different (new) kind of incorrect code,
>                  // versions 4 and 6 each generate OpArrayLength as it has long been done
>     // return 7; // GLSL volatile keyword maps to both SPIR-V decorations Volatile and Coherent
>     // return 8; // switch to new dead block eliminator; use OpUnreachable
>     return 9; // don't include opaque function parameters in OpEntryPoint global's operand list
> }

Agree with version changes. I think it depend on when would change #2277 be merged. It is better to wait if we need to do so.

Besides, I don't make sure whether a new version number should be added into above generator.
Seems this change needs version change at least in MINOR_VERSION for GLSL, but it doesn't make influence to SPIR-V (mostly are built-in limits and an interface change).

ShchchowAMD:
@johnkslang  Hi John.

If I need to declare a change in interface (limitCheck), should I just modify CHANGE.md, adding a new version by modifying last number as following?

## 11.0.0 2020-07-20

### Function interface of limitCheck changes

#### Add default parameter to enable equal check

At the end of function parameter list, add an default value as false, to enable/disable equal check during limitCheck.

johnkslang:
I think we only need to do this if it's possible to break something by the change.

PR #2226, opened 2020-05-19
Add Vertex input attrib check and layout Type Check

ShchchowAMD:
Add related constant:
1. gl_MaxVaryingVectors

Add following checks:
1. vertex varying in input attribs check
2. symbol location check should not be null
3.  Comma op should be folded as aggregate,
otherwise, constant initialization would fail
see new case at the end of 100.frag


johnkslang:
This needs a bit more explanation.

The specification says:

> However, the lowest precedence operators of the sequence operator (,) and the
assignment operators (=, +=, …) are not included in the operators that can create a constant
expression.

Are you are trying to undo this rule?

ShchchowAMD:
Hi John,

I re-check the spec and I think you are right.

From es spec,
Constant expression excludes comma from version > 100 (as sequence operator).

This rule was introduced with version > 100,

So I add version check and also test cases now.

johnkslang:
No, it was fixed in the spec because the spec was wrong, not because version 100 should actually support the comma.

ShchchowAMD:
Thanks for the explanation.
I've removed comma part and test in version 100.

Now we will treat this rule as applying under all versions.

ShchchowAMD:
Hi John, is there any further suggestions for this change please?

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2450,7 +2450,7 @@ TIntermTyped* TIntermediate::addComma(TIntermTyped* left, TIntermTyped* right, c
     commaAggregate->setType(right->getType());
     commaAggregate->getWritableType().getQualifier().makeTemporary();
 
-    return commaAggregate;
+    return commaAggregate->getAsAggregate();
```
johnkslang, line 2453: Is this change still needed?  It seems to add something redundant.

glslang/MachineIndependent/Constant.cpp
```diff
@@ -1047,6 +1047,8 @@ TIntermTyped* TIntermediate::fold(TIntermAggregate* aggrNode)
         objectSize = std::max(children[0]->getAsTyped()->getType().getVectorSize(),
                               children[2]->getAsTyped()->getType().getVectorSize());
         break;
+    case EOpComma:
+        return children.back()->getAsTyped();
```
johnkslang, line 1051: Should this still be included?

ShchchowAMD:
Close current version, comma case needs to be removed with further support.
Will be moved to another new pr as this has been placed in PR list too long.

PR #2227, opened 2020-05-19
Code refine.

Roy-AMD:
Encapsulate repeated code into functions to reduce code length.
Use a more reliable way to obtain the access name.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -579,10 +579,7 @@ TDefaultGlslIoResolver::TDefaultGlslIoResolver(const TIntermediate& intermediate
 
 int TDefaultGlslIoResolver::resolveInOutLocation(EShLanguage stage, TVarEntryInfo& ent) {
     const TType& type = ent.symbol->getType();
-    const TString& name = IsAnonymous(ent.symbol->getName()) ?
-                            ent.symbol->getType().getTypeName()
-                            :
-                            ent.symbol->getName();
+    TString& name = getAccessName(ent.symbol);
     if (currentStage != stage) {
         preStage = currentStage;
         currentStage = stage;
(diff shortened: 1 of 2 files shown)


johnkslang:
Good reduction of replicated code.

glslang/MachineIndependent/iomapper.cpp
```diff
@@ -907,6 +892,17 @@ void TDefaultGlslIoResolver::reserverResourceSlot(TVarEntryInfo& ent, TInfoSink&
     }
 }
 
+TString& TDefaultGlslIoResolver::getAccessName(const TIntermSymbol* symbol)
+{
+    TString name;
+    if (symbol->getBasicType() == EbtBlock) {
```
johnkslang, line 898: The test is changing from `IsAnonymous` to whether the basic type is a block.  Are these equivalent?

Roy-AMD: > The test is changing from `IsAnonymous` to whether the basic type is a block. Are these equivalent?

In a sense, it is not equivalent. When a block is named, the new code will still get the type name, but this is exactly what we expected. GlslIoMapper has not completely stabilized. We are still looking for bugs. When we get a stable version and we will provide test code for it. But now these codes may still be changed more frequently.

Andreyogld3d: @Roy-AMD , @johnkslang I have warning during compilation:

Visual Studio 2017:
glslang\glslang\machineindependent\iomapper.cpp(903): warning C4172: returning address of local variable or temporary: name

johnkslang: Oh, right. This is not good code.

Probably, it should directly return the references it gets, or declare name to be a reference.  I'll look.

johnkslang: I made it simply be;

```
    return symbol->getBasicType() == EbtBlock ?
        symbol->getType().getTypeName() :
        symbol->getName();
```

And also fixed `const` correctness.

PR #2228, opened 2020-05-19
Add missing standard library header include

zoddicus:


zoddicus:
This header missing, specifically the definition of uint32_t missing, is causing one of Windows bots for shaderc to fail when I try to roll DEPS ahead.

FYI -> https://source.cloud.google.com/results/invocations/9af3869e-507c-4464-a040-c1133789d859/targets/graphics_shader_compiler%2Fshaderc%2Fwindows%2Fpresubmit_release_2013/log

johnkslang:
Yes, I found `uint32_t` to not be very portable (different compilers having different rules), so was just using normal `int` stuff, which is actually more portable, despite all efforts to make the situation the reverse.

I'll fix the file instead to avoid the portability pitfalls of `uint32_t`.

johnkslang:
Also note the loop bound this `uint32_t` was used for is actually a `size_t`, so it seems that the right fix is not to include that header to shut up the compiler, but to fix the type to be `size_t`.

PR #2234, opened 2020-05-20
Tests: fix some paths not automapping, and some shaders with no main.

johnkslang:


johnkslang:
@alan-baker this should fix validation failures regarding location and binding.

It adds those to about 244 tests.

alan-baker:
I've been testing this against https://github.com/KhronosGroup/SPIRV-Tools/pull/3308. I think there a couple issues with the validation and a couple with glslang tests. Mesh and task shader validation seems incorrect. Still investigating.

johnkslang:
This was done in #2307

PR #2237, opened 2020-05-20
EXT_ray_tracing requires spv1.4

### LINKED ISSUE #2236: EXT_ray_tracing requires spirv 1.4
As per https://gitlab.khronos.org/vulkan/vulkan/issues/2096 target environment should be SPIRV 1.4

I should have a fix out shortly

alelenv:
Fixes #2236 

Added helpers to error check based on extension


glslang/MachineIndependent/Versions.cpp
```diff
@@ -1201,5 +1223,12 @@ void TParseVersions::requireSpv(const TSourceLoc& loc, const char* op)
         error(loc, "only allowed when generating SPIR-V", op, "");
 #endif
 }
+void TParseVersions::requireSpv(const TSourceLoc& loc, const char *op, unsigned int version)
+{
+#ifndef GLSLANG_WEB
+    if (spvVersion.spv < version)
+        error(loc, "not supported for current targetted SPIR-V version", op, "");
```
dgkoch, line 1230: targeted

alelenv: Thanks, Updated

glslang/MachineIndependent/Versions.cpp
```diff
@@ -828,6 +831,9 @@ void TParseVersions::updateExtensionBehavior(int line, const char* extension, co
     // check if extension is used with correct shader stage
     checkExtensionStage(getCurrentLoc(), extension);
 
+    // check if extension has additional requirements
+    extensionRequires(getCurrentLoc(), extension ,behaviorString);
```
johnkslang, line 835: Beyond the ` ,` -> `, `:

This seems an odd design.  If we have 20 extensions with requirements, you'll call this 20 times, and each time, the call will then check 20 extensions?

I like the sound of this, and believe it could work by just taking a spvVersion that has the minimums for everything, and then the function just does all the compares.  20 such extensions will generate 20 sets of compares (instead of 400), but more importantly, it's a clear declaration for the extension and matches what the name implies.

alelenv: Thanks, agree with your points. I followed checkExtensionStage which does same, but would not be as bad since we rarely have cases of a stage specific extension.

I plan to add a map similar to extensionBehavior which will store minimum SPV version required for an extension. We could keep the map small by just storing entries for extension requiring SPV >= 1.1 since a large number of extensions would be targeting default of 1.0.

Is this in line with what you were proposing or did you mean something else for the 'spvVersion that has the minimums for everything' part?

johnkslang: Maybe it is a similar problem with checkExtensionStage.  These things should be more declarative instead of procedural.

It's a hole in general, in glslang, that we don't check required minimum versions for using an extension.  Generally, we check for either a sufficient core version or an extension enabled.

I don't mind if you don't entirely plug that hole; we can stick to the ray-tracing issue here.

Maybe we just want a simple table declaring stages and minimum versions; just C++ struct-style table that's easy to see and edit.  I suppose that could immediately go into the map you're thinking about.  It could start with just the ray-tracing things of the moment.

I'd be interested in seeing/merging such a thing if you want to try it.

alelenv: I added a prototype to see if I am in the right direction

extensionData is the table (currently added under initializeExtensionData(), ideally wanted that to be declared in Versions.h but having some build issues due to ShaderLang.h dependencies)

We can add more fields to the table (for additional error checks based on extensions)
The map 'extensionMinSpv' removes the N sequential string compares previously done.
We should probably not save entire extensions strings in table but maybe atoms/ints but we can iterate.

If this is not what you were thinking, some exact pointers on where to add and what would be appreciated

Thanks!

johnkslang: Yes, this is the direction I was thinking.  Thanks for fleshing it out.

johnkslang: Currently, we are firmly using the `E_....` macros for the strings, and already have a map that is based on their addresses, not on their content.  This makes things perform better when querying during tokenization, and prevents spelling mistakes in retyping strings.

So, having a map based on the addresses in the `E...` macros is good.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -972,6 +972,7 @@ bool ProcessDeferred(
 #endif
 
     parseContext->initializeExtensionBehavior();
+    parseContext->initializeExtensionData();
```
johnkslang, line 975: This distinction doesn't seem to make much sense at this level, I'd prefer to keep the single call if that works.

glslang/MachineIndependent/Versions.cpp
```diff
@@ -329,6 +337,23 @@ void TParseVersions::initializeExtensionBehavior()
     extensionBehavior[E_GL_EXT_shader_subgroup_extended_types_int64]   = EBhDisable;
     extensionBehavior[E_GL_EXT_shader_subgroup_extended_types_float16] = EBhDisable;
 }
+
+void TParseVersions::initializeExtensionData()
+{
+    typedef struct {
+        const char *const extensionName;
+        EShTargetLanguageVersion minSpvVersion;
+    }extensionData;
```
johnkslang, line 346: Missing space.

glslang/MachineIndependent/Versions.cpp
```diff
@@ -329,6 +337,23 @@ void TParseVersions::initializeExtensionBehavior()
...
+
+    const extensionData exts[] = { {E_GL_EXT_ray_tracing, EShTargetSpv_1_4} };
```
johnkslang, line 348: I think the obvious generalization here is to use `SpvVersion` as the second member.  If not needed yet, it's okay either way.

glslang/MachineIndependent/Versions.h
```diff
@@ -326,6 +327,13 @@ const int Num_AEP_texture_buffer = sizeof(AEP_texture_buffer)/sizeof(AEP_texture
 const char* const AEP_texture_cube_map_array[] = { E_GL_EXT_texture_cube_map_array, E_GL_OES_texture_cube_map_array };
 const int Num_AEP_texture_cube_map_array = sizeof(AEP_texture_cube_map_array)/sizeof(AEP_texture_cube_map_array[0]);
 
+//#include "./../Public/ShaderLang.h"
```
johnkslang, line 330: Of course, this needs to be cleaned up.

If it works having the data locally in the function, that seems better.

glslang/MachineIndependent/ShaderLang.cpp
alelenv: Updated

glslang/MachineIndependent/Versions.cpp
alelenv: SpvVersion has a constructor so cannot use aggregate initialization. Also fine to save space for now so keeping it as is.

alelenv:
Thanks for feedback and updated.
Please squash before merging

glslang/MachineIndependent/Versions.cpp
alelenv: Done

PR #2245, opened 2020-05-23
Change const char* to std::string for extBuiltinMap in GlslangToSpv

amdrexu:
`extBuiltinMap` is defined as the map <libName, libId>. `const char*` as
the key is not correct. Actually, we care the content of the string
rather than pointer value. The problem is not exposed because we always
use pre-defined const string `spv::E_SPV_XXX`. If the usage is changed
to this, the problem of redefinitions occurs because `getExtBuiltins()`
is malfunction.

```
  std::string ext1("XXX");
  getExtBuiltIns(ext1.c_str());

  std::string ext2("XXX");
  getExtBuiltins(ext2.c_str());
```

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -251,7 +251,7 @@ protected:
     bool nanMinMaxClamp;               // true if use NMin/NMax/NClamp instead of FMin/FMax/FClamp
     spv::Id stdBuiltins;
     spv::Id nonSemanticDebugPrintf;
-    std::unordered_map<const char*, spv::Id> extBuiltinMap;
+    std::unordered_map<std::string, spv::Id> extBuiltinMap;
 
     std::unordered_map<int, spv::Id> symbolValues;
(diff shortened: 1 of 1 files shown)


johnkslang:
Right, the current design is that you should always use the declared pointer; the E_SPV_....

Is there a real need to change this?

The best way to ensure no misspellings is to use the header(s) declaring these, and have everyone use the same definition.

amdrexu:
> Is there a real need to change this?

I am doing investigation to support a possible new extension. The shader developer could specify the SPV extension string explicitly in shader. So I have to take this string from source GLSL and generate corresponding declarations of both SPV extension and the imported extended instruction set (`OpExtInstImport`). Currently, it produces duplicated declarations. If we would like to avoid this, I have to change `const char*` to `std::string` like this PR. Otherwise, I have to do manual comparison for existing SPV extensions and take their pre-defined extension strings.

I am OK to drop this PR. But I think maybe we should make some comments in `getExtBuiltins()` to describe this restrictive usage and explain the performance reason.

PR #2254, opened 2020-05-31
HLSL: fix handling of uniform qualifier in entry point parameters

### LINKED ISSUE #2253: [HLSL] Uniform variable in entry point params gets compiled to input variable
Running this command:
```
glslangValidator -e main -H test.vert.hlsl
```
On this shader:
```
void main(uniform float4 somevar,
          out float4 l_position : SV_Position) {

  l_position = somevar;
}
```
Results in `somevar` being compiled as `Input` instead of being added either as `UniformConstant` or being added to the `$Global` block:
```
...
              16:             TypePointer Input 7(fvec4)
     17(somevar):     16(ptr) Variable Input
...
```

This is a major blocker for us.

rdb:
Fixes #2253, which is a rather pressing issue for us.  Feel free to discard this or suggest alternatives if you don't like this approach.

The change in a nutshell: when encountering uniforms in the entry point function parameters, instead of erroneously creating a global input variable for them, it now adds them to the $Global uniform block, from which they are accessed and loaded into the temporary variables in the wrapper function.  Opaque types, like textures, still get added as a global uniform variable of course.

Also contains a fix for use of "uniform in" which is supported in function arguments by FXC and handled as a regular uniform.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -4388,8 +4388,10 @@ bool TGlslangToSpvTraverser::writableParam(glslang::TStorageQualifier qualifier)
     assert(qualifier == glslang::EvqIn ||
            qualifier == glslang::EvqOut ||
            qualifier == glslang::EvqInOut ||
+           qualifier == glslang::EvqUniform ||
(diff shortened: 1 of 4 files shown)


johnkslang:
Looks okay.

Issue #2216, opened 2020-05-14
Planned deprecated event: SPIRV folder location.


johnkslang:
See announcements at the beginning of the readme:

1. SPIRV Folder, 1-May, 2020. Glslang, when installed through CMake, will install a SPIRV folder into ${CMAKE_INSTALL_INCLUDEDIR}. This SPIRV folder is being moved to glslang/SPIRV. During the transition the SPIRV folder will be installed into both locations. The old install of SPIRV/ will be removed as a CMake install target no sooner than May 1, 2020. See issue #1964.

The date has passed, but we will do this soon.

johnkslang:
If people are only using this location to get `spirv.hpp`, I recommend they get that from SPIR-V headers instead.

PR #2256, opened 2020-06-01
Remove install to the SPIRV/ folder.

### LINKED ISSUE #1964: Remove deprecated install path for SPIRV/ folder

dj2:
This CL updates the build scripts to only install to glslang/SPIRV
instead of also installing to the SPIRV/ folder. The deprecation notice
is also removed.

Note, this may cause downstream build issues if include directories have
not been updated

Fixes #1964 #2216


dj2:
Could someone using GLSLang through the install path and using the `GlslangToSpv.h` header, please verify this is OK with your system?

dj2:
I believe the answer to my concern about build paths is, if you care about making them the same, in CMake you'd just add the parent of the `glslang` folder which makes the build paths the same between cmake and system.

johnkslang:
Some of these headers are not an "interface" to glslang, and should never have been installed as such.

johnkslang:
Due to the amount of energy and presumption around the history of this, I thought it worth documenting, as it is an accumulation of changes that each on it's own that does not seem so bad:

- @johnkslang: 2015: Created HEADERS as the private SPIR-V headers needed to build glslang, nothing public, but had a `spirv.hpp` in the list
- @dneto0: Apr 14, 2016: Created SPIRV-Headers `include/spirv/1.0/spirv.hpp` for public use of SPIR-V.
- userA: Jan 29, 2017: added installing glslang HEADERS to the local path `include/SPIRV`
- userB: Apr 28, 2018, changed `include/SPIRV` to `${CMAKE_INSTALL_INCLUDEDIR}/SPIRV/`
- helperC: Oct 3, 2019: moved `${CMAKE_INSTALL_INCLUDEDIR}/SPIRV/`  to `${CMAKE_INSTALL_INCLUDEDIR}/glslang/SPIRV/`, breaking downstream builds

Apparently, people wanted to use glslang's copy of `spirv.hpp` as the public header for doing SPIR-V work, and migrated it there instead of using SPIRV-Headers. Subsequently, a reasonable amount of complaints about glslang "adding copies" and "breaking compatibility", etc. were unintended consequences of the above sequence.

(It still seems strange to me that building glslang libraries should involve installing headers in prominent public locations.)

PR #2257, opened 2020-06-01
Add default descriptorset decoration for acceleration structure variables.

### LINKED ISSUE #2247: Bug: set = 0 cant be omited
Bug: set = 0 cant be omitted for uniform acceleration structure

```
layout(binding = 0) uniform accelerationStructureEXT acc;  // creates error
```
ERROR:
[ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x1ac157ba668, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: UniformConstant id '224' is missing DescriptorSet decoration.
From Vulkan spec, section 14.5.2:
These variables must have DescriptorSet and Binding decorations specified
  %acc = OpVariable %_ptr_UniformConstant_222 UniformConstant

```
layout(binding = 0, set = 0) uniform accelerationStructureEXT acc;  // without error
```

FIX:
Set should be 0 by default for AS like in others

alelenv:
Fixes #2247 

I will file a separate bug for enforcing 'binding' qualifier on this type

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -1292,7 +1292,8 @@ bool IsDescriptorResource(const glslang::TType& type)
     // basically samplerXXX/subpass/sampler/texture are all included
     // if they are the global-scope-class, not the function parameter
     // (or local, if they ever exist) class.
-    if (type.getBasicType() == glslang::EbtSampler)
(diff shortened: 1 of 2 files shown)


alelenv:
@johnkslang to review

PR #2258, opened 2020-06-01
HLSL: Add better diagnostic when using in/out qualifiers in global scope

### LINKED ISSUE #2252: [HLSL] Declaring in/out variables in global scope triggers unhelpful assertion
Running this:
```
glslangValidator -e main -G test.vert.hlsl
```
With this shader code:
```hlsl
out float4 l_position : SV_Position;

void main() {
    l_position = float4(0, 0, 0, 0);
}
```
Triggers an `assert(false)`:
```
glslangValidator: ../SPIRV/GlslangToSpv.cpp:1233: spv::StorageClass {anonymous}::TGlslangToSpvTraverser::TranslateStorageClass(const glslang::TType&): Assertion `0' failed.
```
It would be more helpful to show an appropriate error message.  For example, DXC shows this:
```
test.vert.hlsl:1:1: error: HLSL usage 'out' is only valid on a parameter
out float4 l_position : SV_Position;
^~~
```

rdb:
Fixes #2252

### DIFF
--- a/hlsl/hlslGrammar.cpp
+++ b/hlsl/hlslGrammar.cpp
@@ -382,6 +382,16 @@ bool HlslGrammar::acceptDeclaration(TIntermNode*& nodeList)
     if (forbidDeclarators)
         return true;
 
+    // Check if there are invalid in/out qualifiers
+    switch (declaredType.getQualifier().storage) {
+    case EvqIn:
+    case EvqOut:
+    case EvqInOut:
+        parseContext.error(token.loc, "in/out qualifiers are only valid on parameters", token.string->c_str(), "");
+    default:
+        break;
+    }
+
     // declarator_list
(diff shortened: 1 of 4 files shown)


hlsl/hlslGrammar.cpp
```diff
@@ -382,6 +382,16 @@ bool HlslGrammar::acceptDeclaration(TIntermNode*& nodeList)
     if (forbidDeclarators)
         return true;
 
+    // Check if there are invalid in/out qualifiers
+    switch (declaredType.getQualifier().storage) {
```
johnkslang, line 386: This is a grey area between syntax and semantics, so maybe okay in the grammar file.

Generally, semantic errors should not go here; this file should stay focused on the grammar recognition.

PR #2260, opened 2020-06-05
Use OpFUnordNotEqual for floating-point !=

### LINKED ISSUE #2259: SPIR-V output incorrectly translates float-not-equal to OpFOrdNotEqual
The IEEE not-equal comparison is equivalent to the unordered OpFUnordNotEqual, not to OpFOrdNotEqual. The current translation will give the wrong result for comparison involving a NaN.

gnl21:
The normal IEEE not equal operation tests whether operands are unordered
or not equal (so comparison with a NaN returns true). This corresponds
to the SPIR-V OpFUnordNotEqual, so change to using that.

Fixes #2259


johnkslang:
I think this proposes to make `a != b` the same as `!(a == b)`.  I think we should be quite careful about this.  Is this backed up by the specification (or lack of detail enforcing NaN operation)?  I even recall discussions that SPIR-V not even support the unordered compares, and one reason we did was for compiler transforms computing the `!` of `==` and `!=`, but this seems to invalidate those arguments.

Do we need any specification changes for this?

gnl21:
I think that this is justified by the current spec. It does propose to make `a != b` the same as `!(a == b)`, but that's what's expected of the standard `!=` operator in (all? most?) HLLs. It's a weird quirk of the naming convention that all IEEE comparisons convert to ordered comparison except `!=`, which naturally converts to the *un*ordered version. The arguments about supporting the unordered comparisons (for `<`, `>`, `==`, etc) apply in reverse to `!=`. There is no natural HLL expression of the ordered-not-equal, but we support it in SPIR-V to simplify compiler transforms.

For the purposes of ordered/unordered comparisons, any comparison involving a `NaN` is unordered. The definition of `OpFUnordEqual` says:
> Floating-point comparison for being unordered or not equal.

which implies that, for `OpFUnordNotEqual`, `x != NaN` for all `x`. This matches C and D3D semantics. I'm not sure that GLSL says explicitly, but I think it is intended to match.

gnl21:
Is there anything that I need to do to progress this?

johnkslang:
Do we know if all IHVs are ready for this?  At least it's been posted here for a while.

We also need to bump the code generation number since we are changing a code emission pattern.  That will help protect against surprises too.

See:

```
{
    // return 1; // start
}
```

(We are also in the middle of possibly migrating to a new method via @ben-clayton's PRs, e.g., #2277.)

johnkslang:
BTW, it will be good to have the version bump in its own commit for ease of verification.

gnl21:
> Do we know if all IHVs are ready for this? At least it's been posted here for a while.

I don't know for certain, but this doesn't change the SPIR-V that drivers are being asked to accept, although it obviously changes the balance of how often these different opcodes come up. Both opcodes are currently tested in the CTS (OpFUnordNotEqual and OpFOrdNotEqual), so they should be working in implementations.

> We also need to bump the code generation number

Sure, I'll do that now.

PR #2265, opened 2020-06-10
HLSL: Fix incorrect case in name of DX9 cube sampler type

rdb:
The type is named samplerCUBE in HLSL.  I have verified that FXC does not accept samplerCube.

### DIFF
--- a/hlsl/hlslParseables.cpp
+++ b/hlsl/hlslParseables.cpp
@@ -324,7 +324,7 @@ glslang::TString& AppendTypeName(glslang::TString& s, const char* argOrder, cons
             case 1: s += "1D";                   break;
             case 2: s += (isMS ? "2DMS" : "2D"); break;
             case 3: s += "3D";                   break;
-            case 4: s += "Cube";                 break;
+            case 4: s += (type == 'S'? "CUBE" : "Cube"); break;
             default: s += "UNKNOWN_SAMPLER";     break;
             }
         }
--- a/hlsl/hlslScanContext.cpp
+++ b/hlsl/hlslScanContext.cpp
@@ -317,7 +317,7 @@ void HlslScanContext::fillInKeywordMap()
     (*KeywordMap)["sampler1D"] =               EHTokSampler1d;
     (*KeywordMap)["sampler2D"] =               EHTokSampler2d;
     (*KeywordMap)["sampler3D"] =               EHTokSampler3d;
-    (*KeywordMap)["samplerCube"] =             EHTokSamplerCube;
+    (*KeywordMap)["samplerCUBE"] =             EHTokSamplerCube;
     (*KeywordMap)["sampler_state"] =           EHTokSamplerState;
     (*KeywordMap)["SamplerState"] =            EHTokSamplerState;
     (*KeywordMap)["SamplerComparisonState"] =  EHTokSamplerComparisonState;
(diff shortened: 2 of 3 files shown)


hlsl/hlslParseables.cpp
```diff
@@ -324,7 +324,7 @@ glslang::TString& AppendTypeName(glslang::TString& s, const char* argOrder, cons
             case 1: s += "1D";                   break;
             case 2: s += (isMS ? "2DMS" : "2D"); break;
             case 3: s += "3D";                   break;
-            case 4: s += "Cube";                 break;
+            case 4: s += (type == 'S'? "CUBE" : "Cube"); break;
```
johnkslang, line 327: Two concerns here:

- Does HLSL have multiple spellings? I'm not sure I understand spelling differently for scalar and non-scalar yet.

- The comment for this function says it is to follow GLSL conventions, because these results are parsed by the GLSL parser.  That implies not making a change, unless we actually did move to the HLSL parser and the comment is out of date.

hlsl/hlslScanContext.cpp
```diff
@@ -317,7 +317,7 @@ void HlslScanContext::fillInKeywordMap()
     (*KeywordMap)["sampler1D"] =               EHTokSampler1d;
     (*KeywordMap)["sampler2D"] =               EHTokSampler2d;
     (*KeywordMap)["sampler3D"] =               EHTokSampler3d;
-    (*KeywordMap)["samplerCube"] =             EHTokSamplerCube;
+    (*KeywordMap)["samplerCUBE"] =             EHTokSamplerCube;
```
johnkslang, line 320: What do you think about keeping both keyword spellings, to reduce breaking existing content that might have used this spelling?

hlsl/hlslParseables.cpp
rdb: This function is used to generate the prototypes for the built-in intrinsics—it has to be modified or most of the unit tests fail, presumably because it can't parse the prototypes of the intrinsics.

'S' indicates sampler here.  The ternary here to make sure that it generates `samplerCUBE` for DX9-style samplers but `TextureCube` for DX10-style textures.

hlsl/hlslScanContext.cpp
rdb: Personally I would suggest to just change it.  My intuition is that this sampler type only exists for DX9 support which I imagine is mainly used by people who need to preserve compatibility with an existing body of HLSL shaders, which seems to be unlikely to use a non-existent HLSL type.

But it would be possible to have a transition period in which both are supported.  If you want I can change this PR to support both.  Would you want me to also revert the AppendTypeName change for now then?

hlsl/hlslParseables.cpp
johnkslang: Okay, so I verified that the HLSL parser is handling the built-in declarations, due to:

```
const bool UseHlslTypes = true;
```

Meaning, as per my second bullet, they are following HLSL conventions and the comment is now misleading (HLSL originally used the GLSL parser to construct the built-ins).

hlsl/hlslScanContext.cpp
johnkslang: I agree.

I think we've gone down the path now too of no longer being able to use the GLSL parser anyway.

PR #2267, opened 2020-06-10
Untangle build target cyclic dependencies

ben-clayton:
Fixing the building of shared libraries.

Applications now require a call to `glslang::RegisterHlslLanguage()` to enable support for the HLSL language.
This is implemented in the `HLSL` library.

Related: #1484, #2147

---

Work in progress. Not ready for review.


glslang/Public/ShaderLang.h
```diff
@@ -404,6 +404,13 @@ bool InitializeProcess();
 // Call once per process to tear down everything
 void FinalizeProcess();
 
+#ifdef ENABLE_HLSL
+// Call to enable HLSL language support before doing any other compiler/linker
+// operations.
+// Implemented in the HLSL library.
+void RegisterHlslLanguage();
```
johnkslang, line 411: I'm worried about breaking existing users here.

Given the protection of ENABLE_HLSL, can we also have HLSL auto-register like GLSL did?

StandAlone/StandAlone.cpp
```diff
@@ -1250,6 +1250,10 @@ void CompileAndLinkShaderFiles(glslang::TWorklist& Worklist)
 
 int singleMain()
 {
+#ifdef ENABLE_HLSL
+    glslang::RegisterHlslLanguage();
```
johnkslang, line 1254: We won't need this if we can get HLSL to auto register (review comments below).

glslang/Public/ShaderLang.h
ben-clayton: > can we also have HLSL auto-register like GLSL did

Not without some really hairy compiler / platform specific logic. The problem is that without `RegisterHlslLanguage()`, no compilation units in the HLSL target are directly referenced, so the static initializers for those CUs will not run. This is not true for the `glslang::GlslLanguage` which shares the CU with the `glslang::Language` methods, which is why we can use an auto registration.

With this PR, the target dependency chain looks like this:

```
Application  ←   hlsl
    ↑             ↑    
    └── glslang ──┘
```

There are a few other strategies we could try:

### 1. Flipping the dependency chain:

So that `glslang` depends on `hlsl` (like it used to, but without the cyclic dependency):

```
Application  ←  glslang  ←  hlsl
```

This would involve abstracting all the `glslang` APIs with interfaces or something similar.  Feels like a lot of work as `hlsl` code heavily relies upon the `MachineIndependent` headers and types.

### 2. Merging the `hlsl` target into `glslang`

Just pull the `hlsl` source files into the `glslang` target, and do away with the CMake `HLSL` target.

If we did this, it might make sense to move the `hlsl` folder to `glslang/hlsl` or somewhere similar. Even with this PR, there's still a number of `ENABLE_HLSL` kludgy blocks in the `glslang` target, so this might not be a bad thing.

While this is certainly the easiest option, it would prevent the generation of `hlsl.dll` / `hlsl.so`, which may upset some existing build setups. Given the existing cyclic dependency between `hlsl.so` and `glslang.so`, this might not be a big problem.

### 3. Doing some magic to automatically call `RegisterHlslLanguage()` from `glslang`

If we can tell that glslang is being built statically then we could revert back to an `extern` function call - the symbol defined in the `hlsl` package should be found by the linker, just as `glslang` and `hlsl` have been bi-directionally referencing each other before.

For the shared library case, we theoretically could use `dlsym()` / `GetProcAddress()` to look up dynamic library and call the `RegisterHlslLanguage()`. 
This is venturing into the vomit inducing land of magic, and I'm sure there's a whole sea of issues just waiting to be found if we went this path.

ben-clayton: FWIW - the reason I went the way I did with this PR is that it means we could in theory remove the `ENABLE_HLSL` preprocessor logic entirely. If the application wants to enable HLSL, depend on the `HLSL` library, call `RegisterHlslLanguage()`, done.

This keeps the `glslang` library identical for `hlsl` or `glsl` use cases, avoiding a library versioning issue for `ENABLE_HLSL`.

dj2: Would it be possible as a 4th option to get rid of the static initializers in the HLSL code? Then you can have it auto-register?

ben-clayton: @dj2 - I'm sorry, but I don't follow. I'm not sure what static initializers you're referring to, nor do I see how removing them would allow for automatic registration.
Care to elaborate?

dj2: I'm guessing I mis-understood then, from your comment above:

> The problem is that without RegisterHlslLanguage(), no compilation units in the HLSL target are directly referenced, so the static initializers for those CUs will not run.

If the problem is missing static initializers, can we just get rid of them?

ben-clayton: Ah - no, [the glsl logic uses static initializers for automatically invoking the registration function](https://github.com/KhronosGroup/glslang/pull/2267/files#diff-d91d483f9185cb0ef86c5e96e18b3a74R83-R87
):

```c++
// Automatically register the GLSL language when the static initializers are run
// for this compilation unit.
struct AutoRegisterGlslLanguage {
    AutoRegisterGlslLanguage() { Language::Register(&GlslLanguage::instance); }
} autoRegisterGlslLanguage;
```

My point was we can't use the same mechanism to have the hlsl library self-register, as there's no `glslang` -> `hlsl` dependency (as that's what we're trying to avoid).

johnkslang: Good discussion, and thanks for championing this @ben-clayton.  And, understood about the issue with HLSL apparently being either circular or needing an extra start-up call.

We already have entry points that include saying what source language you want, and I'm hopeful we can still find a backward compatible solution.

(Though, it is also attractive to bump the major version up and do more things "the right way" instead of trying to maintain backward compatibility.)

More discussion...

> Not without some really hairy compiler / platform specific logic.

Agree we don't want to do that.

> the reason I went the way I did with this PR is that it means we could in theory remove the ENABLE_HLSL preprocessor logic entirely

I see the appeal of that, in the abstract, but there are reasons to keep ENABLE_HLSL (or at least convert the to GLSLANG_WEB), and without going into lots of detail on a side discussion, would like to solve this current PR assuming we aren't going down the path of removing ENABLE_HLSL.

So, basically, if there are easier ways, or ways that help once given that ENABLE_HLSL is staying, we should feel free to pursue those.

Regarding global objects relying on start-up time constructors to accomplish things:  This needs to be done with great caution, as there are numerous ways of getting burned by those.  I'd prefer not having them, but if we do, things to think through are:

- multiple different instances of glslang existing in large projects
- lifetime of what's created WRT multiple cycles through these that we already have:
```
   bool InitializeProcess()
   void FinalizeProcess()
```
- confusion between having two sets of things sounding like this too.

I think they might be okay, just easier to not have to be sure they are okay.

> 2. Merging the hlsl target into glslang

This seems like the most obvious natural thing to happen, as long as we don't accidentally bloat GLSLANG_WEB.

But, what's the true fundamental restriction here?  When ENABLE_HLSL is off, there is no dependency, and we don't include all that code.  When it is on, we need both libraries, and I'm not clear at the moment why it's fatal to have a circular dependency between two things that have to be present.

ben-clayton: > Regarding global objects relying on start-up time constructors to accomplish things: This needs to be done with great caution, as there are numerous ways of getting burned by those. 

Sure, there's really no need to keep this for the `GlslLanguage`, as we already have `InitializeProcess()` and `FinalizeProcess()`. Happy to change if we still want to keep the `glslang::Language` abstraction.

> This seems like the most obvious natural thing to happen, as long as we don't accidentally bloat GLSLANG_WEB.
>
> But, what's the true fundamental restriction here? When ENABLE_HLSL is off, there is no dependency, and we don't include all that code. 

We can conditionally pull in the `hlsl` source files into the `glslang` target (in CMake) based on the `ENABLE_HLSL` build flag to save wrapping all the code in preprocessor smell. I notice that [`BUILD.gn`](https://github.com/KhronosGroup/glslang/blob/b919bc889e1d1032fcbad231b444ff96a541615d/BUILD.gn#L150-L164) and [`BUILD.bazel`](https://github.com/KhronosGroup/glslang/blob/b919bc889e1d1032fcbad231b444ff96a541615d/BUILD.bazel#L39) already folds in the `hlsl` source for the `glslang` target.

> When it is on, we need both libraries, and I'm not clear at the moment why it's fatal to have a circular dependency between two things that have to be present.

Having libraries (whether static or dynamic) with circular dependencies is a real pain for build systems and the built libraries.
A couple of examples why:
* Most build systems require target dependencies to be a DAG. Currently, CMake is unaware of the `hlsl` dependency on `glslang` - attempting to add this missing dependency results in [an error](https://github.com/KhronosGroup/glslang/issues/1484#issuecomment-445809573) as CMake forbids circular dependencies. Without the correct dependency information, CMake doesn't know to rebuild `hlsl` if files in `glslang` change, likely leading to broken builds.
*  It's a good idea to build with the `--no-undefined` linker option to ensure that all dependencies can be resolved at build time, otherwise you can end up with unresolved function calls crashing in spectacularly horrible ways at runtime.

Finally, I'm not even sure if Windows allows circular DLL dependencies. All searches I've found are people asking how to get Visual Studio to allow this for building (which it won't, as it requires a DAG again), and most the answers I've seen are along the lines of "don't do this".

I'd strongly recommend keeping your build targets as a simple DAG.

If we want to just fold the `hlsl` target into `glslang`, that works for me.
One day we might want to consider breaking up the multi-language monolithic target, and revisit something like this approach. No urgency.

ben-clayton:
Landed https://github.com/KhronosGroup/glslang/pull/2271 instead. Closing this PR.

PR #2271, opened 2020-06-15
Untangle build target cyclic dependencies (approach #2)

ben-clayton:
Fixes the building of shared libraries.

This PR holds two changes:

1. Split the SPIR-V C interface to own file.

This breaks a cyclic dependency between the `SPIRV` and `glslang` build targets.

2. Fold the `HLSL` source into the `glslang` target
 
This breaks the cyclic dependency between the `glslang` and `hlsl` targets (by essentially removing the `hlsl` target).

The `BUILD.gn` and `BUILD.bazel` build rules already pull the `HLSL` source into the `glslang` target.

`Android.mk` is now the only remaining build config that has a (real) dedicated `HLSL` target, but this is explicity static and does not suffer the same link-time issues with the cyclic dependency. We may wish to stub this `HLSL` target too at a later date.

Related issue: #1484, #2147
Related PR: #2267

### DIFF
--- a/hlsl/stub.cpp
+++ b/hlsl/stub.cpp
@@ -1,0 +1,40 @@
+//
+// Copyright (C) 2020 Google, Inc.
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//
+//    Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//
+//    Redistributions in binary form must reproduce the above
+//    copyright notice, this list of conditions and the following
+//    disclaimer in the documentation and/or other materials provided
+//    with the distribution.
+//
+//    Neither the name of Google, Inc., nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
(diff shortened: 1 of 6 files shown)


hlsl/stub.cpp
```diff
@@ -0,0 +1,41 @@
...
+// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+
+
+// The HLSL source is directly embedded into the glslang target when ENABLE_HLSL
+// is set.
+// The HLSL target is now just a stub that exists for backwards compatibility
+// for projects that referenced this target. As a target requires at least one
+// source file to build, this file acts as that stub.
```
johnkslang, line 41: I don't know if it's an actual problem, or not to be missing the final new line.

johnkslang:
Yes, this seems a reasonable thing to do, I like it.

I don't know if the finer granularity of the libraries is actually needed, but it is better to start this way (post your PR) and then only move to finer grained if the move is done correctly, without circularities.

hlsl/stub.cpp
ben-clayton: New line added.

PR #2274, opened 2020-06-15
Use int instead of enum for EShLanguageMask.

gnoliyil:
The result of bitwise or and bitwise negation of enums should
be int instead of EShLanguageMask. Otherwise it will be an
undefind behavior and will trigger UBsan.


johnkslang:
Is it possible the tool is in error?

C++ allows values that are representable by the type need to represent all values.  These operations do not cause values to go out of range.

Why should we lose real type safety to satisfy this essentially pointless rule?

gnoliyil:
Thanks for reply! 

Just found that change https://github.com/KhronosGroup/glslang/commit/91ac60fc525c78f158541dc412edf7fd9804a00c has already solved the undefined behavior issue by setting the enum type to unsigned and I hadn't synced to origin/master when I saw the error. Sorry and please disregard this pull request.

PR #2275, opened 2020-06-16
Move hlsl/ source to glslang/HLSL/

ben-clayton:
Now that the HLSL source files are part of the `glslang` target (#2271), it makes sense for these to sit in the `glslang` directory.

Changed the case of the directory from `hlsl` to `HLSL` to better match the sibling directories.


johnkslang:
Looks good.  I'm going to bump the middle version number along with this change.

Issue #1754, opened 2019-04-24
Improve Semantic Version access in the source code, to allow for easy extraction of the information by build systems like cmake


danielmrussell:
The FindZLIB.cmake module extracts the following #defines from the ZLIB source code, in order to construct the elements of its current library semantic version:
• ZLIB_VERSION_MAJOR
• ZLIB_VERSION_MINOR
• ZLIB_VERSION_PATCH

The built-in find_package hook for ZLIB that comes with cmake FindZLIB.cmake uses textual extraction (with regexps etc) to get to this information, and put it into CMAKE variables fulfilled by the find_package operation.

I am attempting to create a similar cmake find_package hook for glslang, but finding it to be a pain. The problem is, were have these defined instead:
• No #define for major version but a hard coded constant in GetSpirvGeneratorVersion(), buried in a difficult place to find (in SPIRV/GlslangToSpv.cpp)
•  GLSLANG_MINOR_VERSION
• GLSLANG_PATCH_LEVEL

I propose we standardize to Best Practices and copy the ZLIB names exactly, but with GLSLANG_ of course, and make them #defines so that it is "easy" to extract them textually from  the cmake module (by copying and modifying the built-in cmake module FindZLIB.cmake).

So the following #defines would be easily accessible from a single short header that is PURELY for recording the versioning information, ideally, or **at the very least, all together in one place.**
• GLSLANG_VERSION_MAJOR
• GLSLANG_VERSION_MINOR
• GLSLANG_VERSION_PATCH

It would make it a lot easier.

johnkslang:
Does it work to have a localized/consistent functional interface instead of #defines?

Glslang truly is multiple tools in the same binary, and we are indeed forming the semantic version by mixing the code generator's versioning with the overall patch level.

danielmrussell:
It's very confusing and seem out of alignment with the trending best practice, as it forces difficult or impossible decisions in the build tools department. Little things really making us waste so much time.

It has to be something readable by cmake itself, so sadly, needs to be #defines or similar. If it is embedded in functions using hard coding, then it might as well be #defined as well and then referenced instead of a hard-coded value, for instance. Even if the names were not brought into alignment with best practices around semantic versioning and cmake, it would still be helpful to be able to read the value in a way that is not super fragile.

Technically, I could parse the values out of the multiple files, and I could even parse the hardcoded value out of the function its in, but that super fragile and almost guaranteed to break once we clean up the code.

#defines less so.

[...]

Issue #1255, opened 2018-02-21
Improve versioning


johnkslang:
There was a request to have glslang versioned with a `major.minor.patchlevel` triple. Since this is primarily associated with glslang as a SPIR-V generator, that's essentially what the triple will version, but will also reflect the validator and both GLSL and HLSL inputs to making SPIR-V.

(Something similar was being done by a client-side hook, which stopped working and is fragile because it depends on the client machine.)

So, planning to look at the existing issues around versioning/tagging, as well as industry standard practice, and have some discussion here. Should look something like the following:

In `major.minor.patchlevel`:

- `patchlevel` should go up in value with each bug fix, possibly being the number of commits that have been made to the repository
-  `minor` would go up with added functionality (or fixes that a big enough to be considered adding functionality)
- `major` would likely be the current SPIR-V generator version, which is currently bumped for potentially backward incompatible code generation changes (but maybe is not sufficient to capture backward incompatible changes in how the front-end interprets shader source)

Current behavior of `-v` is to print this, which already includes the number of commits (defunct, currently at `2000`, see https://github.com/KhronosGroup/glslang/blob/master/make-revision) as a patch level, and the `SPIR-V Generator Number` proposed as the major number:

```
Glslang Version: Overload400-PrecQual.2000 12-Apr-2017
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos.Overload400-PrecQual.2000 12-Apr-2017
GLSL Version: 4.60 glslang Khronos.Overload400-PrecQual.2000 12-Apr-2017
SPIR-V Version 0x00010200, Revision 3
GLSL.std.450 Version 100, Revision 1
Khronos Tool ID 8
SPIR-V Generator Version 4
GL_KHR_vulkan_glsl version 100
ARB_GL_gl_spirv version 100
```

Command-line options should change to:

- `--version` as a synonym for `-v`, with the first line of output containing the `major.minor.patchlevel` triple
- `-dumpfullversion` print just the bare `major.minor.patchlevel` triple
- `-dumpversion` synonym for `-dumpfullversion`

Any input?

linuxturtle:
I very much like where you're going with this. Thank you for considering it! At what point would you tag the git repository?  From a packaging standpoint, it would be great if you could simply tag when you think the major/minor numbers should change. That makes it quite simple to create a reproducible version string like you describe above for any commit with:

   `git describe --tags | sed 's/\(.*\)-.*/\1/'`

[...]

PR #2277, opened 2020-06-17
Generate build information from CHANGES.md

ben-clayton:
This PR significantly reworks the way glslang is versioned.

Instead of committing changes to the `GLSLANG_MINOR_VERSION` define in [`glslang/Public/ShaderLang.h`](https://github.com/KhronosGroup/glslang/blob/master/glslang/Public/ShaderLang.h), and using [`make-revision`](https://github.com/KhronosGroup/glslang/blob/master/make-revision) to generate `GLSLANG_PATCH_LEVEL` in [`glslang/Include/revision.h`](https://github.com/KhronosGroup/glslang/blob/master/glslang/Include/revision.h), all version information is now derived from the new `CHANGES` file.

`CHANGES` acts as the single source of truth for `glslang` version information, along with a convenient place to put all release notes for each notable change made.

`CHANGES` is parsed using the new `build_info.py` python script. This script can read basic template files to produce new source files - which it does to read the new `build_info.h.tmpl` to generate (at build time) a glslang private header at `<build-dir>/include/glslang/build_info.h`. I've written generators for each of the CMake, Bazel, gn, and Android.mk build scripts.

The new version code conforms to the [Semantic Versioning 2.0 spec](https://semver.org/).

This new version is also used by the CMake rules to produce versioned shared objects, including a major-versioned `SONAME`.

## Breaking API changes:

* The public defines `GLSLANG_MINOR_VERSION` and `GLSLANG_PATCH_LEVEL` have been entirely removed.
* `glslang/Public/ShaderLang.h` and `glslang/Include/revision.h` have been deleted.
* Instead `<build-dir>/include/glslang/build_info.h` is created in the build directory, and `<build-dir>/include` is a CMake `PUBLIC` (dependee-inherited) include directory for the `glslang` targets.
* `<build-dir>/include/glslang/build_info.h` contains the following new `#define`s:
  * `GLSLANG_VERSION_MAJOR`
  * `GLSLANG_VERSION_MINOR`
  * `GLSLANG_VERSION_PATCH`
  * `GLSLANG_VERSION_FLAVOR`
  * `GLSLANG_VERSION_GREATER_THAN(major, minor, patch)`
  * `GLSLANG_VERSION_GREATER_OR_EQUAL_TO(major, minor, patch)`
  * `GLSLANG_VERSION_LESS_THAN(major, minor, patch)`
  * `GLSLANG_VERSION_LESS_OR_EQUAL_TO(major, minor, patch)`
* The CMake install output directory contains a copy of `build_info.h` at: `include/glslang/build_info.h`

Python3 is now always required to build glslang (likely always required for transitive dependency builds).

There is a new `glslang::GetVersion()` function which returns a `Version` struct with the version major, minor, patch and flavor.

---

## Related issues:

* Shared library use not first-class functionality, not tested #1484
* Improve Semantic Version access in the source code, to allow for easy extraction of the information by build systems like cmake #1754
* Improve versioning #1255


johnkslang:
Couple of questions, before looking in detail.

1. Could we see the new output of `glslangValidator -v`, perhaps next to the previous output?
2. What becomes of this?

```
{
    // return 1; // start
```

ben-clayton:
> 1. Could we see the new output of glslangValidator -v, perhaps next to the previous output?

New (`master` branch, release won't have `-dev`):

```
Glslang Version: 9:9.0.0-dev
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 9.0.0-dev
GLSL Version: 4.60 glslang Khronos. 9.0.0-dev
SPIR-V Version 0x00010500, Revision 3
Khronos Tool ID 8
SPIR-V Generator Version 9
ARB_GL_gl_spirv version 100
```

Old:

```
Glslang Version: 9.15.3802
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 15.3802
GLSL Version: 4.60 glslang Khronos. 15.3802
SPIR-V Version 0x00010500, Revision 3
Khronos Tool ID 8
SPIR-V Generator Version 9
ARB_GL_gl_spirv version 100
```

Note the new version is not final - the major should probably be at least 10.

> 2. What becomes of this?

It rather depends on what is using it.

We could keep it as the PR currently does - the **`9:`** of the `9:9.0.0-dev` is the number returned by `GetSpirvGeneratorVersion()`.

If I understand correctly, you say that some people are using the SPIR-V version for knowing when to invalidate the shader caches? That can still work if we expose the SPIR-V version in the `glslang::Version` structure returned by `glslang::GetVersion()`.
I do wonder however whether they should always be regenerating their caches if they're pulling a new version of this library.

jengelh:
>The public defines GLSLANG_MINOR_VERSION and GLSLANG_PATCH_LEVEL have been entirely removed

Too bad, this is a common idiom seen that saves time implementing function existence checks in one's build description:

```
// mycode.cpp
#if GLSLANG_MINOR_VERSION >= 3559
blah->beginParameterParsing(...)
#else
something else
#endif
```

ben-clayton:
Thanks for the feedback!

> Too bad, this is a common idiom seen that saves time implementing function existence checks in one's build description

Is this idiom in active use for glslang? I took a look around public repos, and I couldn't see use of this pattern. GitHub's code search is awful though, and I clearly cannot inspect closed source projects. 

For what it's worth, the semantic versioning proposed here will guarantee API backwards compatibility for each minor and patch release of the same major. I appreciate your argument can be extended to major versions, and use of newly introduced functions in minor releases.

If there's strong desire, we could make the generated header containing the version defines be public for targets depending on glslang. I didn't start with this, as I prefer to keep the public API small and grow only if needed.

johnkslang:
A few random notes:

- I'd rather not see any of the numbers going backward.  Is that actually critical?
- At least historically, there have been several important private uses of glslang.
- I think the caching issue (which is subtle) is satisfied by keeping all the different version strings, as you have done.

ben-clayton:
> I'd rather not see any of the numbers going backward. Is that actually critical?

I can certainly make the major 10. However when you say "**any** of the numbers", the less significant numbers should be reset to zero when bumping a more significant number - see [7.](https://semver.org/spec/v2.0.0.html#spec-item-7) and [8.](https://semver.org/spec/v2.0.0.html#spec-item-8) of the Semantic Versioning spec. Are you suggesting always having the minor / patch numbers increase, even when a more significant number is bumped?

> At least historically, there have been several important private uses of glslang.

Understandable. I can take a look at making the `build_info.h` public if this is something people want (which it seems @jengelh certainly wants).

> I think the caching issue (which is subtle) is satisfied by keeping all the different version strings, as you have done.

Ack.

jengelh:
>Is this idiom [testing for the values of preproc macros] in active use for glslang?

I do not know about glslang (it is kinda new), but it is something that has made appearances in Linux kernel modules, FUSE modules, software that would have to support libmysqlclient 5.5 & 8 APIs simultaneously, and I think libusb users do something similar.
I am a distro packager, and I "could swear I've seen it quite often", but because it has become daily routine, I would have to do an active search in the package base to deliver further concrete examples. But I'd surely find them in short order.

ben-clayton:
Okay, compelling enough. I'll try and get these versions public.

Thanks @jengelh !

ben-clayton:
@jengelh - I've updated the CMake rules to make the generated header include directory `PUBLIC`. Dependee targets should be able to `#include "glslang/build_info.h"`, and this also is now emitted into the install directory at `include/glslang/build_info.h`.

johnkslang:
> Note the new version is not final - the major should probably be at least 10.

Right, and I think maybe a compromise of first going from (say) `9.15.3802` to `10.15.3802`, and then later get to `11.0.0`, just to give a bit more warning it is coming.

I'm also going to update the README now to announce this.

(Because I recall real push, long ago, to have an increasing number, but do imagine that by now the new scheme should suffice.)

ben-clayton:
> Right, and I think maybe a compromise of first going from (say) 9.15.3802 to 10.15.3802, and then later get to 11.0.0, just to give a bit more warning it is coming.

Done.

I've rebased so that we can have kokoro test the android-ndk and cmake script changes. Note - we still don't have presubmits for the .gn build. I'm going to spend a bit of time today to see if I can get this going.

I've also renamed `CHANGES` to `CHANGES.md` so that this renders nicely in github.

ben-clayton:
I believe this is now in reviewable shape.

SPIRV/GlslangToSpv.cpp
```diff
@@ -5207,10 +5209,10 @@ spv::Id TGlslangToSpvTraverser::createImageTextureFunctionCall(glslang::TIntermO
         assert(builder.isStructType(resultStructType));
 
         //resType (SPIR-V type) contains 6 elements:
-        //Member 0 must be a Boolean type scalar(LOD), 
-        //Member 1 must be a vector of integer type, whose Signedness operand is 0(anchor),  
-        //Member 2 must be a vector of integer type, whose Signedness operand is 0(offset), 
-        //Member 3 must be a vector of integer type, whose Signedness operand is 0(mask), 
+        //Member 0 must be a Boolean type scalar(LOD),
```
johnkslang, line 5212: I'd like to stop stripping unrelated white space in these PRs.

I also prefer not having trailing white-space, but we don't enforce it, and it's not as important as `git blame` giving a simple answer and (this time) making the code deltas bigger with unrelated changes.

(Enforcing should never be done, BTW, on the test files, or generated files, only the true source, if anyone decides to try to enforce it.)

glslang/Public/ShaderLang.h
```diff
@@ -74,11 +74,6 @@
     extern "C" {
 #endif
 
-// This should always increase, as some paths to do not consume
-// a more major number.
-// It should increment by one when new functionality is added.
-#define GLSLANG_MINOR_VERSION 15
```
johnkslang, line 80: I think this will break some people that were using a tool to grep through the source to find the version info.

SPIRV/GlslangToSpv.cpp
ben-clayton: Will put back and try and prevent my IDE from doing this in the future.

glslang/Public/ShaderLang.h
ben-clayton: The version information now exists as a single checked-in file - `CHANGES.md`.

If they want to get this info, they can either:
1. Use the new CMake variables: `GLSLANG_VERSION`, `GLSLANG_VERSION_MAJOR`, `GLSLANG_VERSION_MINOR`, `GLSLANG_VERSION_PATCH`, `GLSLANG_VERSION_FLAVOR` (we might want to consider promoting these to `PARENT_SCOPE` if we find outer projects want to get at this).
2. Use `build_info.py` with a template to generate the required version info.
3. Parse the `CHANGES.md` file themselves, using the regex: `^#* +(\d+)\.(\d+)\.(\d+)(-\w+)? (\d\d\d\d-\d\d-\d\d)? *$` (the one used in `build_info.py`)

I suspect the people most likely wanting to get at this are package maintainers, who can now easily generate the `pkg-config` file as part of the CMake rules.

---

Unfortunately there's no simple way to keep the existing greps going. If we can identify those people relying on these greps, I can guide them on a migration strategy.

SPIRV/GlslangToSpv.cpp
ben-clayton: Done

glslang/Public/ShaderLang.h
johnkslang: I think just updating the README, such that people can basically understand the versioning process, will be sufficient.

ben-clayton: Added some notes to README. Not sure how much detail you want to go into there.

johnkslang: Looks fine.

orbea:
> Is this idiom in active use for glslang?

@ben-clayton  Libplacebo is one.

https://code.videolan.org/videolan/libplacebo/-/blob/44525d4bf6aa63910b7b3eb70d69b0855e889e06/src/meson.build#L86
https://code.videolan.org/videolan/libplacebo/-/blob/44525d4bf6aa63910b7b3eb70d69b0855e889e06/src/glsl/glslang.cc#L26
https://code.videolan.org/videolan/libplacebo/-/blob/44525d4bf6aa63910b7b3eb70d69b0855e889e06/src/glsl/glslang.cc#L37
https://code.videolan.org/videolan/libplacebo/-/blob/44525d4bf6aa63910b7b3eb70d69b0855e889e06/src/glsl/glslang.cc#L81
https://code.videolan.org/videolan/libplacebo/-/blob/44525d4bf6aa63910b7b3eb70d69b0855e889e06/src/glsl/glslang.cc#L37

ben-clayton:
Hi @orbea,

Sorry I didn't spot this use of `GLSLANG_PATCH_LEVEL` while working on these changes.

Looking through the build files you have linked to and I see the `"work-arounds for glslang braindeath"` - argh, not pretty.
I can fully appreciate that this change must be frustrating for projects continually searching for reliable ways to determine `glslang` version, and I'm sorry this change has caused even more trouble.
This change was the start of work to properly semantically version glslang, so that dependencies like this one can confidently assume compatibility of glslang releases with the same major version, and so we can all avoid these sorts of nasty workarounds moving forward. Unfortunately in order to fix things for the long term, we had to break things in the short term.

I'm not overly familiar with the meson build system, but my understanding is that these build rules are trying to determine the `glslang_ver` from the system installed glslang package headers.
If that's correct, then new glslang packages will contain the `glslang/build_info.h` header, which at the time of writing contains:

```c
#ifndef GLSLANG_BUILD_INFO
#define GLSLANG_BUILD_INFO

#define GLSLANG_VERSION_MAJOR 11
#define GLSLANG_VERSION_MINOR 0
#define GLSLANG_VERSION_PATCH 0
#define GLSLANG_VERSION_FLAVOR ""

#define GLSLANG_VERSION_GREATER_THAN(major, minor, patch) \
    (((major) > GLSLANG_VERSION_MAJOR) || ((major) == GLSLANG_VERSION_MAJOR && \
    (((minor) > GLSLANG_VERSION_MINOR) || ((minor) == GLSLANG_VERSION_MINOR && \
     ((patch) > GLSLANG_VERSION_PATCH)))))

#define GLSLANG_VERSION_GREATER_OR_EQUAL_TO(major, minor, patch) \
    (((major) > GLSLANG_VERSION_MAJOR) || ((major) == GLSLANG_VERSION_MAJOR && \
    (((minor) > GLSLANG_VERSION_MINOR) || ((minor) == GLSLANG_VERSION_MINOR && \
     ((patch) >= GLSLANG_VERSION_PATCH)))))

#define GLSLANG_VERSION_LESS_THAN(major, minor, patch) \
    (((major) < GLSLANG_VERSION_MAJOR) || ((major) == GLSLANG_VERSION_MAJOR && \
    (((minor) < GLSLANG_VERSION_MINOR) || ((minor) == GLSLANG_VERSION_MINOR && \
     ((patch) < GLSLANG_VERSION_PATCH)))))

#define GLSLANG_VERSION_LESS_OR_EQUAL_TO(major, minor, patch) \
    (((major) < GLSLANG_VERSION_MAJOR) || ((major) == GLSLANG_VERSION_MAJOR && \
    (((minor) < GLSLANG_VERSION_MINOR) || ((minor) == GLSLANG_VERSION_MINOR && \
     ((patch) <= GLSLANG_VERSION_PATCH)))))

#endif // GLSLANG_BUILD_INFO
```

This header was first introduced at version `10.15.3847` (major 10, minor 15, patch 3847). 

To fix this project, I'd probably suggest:
* Add more logic to `src/meson.build` so that it can recognise the new `glslang/build_info.h` header, and extract version information from it. 
* Pass the build-time recognised glslang version into `glslang.cc` via a compiler command line define.
* Remove the `#include <glslang/Include/revision.h>` line from `src/glsl/glslang.cc`.
* Use the new `src/meson.build` glslang version define where ever `GLSLANG_PATCH_LEVEL` is currently used.

If you'd like to discuss this further, may I suggest opening a new issue? Comments on issues / PRs that are closed/merged can easily go unnoticed.

Many thanks,
Ben

Issue #2276, opened 2020-06-17
Add `-g0` to strip all debug info


ShabbyX:
This is a request to add a `-g0` option to strip debug info. This should add `optimizer.RegisterPass(spvtools::CreateStripDebugInfoPass());` to the optimization passes.

The reason for this request is to be able to reduce the binary size of compiled shaders when debug info doesn't really need to be shipped to users.

Note: I took a hack at adding this flag, but I see `Options` being an `int` and has all 32 bits defined. It's passed around in a lot of places but not used at all? I would have changed the type / removed it had it not been exposed in `glslang/Public/ShaderLang.h`. Not sure what's the way forward.

johnkslang:
If you just want an option for controlling logic in `StandAlone.cpp`, just add a `bool` at the top.  E.g., see `targetHlslFunctionality1`.  The `TOptions` is what I think you are referring to as being full, but it is local to that file, for managing command options.

There is a struct `SpvOptions` that looks like how you might want to communicate this request across components within glslang.

johnkslang:
I does sound like a good thing to add, do you still want to try, given the information in the previous comment?

ShabbyX:
Ok, I see, I can use that to set an option in `spvOptions`, so `SpvTools.cpp` can use it. That's what you meant right?

I'll do that some time today.

johnkslang:
Yes, if that works out, thanks.

PR #2278, opened 2020-06-17
Add -g0 command line argument

ShabbyX:
Analogous to gcc, -g0 would strip all debug info.  This is done
regardless of whether optimizations are enabled.

Fixes: #2276

### DIFF
--- a/SPIRV/SpvTools.h
+++ b/SPIRV/SpvTools.h
@@ -52,9 +52,10 @@
 namespace glslang {
 
 struct SpvOptions {
-    SpvOptions() : generateDebugInfo(false), disableOptimizer(true),
+    SpvOptions() : generateDebugInfo(false), stripDebugInfo(false), disableOptimizer(true),
         optimizeSize(false), disassemble(false), validate(false) { }
     bool generateDebugInfo;
+    bool stripDebugInfo;
     bool disableOptimizer;
     bool optimizeSize;
     bool disassemble;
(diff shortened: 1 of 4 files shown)


SPIRV/SpvTools.h
```diff
@@ -70,11 +71,17 @@ void SpirvToolsDisassemble(std::ostream& out, const std::vector<unsigned int>& s
 void SpirvToolsValidate(const glslang::TIntermediate& intermediate, std::vector<unsigned int>& spirv,
                         spv::SpvBuildLogger*, bool prelegalization);
 
-// Apply the SPIRV-Tools optimizer to generated SPIR-V, for the purpose of
-// legalizing HLSL SPIR-V.
-void SpirvToolsLegalize(const glslang::TIntermediate& intermediate, std::vector<unsigned int>& spirv,
```
ShabbyX, line 75: I suspect this was a relic of the past, and changed the function name to reflect what it actually does. HLSL SPIR-V legalization is just a side effect.

johnkslang: A bit that way, but it's an overstatement.

Yes, there is a call for size transforms, but otherwise, its the set of transforms needed for legalization, not a general optimizing.  Mostly, optimizing is the side effect of legalization or size reduction.

SPIRV/SpvTools.h
```diff
@@ -70,11 +71,17 @@ void SpirvToolsDisassemble(std::ostream& out, const std::vector<unsigned int>& s
...
+// Apply the SPIRV-Tools optimizer to generated SPIR-V.  HLSL SPIR-V is legalized through these
+// optimizations as a side effect.
+void SpirvToolsOptimize(const glslang::TIntermediate& intermediate, std::vector<unsigned int>& spirv,
```
johnkslang, line 76: Let's call it what it really does, `SpirvToolsTransforms`

johnkslang:
Looks good, modulo the name change and comment change.  Would like those to be more in the direction of what was already there.

SPIRV/SpvTools.h
ShabbyX: I guess it really depends on what your source is. We use this with GLSL, and this is what's called for -Os, and there is nothing illegal to begin with.

Will update the comment.

SPIRV/SpvTools.h
ShabbyX: Done

PR #2281, opened 2020-06-19
Add more Kokoro configurations

ben-clayton:
Add three new kokoro configurations that can be used to test the CMake and `Android.mk` build rules.

Both `CI-ubuntu-clang-cmake` and `CI-ubuntu-gcc-cmake` take around 2 minutes to complete.

`CI-android-ndk-build` uses the `ndk_test` build script which does not define `APP_ABI`, so builds a [bunch of ABIs by default](https://developer.android.com/ndk/guides/application_mk), and takes around 6 minutes. If we want to shorten this, we can reduce the number of ABIs built by defining the `APP_ABI` parameter in the `Application.mk` file (assuming nothing external is depending on this).

Note: landing this PR does not automatically enable these new presubmit tests - we also need to enable another google-private config (cl/317307422) around the same time that this PR lands.

### DIFF
--- a/kokoro/android-ndk-build/build-docker.sh
+++ b/kokoro/android-ndk-build/build-docker.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Copyright (C) 2020 Google, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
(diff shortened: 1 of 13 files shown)


ben-clayton:
Note: new kokoro tests pass _long_ before AppVeyor even starts for me...

johnkslang:
Looks good.

Issue #2283, opened 2020-06-22
Tracker for adding glslang support for versioned shared objects / DLLs


ben-clayton:
This issue tracks the work being done to add support for versioned shared objects / DLLs for glslang.

- [x] Untangle cyclic build target dependencies. **Status: Done** - #2271
- [ ] Switch to using [Semantic Versioning](https://semver.org/). **Status: WIP** - #2277 
- [ ] Add `SONAME` library versioning. **Status: WIP** - #2277 
- [ ] Reduce exported symbols to public API. **Status: WIP** - #2282
- [ ] Export DLL public API symbols. **Status: WIP** - #2282
- [x] Add presubmit builds for CMake, Android.mk. **Status: Done** - #2281 
- [ ] Add presubmit builds for shared objects. **Status: Not started**
- [ ] Presubmit detection & warnings when making breaking API changes without bumping major version. **Status: Not started**

---

Related issues: #1255, #1421, #1484, #2147, #2267

johnkslang:
Note that #2277 is still WIP, but now #2282 is not.  Is there any dependency?

ben-clayton:
I want to add presubmit builds for SOs / DLLs to check #2282 is safe to land.
I've locally tested #2282, but I always like to get a second opinion.

I'll do that now.

jspanchu:
Few symbols that are exported from glslang shared library are not at all defined. For example, symbols of `TShader` are exported in `glslang`, however they are declared in `MachineDependent` which is a hard-coded static library built without `GLSL_EXPORT`.

It is probably simpler by making your implemantation modules `GenericCodeGen`, `MachineDependent`, `OSDependent` object libraries. Does anyone downstream really link to them?

arcady-lunarg:
The symbols tagged with GLSLANG_EXPORT in #3705 are the proposed public interface to the `SPIRV` library (which got merged into glslang proper recently). If anyone has any comments about that, I'm going to leave the PR open for a week or so.

arcady-lunarg:
This is basically done, there are only a few pieces missing, namely testing with the DLL build on Windows in CI, and testing the ABI for backwards compatibility.

PR #2286, opened 2020-06-24
Kokoro: Split linux cmake cfgs into static/shared

ben-clayton:
Allows for testing of generation of both static libraries and shared objects.

The old configs are staying in place until I'm confident everything is working correctly.

Issues: #1421, #1484, #2283


ben-clayton:
[New tests pass](https://github.com/KhronosGroup/glslang/pull/2287)

These new tests are currently only configured for the `kokoro-testing` branch. Once this PR lands, I'll update Kokoro to test `master` instead, and put up a second PR to remove the old configs.

PR #2307, opened 2020-07-01
Location validation, with update to SPIR-V tools validating location

johnkslang:
1. Change glslang test framework to more broadly use automapping binding/location.

    This adds or changes binding/location decorations in 100s of shaders.
    It also allows more output (spv.register.autoassign.rangetest.frag)
    due to allowing ioMap() to fail.

2. Update SPIRV-Tools: Update to location-validation in SPIRV-Tools.

    This introduces five new "Validation failures":
    - baseResults/hlsl.semantic.vert: issue with gl_ClipDistance/CullDistance
    - baseResults/spv.430.vert: issue gl_ClipDistance
    - baseResults/spv.450.tesc: still unknown
    - baseResults/spv.dataOut.frag: gl_FragData should not be supported, problem with front end
    - baseResults/spv.meshShaderPerViewUserDefined.mesh: seems okay, maybe a problem with SPIRV-Tools

So, rather than hundreds of failures we might have gotten, there are these 5.  Will look more at root causes to those before merging, but this can be broken into separate steps.


johnkslang:
Because this blocks combining latest of glslang and SPIRV-Tools, I will merge now and look at the 5 questionable tests in parallel. To be clear, there are no regressions.

PR #2314, opened 2020-07-03
Add --quiet option.

marcinslusarz:
Being quiet should have been the default, but I guess it's too late
now to change it.

---
This change is [Reviewable](https://reviewable.io/reviews/khronosgroup/glslang/2314)

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -107,6 +107,7 @@ bool SpvToolsDisassembler = false;
 bool SpvToolsValidate = false;
 bool NaNClamp = false;
 bool stripDebugInfo = false;
+bool beQuiet = false;
 
 //
 // Return codes from main/exit().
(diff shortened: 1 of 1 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -69,38 +69,39 @@ extern "C" {
...
+    EOptionDebug                = (1LL << 26),
+    EOptionStdin                = (1LL << 27),
+    EOptionOptimizeDisable      = (1LL << 28),
+    EOptionOptimizeSize         = (1LL << 29),
+    EOptionInvertY              = (1LL << 30),
+    EOptionDumpBareVersion      = (1LL << 31),
+    EOptionQuiet                = (1LL << 32),
 };
 bool targetHlslFunctionality1 = false;
 bool SpvToolsDisassembler = false;
```
johnkslang, line 107: New options have been getting added to this list here, rather than needing another bit.  I think it's actually better, the reasons for having a bitmask now obsoleted.

Can you just add another bool here instead?

marcinslusarz: Done

PR #2317, opened 2020-07-03
gn: Optionally disable optimizations and HLSL

ShabbyX:
To reduce the binary size of ANGLE, a gn override is added
(glslang_angle) which:

- Controls whether ENABLE_OPT=1 is set
- Customizes the build for the Vulkan backend of ANGLE.  As a first
  step, this removes HLSL functionality which together with no
  optimization shave ~2.5MB off of ANGLE's binary size.

Upcoming changes will add a macro for GLSLANG_ANGLE similar to
GLSLANG_WEB that will strip features from glslang to support only what
ANGLE needs.

Signed-off-by: Shahbaz Youssefi <ShabbyX@gmail.com>

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -51,164 +51,183 @@ spirv_tools_dir = glslang_spirv_tools_dir
 config("glslang_public") {
   include_dirs = [ "." ]
 
-  defines = [ "ENABLE_HLSL=1" ]
+  if (!glslang_angle) {
+    defines = [ "ENABLE_HLSL=1" ]
+  }
 }
 
-source_set("glslang_sources") {
-  public_configs = [ ":glslang_public" ]
+template("glslang_sources_common") {
+  source_set(target_name) {
+    public_configs = [ ":glslang_public" ]
 
-  sources = [
-    "OGLCompilersDLL/InitializeDll.cpp",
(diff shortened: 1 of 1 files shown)


ShabbyX:
Open to suggestions regarding whether:

- Should I make the changes less granular? More granular? I can disable optimizations with `glslang_angle`, so there is only one new parameter. On the other hand, I can add another parameter like `glslang_enable_hlsl` to control whether HLSL should be included. As far as ANGLE and chromium are concerned, having `glslang_angle` is enough. Are there other users of `BUILD.gn`?
- Are names `glslang_angle` and `GLSLANG_ANGLE` sound good to you?

ShabbyX:
I removed `glslang_enable_opt` for now and made it take its value from `!glslang_angle`. Existing users, if any, would be unaffected.

ShabbyX:
Travis failures are a cmake build failure.

ben-clayton:
Sorry for the Travis failures.
https://github.com/KhronosGroup/glslang/pull/2316 has landed which would fix this. If you rebase this change, it should turn green.

ben-clayton:
I'm also not sure why the kokoro presubmits didn't kick in...

PR #2320, opened 2020-07-05
gn: Fix `gn gen --check` by adding missing dependency

ShabbyX:


ShabbyX:
@ben-clayton @johnkslang caught by `gn gen --check` which is run by Chromium bots.

Blocks roll into ANGLE and Chromium.

PR #2321, opened 2020-07-05
Emit OpLine before OpFunction

inequation:
This ensures useful source line association for other tools, such as [AMDVLK's fork of SPIRV-LLVM-Translator](https://github.com/GPUOpen-Drivers/llpc/blob/2bd4eb104a93a61614f398826814e60cfc312fbc/llpc/translator/lib/SPIRV/SPIRVToLLVMDbgTran.cpp#L64). Emitting `OpLine` before `OpFunction` enables instructions generated automatically _in support of a function_ (such as setting a hypothetical stack frame, or - in the case of AMD GPUs - moving data out of implicitly pre-populated registers) to be associated with source code, rather than being untraceable.

After this change, source file name information survives the entire compilation pipeline, from source to the DWARF info in the AMDGPU ELF.

### DIFF
--- a/SPIRV/SpvBuilder.h
+++ b/SPIRV/SpvBuilder.h
@@ -351,6 +351,10 @@ public:
     Function* makeFunctionEntry(Decoration precision, Id returnType, const char* name,
         const std::vector<Id>& paramTypes, const std::vector<std::vector<Decoration>>& precisions, Block **entry = 0);
 
+    // Update an existing function made with makeFunctionEntry() to given
+    // source-line association. Improves debug information quality.
+    void updateFunctionEntrySourceLine(const char* name, const char* filename, int line);
+
     // Create a return. An 'implicit' return is one not appearing in the source
     // code.  In the case of an implicit return, no post-return block is inserted.
     void makeReturn(bool implicit, Id retVal = 0);
@@ -853,6 +857,9 @@ public:
 
     // map from include file name ids to their contents
     std::map<spv::Id, const std::string*> includeFiles;
+
+    // map from entry point names to their function indices
+    std::unordered_map<std::string, size_t> functionNames;
 
     // The stream for outputting warnings and errors.
     SpvBuildLogger* logger;
--- a/Test/baseResults/hlsl.pp.line2.frag.out
+++ b/Test/baseResults/hlsl.pp.line2.frag.out
@@ -9,7 +9,7 @@ hlsl.pp.line2.frag
                               EntryPoint Fragment 5  "MainPs" 71 75
                               ExecutionMode 5 OriginUpperLeft
                1:             String  "hlsl.pp.line2.frag"
-              17:             String  "foo.frag"
+              11:             String  "foo.frag"
               32:             String  "foo.h"
               42:             String  "foo2.h"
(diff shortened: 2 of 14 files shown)


inequation:
Apologies for taking so long, I just resurrected the project that I need this for after a hiatus. I've updated the test suite to account for the newly added line statements.

greg-lunarg:
Sorry for the delay. Looking now...

Test/baseResults/hlsl.pp.line2.frag.out
```diff
@@ -124,32 +124,35 @@ PS_OUTPUT MainPs ( PS_INPUT i )
...
-              79:   11(fvec4) CompositeExtract 78 0
+              78:13(PS_OUTPUT) FunctionCall 16(@MainPs(struct-PS_INPUT-vf21;) 76(param)
+              79:   12(fvec4) CompositeExtract 78 0
                               Store 75(@entryPointOutput.vColor) 79
                               Return
                               FunctionEnd
-15(@MainPs(struct-PS_INPUT-vf21;):12(PS_OUTPUT) Function None 13
-           14(i):     10(ptr) FunctionParameter
-              16:             Label
+                              Line 1 23 0
```
greg-lunarg, line 149: Shouldn't this be Line 11 23 0?

Test/baseResults/hlsl.pp.line2.frag.out
```diff
@@ -124,32 +124,35 @@ PS_OUTPUT MainPs ( PS_INPUT i )
...
+              74:             TypePointer Output 12(fvec4)
 75(@entryPointOutput.vColor):     74(ptr) Variable Output
+                              Line 11 23 0
        5(MainPs):           3 Function None 4
                6:             Label
            69(i):     10(ptr) Variable Function
        76(param):     10(ptr) Variable Function
-                              Line 17 23 0
+                              Line 11 23 0
+                              Line 11 23 0
```
greg-lunarg, line 138: Curious why we are getting two OpLines here now? Not a big deal, but this is coming through in the actual SPIR-V as well, so if this could be cleaned up easily, that would be nicer.

SPIRV/SpvBuilder.h
```diff
@@ -854,6 +858,9 @@ class Builder {
     // map from include file name ids to their contents
     std::map<spv::Id, const std::string*> includeFiles;
 
+    // map from entry point names to their function indices
+    std::unordered_map<std::string, size_t> functionNames;
```
greg-lunarg, line 862: Maybe a better name: functionIndex?

greg-lunarg (changes requested):
Thanks for the submission! One possible correctness issue, one small inefficiency and one variable rename.

qingyuanzNV:
Hi, is this PR still going? We have an internal patch which aims the exactly same thing with a slightly different approach (we set the line info in TGlslangToSpvTraverser::visitAggregate). What's the expected direction on this?

inequation:
Hi @qingyuanzNV! Personally, I have no resources right now to bring this to fruition. If you do, please go ahead and get yours integrated!

arcady-lunarg:
This was implemented in #2950.

PR #2322, opened 2020-07-05
gn: Fix dawn tests in Chromium

ShabbyX:
Dawn tests use shaderc, which assumes glslang has HLSL support.  This
change makes HLSL support also follow template arguments, and changes
the target names such that glslang_sources will remain the "has all
features" target and the new glslang_lib_sources would be what ANGLE
would use.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -50,15 +50,19 @@ spirv_tools_dir = glslang_spirv_tools_dir
 
 config("glslang_public") {
   include_dirs = [ "." ]
+}
 
-  if (!glslang_angle) {
-    defines = [ "ENABLE_HLSL=1" ]
-  }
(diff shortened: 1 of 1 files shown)


johnkslang:
Does this assumption make glslang for Dawn bigger than necessary?  Should shaderc be changed instead to not assume HLSL support?

ShabbyX:
glslang for dawn is used only for testing. It's not shipped to anyone. Before my changes to `glslang/BUILD.gn`, HLSL and OPT support where both unconditionally enabled. I was trying to make glslang smaller for ANGLE *as well as* dawn testing which didn't work out.

With this change, I keep glslang in dawn testing as it previously was. Future work with shaderc could take advantage of this reduction in size, but there's really not a big reason.

PR #2326, opened 2020-07-07
Remove GLSLANG_BUILD_PIC flag

ben-clayton:
On closer inspection, it appears that nearly all the targets use the `POSITION_INDEPENDENT_CODE` target option anyway.
Simplify all this away by always being PIC.


ben-clayton:
Based on https://github.com/KhronosGroup/glslang/pull/2325.

PR #2336, opened 2020-07-13
Kokoro: Run tests for CMake builds

ben-clayton:


ben-clayton:
Tests are failing with:

```
Testing SPV Debug Information
2,3d1
< error: SPIRV-Tools Validation Errors
< error: Invalid SPIR-V binary version 1.3 for target environment SPIR-V 1.0 (under OpenGL 4.5 semantics).
```

Which I also observe locally.

johnkslang:
I randomly checked CI-ubuntu-gcc-cmake-static.  It keeps reporting:

```
../build/install/bin/glslangValidator: error while loading shared libraries: libglslang.so.10
```

when trying to execute the tests.  It earlier said:

```
CMake Warning (dev) at /bin/cmake-3.17.2/share/cmake-3.17/Modules/GNUInstallDirs.cmake:225 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
```

I see CI-ubuntu-gcc-cmake-static reporting the 1.3 vs. 1.0 thing.  The debug block runs 3 tests, and the middle one is indeed making SPIR-V 1.3.  Maybe it's telling the validator it made 1.0, and the validator complains?  But, I don't see that locally, nor in the last commit to master, so it must also be about something being out of date somewhere.

ben-clayton:
> I randomly checked CI-ubuntu-gcc-cmake-static. It keeps reporting:

You mean CI-ubuntu-gcc-cmake-shared, right? Good spot - I need to set `LD_LIBRARY_PATH` due to the non-standard install path.

As for the `CMAKE_INSTALL_LIBDIR` warning - I'll take a look. I vaguely remember this being raised before I started working on the shared libs, but I'll investigate all the same.

ben-clayton:
As for the:

```
Testing SPV Debug Information
2,3d1
< error: SPIRV-Tools Validation Errors
< error: Invalid SPIR-V binary version 1.3 for target environment SPIR-V 1.0 (under OpenGL 4.5 semantics).
```

This is failing because [these two lines](https://github.com/KhronosGroup/glslang/blob/f5add0b20d2a58957164fe48fefd63d94de44ea9/Test/baseResults/spv.debugInfo.1.1.frag.out#L2-L3) are expected but aren't actually emitted any more. Has something been relaxed recently?

ben-clayton:
Oh hold on - these Kokoro tests aren't calling `update_glslang_sources.py` - by not having these in `External/`, are we disabling validation or something?

johnkslang:
Yes, if SPIR-V tools are not present, glslang operates anyway, without them, or links to them from another location, which might pick up the wrong version.

ben-clayton:
Switched to draft due to #2361

PR #2338, opened 2020-07-13
Regenerate grammar with Bison 3.0.4

ben-clayton:


ben-clayton:
Clearly not the cause of the mac failures. Closing.

PR #2339, opened 2020-07-13
CMake: Move project() to top of CMakeLists.txt

ben-clayton:
And declare project `LANGUAGES` - should fix the following warning:

```
CMake Warning (dev) at /bin/cmake-3.17.2/share/cmake-3.17/Modules/GNUInstallDirs.cmake:225 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
```

Also remove `NOTICE` from message() about PCHs - it seems to print this in the actual message, contrary to the documentation where it is used as a severity.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,9 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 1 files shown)


ben-clayton:
Confirmed `CMAKE_INSTALL_LIBDIR` warning has now disappeared.

PR #2340, opened 2020-07-13
Fix failing gtest

zoddicus:
This test is failing for me at HEAD on MacOS.


zoddicus:
This is failing on appveyor

johnkslang:
Just changing test results will certainly cause all the test runs that are currently passing to start failing, yes?

If this is nondeterministic, we'll have to find the cause.

johnkslang:
#2343 fixes the real problem.

PR #2345, opened 2020-07-14
also search global sequences for live variables

mbechard:
when traversing the AST to find live UBOs etc, also traverse
module-level variable assignment, not just function calls.

Otherwise for cases such as this at the module level:

`float f = SomeUniform;`

SomeUniform won't be determined to be 'live'.

Is checking for EOpSequence the correct OpType here, or is there something more appropriate?

### DIFF
--- a/glslang/MachineIndependent/LiveTraverser.h
+++ b/glslang/MachineIndependent/LiveTraverser.h
@@ -74,14 +74,33 @@ public:
         for (unsigned int f = 0; f < globals.size(); ++f) {
             TIntermAggregate* candidate = globals[f]->getAsAggregate();
             if (candidate && candidate->getOp() == EOpFunction && candidate->getName() == name) {
-                functions.push_back(candidate);
+                destinations.push_back(candidate);
                 break;
             }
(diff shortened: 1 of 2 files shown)


johnkslang:
Why limit it to a nested EOpSequence?  The code was looking inside each function, and you want to add looking outside each function, it just seems like you'd want to traverse anything that's not a function.  It could be you picked them all up, just possibly fragile, if not everything stays two levels deep in the tree at global scope.

mbechard:
Should it just traverse everything then and remove the code that is selectively just doing functions?

johnkslang:
That sounds reasonable.  I don't know if there's a subtlety, but the obvious thing now is having a traversal the just selects functions, and a traversal that just skips functions, wondering if it should all just be one traversal.  But, whatever solves the actual problem.

mbechard:
Went a different route. Now when global variables are visited it adds those to the list of things to traverse. That way globals that aren't referenced anywhere aren't going to end up tagging the variable source as 'live'.

PR #2360, opened 2020-07-29
Update some language usage.

dj2:
This CL updates various bits of language to be more inclusive. The
updates are from the Android respectful code list
(https://source.android.com/setup/contribute/respectful-code)


johnkslang:
This includes too many changes to spaces.  Please separate those (or better, skip them, to help git blame work a bit better).

"Native" is a perfectly nice respectful and highly communicative word, those information-losing changes don't need to be made.

Test/baseResults/hlsl.intrinsics.comp.out
```diff
@@ -715,7 +715,6 @@ local_size = (1, 1, 1)
 0:?     'inU0' (layout( location=3) in 4-component vector of uint)
 0:?     'inU1' (layout( location=4) in 4-component vector of uint)
 
-Validation failed
```
johnkslang, line 718: Why are you changing test results?

dj2: I ran the update results flag to glslangtests and this is what it spit out? Was that the wrong thing to do?

dj2:
Closing this due to the extraneous whitespace changes and I don't know what happened with the test suite.

PR #2362, opened 2020-07-31
Use GLSLANG_ANGLE to strip features to what ANGLE requires

ShabbyX:
This change strips a few features similar to GLSLANG_WEB but doesn't
remove every detail like the latter.  It also hardcodes profile/version
to core/450.

In particular, TBuiltIns::initialize is specialized to remove most of
what is not supported or won't be supported by ANGLE.  The result of
this function is parsed with TParseContext::parseShaderStrings which is
a performance bottleneck.

This change shaves about 300KB off of ANGLE's binary size and reduces
the cost of SetupBuiltinSymbolTable to nearly a sixth.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -109,7 +109,6 @@ template("glslang_sources_common") {
       "SPIRV/SpvBuilder.cpp",
       "SPIRV/SpvBuilder.h",
       "SPIRV/SpvPostProcess.cpp",
-      "SPIRV/SpvTools.cpp",
       "SPIRV/SpvTools.h",
       "SPIRV/bitutils.h",
       "SPIRV/disassemble.cpp",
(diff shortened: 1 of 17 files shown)


ShabbyX:
@johnkslang This is a resubmit of #2330 with only the first of the two commits (which is higher priority). Regarding the four issues with that PR:

> 1. GitHub reported it passing all tests, but this is not true, as almost no tests actually ran. They were later caught when tests were running again, and we traced things back to this PR.
> 2. The grammar does not build without generating warnings that should not be generated; this needs to be cleaned up.
> 3. The built-in symbol table parsing reports problems on macos, which might need addressing, if not the same issue is in the previous bullet.
> 4. I also saw this time (missed the first time) changing some 'int' to 'bool', making types dependent of flags, and this concerns me too. (There was a reason to use 'int' even though it seems like it should be 'bool'.)

1. It seems that the tests run fine now AFAICT.
2. This PR doesn't include the grammar change
3. Unknown (I don't have macs to test this). Hopefully the same issue as 2. Will you be able to verify that there are no issues with this PR?
4. Changed the `bool`s back to `int`. Had to add a few `!= 0`s to silence a warning regarding `a` being `int` in `if (a || b)` (with clang suggesting `|` instead of `||`).

Issue #2366, opened 2020-08-06
Move to correct semantic checking for extensions for buffer references and the 8/16 int/float types.


johnkslang:
The architecture should be:  "write for all features, when a feature gets used, see if it's allowed", not "decide what features to recognize based on what extensions are enabled."  This isn't just glslang, it's pretty fundamental to what's actually going with a massive combinatorial set of features, and has these advantages:

- simpler control flow: write for one superset version of all variants, not choosing different paths through the code base based on the variant
- better error messages: "to use feature XYZ, you need extension ABC"
- better performance: the extension map is only searched when the feature gets used, not at every use of an operator/keyword/...
- separates parsing semantic checking into the language-specific front ends, outside the common AST builder they use
- essentially, it's declarative instead of procedural, and that has a variety of advantages

Mostly, parsing already does this the correct way, but just a couple new features flipped it around and started asking "what's enabled, I'll go recognize for that", pushing this (GLSL) questioning into the (common) AST builder, and missing out on the list above.

What's worse is that even though we had a robust way of asking if a feature had the required extensions enabled in the GLSL front end, a parallel yet-lower-quality version of that was added to the AST (which should not be dynamically tracking extension state changes), and as a result lead to issue #2358, where we (re)discovered this.  It wasn't too obvious code reading due to the very similar names used.

johnkslang:
This is a focal point of the issue, from the code base:

```
    // JohnK: I think this function should go away.
    // This data structure is just a log to pass on to back ends.
    // Versioning and extensions are handled in Version.cpp, with a rich
    // set of functions for querying stages, versions, extension enable/disabled, etc.
    bool extensionRequested(const char *extension) const {
        auto it = requestedExtensions.find(extension);
        if (it != requestedExtensions.end()) {
            return (it->second == EBhDisable) ? false : true;
        }
        return false;
    }
```

PR #2367, opened 2020-08-07
Remove incorrect style of extension-based semantic checking.

johnkslang:
Correct style is to recognize all of GLSL, and for optional features, after the feature is recognized, put a check for the right extensions being present in that path.  (Not to see what features are present and then go down different paths while parsing.)

See #2366 for more detail.

There were two main cases, a few checks for GL_EXT_buffer_reference2, and then a whole bunch of checks for what conversions are allowed.

This is draft until getting them all fixed.


johnkslang:
The numeric extensions are written to heavily rely on switching semantics at the last moment based on the extensions, rather than just seeing if the requested action is allowed by the extensions.  Am fixing what can be fixed locally, especially:

- fixing obvious bugs, like where HLSL code was moved under GLSL extension code (meaning it was dead and lost its original meaning)
- deferring checking the extensions until it the actual situations arise that need the checks
- documenting other likely bugs I'm uncovering in the code

Final step is still hoping to decouple (or fully unify) the data needed for querying these extensions from the list of extensions to be passed to back ends.

PR #2369, opened 2020-08-10
GLSLANG_EXPORT for C APIs.

ezdiy:
Fixes FFI breakage introduced in #2283

### DIFF
--- a/SPIRV/CInterface/spirv_c_interface.cpp
+++ b/SPIRV/CInterface/spirv_c_interface.cpp
@@ -79,7 +79,7 @@ static EShLanguage c_shader_stage(glslang_stage_t stage)
     return EShLangCount;
 }
 
-void glslang_program_SPIRV_generate(glslang_program_t* program, glslang_stage_t stage)
+GLSLANG_EXPORT void glslang_program_SPIRV_generate(glslang_program_t* program, glslang_stage_t stage)
 {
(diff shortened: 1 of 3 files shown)


johnkslang:
We normally have 11 or twelve checks, but this only saw 2.  This PR though is a good candidate to see running on more bots, somehow, before merging.

@ben-clayton is taking a break right now, but would also be nice to have his input on the change itself.

ben-clayton:
@johnkslang - you need to apply the "kokoro: run" label for the other bots to run. 
The change LGTM.

ezdiy:
One issue currently not taken into account is https://github.com/KhronosGroup/glslang/blob/master/StandAlone/resource_limits_c.h

This compiles into `libglslang-default-resource-limits.so` currently not under `-fvisibility=hidden` regime in the makefiles, hence exports still working (at least on linux, not so sure about win32). Making it export properly would ideally involve collapsing all that into `glslang_c_interface.c` / `glslang_c_interface.h`, as the separation to additional library looks to me like some temporary workaround to deal with circular dependency yet to be refactored. Alternatively turn whole C api into separate library that has its own linking mode - ie even if c++ is static, c-export may be dynamic, and perhaps make that a cmake default.

The reasoning here is that FFI from non-C languages (think LWJGL...) *demand* shared c-export library, yet don't care about nature of t he c++ side whatsoever.

PR #2371, opened 2020-08-14
Use --test-root to pass files to Bazel tests.

RafaelMarinheiro:
The current implementation makes tests fail when it is imported from a different Bazel workspace.
We fix that by using the --test-root flag to pass the rootpath to the tests.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -51,7 +51,10 @@ py_binary(
 
 genrule(
     name = "gen_build_info_h",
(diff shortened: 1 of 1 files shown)


johnkslang:
This seems to have broken the Bazel test.  See https://source.cloud.google.com/results/invocations/d0633006-b45c-4f57-b00a-66c2153ba1d8/targets.

Is this legitimate?  Should we revert it?

PR #2373, opened 2020-08-18
build: set SOVERSION on all libraries

jengelh:
Commit fbe9a23baf2cb020fe3f80d68d972e2a97d9b954 erroneously
only added SOVERSION to libglslang.so, but none of the others
that are produced.

Signed-off-by: Jan Engelhardt <jengelh@inai.de>


jeremy-lunarg:
This shouldn't have happened. We renamed the default branch yesterday per Khronos policy. I didn't explicitly close this PR and it won't let me reopen this PR. Would you be willing to reapply your changes onto main and open a new PR?

PR #2377, opened 2020-08-23
4.array size fix and related

ShchchowAMD:
1. Expand meaning of implicit array size.

    1.1 The purpose of this change is to allow us to determine whether an array size is implicitly assigned or explicitly assigned. Currently, both ways will use changeOuterArraySize(), so we don't know where the front-size comes from.

    1.2 Now when implicitArraySize == 0, it means we have assigned an explicit array size to the identifier.

    1.3 This is introduced for some combined array size check, like gl_MaxCombinedClipCullDistance. We need to know whether gl_ClipDistance / gl_CullDistance has been redeclared. And only explicit array size will be used to check the limitation.

2. Add extension ARB_cull_distance and GL_EXT_clip_cull_distance. Checked as both arrays are supported in ES 320.

3. Add array size merging and primitive layout merging when using multi-shader mode in single stage(GS).

    3.1 Implicit size of declared input array should be less or equal to same explicitly declared identifier/primitive layout.

    3.2 Explicit array size of input array should be exactly equal to primitive layout.

4. Sample mask array should depend on implementation. One bit for one sampler.

5. member "resources" should not be redeclared in parseVersion, otherwise there would be wrong match / empty "resources" when using parent class.


johnkslang:
I don't think we should be reporting arrays as having 0 size.

I don't think ES 320 has clipdistance and culldistance.

ShchchowAMD:
For array size, would it be better if I use a new variable isImplicitlySized to check this? Need to know sizing type when checking combined size.

For clip/cull distance, I found it in GLSL ES 3.2 Spec 14.2.5, should they only depend on extension EXT_clip_cull_distance?

johnkslang:
> We need to know whether gl_ClipDistance / gl_CullDistance has been redeclared. And only explicit array size will be used to check the limitation.

Why shouldn't implicitly sized arrays be checked?

> For clip/cull distance, I found it in GLSL ES 3.2 Spec 14.2.5, should they only depend on extension EXT_clip_cull_distance?

This looks like a mistake in the asciidoc conversion.  They got partially listed, but are never discussed or specified, and did not appear in (at least some) earlier revisions of the same version.  It would be new functionality, so I think you found a specification bug.

I reported this just now in https://github.com/KhronosGroup/GLSL/issues/132.

ShchchowAMD:
Hi John, if there's a fragment shader case as :
 (gl_MaxClipDistance == gl_MaxCullDistance ==gl_MaxCombinedClipCullDistance == 8)

"
    #version 320 es
    #extension GL_EXT_clip_cull_distance : enable
    in  highp float gl_ClipDistance[8];
    void main() {
    ....
"

In this case , we only redeclare clip distance with max size when cull distance is unsized (implicit size == 1).

I treat this case as right, so that combined size should be 0 + 8 <= gl_MaxCombinedClipCullDistance.

That's the reason I want to exclude implicit size in this case.

Otherwise, gl_ClipDistance.implicitArraySize + gl_CullDistance.implicitArraySize > gl_MaxCombinedClipCullDistance.

johnkslang:
> For clip/cull distance, I found it in GLSL ES 3.2 Spec 14.2.5, should they only depend on extension EXT_clip_cull_distance?

>  I reported this just now in KhronosGroup/GLSL#132.

Update: This is getting removed from the GLSL ES specification.  They were actually only in the non-normative SPIR-V mapping section.

ShchchowAMD:
Hi john, as you mentioned, both gl_ClipDistance and gl_CullDistance has been removed from normative ES version.

Now they would only depend on EXT_clip_cull_distance.

ShchchowAMD:
Hi john,

Is there any concerns for this PR?

I'm looking at issues and feedback for those waiting PRs currently.

PR #2378, opened 2020-08-24
Preprocessor related issue fix

ShchchowAMD:
1. Accoding to ESSL spec : All macro names containing two consecutive underscores ( __ ) are reserved for future use as predefined macro names. (Report warning/error if relaxedError/Default).

2. This part isn't 100% sure. (Whether we should treat 1.23 in following statement as unrecognized).
According to spec: If an implementation does not recognize the tokens following #pragma, then it will ignore that pragma, so report a compile-time warning intead of error for the following statement:
#pragma debug(1.23)

3. The 'defined' macro should be allowed to expand (if relaxed error) and \'__LINE__\' should be allowed to be replaced with its original line number (otherwise, other expanding macros may change this value).

5. Add missing type int16 and uint16.


johnkslang:
I think you'll find the "__" rules to be version dependent.

The general intended case is for use in middleware, so the point was to discourage use in a shader, but not make it completely illegal.  However, that got expressed slightly different ways, and there were people who wanted to enforce it differently for different version, to match.

I don't want to go figure out all the history again until after you're sure you are correcting something that is incorrect with respect to that. Let me know, thanks.

ShchchowAMD:
Fixed version dependent issue. Now would still report errors for default setting. But using relaxedError to control behaviour of error reporting.

Those cases are from CTS group:

KHR-GL46.shaders.preprocessor.invalid_conditionals.tokens_after_*
dEQP-GLES2.functional.shaders.preprocessor.pragmas.pragma_unrecognized_debug_*
dEQP-GLES2.functional.shaders.preprocessor.basic.identifier_with_double_underscore_*
dEQP-GLES3.functional.shaders.preprocessor.conditional_inclusion.define_defined_*

#extension macro declaration now should allow to be after non-pp tokens, as discussed in:
https://github.com/KhronosGroup/glslang/pull/1982
This part will be remained until adding state table mentioned as above.

ShchchowAMD:
Hi John, thanks for your reviews. I've fixed mistakes with version dependency.

Now I'm trying to use relaxedError to determine whether glslang should report an error or waning in pp process.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -246,7 +246,11 @@ void TParseContext::handlePragma(const TSourceLoc& loc, const TVector<TString>&
         else if (tokens[2].compare("off") == 0)
             contextPragma.optimize = false;
         else {
-            error(loc, "\"on\" or \"off\" expected after '(' for 'optimize' pragma", "#pragma", "");
+            if(relaxedErrors())
+                //  If an implementation does not recognize the tokens following #pragma, then it will ignore that pragma.
```
johnkslang, line 250: If we really follow this rule, there should never be an error, yes?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -270,7 +274,11 @@ void TParseContext::handlePragma(const TSourceLoc& loc, const TVector<TString>&
         else if (tokens[2].compare("off") == 0)
             contextPragma.debug = false;
         else {
-            error(loc, "\"on\" or \"off\" expected after '(' for 'debug' pragma", "#pragma", "");
+            if(relaxedErrors())
+                //  If an implementation does not recognize the tokens following #pragma, then it will ignore that pragma.
```
johnkslang, line 278: Same comment.

I do think the warning is a good idea, just perhaps unconditionally.

ShchchowAMD:
As suggested, both warnings are now unconditional.

PR #2379, opened 2020-08-25
Fix scope definition in ES 100.

ShchchowAMD:
1. Add tokens related to IMAGE SHADOW.

2. Allow redefinition with a same name in parameter list. Only in ES 100.
" If it is declared as a parameter in a function definition, it is scoped until the end of that function definition.  A function body has a scope nested inside the function’s definition."

### DIFF
--- a/Test/baseResults/100scope.vert.out
+++ b/Test/baseResults/100scope.vert.out
@@ -1,5 +1,4 @@
 100scope.vert
-ERROR: 0:5: 'a' : redefinition 
 ERROR: 0:17: 'b' : function name is redeclaration of existing name 
 ERROR: 0:20: 'c' : redefinition 
 ERROR: 0:22: 'f' : redefinition 
@@ -13,7 +12,7 @@ ERROR: 0:57: 'z' : undeclared identifier
 ERROR: 0:57: 'z' : redefinition 
 ERROR: 0:73: 'degrees' : can't use function syntax on variable 
 ERROR: 0:76: 'vertex-shader struct output' : not supported for this version or the enabled extensions 
-ERROR: 14 compilation errors.  No code generated.
+ERROR: 13 compilation errors.  No code generated.
 
 
 Shader version: 100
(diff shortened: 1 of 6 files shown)


johnkslang:
But, these types don't exist.  It will be backward incompatible.  What's missing here?

ShchchowAMD:
I‘ll do further check.
Seems they should be all reserved.

ShchchowAMD:
Hi John,

I found image2DShadow and other 3 tokens were in the reserved key words list from version 140 to 410.

If I understand right, those four removed tokens should not be added back to grammar right?

ShchchowAMD:
Remove those 4 tokens.

Test case for scope definition in Version 100 is : 

```
 #version 100 es
 int func (int inp, int x)
  { 
 
 int x = 5; return inp + x - 5; 
 
 }
``` 

should be treated as 

```
#version 100 es
int func (int inp, int x) 
{ 
     {
           int x = 5; return inp + x - 5; 
     }
}
```

johnkslang:
Can I ask what's motivating changing the way scoping works?  Is it only from reading the specification, or is it something else as well?

ShchchowAMD:
Hi John,

Those issues are found during debugging with CTS test cases. The motivation is to pass the test if something against the specification.

Related part of cases' shader sources are attached in previous comment.

ShchchowAMD:
Above shaders could be found in CTS cases:
dEQP-GLES2.functional.shaders.scoping.valid.local_variable_hides_function_parameter_vertex
dEQP-GLES2.functional.shaders.scoping.valid.local_variable_hides_function_parameter_fragment

Test/baseResults/100scope.vert.out
```diff
@@ -1,5 +1,4 @@
 100scope.vert
-ERROR: 0:5: 'a' : redefinition
```
johnkslang, line 2: Do we still have other tests that are testing this for the normal case?  I don't want to lose this test.  Is ES 100 the only exception to the normal scoping rules?

ShchchowAMD:
If I understand right, this only occurs when version = 100 ES, same as declared version in cts case.

Maybe I can copy this point to version 300 and add comments, which could also show the difference.

johnkslang:
Yes, or somehow verify we test the other scoping rule.

ShchchowAMD:
Hi john, we could found same codes in 110scope.vert.

They could be well compared to see the difference of redefinition scoping in both versions.

PR #2380, opened 2020-08-26
Suppress two override suggestion warnings.

null77:
We're turning these on in ANGLE and want to keep the build clean.


null77:
@johnkslang PTAL

Note, only applies to the GN file.

PR #2381, opened 2020-08-27
Add missing common constants/operations.

ShchchowAMD:
Add missing common constants/operations.

Refine extension dependency for common operations.

Add error check for SampleMask.


ShchchowAMD:
1. For interpolation part, those issues are found through cts case group:
dEQP-GLES31.functional.shaders.multisample_interpolation.*
 Interpolation builtin function should use input variables, excluding their members.

2. For constant operation and missing types, they are found throug cts group:
dEQP-GLES3.functional.shaders.constant_expressions.builtin_functions.*
Add the missing built-in functions to fold some functions operations to the constant:
2.1. Add the angle and trigonometry functions: sinh, cosh, tanh, asinh, acosh, atanh;
2.2. Add the bit conversion functions: floatBitsToInt, floatBitsToUint, intBitsToFloat, uintBitsToFloat;
2.3. Add the pack/unpack functions: packUnorm2x16 , packSnorm2x16 , packHalf2x16, unpackUnorm2x16 , unpackSnorm2x16, UnpackHalf2x16;
2.4. Add the matrix operation functions: transpose , inverse , determinant , multiply;

3. For initialize.cpp ( or the extension dependency part), basic constant expression / builtin functions,
    This is also a refine version for common constant expressions and their dependency to extensions.
    The original motivation is to fix cts group:
    KHR-GL46.shaders45.shader_integer_mix*

What happened here is:
3.1. Add extension GL_EXT_shader_integer_mix related variables
3.2. Modify BaseFunctions to bind mix function to correct extensions
3.3. Add function setTabledBuiltinExtensions to bind extensions for all functions in builtin tables
3.4. Add function setFunctionExtensions, it only bind extension when both function name and argument types are matched.
3.5. Add function GetTableBuiltinArgTypes to get function argument types according to builtin function definition.

4. Fix array size check of gl_FragColor and Sample Mask.

5. Other error check are added for negative cts cases.

johnkslang:
I think things might be faster if:  when you plan to change the semantics of the language, via glslang, we first start with the specification language that is contradictory to glslang.  For semantics, the spec and glslang are on similar ground, and it might be one or the other that should change.

So, let's start with where the specification prevents interpolation of a member of a struct?  Maybe it is a mistake?  The point about interpolation was that it needed a "normal" thing to interpolate, like a scalar or a vector.

ShchchowAMD:
Get it. Please allow me several days to organize those recent commits/changes, especially error check or limit check, based on specification and their working versions.

I agree with :

> For semantics, the spec and glslang are on similar ground, and it might be one or the other that should change.

Some kinds of error checks or issues found with the help of CTS cases, need to be carefully double checked again. I'll list them soon.

ShchchowAMD:
Hi John, as you mentioned, I agree with:

> The point about interpolation was that it needed a "normal" thing to interpolate, like a scalar or a vector.

We already initialize those functions with parameter type list like:

>             "float interpolateAtCentroid(float);"
>             "vec2  interpolateAtCentroid(vec2);"
>             "vec3  interpolateAtCentroid(vec3);"
>             "vec4  interpolateAtCentroid(vec4);"
> 
>             "float interpolateAtSample(float, int);"
>             "vec2  interpolateAtSample(vec2,  int);"
>             "vec3  interpolateAtSample(vec3,  int);"
>             "vec4  interpolateAtSample(vec4,  int);"
> 
>             "float interpolateAtOffset(float, vec2);"
>             "vec2  interpolateAtOffset(vec2,  vec2);"
>             "vec3  interpolateAtOffset(vec3,  vec2);"
>             "vec4  interpolateAtOffset(vec4,  vec2);"

For the parameter types of those interpolation functions, (8.14.2)

> this can be either a variable, or an array element. Component selection operators (e.g. .xy) may not be used when specifying interpolant.

So, maybe the atomicity of those parameters are based on a normal in/out scalar/vector, no matter it is a member of struct or a normal variable. Also we still prevent component selection operators.

For qualifier part, those keywords needed in interpolation functions, like flat, centroid, etc,. can be applied to a member of struct. So I think there's no grammar problems currently.

I'll remove related error check.

ShchchowAMD:
Additional change list:

Constant part:
new float type and functions/operators implementation:
1. new type SP_IEEE, float16 :  represents float32 and float16, also conversion between fp16 and fp32;
2. matrix operation functions: transpose , inverse , determinant , multiply;
3. pack/unpack functions: packUnorm2x16 , packSnorm2x16 , packHalf2x16, unpackUnorm2x16 , unpackSnorm2x16, UnpackHalf2x16;
4. angle and trigonometry functions: sinh, cosh, tanh, asinh, acosh, atanh;
5. bit conversion functions: floatBitsToInt, floatBitsToUint, intBitsToFloat, uintBitsToFloat;


initialize part:
1. new types:
   "float16_t", "f16vec2","f16vec3", "f16vec4", 
   "double", "dvec2", "dvec3", "dvec4", 
   "int8_t", "i8vec2", "i8vec3", "i8vec4", 
   "int16_t", "i16vec2", "i16vec3", "i16vec4", 
   "int64_t", "i64vec2", "i64vec3", "i64vec4", 
   "uint8_t", "u8vec2", "u8vec3", "u8vec4", 
   "uint16_t", "u16vec2", "u16vec3", "u16vec4", 
   "uint64_t", "u64vec2", "u64vec3", "u64vec4",

2. new function setTabledBuiltinExtensions:
Refine some basic functions/operators implementation (mostly double and dvec), following previous design.
We can use basic type and function tables to set built-in functions/operators.

3. Refine sampleMask related array size. SampleMask related built-in arrays' size only depend on implementation (like maxSamples) rather than run-time determined.

4.  Operator of Grad/Proj/Lod Offset and Atomic Load/Store: add parameters type check
        
5. Add GL_EXT_shader_integer_mix extension declaration.

Test/400.frag
```diff
@@ -126,7 +126,7 @@ void interp()
 
     interpolateAtSample(sampInArray, 1);        // ERROR
     interpolateAtSample(sampInArray[i], 0);
-    interpolateAtSample(s1.x, 2);
+    interpolateAtSample(s1.x, 2);   // ERROR
```
johnkslang, line 129: Is this out of date now?

Test/460.frag
```diff
@@ -9,7 +9,7 @@ in S s;
 
 void main()
 {
-    interpolateAtCentroid(s.v);
+    interpolateAtCentroid(s.v); // ERROR interpolate arguments should be input variables, not their members
```
johnkslang, line 12: Out of date?

glslang/MachineIndependent/Constant.cpp
```diff
@@ -75,6 +75,206 @@ bool isInf(double x)
...
+
+    return (out);
+}
+
+// Convert from float16 to float32
+ float float16ToFloat32(float16 in)
+{
+    SP_IEEE out;
+
+    if (in.u16 == 0)
```
johnkslang, line 155: Let's stick to one convention for this:

```
if (...) {
} else {
}
```

glslang/MachineIndependent/Constant.cpp
```diff
@@ -75,6 +75,206 @@ bool isInf(double x)
...
+}
+
+double clamp(const double& x, const double& minval, const double& maxval)
+{
+    return Min(Max(x, minval), maxval);
+}
+
+double determinant3(const double* col1, const double* col2, const double* col3)
+{
+    return  col1[0] * col2[1] * col3[2]
```
johnkslang, line 176: What is the need driving this?  Not all drivers can do it, and I thought the user should get some warning about that.  Maybe if both happen it's okay (haven't read through all this yet).

Test/400.frag
ShchchowAMD: Fixed

Test/460.frag
ShchchowAMD: Fixed

glslang/MachineIndependent/Constant.cpp
ShchchowAMD: Fixed

glslang/MachineIndependent/Constant.cpp
ShchchowAMD: The purpose is to fix issues and add missing constant expressions reported by GLES-CTS, such as group: 

> dEQP-GLES3.functional.shaders.constant_expressions.builtin_functions.matrix.*

I found this op would be finally converted to spv::GLSLstd450Determinant：
>     case glslang::EOpDeterminant:
>         libCall = spv::GLSLstd450Determinant;

Would that be OK if I add warning based on:
> (isEsProfile() && version >= 300) || (!isESProfile() && version >= 150)

and also checking type **DOUBLE** for those new added built-in funtions?

ShchchowAMD: As an example, determinant is supported from version **300 (ES, float), 150 (float), 400 (double)**. They are supposed to be implemented for users

If we need to add version checks for those matrix determinant or inverse functions, I'll check it and commit if anything missing.

Besides, if you mean that those functions seems to be too large (or ugly maybe), I can refine them as possible as I can.

Function determinant3() here is an example to help reduce complexity, which will be used for determinant(mat4):

> case 16:// mat4
>             det = mat[3]  * determinant3(mat + 4, mat + 8, mat + 12)
>                 + mat[7]  * determinant3(mat + 0, mat + 8, mat + 12)
>                 - mat[11] * determinant3(mat + 0, mat + 4, mat + 12)
>                 + mat[15] * determinant3(mat + 0, mat + 4, mat + 8);
>             break;

johnkslang: I thought this was driven by some implementations not implementing, or certainly not wanting to support, ongoing, this functionality.  That's why later specs backed off.  The idea was to save this kind of effort.  So, it is your choice to do it, that's fine.  However, at minimum, we want glslangValidator communicating to the user that this is non-portable (at the very least, non-portable to future versions).

Can you simultaneously do the constant folding and then give a warning if the result is needed to be consumed as a constant?  I think that might be difficult. So, it will silently be okay for early versions, and then give an error on later versions?  Let's make sure it does not silently work on anything that CTS didn't require.

ShchchowAMD: I found description as:

> Built-in functions not listed above and not defined as equations of the above have undefined precision.  These include, for example, the trigonometric functions and determinant. 

So as you say, would that be okay if I add some warnings and update test results of glslangValidator to make sure those functions won't work silently?

Also, maybe try to add a new option to allow those overloaded functions and let users make choice depending on their needs (it seems to be better as they can directly query those information from getVersion)?

johnkslang: > So as you say, would that be okay if I add some warnings and update test results of glslangValidator to make sure those functions won't work silently?

Yes, I think a warning is sufficient.

> Also, maybe try to add a new option to allow those overloaded functions and let users make choice depending on their needs (it seems to be better as they can directly query those information from getVersion)?

No, I don't think it needs controlled, it is okay to do the folding.  I'm a bit concerned about invariance, but I simply don't think this is a real issue, in practice.

ShchchowAMD: Okay, will update soon after verifying with CTS.

ShchchowAMD: Besides, the warning scope we are discussing would include
1. **determinant** 
2. **functions which call determinant** like **inverse**.

Do I miss something?

johnkslang: These from the desktop spec are the required ones.  It's the ones not in this list I don't want to make start working without a warning, if they previously were not working.  Nicer still to have glslang understand this list, but that can be a separate thing.  (I think there is an issue for this.)

> Angle and Trigonometric Functions
> 
> - radians
> - degrees
> - sin
> - cos
> - asin
> - acos
>  
> Exponential Functions
>  
> - pow
> - exp
> - log
> - exp2
> - log2
> - sqrt
> - inversesqrt
>  
> Common Functions
>  
> - abs
> - sign
> - floor
> - trunc
> - round
> - ceil
> - mod
> - min
> - max
> - clamp
>  
> Geometric Functions
>  
> - length
> - dot
> - normalize

ShchchowAMD: Thanks. I'll make commits for this part soon.

ShchchowAMD: @johnkslang I checked the list. Seems functions in list 'CustomFunctions', including determinant and matrixInverse, are not in this list.

Should we add warning for all those custom functions if not implemented before (including pack functions, bit encoding and matrixCompMult ?

johnkslang: Yes, I think we should warn, but what do we warn on?

I think this is complicated, because later when a constant is used, it has forgotten what computations lead to its value.

That is, it is not a warning to constant fold `determinate`, that's a choice by the compiler, not something the user did, but the warning should come later, if the user depended on the constant folding happen, since that is not portable.

(This is why I didn't add the functionality, because for no effort, the portability issue gets raised automatically.)

ShchchowAMD: I make a test.
```
	const float cval = determinant(mat2(2.0)); // fold()
        float cval = determinant(a);               // not parsed to fold(), and mat2 a = mat2(2.0)
```


Those new added custom functions would trigger fold() only in **constant expression**. Would not make difference with non-constant usage.

As they are all in the custom functions list, and we could find those **operators** in the fold() function, can we add warning **inside fold()** with **loc**?

I think custom functions are needed by customers, and as you say, we can info them those portability risks.
Or, maybe, a new option like SSBOstd140 in glslang (not prefer this)?

ShchchowAMD: Seems a better idea is to use list **customFunctions** to check those custom functions within **addBuiltInFunctionCall**. (Make it a more general list, not only in initialization)

We could add warning here and control constant expression usage if needed.

New flags or control options can also be added if we want to prevent potential portability issues.

ShchchowAMD: Seems a better idea is to use list **customFunctions** to check those custom functions within **addBuiltInFunctionCall**. (Make it a more general list, not only in initialization)

We could add warning here and control constant expression usage if needed.

New flags or control options can also be added if we want to prevent potential portability issues.

johnkslang: This issue would be with something like

```
int a[floor(10 * determinant(mat2(2.0))[0].x)];
```

Is it legal, or not legal?

ShchchowAMD: This is an illegal case, determinant will return a float / double, so the following indexing and swizzle should be illegal.

I think another case may show the issues:
    
1. how we treat following right-hand value, '10' seems not to be involved in right-hand folding.

>                                         int a[10 * inverse(mat2(2.0))[0].x];

> // wrong operand types: no operation '*' exists that takes a left-hand operand of type ' const int' and a right operand of type ' const float' (or there is no acceptable conversion)

2. the index should be a constant type

>                                          int a[inverse(mat2(2.0))[0].x];

> '' : array size must be a constant integer expression

I‘m looking at those issues now.
Can we add a defaultly disabled option in glslang and add another process to fix those problems, like what we do for 'convert'.
Just want to implement those functionality for users.

johnkslang: Yes, for inverse, that's the example I intended when I said determinant.

Yes, I think we can do something with an option `--full-fold` or something to say you know the target has broader folding than what is globally portable.

ShchchowAMD:
Option 'full-fold' has been added (default disabled).
When we enable this option, functions not in above normal list will be allowed to fold, and then a global warning will be reported if any of them trigged full-fold.

Instead of modifying existed 'fold' functions or any other interfaces, we will use 'fullFoldUnary' and 'fullFoldBinary' to check operator types. Both at intermediate level and constant node level, folding check has been added.

ShchchowAMD:
An example could be observed in 300.frag.out :

outerProduct won't be folded as it is not in the list and also the 'full-fold' option is disabled.

PS. For the discussion we made before, the error report should be correct, expression "10 * inverse(mat2(2.0))[0].x " still has a type 'float', which is illegal as array's length. The point is to avoid automatically folding all the built-in functions.

glslang/MachineIndependent/Constant.cpp
```diff
@@ -75,6 +75,187 @@ bool isInf(double x)
            (bitPatternH & 0xFFFFF) == 0 && bitPatternL == 0;
 }
 
+typedef union
+{
+    unsigned int u32;
+    float f;
+    struct
+    {
+#if defined(__BIG_ENDIAN__)
```
johnkslang, line 84: I don't believe endianness effects the internal contents of an IEEE layout.  Endianness effects byte ordering in memory, not bit ordering in a register.

ShchchowAMD: I'll refine this part with templates in hex_float.h

clang-gn will also be fixed according to last CI build.

ShchchowAMD:
Refine fp types according to review.
Now using hex floats instead.

ShchchowAMD:
May I know where to check failure logs of those additional CI builds?

johnkslang:
Can you see if you click on "Details" and then click then on the listed path?

ShchchowAMD:
The fusion site could still not be accessed.

Build status on google cloud is now reachable if I signed in with an account or with Guest mode using Chrome.

greg-lunarg:
Looks like you are missing some kind of dependency between Constant.cpp and bitutils.h for the bazel builds.

ShchchowAMD:
Thanks ! Fail log has been found.

Issue #2357, opened 2020-07-24
Not setting the NonUniform decoration correctly for SSBO stores


jekstrand:
The dEQP-VK.descriptor_indexing.storage_buffer_in_loop CTS test has the following vertex shader:
```glsl
#extension GL_EXT_nonuniform_qualifier : require	
layout(push_constant)     uniform Block { int lowerBound, upperBound; } pc;
layout(set=1,binding=12) uniform isamplerBuffer iter;	
layout(set=0,binding=2) buffer Data { vec4 cnew, cold; } data[]; 
layout(location = 0) in  vec4  in_position;	
layout(location = 1) in  vec2  in_normalpos;	
layout(location = 2) in  ivec4 index;			
layout(location = 0) out vec4  position;	
layout(location = 1) out vec2  normalpos;	
layout(location = 2) out int   vIndex;		
layout(location = 3) out int   rIndex;		
layout(location = 4) out int   gIndex;		
layout(location = 5) out int   bIndex;		
layout(location = 6) out int   aIndex;		
void main()							
{										
    gl_PointSize = 0.2f;				
    position = in_position;			
    normalpos = in_normalpos;			
    gl_Position = position;			
    vIndex = gl_VertexIndex;			
    rIndex = index.x;					
    gIndex = index.y;					
    bIndex = index.z;					
    aIndex = index.w;					
  if (gIndex != 0) data[nonuniformEXT(gIndex)].cnew = data[nonuniformEXT(rIndex)].cold;	
}											
```
for which GLSLang generates the following SPIR-V
```spirv
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 80
; Schema: 0
OpCapability Shader
OpCapability SampledBuffer
OpCapability ShaderNonUniform
OpCapability RuntimeDescriptorArray
OpCapability StorageBufferArrayNonUniformIndexing
OpExtension &quot;SPV_EXT_descriptor_indexing&quot;
%1 = OpExtInstImport &quot;GLSL.std.450&quot;
OpMemoryModel Logical GLSL450
OpEntryPoint Vertex %4 &quot;main&quot; %13 %20 %22 %26 %28 %34 %36 %38 %41 %45 %48 %52
OpMemberDecorate %11 0 BuiltIn Position
OpMemberDecorate %11 1 BuiltIn PointSize
OpMemberDecorate %11 2 BuiltIn ClipDistance
[...]

PR #2387, opened 2020-09-03
NonUniform SPIR-V fixes.

BNieuwenhuizen:
This fixes 3 places where glslang forgets to emit NonUniform decorations:

- OpImage in the texture instruction lowering
- constructors
- stores

Missing:
- other early returning cases in binary and unary ops

The qualifier inference seemed to be working ok, but given that we also set the decoration on the result of a nonuniform load as a documented workaround for .. something, there might still be something suspicious there.

Not very successful testing the change for new
- ctest glslang-testsuite fails for me both on master and this branch and I haven't found the output yet
- Test/runtests fails for me both on master and this branch though the test is clearly unrelated.

Haven't added new tests yet.

Should fix the issues noticed in https://github.com/KhronosGroup/glslang/issues/2357


greg-lunarg:
I am somewhat familiar with glslang. I will see if I can figure out test failure there.

greg-lunarg:
For the glslangtests "failure", you will need to update Test/baseResults/spv.nonuniform.frag.out.

I think all the new instructions are valid and "correct", at least as correct as the previous SPIR-V.

Certainly the new decoration on the OpImage instruction is fine.

The new nonuniform decoration on 18(param) is probably fine, especially since there are nonuniform decorations on a number of other local variables as well. It is a little asymmetric that 20(param) is not decorated, but that is another issue, if one at all.

The new duplicated decoration on 17(nu_li) should be fine since there is already two of them. Three is probably just as valid at two.

greg-lunarg:
Also seeing a failure with spv_meshShaderPerViewUserDefined_mesh, but this is mentioned by johnk as having started to fail July 1, suspecting it might be a spirv-val problem.

greg-lunarg:
If you use the latest spirv-tools and spirv-headers, all non-related failures no longer occur.

greg-lunarg:
I have created a branch which fixes the regressions and adds a buffer store to the nonuniform tests.

I will add a "constructor" test next and then I think these can go in.

greg-lunarg:
I have now added a texture sample to the uniform test to verify that NonUniform does not propagate to the constructor (OpSampledImage here).

I believe my branch now addresses #2357 as well as updates and adds the necessary regression tests.

greg-lunarg:
Any additional shortcomings vis-a-vis NonUniform can be addressed in followup PRs.

@BNieuwenhuizen How do you want to proceed? I suggest we proceed with this PR and you cherry-pick the last three commits from my branch into your branch. Let me know if you would rather do something different.

BNieuwenhuizen:
Hi Greg, I cherry-picked your test fixes and squashed the fix of the test to my relevant patch to keep history bisectable. Let me know if you want me to roll that back. Thanks a lot!

johnkslang:
Wondering if you could explain the issue/fix context more.  This is certainly more than just adding extra NonUniform decorations, as there are also more instructions now.  I think some context might be missing here.  Is this 5 different changes, or should they be squashed?  Thanks.

BNieuwenhuizen:
The first 3 patches fix GLSLang and fix up the tests to still pass with these changes. The last two add two patches expand the testcase for more test coverage of the the things fixed in the first 3 patches. The extra instructions come solely from the expanded testcases.

The patches are ordered such that they're bisection safe (under the assumption that they wouldn't be squashed), so either way is fine.

BNieuwenhuizen:
Friendly ping on review/merging. Thanks!

johnkslang:
cc @jeffbolznv, do you have input on the nonUniform changes here?

greg-lunarg:
FWIW, @jeffbolznv fed me some ideas on how to improve nonUniform propagation in glslang, but as @BNieuwenhuizen had already created this PR and it addressed #2357, I thought I would wait for this to move forward first and then evaluate and consider the suggestions from @jeffbolznv.

greg-lunarg:
I would be willing to do a review instead of @jeffbolznv since I have agreed to take on his larger concern and have done some ground work on it already.

greg-lunarg:
Just considering the regression test spv.nonuniform.frag and its .out file, I think this change is fine.

You asked about extra instructions. I am presuming you are referring to spv.nonuniform.frag.out. I also presume you are referring to instructions other than the nonUniform decoration instructions which would be expected. I believe these are totally due to the additional two statements I added to spv.nonuniform.frag. One was added to test that decorations were correctly added for stores, which was previously not tested by the regression test and fixed by this PR. The other was to test that nonuniform was not propagated through Sampler2D(), which @jeffbolznv has noted is expressly prohibited by the spec.

This PR supplies all of the changes requested by #2357. It does not create any decorations which are invalid or inconsistent with the GLSL spec.

So again, just based on the files spv.nonuniform.frag and its .out, I think this change is fine. It is consistent with the goals of @jeffbolznv with regard to recent issues around nonUniform. I have not determined if it is sufficient for all cases. That will take me longer, but I am planning to do that.

I have not evaluated the actual additional compiler code. That would take me longer. Let me know if you want me to do that.

Once we are satisfied with the actual compiler changes, I think we should accept the PR since it fixes a current issue #2357 and my work will take longer.

Let me know if there are additional concerns I have not addressed.

johnkslang:
> I have not evaluated the actual additional compiler code. That would take me longer. Let me know if you want me to do that.

What part is this referring to?  Sanity checking the change, it looks okay, provided you see the changes made as technically correct.

So would like to see:

- complete sign off by Greg
- the result of all the bot testing, not just 2, maybe we need some form of retrigger

greg-lunarg:
Reviewing now...

SPIRV/GlslangToSpv.cpp
```diff
@@ -2611,6 +2613,10 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
         else
             constructed = builder.createConstructor(precision, arguments, resultType());
 
+        if (node->getType().getQualifier().isNonUniform()) {
+            builder.addDecoration(constructed, spv::DecorationNonUniformEXT);
+        }
+
```
greg-lunarg, line 2619: This code is probably unnecessary. From the GLSL spec:
```
Constructors and builtin functions, which all have return types that
are not qualified by nonuniformEXT, will not generate nonuniform results.
```
I added a sample of sampler2D() constructor of a nonuniform texture in the test file spv.nonuniform.frag. You will see that the OpSampledImage was NOT decorated nonuniform because glslang correctly did not propagate the qualifier to the constructor type. When I commented this code out, no change occurred in the test results. This code probably won't hurt anything, but I don't think it will ever get executed and so should probably just be deleted.

greg-lunarg:
In general, I am going to sign off on this change even though there is a slight flaw: it could mark some uniform pointers used by Stores as nonuniform. I think this flaw is ok to allow because: 1) It should not cause incorrect code and behavior. At worst this might only cause less efficient code to be generated. 2) It is not clear it will occur in practice. I can write a piece of code where it happens (a reference into an array of nonuniforms with a uniform index) but it is not clear anyone would ever write such code. It doesn't happen in any of the statements in the spv.nonuniform.frag test file. 3) This problem also exists for Loads and there haven't been any complaints yet, and 4) we would need to overhaul the logic that determines if an lvalue is nonuniform. Given the first three points, its not clear if its worth waiting to overhaul it since 5) this code does fix cases where truly incorrect code is being generated at this moment.

The flaw is that the code is assuming that if the rvalue type of a Store (or Load) is nonuniform, the lvalue is nonuniform. This is not always true, but it seems for all useful applications of nonuniform it is a reasonable assumption, and even if wrong is not fatal.

The true fix would be to add infrastructure to analyze and remember which lvalues are truly nonuniform.

In the meantime, I recommend we accept this PR.

PR #2388, opened 2020-09-03
Error when initializing rayQuery with assignment

Tobski:
This PR adds an error check to prevent initialization of ray query objects via assignment, since there are no valid constructors for the ray query object, and ray query objects cannot be copied. Without this change, the code in the new test generates an OpLoad/OpStore pair, which is invalid.


dgkoch:
cc @alelenv @johnkslang 

@Tobski you need to sign the CLA yet.

PR #2393, opened 2020-09-12
Fix build with android ndk r16b

nihui:
For some reasons, I want to build glslang with android ndk r16b

I tried ```#include <cmath>```, but it can not fix this error.
So I fix it by just dropping the std namespace.

I think log2() shall be fine here.

```
glslang/glslang/MachineIndependent/ParseHelper.cpp:5421:77: error: no member named 'log2' in namespace 'std'; did you mean simply 'log2'?
            publicType.qualifier.layoutBufferReferenceAlign = (unsigned int)std::log2(value);
                                                                            ^~~~~~~~~
                                                                            log2
android-ndk-r16b/sysroot/usr/include/math.h:168:8: note: 'log2' declared here
double log2(double __x) __INTRODUCED_IN(18);
       ^
1 error generated.
```


lance-loong:
@jackoalan hi, Is there any problem with this repair? Hope it can be repaired as soon as possible. Thank you very much for your efforts

lance-loong:
Maybe we can fix it with GLM:

```
#	if GLM_HAS_CXX11_STL
		using std::log2;
#	else
		template<typename genType>
		genType log2(genType Value)
		{
			return std::log(Value) * static_cast<genType>(1.4426950408889634073599246810019);
		}
#	endif
```

greg-lunarg:
I have just started maintaining glslang. I will take a look and get back shortly.

greg-lunarg:
Here is how Google handled in this in spirv-tools, const_folding_rules.cpp:

```
#ifdef __ANDROID__
    // Android NDK r15c tageting ABI 15 doesn't have full support for C++11
    // (no std::exp2/log2). ::exp2 is available from C99 but ::log2 isn't
    // available up until ABI 18 so we use a shim
    auto log2_shim = [](double v) -> double { return log(v) / log(2.0); };
    ext_rules_[{ext_inst_glslstd450_id, GLSLstd450Exp2}].push_back(
        FoldFPUnaryOp(FoldFTranscendentalUnary(::exp2)));
    ext_rules_[{ext_inst_glslstd450_id, GLSLstd450Log2}].push_back(
        FoldFPUnaryOp(FoldFTranscendentalUnary(log2_shim)));
#else
    ext_rules_[{ext_inst_glslstd450_id, GLSLstd450Exp2}].push_back(
        FoldFPUnaryOp(FoldFTranscendentalUnary(std::exp2)));
    ext_rules_[{ext_inst_glslstd450_id, GLSLstd450Log2}].push_back(
        FoldFPUnaryOp(FoldFTranscendentalUnary(std::log2)));
#endif

```

Can we do something more along these lines?

greg-lunarg (changes requested):
See my latest comment in conversation.

nihui:
patch updated

greg-lunarg:
I approve of the change. Thank you!

I can't see why the ubuntu-clang-gn build failed; the details are missing. Can you do this build manually and look for errors? Or can you push a trivial change to retrigger?

PR #2395, opened 2020-09-12
Find python interpreter from host when cross-compiling

nihui:
glslang build failed for missing python interpreter when cross-compile glslang for ios

https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake defines a handy find_host_package macro
If it exsits, use find_host_package, like how SPIRV-Tools handle it here

https://github.com/KhronosGroup/SPIRV-Tools/blob/726af6f78f80988271c8b558ae9cc84fa5a65016/CMakeLists.txt#L181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,9 +228,15 @@ function(glslang_set_link_args TARGET)
     endif()
 endfunction(glslang_set_link_args)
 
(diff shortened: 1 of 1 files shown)


juan-lunarg:
FWIW we have now added iOS to CI

PR #2396, opened 2020-09-14
8. io mapping refine & qualifier member check & resolver expand

ShchchowAMD:
Fix io mapping issue:
1. refine usage of access name.
2. add validation of interfaces & buffer & uniforms
3. split nest level, for member check of invariant qualifier
4. use new id level setting, help to identify builtins
5. fix issues introduced by vector and usage of for_each in io mapping

### DIFF
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -113,7 +113,7 @@ TIntermSymbol* TIntermediate::addSymbol(const TType& type, const TSourceLoc& loc
 //
 // Returns nullptr if the working conversions and promotions could not be found.
 //
-TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIntermTyped* right, const TSourceLoc& loc)
+TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIntermTyped* right, TSourceLoc loc)
 {
     // No operations work on blocks
     if (left->getType().getBasicType() == EbtBlock || right->getType().getBasicType() == EbtBlock)
@@ -227,8 +227,7 @@ TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIn
 //
 // Low level: add binary node (no promotions or other argument modifications)
 //
-TIntermBinary* TIntermediate::addBinaryNode(TOperator op, TIntermTyped* left, TIntermTyped* right,
-    const TSourceLoc& loc) const
+TIntermBinary* TIntermediate::addBinaryNode(TOperator op, TIntermTyped* left, TIntermTyped* right, TSourceLoc loc) const
 {
     // build the node
     TIntermBinary* node = new TIntermBinary(op);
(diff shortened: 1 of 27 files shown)


ShchchowAMD:
6. Validate uniforms qualifiers, members, types.
7. LH / RH value check.

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -113,7 +113,7 @@ TIntermSymbol* TIntermediate::addSymbol(const TType& type, const TSourceLoc& loc
 //
 // Returns nullptr if the working conversions and promotions could not be found.
 //
-TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIntermTyped* right, const TSourceLoc& loc)
+TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIntermTyped* right, TSourceLoc loc)
```
johnkslang, line 116: Why this change?  Is it just because of needing to rebase?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4632,14 +4647,14 @@ void TParseContext::paramCheckFix(const TSourceLoc& loc, const TQualifier& quali
 
 void TParseContext::nestedBlockCheck(const TSourceLoc& loc)
 {
-    if (structNestingLevel > 0)
+    if (structNestingLevel > 0 || blockNestingLevel > 0)
```
johnkslang, line 4650: Why do we need two counters instead of 1?

ShchchowAMD: From m4 file:

>         parseContext.memberQualifierCheck($1);
>         parseContext.voidErrorCheck($2.loc, (*$3)[0].type->getFieldName(), $2.basicType);
>         parseContext.mergeQualifiers($2.loc, $2.qualifier, $1.qualifier, true);
>         parseContext.precisionQualifierCheck($2.loc, $2.basicType, $2.qualifier);

Glslang will check a member (of a block or struct) directly with memberQualifierCheck function.

Within memberQualifierCheck, we can only know it is a member of a block or struct by checking structNestingLevel (**parent's storage qualifier could not be found here**).

In some cases, we need to exactly its **parent struct's** information.

For example, in this PR, we consider a "invariant" qualifier check case (GS):

> out GSOutputBlock {
>    highp flat invariant vec4 gs_fs_test;
> } output_block;

This is **legal** because it is inside an output block. But its storage is **EvqTemporary** here. If we do **invariantCheck**, it would be reported as **illegal**. **blockNestingLevel** is added to help us ensure that current point is inside a global scope block.

In the other hand, if we don't split this nesting level, member variable in following case will be treated as same as above one.

> struct sp {
>     highp flat invariant vec4 gs_fs_test;;      // ERROR
> };

glslang/MachineIndependent/Intermediate.cpp
ShchchowAMD: I think this is a mistake.

I was trying to keep

> 
>             if (loc.line == 0)
>                 loc = node->getLoc();

when setting aggregate operator so I keep loc non-const before, but seems following 

> statement
>     if (loc.line != 0 || node != nullptr)
>         aggNode->setLoc(loc.line != 0 ? loc : node->getLoc());

could finish this job.

Now I've reversed to original version.

ShchchowAMD:
Hi john,

Is there any concerns for this PR?

I'm looking at issues and feedback for those waiting PRs currently.

mbechard:
@ShchchowAMD I think this commit breaks combining multiple units together in the same stage. Specifically TIntermediate::mergeTrees() does seedIdMaps and remapIds, with maxId + 1 essentially being used to offset all local ids within a compilation unit into their own numeric space so there are no overlaps. Ie. all ids in unit[0] < all ids in unit[1] < all ids in unit[2].
However I think the new unique Id level system in updateUniqueIdLevelFlag() doesn't work with this workflow.

ShchchowAMD:
@mbechard  I'll work on this. Thanks for reminding.
Found ref : #2156

ShchchowAMD:
@mbechard TGlslangToSpvTraverser::rValueParameters dependents on unique IDs. but  symbolTable->adoptLevels will reset the unique ID, so if there are multiple shader in single shader stage, we can't make sure the ID is unique for all symbols.
I'll make a fix PR and seems it is better to set unique id again after using it. I'll make comments sooner.

Issue #2405, opened 2020-09-26
SpirvToolsDisassemble does not allow choosing alternative `spv_target_env`.


phniix:
#### ;tldr

The utility function, `glslang::SpirvToolsDisassemble`, does not allow choosing an alternative SPIR-V context. We need to expand this.

Description
--- 
The current interface does not provide the flexibility to allow the evaluation to be performed in other SPIR-V environments. It would be good if a new interface was provided, to allow for this.

Additionally, the current API's fixed choice, `SPV_ENV_UNIVERSAL_1_3` is not documented, and requires looking into the source code to determine this fact. It would be good if the documentation was improved at the API level, to at least not necessitate digging into the source code to know this.

Proposal
---
I propose a new API interface, as opposed to leveraging the existing interface. I feel that such a distinction is worthwhile.

The new API interface would be akin to the existing one, with an additional parameter.
```
void SpirvToolsDisassemble(std::ostream& out, const std::vector<unsigned int>& spirv,
           spv_target_env requested_context);
```

I'm not overly stressed should a preference lean more heavily towards only on a single interface:
```
void SpirvToolsDisassemble(std::ostream& out, const std::vector<unsigned int>& spirv,
           spv_target_env requested_context=SPV_ENV_UNIVERSAL_1_3);
```

The existing API will also get a brand new coat of "documentation paint" for the `SPV_ENV_UNIVERSAL_1_3` usage.

Reference Implementation
---
A reference implementation for the above proposal exists here: https://github.com/phniix/glslang/tree/SpirvToolsDisassemble-Improvements

phniix:
Updates made based on feedback from @johnkslang. Thanks!!
New PR #2408 tidies up the reference implementation.

phniix:
@johnkslang Should this get opened up again, because of the situation outlined in https://github.com/KhronosGroup/glslang/issues/2413

I asked a bunch of questions to yourself, and @timvpGoogle there upon discovery of that issue.
There is here: https://github.com/KhronosGroup/glslang/issues/2413#issuecomment-706869986

Following all those questions, and whether this opens up again or we spin off a new one...

What's the best way to proceed ?

johnkslang:
I reopened #2413 for further discussing this.

phniix:
Issue https://github.com/KhronosGroup/glslang/issues/2434 addressed what caused this issue to be reverted a couple of times.

[...]

PR #2406, opened 2020-09-26
Improved SpirvToolsDisassemble interface and documentation.

phniix:
This addresses the problems highlighted in: https://github.com/KhronosGroup/glslang/issues/2405

### DIFF
--- a/StandAlone/CMakeLists.txt
+++ b/StandAlone/CMakeLists.txt
@@ -70,6 +70,12 @@ target_include_directories(glslangValidator PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External>)
 
+if(ENABLE_OPT)
+target_include_directories(glslangValidator PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External/spirv-tools/include>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External/spirv-tools/include>)
+endif()
+
 if(ENABLE_SPVREMAPPER)
     set(REMAPPER_SOURCES spirv-remap.cpp)
     add_executable(spirv-remap ${REMAPPER_SOURCES})
--- a/SPIRV/SpvTools.cpp
+++ b/SPIRV/SpvTools.cpp
@@ -44,7 +44,6 @@
 
 #include "SpvTools.h"
 #include "spirv-tools/optimizer.hpp"
-#include "spirv-tools/libspirv.h"
 
 namespace glslang {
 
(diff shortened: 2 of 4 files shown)


StandAlone/CMakeLists.txt
```diff
@@ -70,6 +70,12 @@ target_include_directories(glslangValidator PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External>)
 
+if(ENABLE_OPT)
+target_include_directories(glslangValidator PUBLIC
```
johnkslang, line 74: What is the purpose in doing this?

SPIRV-Tools is where this source should be looked at, developed with, etc., not glslang.  I'd rather be going the other direction, of just linking to SPIRV-Tools libraries, and having a cleaner separation between glslang and SPIRV-Tools.

phniix: So the reason I did this was to overcome a couple of compile time errors that came in as a result of moving `#include "spirv-tools/libspirv.h"` from `SpvTools.cpp` to `SpvTools.h`.

I was getting the following errors after this change was made. There were:
```
  Building Custom Rule D:/gitrepos/phx-glslang/StandAlone/CMakeLists.txt
  StandAlone.cpp
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\StandAlone\glslangValidator.vcxproj]
  Building Custom Rule D:/gitrepos/phx-glslang/External/googletest/googlemock/CMakeLists.txt
```
and
```
  Building Custom Rule D:/gitrepos/phx-glslang/gtests/CMakeLists.txt
  cmake_pch.cxx
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\gtests\glslangtests.vcxproj]
  Building Custom Rule D:/gitrepos/phx-glslang/External/spirv-tools/tools/CMakeLists.txt
```
As a result, the two changes I made are specific to these two build targets, and only when `ENABLE_OPT` is defined, as that is when the header include is made known to the compiler.

I've just checked now, and it doesn't happen when `-DENABLE_OPT=ON` is specified on current master @ https://github.com/KhronosGroup/glslang/commit/1815d86a224b75abc1dd084789c20f7da3c9fed0 . I'm a bit thrown off myself to see that the previous implementation had `#include ""` vs `#include <>` to be honest. I'm adding these changes here so that I'm not breaking the builds, and keeping existing `#include ""` syntax to not ruffle feathers or loosen mortar.

The way that the header include paths are set at the moment, SpvTools can't find the specific file.

More than happy to make amendments to rectify any dramas here. 
Please let me know what's needed.

johnkslang:
I reverted this.  Not enough tests ran on the PR to uncover that the header changes really don't work.  Maybe you should post an issue about this with SPIRV-Tools, if you think they are including incorrectly.

phniix:
No dramas. I understand.

Sounds like what my next step would be is to resubmit this, sans CMake changes, as the PR tests should go through without any dramas as they have now. My CMake changes are for build config files in KhronosGroup/glslang so I don't see it would require reaching out to KhronosGroup/SPIRV-Tools.

I'll resubmit another branch for resovling https://github.com/KhronosGroup/glslang/issues/2405
Will loop back in on this when time allows.

phniix:
I've created a new PR to address the above: https://github.com/KhronosGroup/glslang/pull/2408
I'm getting local build failures as before:
```
  glslang-default-resource-limits.vcxproj -> D:\gitrepos\phx-glslang\build\StandAlone\Release\glslang-default-resource-limits.lib
  StandAlone.cpp
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\StandAlone\glslangValidator.vcxproj]
  gmock.vcxproj -> D:\gitrepos\phx-glslang\build\lib\Release\gmock.lib
  cmake_pch.cxx
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\gtests\glslangtests.vcxproj]
```
Let's pick up the convo in #2408 :)

PR #2408, opened 2020-09-28
Add new SpirvToolsDisassemble API interface + Improve Doc on existing API interface

phniix:
This addresses the problems highlighted in: #2405

- Add more flexible SpirvToolsDisassemble interface to allow specifying spv_target_env for disassembly output.
- Improve documentation on existing SpirvToolsDisassemble interface.

### DIFF
--- a/SPIRV/SpvTools.cpp
+++ b/SPIRV/SpvTools.cpp
@@ -44,7 +44,6 @@
 
 #include "SpvTools.h"
 #include "spirv-tools/optimizer.hpp"
-#include "spirv-tools/libspirv.h"
 
 namespace glslang {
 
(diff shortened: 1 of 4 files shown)


phniix:
Based on existing discussion in #2406, CMake changes were not included in this PR.

I still get local build failures, as shown here, with the changes:
```
  glslang-default-resource-limits.vcxproj -> D:\gitrepos\phx-glslang\build\StandAlone\Release\glslang-default-resource-limits.lib
  StandAlone.cpp
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\StandAlone\glslangValidator.vcxproj]
  gmock.vcxproj -> D:\gitrepos\phx-glslang\build\lib\Release\gmock.lib
  cmake_pch.cxx
d:\gitrepos\phx-glslang\spirv\SpvTools.h(47): fatal error C1083: Cannot open include file: 'spirv-tools/libspirv.h': No such file or d irectory [D:\gitrepos\phx-glslang\build\gtests\glslangtests.vcxproj]
  spirv-as.vcxproj -> D:\gitrepos\phx-glslang\build\External\spirv-tools\tools\Release\spirv-as.exe
```

I figure we will just look at the results of appveyor and take it from there.

phniix:
Appveyor has returned the types of results we anticipated without the header path inclusions.

My previous approach in https://github.com/KhronosGroup/glslang/pull/2406 was correct. 
However, a subtlety did exist, and it lays in two things:
- `ENABLE_OPT` is defined, as the CMAKE configs perform `ENABLE_OPT=ON/OFF`.. and,
- There is a `${spirv-tools_SOURCE_DIR}` defined that points to the path needed for the header locations.

This subtley is now implemented as:

- `ENABLE_OPT` is defined, as the CMAKE configs perform `ENABLE_OPT=ON/OFF`

I changed the `#ifdef ENABLE_OPT`  to `#if ENABLE_OPT`, and it seems to match up with how the rest of the library is handling this particular option. 

- There is a `${spirv-tools_SOURCE_DIR}` defined that points to the path needed for the header locations.

The `CMakeLists.txt` definition in the SPIRV directory already uses `${spirv-tools_SOURCE_DIR}` in an `if(ENABLE_OPT)` block for ensuring the headers are available when it compiles. In contrast to my previous solution, I use this method as it is much cleaner, and does exactly what we need with having to reference build or install interfaces. 

(@johnkslang )

PR #2410, opened 2020-09-28
Disable -Wno-conversion on MSVC compiler

jonahryandavis:


jonahryandavis:
@johnkslang Can you take a look at this PR? Sorry, I'm not familiar with the process for getting eyes on pull requests in this repository.

jonahryandavis:
For context, this is breaking glslang's roll into ANGLE: https://chromium-review.googlesource.com/c/angle/angle/+/2435250

BUILD.gn
```diff
@@ -76,7 +76,9 @@ if (!defined(glslang_angle)) {
 
 config("glslang_public") {
   include_dirs = [ "." ]
-  cflags = [ "-Wno-conversion" ]
+  if (!is_win || is_clang) {
```
ShabbyX, line 79: Does it really need !is_win? Clang should accept this flag regardless of platform.

jonahryandavis: gcc also can accept this flag, this is the way it's done in other parts of the gn files

PR #2419, opened 2020-10-12
CMake: Use VERSION/SOVERSION for all shared libs, make glslang-default-resource-limits STATIC

akien-mga:
The first patch is taken from the Debian and Fedora packaging (https://salsa.debian.org/xorg-team/vulkan/glslang/-/blob/debian-unstable/debian/patches/glslang-default-resource-limits_staticlib.patch). I did not double check if it's correct for `glslang-default-resource-limits` to be static, but it does seem sensible.

The second patch applies the new shared library versioning from `libglslang.so` to the other shared libraries built with glslang:
https://github.com/KhronosGroup/glslang/blob/2067d1a93e6edc17f2a6b7e3e5138a9bbcd35ef9/glslang/CMakeLists.txt#L165-L170

The output of this PR is (on Linux, with `-DBUILD_SHARED_LIBS=ON`):
```
$ ls -l /usr/lib64/lib{glslang,HLSL,SPIRV,SPVRemapper}.so*
lrwxrwxrwx 1 root root      16 Oct 12 11:29 /usr/lib64/libglslang.so -> libglslang.so.11*
lrwxrwxrwx 1 root root      20 Oct 12 11:29 /usr/lib64/libglslang.so.11 -> libglslang.so.11.0.0*
-rwxr-xr-x 1 root root 2238616 Oct 12 11:29 /usr/lib64/libglslang.so.11.0.0*
lrwxrwxrwx 1 root root      13 Oct 12 11:29 /usr/lib64/libHLSL.so -> libHLSL.so.11*
lrwxrwxrwx 1 root root      17 Oct 12 11:29 /usr/lib64/libHLSL.so.11 -> libHLSL.so.11.0.0*
-rwxr-xr-x 1 root root   14656 Oct 12 11:29 /usr/lib64/libHLSL.so.11.0.0*
lrwxrwxrwx 1 root root      14 Oct 12 11:29 /usr/lib64/libSPIRV.so -> libSPIRV.so.11*
lrwxrwxrwx 1 root root      18 Oct 12 11:29 /usr/lib64/libSPIRV.so.11 -> libSPIRV.so.11.0.0*
-rwxr-xr-x 1 root root  549392 Oct 12 11:29 /usr/lib64/libSPIRV.so.11.0.0*
lrwxrwxrwx 1 root root      20 Oct 12 11:29 /usr/lib64/libSPVRemapper.so -> libSPVRemapper.so.11*
lrwxrwxrwx 1 root root      24 Oct 12 11:29 /usr/lib64/libSPVRemapper.so.11 -> libSPVRemapper.so.11.0.0*
-rwxr-xr-x 1 root root  181280 Oct 12 11:29 /usr/lib64/libSPVRemapper.so.11.0.0*
```

CC @tjaalton @airlied to review as this will impact Debian and Fedora packaging (positively AFAIK your packaging policies :)).


ben-clayton:
Hi @akien-mga,

Related bug: https://github.com/KhronosGroup/glslang/issues/2283

I'm not entirely sure about versioning the libraries aside from `glslang` - the reason being there's no clearly defined public interface for these libraries. If we're going to take semantic versioning seriously, then we'll need to ensure that the major version is bumped whenever we make a non-backwards compatible change to any public API. As the libraries `libHLSL`, `libSPIRV`, `libSPVRemapper` have no public defined API, **everything** is treated as public, and so **any** changes to these libraries may be considered breaking.

It's my plan to add presubmit checks to verify that the API / ABI is backwards compatible when landing changes without a major version bump.

I've been scratching my head over how to best deal with these libs, and your expertise would be helpful here. I see a couple of options:
* (a) we force these to all be static, and maybe remove them from distro packages. This might upset some developers, but they'll probably be equally upset whenever they update their packages and find a bunch of build failures. As far as I'm aware these libraries were considered 'private' by the authors and were never really expected to be included in distro packages in the first place.
* (b) we take the effort in classifying a public API for each library. This is a whole bunch of work, and will likely end up in breaking changes due to code refactoring. The public API will likely miss symbols currently in use, which might upset others again.

I had been thinking of (a) and then maybe someday (b). Happy to hear other thoughts and opinions though!

arcady-lunarg:
The change to use VERSION/SOVERSION has already been done in a different PR, but the change to make glslang-default-resource-limits STATIC has not and I think still has value, as does the refactoring to use `set_target_properties`. I'm going to rebase this and try to get it merged.

jeremy-lunarg (approved):
LGTM

akien-mga:
Thanks for picking this up! :)

Issue #2421, opened 2020-10-16
Add .gn standalone build configs & presubmits


ben-clayton:
#2413 has highlighted the lack of GN build presubmits, and that downstream build breakages can be painful.

Let's consider adding a standalone GN build configuration, and Kokoro presubmits.

PR #2422, opened 2020-10-16
Add basic GN configurations

ben-clayton:
This allows glslang to be build standalone using the gn build system.

To build with gn:

```
gclient sync --gclientfile=standalone.gclient
gn gen out/default
cd out/default
ninja
```

Issue: #2421

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.a
 *.so
 *.exe
+*.gclient_entries
 .vscode/
 tags
 TAGS
@@ -11,3 +12,4 @@ Test/localResults/
 External/googletest
 External/spirv-tools
 out/
+third_party/llvm-build
--- a/.gn
+++ b/.gn
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 Google, Inc.
+# Copyright (C) 2020 The Khronos Group Inc.
 #
 # All rights reserved.
 #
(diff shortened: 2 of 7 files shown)


ben-clayton:
@null77 - I'm a bit no-idea-dog about all of this. I've taken inspiration from a few other projects, and this seems to work.

Given my lack of expertise in GN, I'm concerned that this may (somehow) cause side effects / breakages for existing gn builds (chrome, fuchsia, etc). In particular, I'm unsure about:

* The addition of new files - does anything change behaviour if these files are added?
* The change in `build_overrides/glslang.gni` - `Externals` does not exist in glslang, but maybe it was named this for some other project?

Is there some way to test whether this upsets the chromium build without landing?

Cheers,
Ben

null77:
Just from looking at the CL it should be harmless. If you want to test, follow the usual ANGLE build setup in our DevSetup page, then after you finish gclient sync, manually cherry-pick this CL into third_party/glslang/src and try building again. Also run 'git cl pre' to exercise local presubmit. If everything builds and presubmit passes you'll have done most of the checks.

ben-clayton:
@null77 - yup all of that seemed to work without error. Thanks.

dj2 (approved):
LGTM.

dj2:
@johnkslang would it be possible to get this merged as it makes working with the GN build a lot simpler.

PR #2428, opened 2020-10-20
Add GL_EXT_fragment_shading_rate

Tobski:
Pending https://github.com/KhronosGroup/SPIRV-Tools/pull/3943 (known_good.json will likely need updating)


SPIRV/spirv.hpp
```diff
@@ -1939,6 +1959,7 @@ inline void HasResultAndType(Op opcode, bool *hasResult, bool *hasResultType) {
     case OpPtrEqual: *hasResult = true; *hasResultType = true; break;
     case OpPtrNotEqual: *hasResult = true; *hasResultType = true; break;
     case OpPtrDiff: *hasResult = true; *hasResultType = true; break;
+    case OpTerminateInvocation: *hasResult = false; *hasResultType = false; break;
```
Tobski, line 1962: This is kind of added as a side effect but it should be harmless; OpTerminateInvocation was added to spir-v headers at the same time.

johnkslang: PRs should not change the content of spirv.hpp, which instead should be a copy of the official one from SPIRV-Headers.  If it works without, just remove it from the PR.  If not, we need to wait for the official content to be available in SPIRV-Headers.

johnkslang: Ah, I had this ready locally, just need to push it.  Doing that now.

johnkslang: Yep, this file now disappears from the diff, so nothing more needs to be done about that.

Tobski: Yea it was just a copy - next time I'll push that in a separate PR. Thanks!

PR #2435, opened 2020-10-23
HLSL: Add support for printf().

amdrexu:
Translate printf() to what GL_EXT_debug_printf has done. HLSL could
define non-constant string variable and we don't have such features
in SPIR-V, so just support constant string variable.


johnkslang:
Change looks good.

@ben-clayton maybe you know, and sorry if I'm being dense about this, but why do these PRs only get results from 2 checks instead of the usual ~12 (or more now)?  How do we trigger getting a fuller set reported?  Thanks.

ben-clayton:
For non-googler accounts I believe you need to apply the `kokoro:run` label. This may not exist, in which case create it with that exact string.

johnkslang:
I see.  I can do that, but am concerned.

We shouldn't be in the situation that non-Googlers don't get their PRs checked automatically. I don't think most people are allowed to add labels.

Historically, we had 6 or 7 checks with each PR, for everyone.  Somehow, we lost something here.  Can we at least get back to the amount of auto-checking that we used to have?  It used to always report Travis, which verified ~7 different builds.

ben-clayton:
> We shouldn't be in the situation that non-Googlers don't get their PRs checked automatically. I don't think most people are allowed to add labels.

Kokoro's PR trigger policy is:
> * Pull request is created by a user that is in the same organization as the repository.
> * Pull request is attached with a `kokoro:run` label. (Pull request authors that do not belong to the organization need to find an authorized user to attach the label for them)

This policy is to avoid running arbitrary executable code on Google owned machines (despite them being run in a very standard GCP sandboxed VM). I do not believe there's any way around this.

> Historically, we had 6 or 7 checks with each PR, for everyone. Somehow, we lost something here. Can we at least get back to the amount of auto-checking that we used to have?

How historically are we talking here? Was this with Kokoro presubmits? Do you have any example PRs you can point me to?

> It used to always report Travis, which verified ~7 different builds.

I haven't removed any Travis bots. @ehsannas - when you set up the original Kokoro bots for this project, did you migrate some from Travis?

Short of migrating all the Kokoro presubmits and CI tests over to something else, I'm afraid this is a policy we cannot change.

johnkslang:
We mostly need to get Travis results displayed again, blocking the PR.  I'm looking now, and see it showing up in the PR checks less and less around June/July, in the middle of a bunch of build changes, but maybe that's a coincidence.  It still seemed to execute, until 11 days ago, it just hasn't been contributing to the list of checks on the PR.  Maybe it's two problems, why nothing in 11 days and why before that it stopped showing up in PRs.

Looking at change history, one issue I see is 4aeca2df where @dneto0 migrated `.travis.yml` to use `https://github.com/dneto0/android-ndk.git`, but when I look there, it says

> This repository will be removed in the near future. Please use `echo y |sdkmanager "ndk-bundle"` instead.

So, that's something that should be sorted out. Maybe the NDK part of Travis need no longer exist.

johnkslang:
It seems to be happening to others, in a similar time frame:  https://travis-ci.community/t/travis-ci-build-status-doesnt-show-up-anymore-in-pull-requests/8718

greg-lunarg:
@johnkslang Could this be the reason for the missing Travis runs?: https://docs.travis-ci.com/user/migrate/open-source-repository-migration

johnkslang:
Travis is working, at least for the moment.

PR #2440, opened 2020-10-28
Do not use PropagateLineInfoPass and RedundantLineInfoElimPass

jaebaek:
Since spirv-opt will remove PropagateLineInfoPass and
RedundantLineInfoElimPass, glslang should not use it. spirv-opt will
propagate the line instructions and eliminate the redundant lines by
default in IR loading/emission.

### DIFF
--- a/SPIRV/SpvTools.cpp
+++ b/SPIRV/SpvTools.cpp
@@ -174,10 +174,7 @@ void SpirvToolsTransform(const glslang::TIntermediate& intermediate, std::vector
     // line information into all SPIR-V instructions. This avoids loss of
     // information when instructions are deleted or moved. Later, remove
     // redundant information to minimize final SPRIR-V size.
-    if (options->generateDebugInfo) {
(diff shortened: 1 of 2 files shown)


jaebaek:
This PR is related to https://github.com/KhronosGroup/SPIRV-Tools/pull/3951.
https://github.com/KhronosGroup/SPIRV-Tools/pull/3951 will remove `PropagateLineInfoPass` and `RedundantLineInfoElimPass`. It will propagate the line instructions and eliminate the redundant lines by
default. This change will fix the shaderc smoke test failure of https://github.com/KhronosGroup/SPIRV-Tools/pull/3951.

johnkslang:
It looks like you didn't update the test results, or if they should have been the same, something more subtle is going on.

jaebaek:
Yes. I will update this PR after submitting https://github.com/KhronosGroup/SPIRV-Tools/pull/3951 first. It's a chicken and an egg issue. They both fail because of each other.

Issue #2445, opened 2020-11-03
Recent BUILD.gn changes broke chromium build


courtney-g:
Changes in https://github.com/KhronosGroup/glslang/pull/2442 broke the chromium build:
ERROR at //third_party/glslang/src/BUILD.gn:320:3: Undefined identifier.
  include_dirs += [ "${spirv_tools_dir}/include" ]
  ^-----------
See //third_party/angle/src/tests/BUILD.gn:286:9: which caused the file to be included.
        "$angle_glslang_dir:glslang_lib_sources",
        ^---------------------------------------
ninja: error: rebuilding 'build.ninja': subcommand failed

include_dirs is not defined at that point. The "+" is needs to be dropped.

phniix:
I love it when the world comes together to solve problems!

@courtney-g , thanks a bunch for taking point on this and solving this little drama! It's been popping up, and addressed here: https://github.com/KhronosGroup/glslang/issues/2405
and also manifested itself earlier here: https://github.com/KhronosGroup/glslang/issues/2413

I did my best to handle and canvas opinion on how to do apply what I thought needed to be done to solve it.. given my utter lack of know-how for the GN build system. See here: https://github.com/KhronosGroup/glslang/issues/2413#issuecomment-711035068 , and here: https://github.com/KhronosGroup/glslang/issues/2405#issuecomment-719745732

So it would be lovely to get and gauge an understanding of how best to combat the dramas that can come up with the GN build system, particularly given the context described in the links above.

@johnkslang has now turned that original issue https://github.com/KhronosGroup/glslang/issues/2405 into a question about GN infrastructure changes. It would be lovely if you, and perhaps your teams, can contribute to that discussion. It will really help the community out.

PR #2442, opened 2020-10-31
Add new SpirvToolsDisassemble API interface + Improve Doc on existing API interface

phniix:
This addresses the problems highlighted in: https://github.com/KhronosGroup/glslang/issues/2405

Add more flexible SpirvToolsDisassemble interface to allow specifying `spv_target_env` for disassembly output.
Improve documentation on existing SpirvToolsDisassemble interface.

PR #2446, opened 2020-11-03
Fix build error with Chromium & ANGLE

courtney-g:
Getting error about undefined symbol (include_dir) at line 320.
Was trying to append to a non-existant variable.

Bug #2445


phniix:
Thanks for tackling this! Not many GN experts were able to come together for the PR: https://github.com/KhronosGroup/glslang/pull/2442 before it was merged.

Perhaps this will provide a use case and example of how the GN team can help mitigate such little dramas before they occur. If not, then your approach here, by being part of the solution, is another terrific example of collaboration!

PR #2451, opened 2020-11-05
Don't compile round() to roundEven in DX9 compatibility mode

rdb:
The [round_ne](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/round-ne--sm4---asm-) instruction was introduced in SM 4.0.  When targeting SM 3.0, FXC actually generates the equivalent of `floor(x + 0.5)` (via the frc instruction), whereas it switches to round_ne when targeting SM 4.0.

However, glslang maps round() to roundEven() even in DX9 compat mode.  This means that DX9 shaders that use round() can't actually be compiled by SPIRV-Cross to run on DX9, because SPIRV-Cross (rightly) refuses to compile shaders containing GLSLstd450RoundEven to SM 3.0.

This changes that so that round() maps to GLSLstd450Round (which has unspecified rounding semantics) in DX9 compat mode, and continues to map to GLSLstd450RoundEven otherwise.


greg-lunarg:
Can no longer see the details of the CI failure. Could you retrigger?

rdb:
I don't think the CI failure was related to this change (though I wasn't able to see the details so I can't know for sure).  I just rebased this from latest master so let's see if it happens again.

PR #2458, opened 2020-11-12
Fix issue for new unique id system.

ShchchowAMD:


ShchchowAMD:
The root cause is within TGlslangToSpvTraversery:visitSymbol(), rValueParameters needs to check unique id in symbol table.

But in function ProcessDeferred() (ShaderLang.cpp),

symbolTable->adoptLevels(*cacheTable) will reset the unique ID, using max id in cache at that moment, and previous ids will be no longer effective. So the fix is to re-set those ids following above call.

Besides, it is needed to explicitly using setUniqueId() before parse() currently. Which could ensure id has been initialized successfully.

FYI. @mbechard .
Ref: #2396  #2156

mbechard:
Thanks, trying this out against my case that's currently failing

ShchchowAMD:
Would you mind to attach example shaders showing the case we discussed before here?

So that I could debug with it.

mbechard:
The shaders that it's failing on requires:
https://github.com/mbechard/glslang/tree/GL_EXT_vulkan_glsl_relaxed
So it would fail to compile on master branch. Would you still be interested in them?

ShchchowAMD:
Ok, I see those shaders added in latest commit.

It is ok, I can have a try if above issue still cause failing.

mbechard:
The shaders that show the error aren't in the test cases actually. The one I see failing are far more complex.

This fix doesn't solve the issue though. The error occurs during the link stage when the multiple compilation units are merged together into a single one. Doesn't this fix occur before that?
I think the solution needs to occur at the point where seedIdMaps() and remapIds(). The point of those functions is to ensure that within the single combined compilation unit all the globals variables used between the multiple units now point to the same id, and any local variables all have unique ids. I think it's important that stays true at that point so any routines making use of the unique ids afterwards have that state setup for them. I think limiting to the usage cases of TGlslangToSpvTraversery:visitSymbol is too late.

The fact that the new updateUniqueIdLevelFlag() clamps at a max of 7 seems questionable to me. Why not just increase the uniqueIds to 64-bit. Give more bits for the 'level', and then some other bits can be used for the compilation unit index when doing the merging, which would make things unique.

ShchchowAMD:
I agreed to increase the uniqueIds to 64-bits.

Three bits for levels may lead to issues of same unique id, as built-in symbols will also use some of them.

Currently I don't have a test case to re-produce this issue and verify my fix. I can make a change for this but can't make sure whether it could fix your local problems.

PS. This PR is for the problem mentioned in comments. We could create another PR for 64-bits uniqueIds.

mbechard:
I'll be happy to test any fix you make. I can also try to pull out the sample shaders, it's just a little tricky cause they are procedurally generated in code.

ShchchowAMD:
I fully understand that. It would be much helpful if I could get some simple sample shaders for local verifying.

Anyway, I'll make another change for this, expanding uniqueIds to 64 bits.

mbechard:
This is the opposite of a minimal case, but it does reproduce it if you are using the vulkan_relaxed extension in the above branch.
Remove the .txt from the filenames and a
dd these three files to the test cases in VkRelaxed.FromFile.cpp and notice the generated spirv. I've attached it here as well (output.spirv.txt). Specifically on line 698 in the spirv it's trying to store to variable 213(t), which hasn't been declared above in the Variable Function section.


[vs3.vert.txt](https://github.com/KhronosGroup/glslang/files/5539626/vs3.vert.txt)
[vs4.vert.txt](https://github.com/KhronosGroup/glslang/files/5539627/vs4.vert.txt)
[vs1.vert.txt](https://github.com/KhronosGroup/glslang/files/5539628/vs1.vert.txt)
[output.spirv.txt](https://github.com/KhronosGroup/glslang/files/5539632/output.spirv.txt)

ShchchowAMD:
I'll have a try.

mbechard:
Hey @ShchchowAMD, any chance to look at this again? Thanks

ShchchowAMD:
Hi mbechard,

I'll try to fix above issue this week.

BR.

johnkslang:
Any update on this?  It sounds like it might relate to the recent issue #2495.

greg-lunarg:
So to confirm: it is believed that this change fixes a previous problem and does not cause any regressions.

mbechard:
Merging multiple compilation units is still broken because the ids being assigned to variables are not made unique during the merging. This pull describes the potential solution: upping the ids to 64-bit, reserving some bits for the module #, some of the level # and the rest as the ID.

greg-lunarg:
Yes, but while we are waiting for the 64-bit fix, I am wondering if there is benefit (and no harm) in merging the current change. We could open another issue pointing at this PR and the outstanding issue you have outlined.

mbechard:
I'm not sure if there were other issues that ShchchowAMD was looking to fix with this. I think this PR was meant to fix a regression I brought up in PR #2396, but it doesn't quite do the trick in it's current form.
Maybe there are other cases not mentioned in the thread though. Anyway, I'll let @ShchchowAMD answer.

ShchchowAMD:
When I changed it to 64 bits (currently I set 8 bits for level), I met some issues in GlslangToSPIRV. One of them is : some variables are compiled to wrong types (seems to be a new issue caused by id), I am now debugging with those issues.

mbechard:
This new commit seems to fix the issue for all of test cases I'm running through.

VadikLeshy:
This PR fixes #2509 

Thanks @ShchchowAMD !

ShchchowAMD:
@greg-lunarg  Hi greg, I think this would be all for fixing issues mentioned above. Please take a look whether it is ok to merge, Thanks!

greg-lunarg:
@ShchchowAMD Will review presently. Thanks!

mbechard:
Sorry, looking at the code closer now, I think this fix still doesn't fix the root issue. The test cases are just luckily passing due to the larger numeric space.
The core issue is that when merging compilation units the ids of the symbols need to be made unique. Previous to the id overhaul this way done by traversing unit0, keeping track of it's highest id value. Then when merging unit1 with unit0, each id in unit1 would be offset by unit0Max. Then ids in unit2 would be offset by unit1Max. Ensuring they never used the same values.
The introduction of using the upper bits in the id for the 'level' breaks this logic. My suggestion of upgrading to 64-bit included the need to reserve some bits for 'unit index', which allows the ids to be made unique by virtue of the 'unit index' bits being unique per unit, and doing away with the maxId logic.

ShchchowAMD:
Thanks for testing and analyzing, I'll try to fix it !

mbechard:
No problem. To be specific, the functions that will likely need to be re-worked are seedIdMap() and remapIds()

ShchchowAMD:
Seems there's an error occured in CI during git clone.

RuntimeError: Failed to run ['git', 'clone', 'https://github.com/KhronosGroup/SPIRV-Tools', '.'] in External/spirv-tools
The command "./update_glslang_sources.py" failed and exited with 1 during .

greg-lunarg:
Probably just an intermittent failure. I have retriggered the failing build.

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -344,12 +345,12 @@ class TBuiltInIdTraverser : public TIntermTraverser {
         }
         maxId = std::max(maxId, symbol->getId());
```
mbechard, line 346: This line is doing a max() using the 'level' included still. This will result in incorrect results if the level is different for any symbols.

mbechard:
Can I suggest a different algorithm? 
* Reserve 10-bits next to your level bits to be the unit id.
* During the seedIdMap inspect these bits and keep track of either 1. the highest value or 2. keep a std::set of each value found. * After seedIdMap, determine a value that is not used in the existing symbols, which is either the highest value + 1, or a value not present in the std::set.
* Pass that value into remapIds to be placed into those reserved 10-bits for any ids that need to be made unique

This avoids errors where bitmasking in/out the level is missed. This also avoids future cases if the ids for a symbols are assigned more randomly and not always closish to 0 in value. The current 'max' algorithm fails if for some reason IDs get assigned large values during initial compilation.

glslang/MachineIndependent/linkValidate.cpp
ShchchowAMD: I'll check this part.

ShchchowAMD:
The new BitMask was supposed to resolve regression caused by level id and keep same behaviors as before. I'll add missing part for built-in variables.

I checked some history versions. Seems that the issue mentioned above would happen the same way when "IDs get assigned large values during initial compilation." because of maxId.

I think it is better to make another new PR for this issue. I'd submit one later and send review request.

ShchchowAMD:
To double check, I think the idea you shared above is to replace maxId shift with reserved unit bits shift.

For example, during remapping, maxUnitId will be used for idshift to keep symbol ids unique.

As usually there won't be too many units during merging, this may fix current issue when there're large number of symbols.

mbechard:
Correct, the "IDs get assigned large values during initial compilation." issue exists from previous code, and only in theory anyways. I don't think the compiler actually generates IDs that would be an issue right now. It was just mentioned as a 'bonus' of this different workflow, as a form future proofing.

ShchchowAMD:
Just in case, would current version fix your previous issues?

I may need several days to locally test unit ids, to make sure it won't meet issues like redeclared symbols with level bits.

Anyway, I'll make another CL for potential unit bit solution. Let's merge this commit for other two issues' fix currently.

mbechard:
Yes, if the max() issue is fixed I think the current solution does fix the current issues.

greg-lunarg:
@mbechard Are you suggesting that one more fix is required before these commits can be pushed, or that these commits can be pushed now even though a final fix will be needed afterwards to fix all issues?

mbechard:
In it's current state this PR still has one defect I see, which is the std::max() line issue I pointed out in the review note. If you're looking to merge this ASAP, then this PR is certainly better than the current state of the master branch, so go ahead. 
Merging compilation unit still can potentially fail due to that remaining defect, but so far my test cases don't run into that defect, so it may be rare.

greg-lunarg:
@mbechard Do these recent two commits address the remaining problems?

mbechard:
Yep looks like it. Thanks

Test/baseResults/link.vk.multiBlocksValid.1.0.geom.out
```diff
@@ -328,14 +328,14 @@ output primitive = triangle_strip
...
-                              Decorate 68(iV) Location 1
+                              Decorate 68(iV) Location 0
                               MemberDecorate 95(BufferBlock) 0 ColMajor
                               MemberDecorate 95(BufferBlock) 0 Offset 0
                               MemberDecorate 95(BufferBlock) 0 MatrixStride 16
                               Decorate 95(BufferBlock) BufferBlock
                               Decorate 97(uBuf) DescriptorSet 0
                               Decorate 97(uBuf) Binding 1
-                              Decorate 100(P) Location 0
+                              Decorate 100(P) Location 2
```
greg-lunarg, line 338: A little concerned that we are leaving a gap here ie that linking has left a hole in the assigned input slots. Is this a potential problem? Aren't there a fixed number of these? Might an application that previously didn't run out of slots now run out?

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -375,12 +378,12 @@ class TUserIdTraverser : public TIntermTraverser {
 };
 
 // Initialize the the ID map with what we know of 'this' AST.
-void TIntermediate::seedIdMap(TIdMaps& idMaps, int& maxId)
+void TIntermediate::seedIdMap(TIdMaps& idMaps, long long& maxId)
```
greg-lunarg, line 381: Should we rename maxId to idShift since it is not a true id (ie it has its level stripped off) and it is really going to be used to do a shift? Should we also improve the function comment to explain this?

glslang/MachineIndependent/localintermediate.h
```diff
@@ -926,14 +929,14 @@ class TIntermediate {
...
-    TIntermSymbol* addSymbol(int Id, const TString&, const TType&, const TConstUnionArray&, TIntermTyped* subtree, const TSourceLoc&);
+    TIntermSymbol* addSymbol(long long Id, const TString&, const TType&, const TConstUnionArray&, TIntermTyped* subtree, const TSourceLoc&);
     void error(TInfoSink& infoSink, const char*);
     void warn(TInfoSink& infoSink, const char*);
     void mergeCallGraphs(TInfoSink&, TIntermediate&);
     void mergeModes(TInfoSink&, TIntermediate&);
     void mergeTrees(TInfoSink&, TIntermediate&);
-    void seedIdMap(TIdMaps& idMaps, int& maxId);
-    void remapIds(const TIdMaps& idMaps, int idShift, TIntermediate&);
+    void seedIdMap(TIdMaps& idMaps, long long& maxId);
```
greg-lunarg, line 938: Rename maxId to idShift here too?

glslang/MachineIndependent/SymbolTable.h
```diff
@@ -691,6 +694,16 @@ class TSymbolTable {
         return table[currentLevel()]->amend(symbol, firstNewMember);
     }
 
+    // Update the level info in symbol's unique ID to current level
+    void amendSymbolIdLevel(TSymbol& symbol)
+    {
+        // clamp level to avoid overflow
+        uint64_t level = currentLevel() > MaxLevelInUniqueID ? MaxLevelInUniqueID : currentLevel();
+        uint64_t symbolId = symbol.getUniqueId();
+        symbolId &= ((1LL << LevelFlagBitOffset) - 1);
```
greg-lunarg, line 703: Use uniqueIdMask here?

glslang/MachineIndependent/SymbolTable.h
```diff
@@ -876,19 +889,24 @@ class TSymbolTable {
     // Add current level in the high-bits of unique id
     void updateUniqueIdLevelFlag() {
         // clamp level to avoid overflow
-        uint32_t level = currentLevel() > 7 ? 7 : currentLevel();
-        uniqueId &= ((1 << LevelFlagBitOffset) - 1);
+        uint64_t level = currentLevel() > MaxLevelInUniqueID ? MaxLevelInUniqueID : currentLevel();
+        uniqueId &= ((1LL << LevelFlagBitOffset) - 1);
```
greg-lunarg, line 893: Use uniqueIdMask here?

greg-lunarg (changes requested):
Just one concern regarding new(?) gap in slot assignment.

glslang/MachineIndependent/SymbolTable.h
ShchchowAMD: Fixed.

glslang/MachineIndependent/SymbolTable.h
ShchchowAMD: Fixed.

glslang/MachineIndependent/localintermediate.h
ShchchowAMD: Fixed.

Test/baseResults/link.vk.multiBlocksValid.1.0.geom.out
ShchchowAMD: Glslang uses symbol ids in priority queue during io mapping / reflection.
So for IO symbols, they now have different order when we auto assigning them locations, because their ids have changed.

For this part, symbol P was the first one in master branch, so its location is 0 (same with old version before my io array change).
But seems its id now gets behind iVV, so the location is 2 now (after iVV not listed here).

glslang/MachineIndependent/linkValidate.cpp
ShchchowAMD: Fixed.

Test/baseResults/link.vk.multiBlocksValid.1.0.geom.out
greg-lunarg: My mistake. I thought a gap was created but it wasn't. Because iV is a struct with two vec4s, it takes two slots. P takes one slot. So with P at 0, iV is at 1. But when they are reordered, when iV is at 0, P is at 2.

greg-lunarg:
This all looks good. Could you please squash into onc commit? Then I will merge.

greg-lunarg:
@ShchchowAMD Can you please squash your 8 commits down to 1 so I can merge? Thanks!

greg-lunarg:
Fixes #2495

greg-lunarg:
Fixes #2509

ShchchowAMD:
OK

greg-lunarg:
I believe the failed ubuntu-clang-gn was due to how far behind the branch was. I am going ahead with merge.

greg-lunarg:
ubuntu-clang-gn pass confirmed with #2534

PR #2466, opened 2020-11-23
Updates for final Vulkan ray tracing extensions

dgkoch:
GLSL specs updates: https://github.com/KhronosGroup/GLSL/pull/144
SPV spec updates: https://github.com/KhronosGroup/SPIRV-Registry/pull/86

Depends on the following changes landing first: 
SPIRV-Headers: https://github.com/KhronosGroup/SPIRV-Headers/pull/182
SPIRV-Tools: https://github.com/KhronosGroup/SPIRV-Tools/pull/4027


dgkoch:
cc @johnkslang

glslang/MachineIndependent/glslang_tab.cpp
```diff
@@ -1097,129 +974,122 @@ static const yytype_int16 yytranslate[] =
...
-/* The user-facing name of the symbol whose (internal) number is
-   YYSYMBOL.  No bounds checking.  */
-static const char *yysymbol_name (yysymbol_kind_t yysymbol) YY_ATTRIBUTE_UNUSED;
-
+#if YYDEBUG || YYERROR_VERBOSE || 1
 /* YYTNAME[SYMBOL-NUM] -- String name of the symbol SYMBOL-NUM.
    First, the terminals, then, starting at YYNTOKENS, nonterminals.  */
 static const char *const yytname[] =
 {
-  "\"end of file\"", "error", "\"invalid token\"", "CONST", "BOOL", "INT",
```
dneto0, line 1184: This change is causing a difference in the error message, with the new message being worse.


-ERROR: 0:66: '' :  syntax error, unexpected end of file
+ERROR: 0:66: '' :  syntax error, unexpected $end

glslang/MachineIndependent/glslang_tab.cpp
```diff
@@ -1,9 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.7.2.  */
+/* A Bison parser, made by GNU Bison 3.0.4.  */
```
dneto0, line 1: Looks like the new version of the grammar was generated with an older Bison.
 Is this the cause of the change in the erorr message noted below?

dneto0 (changes requested):
Some failures are caused by using an older bison.
Using bison 3.7 will fix the "end of file" error messages

Test/baseResults/rayQuery-global.rgen.out
```diff
@@ -0,0 +1,76 @@
+rayQuery-global.rgen
+// Module Version 10000
+// Generated by (magic number): 8000a
+// Id's are bound by 44
+
+                              Capability RayQueryProvisionalKHR
```
dneto0, line 6: This should be RayQueryKHR and RayTracingKHR

Test/baseResults/spv.ext.ClosestHitShader_Subgroup.rchit.out
```diff
@@ -0,0 +1,114 @@
...
+              20:   16(fvec3) ConstantComposite 19 19 19
+              21:   15(float) Constant 1061158912
+              22:             TypeInt 32 1
+              23:     22(int) Constant 1
+              24:             TypeVector 15(float) 4
+              25:             TypePointer IncomingRayPayloadKHR 24(fvec4)
+26(incomingPayload):     25(ptr) Variable IncomingRayPayloadKHR
+              27:             TypePointer Input 10(int)
+28(gl_SubgroupInvocationID):     27(ptr) Variable Input
+              31:             TypeVector 10(int) 4
```
dneto0, line 70: I'm getting a test difference around here:

-              31:             TypeVector 10(int) 4
-              32:             TypeInt 64 0
-              33:             TypePointer Input 31(ivec4)
+              31:             TypeInt 64 0
+              32:             TypeVector 10(int) 4
+              33:             TypePointer Input 32(ivec4)

dneto0:
I also recommend updating the known_good.json file in the root source directory, to ensure we validate against updated SPIRV-Tools.

johnkslang:
I get a bunch of these "Validation failed" messages locally, when the only thing I change is spriv.hpp, and the corresponding spelling changes in glslang.  It should make no difference at all in the generated result, and I verified that for at least one result, but the rebuilt validator (same code base, just the new header) says "Validation failed" on an extra 16 ray-tracing tests.

glslang/MachineIndependent/glslang_tab.cpp
dgkoch: Regenerated the grammar with bison 3.7.

Test/baseResults/rayQuery-global.rgen.out
dgkoch: fixed.

glslang/MachineIndependent/glslang_tab.cpp
dgkoch: hopefully fixed by updating grammar with bison 3.7.

Test/baseResults/spv.ext.ClosestHitShader_Subgroup.rchit.out
dgkoch: I haven't seen this issue...

dgkoch:
@johnkslang @dneto0 I think I've addressed all the outstanding issues in the latest push, including updated spirv-headers and spirv-tools.

johnkslang:
I think this PR needs to move to use the latest spirv.hpp from SPIRV-Headers. (That's what I was doing locally, trying to make a PR to move to the latest header, but having a validator issue.  In the meantime, discovered the header wouldn't match what's here anyway.)

E.g., see `OpConvertUToAcclerationStructureKHR` which is spelled more correctly in SPIRV-Headers as `OpConvertUToAccelerationStructureKHR`.

dneto0:
(Thumbs up to using latest spirv-headers and spirv-tools. I  haven't looked at the spelling of the acceleration structure)

johnkslang:
There are also differences in the switch statement for `OpTerminateInvocation` in spirv.hpp.

dgkoch:
> There are also differences in the switch statement for `OpTerminateInvocation` in spirv.hpp.

That seems to be a previous manual addition to spirv.hpp that seems to be missing in spriv-headers?

dneto0:
I think I found a bug resulting in indeterminate ordering of the output of the 64-bit int type vs. the ivec4 type.  
It doesn't seem related to this PR?!  
I'll file a separate PR for it.

dneto0:
Ah, this PR *exposed* the pre-existing bug with indeterminate order.

This is the fix:

$ git diff HEAD^
diff --git a/SPIRV/GlslangToSpv.cpp b/SPIRV/GlslangToSpv.cpp
index ba4264c2..a2494ec7 100644
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2122,8 +2122,9 @@ std::pair<spv::Id, spv::Id> TGlslangToSpvTraverser::getForcedType(glslang::TBuil
             // these require changing a 64-bit scaler -> a vector of 32-bit components
             if (glslangType.isVector())
                 break;
-            std::pair<spv::Id, spv::Id> ret(builder.makeVectorType(builder.makeUintType(32), 4),
-                                            builder.makeUintType(64));
+            spv::Id ivec4_type = builder.makeVectorType(builder.makeUintType(32), 4);
+            spv::Id uint64_type = builder.makeUintType(64);
+            std::pair<spv::Id, spv::Id> ret(ivec4_type, uint64_type);
             return ret;
         }
         // There are no SPIR-V builtins defined for these and map onto original non-transposed

The problem is that evaluation order of arguments is unspecified.  It's compiler-dependent.  So enforce an order.

@dgkoch ^

dgkoch:
Updated to the latest spirv.hpp in this CL (early on).

johnkslang:
So, #2468 passes the bot tests and has the same base known_good as this one.

Maybe it should go in first.  (After looking into the non-determinism.)

dgkoch:
> So, #2468 passes the bot tests and has the same base known_good as this one.
> 
> Maybe it should go in first. (After looking into the non-determinism.)

I'd rather not put that one in. It'll conflict with all the test updates in this one.

Test/baseResults/spv.ext.ClosestHitShader_Subgroup.rchit.out
johnkslang: I also have not seen this.

Maybe that's because we're not watching Travis anymore and not seeing multi-platform test differences?

johnkslang: Which bot tests are actually running all the tests and comparing results?  Only Appveyor?

We lost Travis and added like 7 others instead, but I'm still unclear if they are all just build tests, whereas Travis tests 6 different execution environments.

dneto0:
(Sorry, I should have been clear.  The indetrminate order issue is not a bug: generated code was always correct.  But it interferes with maintaining checked-in tests.  This PR was the first to exercise the issue.)

johnkslang:
> (Sorry, I should have been clear. The indetrminate order issue is not a bug: generated code was always correct. But it interferes with maintaining checked-in tests. This PR was the first to exercise the issue.)

But, historically (until a few months ago?), we would also have considered this an error, because we require getting the same test results on multiple compilers to pass the bot tests and turn green.  I have fixed similar things before, due to that.

I'm worried this means we've dropped that kind of testing barrier in recent times, at least in part due to Travis CI not running.

johnkslang:
> I'd rather not put that one in. It'll conflict with all the test updates in this one.

If we put it in, this one would need a rebase.  If we don't put it in, we need to manually sign off that we verified the header is identically (what I was doing that discovered some of these issues), which is probably acceptable.

For normal process in the future I highly recommend:

1. a PR that updates to the latest headers and tools, both the actual `SPIRV/spirv.hpp` and in the "known_good"
2. a follow-on PR that adds the new functionality, where the headers/tools are not changed

johnkslang:
The non-determinism made appveyor get the other result.

Let's fix that as well, and see all the bots passing, before doing the merge.  I think it's okay to do that as part of this PR; just one more small update to it.

dgkoch:
Added David's patch from above. 
The included spirv.hpp should exactly match what's in SPIRV-Headers.

PR #2468, opened 2020-11-23
Move to the most recent SPIRV-Headers and SPIRV-Tools.

johnkslang:


johnkslang:
This moves to the latest SPIRV-Headers and SPIRV-Tools, which includes additional "Validation failed" from my test results.  I don't know if that will be replicated here, but is of potential value in comparing to the ray-tracing changes in #2466.

johnkslang:
Superseded by #2466, which moved to the same content.

PR #2469, opened 2020-11-24
Add ray query capability if acceleration structure or ray query types declared

### LINKED ISSUE #2430: Fragment shader ray query, invalid SPIR-V
Hello,

I've run into what I believe is a bug with glslValidator - although it could be the khronos validation layer too: Having tested it with functional shader code, the shader program runs fine, despite the validation error message.
I've got the following fragment shader:
```glsl
#version 460
#extension GL_ARB_separate_shader_objects : enable
#extension GL_EXT_nonuniform_qualifier : enable
#extension GL_GOOGLE_include_directive : enable
#extension GL_EXT_scalar_block_layout : enable
#extension GL_EXT_ray_tracing : enable
#extension GL_EXT_ray_query : enable

layout(location = 0) out vec4 outColor;

layout(binding = 1, set = 0) uniform accelerationStructureEXT topLevelAS;

void main() {
  outColor = vec4(0.0);
}
```
It compiles without issue, but upon loading the spir-v via `vkCreateShaderModule`, I get a validation error:
```
Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x1c6cf1057b8, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: Opcode TypeAccelerationStructureKHR requires one of these capabilities: RayQueryProvisionalKHR RayTracingNV RayTracingProvisionalKHR
  %49 = OpTypeAccelerationStructureKHR
```
The error is gone if I remove the declaration of topLevelAS. The error is also gone as soon as I actually make use of the topLevelAS in the code somewhere, so it's not exactly critical. I've tested it with the current master branch of glslang. Maybe some optimizing happens, the code compiles without the mentioned capabilities, which results in the error?

dgkoch:
Fixes #2430

in non-ray tracing stages and the extension is enabled


dgkoch:
cc @alelenv @johnkslang @dneto0

alelenv:
LGTM

PR #2471, opened 2020-11-28
update spirv-headers and fix handling of gl_HitTEXT

dgkoch:
Update spirv-headers known_good to f027d53ded7e230e008d37c8b47ede7cd308e19d
and update SPIRV/spirv.hpp to copy from that version as well.

In GLSL gl_HitTNV/gl_HitTEXT is defined as an alias of gl_RayTmaxNV/gl_RayTmaxEXT
SPV_NV_ray_tracing has a dedicated HitTNV which gl_HitTNV maps to.
For SPV_KHR_ray_tracing, gl_HitTEXT gets mapped to a RayTmaxKHR decoraged variable
to simplify the SPIRV consumer.
This change fixes the mapping for the GL_EXT_ray_tracing extension, and updates
the test results to match.

Update MissNV shader test to not use ObjectRay builtins
- They shouldn't exist in the miss stage because there is no object intersected


dgkoch:
cc @johnkslang @dneto0

This is needed after https://github.com/KhronosGroup/SPIRV-Tools/pull/4040 and should probably be landed as a precursor to https://github.com/KhronosGroup/SPIRV-Tools/pull/4041

johnkslang:
Okay.  Will like to update SPIRV-Headers first.  We should not be seeing individual edits here.

Maybe it's really time to stop using a copy.  It was needed early on, but now submodule might be better.  A submodule can still control which commit is being relied on, but otherwise does all the rest automatically (when defaults are updated to operate recursively).

johnkslang:
Ah, this was a backward incompatible change to the header.  That should like almost never happen.  Will again bypass normal process of first copying in the header to glslang.

PR #2478, opened 2020-12-07
Update spirv-tools known-good #2 - Pick up ray tracing terminator fix

greg-lunarg:


greg-lunarg:
It appears that this PR is running into googletest build failures in the Travis CI seen in previous commits. I don't think anything in this PR should cause the failures being seen here.

PR #2481, opened 2020-12-10
Fix Travis to use pre-breakage googletest

### LINKED ISSUE #2480: Travis CI for glslang broken by latest googletest
The current googletest top-of-tree is breaking Travis CI for glslang. See https://github.com/KhronosGroup/glslang/pull/2472

The googletest problem is documented by issue https://github.com/google/googletest/issues/3128

The googletest problem was introduced by https://github.com/google/googletest/commit/fbef0711cfce7b8f149aac773d30ae48ce3e166c

There are a number of ways to fix this problem.

1) hack .travis.yml to checkout a known good version of googletest after cloning it

2) make googletest part of the glslang known-good.json mechanism. These versions are fetched by update_glslang_sources.py, which is currently called by .travis.yml.

3) Another is some solution involving submodules. This would be a longer term solution.

#1 is probably the easiest and fastest solution. That is my vote. I will submit a PR shortly if I don't hear any concerns.

greg-lunarg:
Fixes #2480

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,17 @@ install:
     fi
 
 before_script:
-  - git clone --depth=1 https://github.com/google/googletest.git External/googletest
+  # check out pre-breakage version of googletest; can be deleted when
+  # issue 3128 is fixed
+  # git clone --depth=1 https://github.com/google/googletest.git External/googletest
+  - mkdir -p External/googletest
+  - cd External/googletest
+  - git init
+  - git remote add origin https://github.com/google/googletest.git
+  - git fetch --depth 1 origin 0c400f67fcf305869c5fb113dd296eca266c9725
+  - git reset --hard FETCH_HEAD
+  - cd ../..
+  # get spirv-tools and spirv-headers
   - ./update_glslang_sources.py
 
 script:


greg-lunarg:
@ben-clayton Would you mind sanity checking what I am doing here?

Essentially, googletest recently committed some code that caused the glslang Travis build to quit building. See #2480 for details.

I have hacked .travis.yml to use the last known good version of googletest until it is fixed.

Thanks!

ben-clayton:
Looks okay - we'll have to remember to roll it forward again once it is fixed.

It's more efficient to just grab the particular change you want than having to fetch everything since a particular date.
You can do that with:

```
mkdir -p External/googletest
cd External/googletest
git init
git remote add origin https://github.com/google/googletest.git
git fetch --depth 1 origin 0c400f67fcf305869c5fb113dd296eca266c9725
git reset --hard FETCH_HEAD
```

greg-lunarg:
@ben-clayton Thanks! Will give that code a try.

johnkslang:
Can you also update the README to say what commit to check out (so humans also avoid the problem)?

greg-lunarg:
@johnkslang Good idea. Thanks!

PR #2500, opened 2021-01-05
Prevent HLSL input under OpenGL semantics

### LINKED ISSUE #2497: Combining -G and -D is currently not supported and should generate error.
It was not the intention that -G would work with -D, that is, that glslangValidator would create SPIR-V from HLSL source that would work with OpenGL. Specifically, it would not join separate texures and samplers into single sampled textures.

The man page should be changed to point out this is not allowed and an error should be generated from glslangValidator when attempted.

FWIW, it is reported that running the resultant SPIR-V from HLSL into spirv-cross to get GLSL and running that back through glslangValidator can create the desired result.

jeremy-lunarg:
Fix #2497.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -844,6 +844,10 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
         (Options & EOptionReadHlsl) == 0)
         Error("uniform array flattening only valid when compiling HLSL source.");
 
+    if ((Options & EOptionReadHlsl) && (Client == glslang::EShClientOpenGL)) {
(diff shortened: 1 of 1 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -1561,7 +1565,8 @@ void usage()
            "              'ver', when present, is the version of the input semantics,\n"
            "              which will appear in #define GL_SPIRV ver;\n"
            "              '--client opengl100' is the same as -G100;\n"
-           "              a '--target-env' for OpenGL will also imply '-G'\n"
+           "              a '--target-env' for OpenGL will also imply '-G';\n"
```
greg-lunarg, line 1568: Good consistency addition :)

greg-lunarg (approved):
LGTM.

PR #2511, opened 2021-01-25
Implement GL_EXT_null_initializer

alan-baker:
Adds null initializer syntax (empty braces)
Allows null initialization of shared variables

See also https://github.com/KhronosGroup/GLSL/pull/151 (GLSL extension).
See also https://github.com/KhronosGroup/SPIRV-Tools/pull/4124 (SPIRV-Tools validation).

Ported and squashed from internal MR 76.


alan-baker:
@johnkslang moving this into the public.

Issue #2400, opened 2020-09-22
SPIR-V validation fails in scalar block layouts


godlikepanos:
Normally, the SPIR-V validator (spirv-val) requires the `--scalar-block-layout` argument in order to relax the block layout rules and to be able to work with VK_EXT_scalar_block_layout. Unfortunately SPIR-V doesn't have a construct to indicate which block layout is in use and that's probably the reason why `--scalar-block-layout` exists. The problem is that glslang invokes the spirv-val without the equivalent of `--scalar-block-layout` and in some cases that throws incorrect errors. 

The following glsl code:
```
#version 460 core
#extension GL_EXT_scalar_block_layout : enable

struct Foo
{
	vec3 m;
	vec3 f;
};

layout(binding = 0, scalar) buffer b
{
	Foo f;
};

void main()
{
}
```

Will throw the following error:
```
error: Structure id 8 decorated as BufferBlock for variable in Uniform storage class must follow standard storage buffer layout rules: member 1 at offset 12 is not aligned to 16
  %Foo = OpTypeStruct %v3float %v3float
```

PR #2512, opened 2021-01-27
Add support for GL_EXT_shared_memory_block

cmarcelo:
This will use generate SPIR-V that uses SPV_KHR_workgroup_memory_explicit_layout extension.

The first patch is a fix for #2400 that I can move to a PR of its own if necessary. It is here because we build on top of that in a later patch.


known_good.json
```diff
@@ -5,14 +5,14 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit" : "ad898cb9498c65cac7c5b6f9ebddb610bd0852ad"
+      "commit" : "28b7ca928692ecaaa8558357f9f9ce66e14ac8ef"
```
cmarcelo, line 8: NOTE: This will be updated once https://github.com/KhronosGroup/SPIRV-Tools/pull/4128 lands.

known_good.json
```diff
@@ -5,14 +5,14 @@
...
     },
     {
       "name" : "spirv-tools/external/spirv-headers",
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Headers",
       "subdir" : "External/spirv-tools/external/spirv-headers",
-      "commit" : "f027d53ded7e230e008d37c8b47ede7cd308e19d"
+      "commit" : "38fe9b08508c876ba830f250a488c51003531e7e"
```
cmarcelo, line 15: NOTE: This will be updated once https://github.com/KhronosGroup/SPIRV-Headers/pull/192 lands.

cmarcelo:
@greg-lunarg could you take a look at this?

greg-lunarg:
I am not seeing 28b7ca as a valid commit for SPIRV-Tools on github. This new value in known_good.json appears to be the reason for the CI failures. You indicate that you are going to update these anyway once SPIRV-Tools and SPIRV-Headers are updated. Once valid github commits are put in here, things should go better.

cmarcelo:
> I am not seeing 28b7ca as a valid commit for SPIRV-Tools on github. This new value in known_good.json appears to be the reason for the CI failures. You indicate that you are going to update these anyway once SPIRV-Tools and SPIRV-Headers are updated. Once valid github commits are put in here, things should go better.

@greg-lunarg Correct, these are the hashes of my PRs, but they don't get propagated in a regular clone. SPIRV-Headers already merged and SPIRV-Tools is in the process, I'll update this PR once the latter lands.

I did push this PR earlier because some of the new changes can be reviewed already, e.g. the first patch that fixes #2400. I can also move it to a separate PR of its own if it helps, just let me know.

greg-lunarg:
I looked at most of this already but would like to look at the #2400 fix.

alan-baker:
I merged the spirv-tools change, but it also includes validation changes for workgroup variable initializers and stricter validation for offset. The test expectations in glslang will need updated as a result.

cmarcelo:
Thanks Alan. I'll try to update the test expectations and include that in this PR.

known_good.json
cmarcelo: Updated.

known_good.json
cmarcelo: Updated.

cmarcelo:
@alan-baker @greg-lunarg updated.

greg-lunarg:
As a whole, these changes look good.

Do we need to squash some of these? I just want to make sure we don't leave points at which the repo is not coherent.

cmarcelo:
> As a whole, these changes look good.
> 
> Do we need to squash some of these? I just want to make sure we don't leave points at which the repo is not coherent.

Updated. I've squashed:

- The known_good and the copy of header update, since kind of make sense.
- The three last patches.

PR #2522, opened 2021-02-03
Add IntLog2 and use it

dneto0:
Replace uses of floating point log2 when we want an integer result from
an integer operand.  This avoids concerns about accuracy of floating
point library functions.

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -286,6 +286,18 @@ template <class T> bool IsMultipleOfPow2(T number, int powerOf2)
     return ! (number & (powerOf2 - 1));
 }
 
+// Returns log2 of an integer power of 2.
+// T should be integral.
+template <class T> int IntLog2(T n)
+{
+    assert(IsPow2(n));
+    int result = 0;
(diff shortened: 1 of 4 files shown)


greg-lunarg:
Retrigger?

dneto0:
> Looks like Appveyor just died. Could you please retrigger? 

I did "git commit --amend" without changing anything, and re-pushed.
That did the trick.

greg-lunarg:
@dneto0 Thanks David. Unfortunately it seems the Debug Appveyor run is still failing.  But it doesn't say why, it just goes away and reports Cancelled. 

BUT looking at the Release Appveyor, there is a failure in glslangtests that looks very related to your change. This failure may be what the Debug run is dying on as well. Can you please take a look?

jeremy-lunarg:
Fixed the windows test failure by replacing the -2147483648 literal with INT_MAX (which is expressed as -2147483647-1) in Int_Negative test.

jeremy-lunarg:
I think this is GTG if you're still good with the changes.

PR #2523, opened 2021-02-03
Remove useless semicolon

dneto0:
The extra semicolon causes a build failure if warnings are turned up high, and
warnings-as-errors is on. (-Werror=extra-semi)


dneto0:
This is an example error I get without this change:
```
/glslang/MachineIndependent/localintermediate.h:867:6: error: extra ‘;’ after in-class function definition [-Werror=extra-semi]
  867 |     };
      |      ^
      |      -
```

PR #2532, opened 2021-02-14
Pass environment through PreprocessDeferred

### LINKED ISSUE #2374: preprocess does not forward environment to PreprocessDeferred
When calling preprocess, the environment does not get passed on to PreprocessDeferred which in turn forces ProcessDeferred to use default environment options. This results in shaders using the **GL_EXT_ray_tracing** extension not being able to be preprocessed because the default target SPIR-V version is set to EShTargetSpv_1_0 in TranslateEnvironment.
```
int32 clientInputSemanticsVersion						= 100;
glslang::EShTargetClientVersion vulkanClientVersion				= glslang::EShTargetVulkan_1_2;
glslang::EShTargetLanguageVersion targetVersion					= glslang::EShTargetSpv_1_5;

shader.setEnvInput(glslang::EShSourceGlsl, shaderType, glslang::EShClientVulkan, clientInputSemanticsVersion);
shader.setEnvClient(glslang::EShClientVulkan, vulkanClientVersion);
shader.setEnvTarget(glslang::EShTargetSpv, targetVersion);

const TBuiltInResource* pResources	= GetDefaultBuiltInResources();
EShMessages messages				= static_cast<EShMessages>(EShMsgSpvRules | EShMsgVulkanRules | EShMsgDefault);
const int defaultVersion			= 110;

std::string preprocessedGLSL;
DirStackFileIncluder includer;

includer.pushExternalLocalDirectory(directoryPath);
if (!shader.preprocess(pResources, defaultVersion, ENoProfile, false, false, messages, &preprocessedGLSL, includer))
{
	LOG_ERROR("[ResourceLoader]: GLSL Preprocessing failed for: \"%s\"\n%s\n%s", filepath.c_str(), shader.getInfoLog(), shader.getInfoDebugLog());
	return false;
}
```

juliusikkala:
This PR fixes #2374 by passing environment through `TShader::preprocess` -> `PreprocessDeferred` -> `ProcessDeferred`. Previously, a default environment was used which causes preprocessing to fail when the shader uses extensions which aren't available in that environment.

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -1270,14 +1270,15 @@ bool PreprocessDeferred(
     EShMessages messages,       // warnings/errors/AST; things to print out
     TShader::Includer& includer,
     TIntermediate& intermediate, // returned tree, etc.
-    std::string* outputString)
+    std::string* outputString,
+    TEnvironment* environment = nullptr)
 {
     DoPreprocessing parser(outputString);
     return ProcessDeferred(compiler, shaderStrings, numStrings, inputLengths, stringNames,
                            preamble, optLevel, resources, defaultVersion,
                            defaultProfile, forceDefaultVersionAndProfile,
                            forwardCompatible, messages, intermediate, parser,
-                           false, includer);
+                           false, includer, "", environment);
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


johnkslang:
Looks good to me.

PR #2544, opened 2021-02-23
Replace log2 and exp2 math func.

ZhiqianXia:
Replace the old version math func with C++11 to prevent some presicion issue.

reference link:
https://www.cplusplus.com/reference/cmath/

Signed-off-by: ZhiqianXia <xzq0528@outlook.com>

### DIFF
--- a/glslang/MachineIndependent/Constant.cpp
+++ b/glslang/MachineIndependent/Constant.cpp
@@ -599,17 +599,11 @@ TIntermTyped* TIntermConstantUnion::fold(TOperator op, const TType& returnType) 
             newConstArray[i].setDConst(log(unionArray[i].getDConst()));
             break;
         case EOpExp2:
-            {
-                const double inv_log2_e = 0.69314718055994530941723212145818;
-                newConstArray[i].setDConst(exp(unionArray[i].getDConst() * inv_log2_e));
-                break;
-            }
+            newConstArray[i].setDConst(exp2(unionArray[i].getDConst());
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/Constant.cpp
```diff
@@ -599,17 +599,11 @@ TIntermTyped* TIntermConstantUnion::fold(TOperator op, const TType& returnType)
             newConstArray[i].setDConst(log(unionArray[i].getDConst()));
             break;
         case EOpExp2:
-            {
-                const double inv_log2_e = 0.69314718055994530941723212145818;
-                newConstArray[i].setDConst(exp(unionArray[i].getDConst() * inv_log2_e));
-                break;
-            }
+            newConstArray[i].setDConst(exp2(unionArray[i].getDConst());
```
greg-lunarg, line 602: You have a syntax error here. You need one more right paren at the end.

Please make sure your branch builds and it passes all tests. Particularly, build INSTALL and run glslang/Test/runtests and glslang/build/install/bin/glslangtests.exe.

greg-lunarg (changes requested):
Please fix syntax error and make sure it builds and passes all tests.

glslang/MachineIndependent/Constant.cpp
ZhiqianXia: @greg-lunarg Thank you for your friendly reminder.

PR #2242, opened 2020-05-21
GL_ext_vulkan_glsl_relaxed extension support, and cross stage aware IO mapper

mbechard:
This merge implements https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt.
Additionally, it implements fixes to IO remapping that solve issues from #2151. Specifically, auto-binding is aware of binding usage across stages so different stages don't use the same bindings for different resources.

PR #2579, opened 2021-03-17
Fix issue with remapping global uniform blocks

mbechard:
Avoid adding global uniform blocks to stages that don't already have it.
Otherwise multiple stages point to the same block object, and a
remapping that occurs later on will change the mapping on multiple
stages.

This fixes an issue reported by @lance-loong where setResourceSetBinding() wasn't being respected due to one staging having default uniforms and another having none.

### DIFF
--- a/Test/baseResults/vk.relaxed.changeSet.vert.out
+++ b/Test/baseResults/vk.relaxed.changeSet.vert.out
@@ -1,0 +1,281 @@
+vk.relaxed.changeSet.vert
+Shader version: 460
+0:? Sequence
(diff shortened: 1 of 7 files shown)


greg-lunarg:
Is the issue from @lance-loong that you mention an existing issue against glslang or another repo? If so, could you give the URL here?

mbechard:
It was a comment in the original pull request, here's a link:
https://github.com/KhronosGroup/glslang/pull/2242#issuecomment-774646600

greg-lunarg (approved):
LGTM. Thanks!

Issue #2578, opened 2021-03-17
Issues with gl_PerVertex matching between stages


mbechard:
This issue is related to issue #2243 and issue #1726. It seems that the contents of gl_PerVertex is very inconsistency filled in glslang, and the extensions that change this block don't seem to be consistently following the specs. From everything I've read, the interface matching rule from the OpenGL spec, 7.4.1 requires the gl_PerVertex structure to match between stages except for gl_PointSize. Various extensions break this rule (GL_NVX_multiview_per_view_attributes, GL_ARB_shader_viewport_layer_array, GL_NV_stereo_view_rendering) without explicitly stating new exceptions.
But, even if they had exceptions, in glslang many of these variables aren't included in gl_PerVertex consistently. gl_ViewportIndex and gl_Layer aren't included in gl_PerVertex at all, even though the extension states they should be. In other cases such as gl_SecondaryPositionNV, it's not included in the Vertex shader's output gl_PerVertex block, but is included in the Geom/Tess/Control shader's input gl_PerVertex block. It is then left outside the output gl_PerVertex blocks for the Geom/Tess/Control shader.

Now, with the more robust block checking that is happening with the recent vulkan_relaxed merge, the gl_PerVertex interface is going to fail to match a lot. Attached is a minimal example, if you run it with 
`glslangValidator -V --aml --amb cross.vert cross.geom`
You'll see it's failing due to the gl_PerVertex blocks not matching.

Should we just disable interface matching validations for gl_PerVertex?

[cross.zip](https://github.com/KhronosGroup/glslang/files/6152913/cross.zip)

greg-lunarg:
It is certainly unacceptable to report correct user code as erroneous. That must be remedied immediately.

I would prefer a slightly more surgical remedy ie. one that only suppresses errors around inconsistent variables and extensions. In other words, I am ok if the missing or lost error messages involve inconsistent variables and extensions.

One thing I would like to avoid is losing valid error messages around consistent variables and extensions that we were getting before the vulkan_relaxed merge. If no quick remedy can avoid this, such regressions should be noted in an issue and addressed with urgency.

Any other missing error messages, including those involving inconsistent variables and extensions, should be noted in a second issue and can be addressed with less urgency.

Would this be possible?

mbechard:
Ya for sure. There was a time during vulkan_relaxed where it started to be fine, due to commit 8111268, which blindly avoids some gl_PerVertex checking. I'll investigate why the errors came back more recently.
[...]

PR #2581, opened 2021-03-18
Partial fix for inconsistencies re: #2578

mbechard:
gl_SecondaryPositionNV and gl_PositionPerViewNV are inconsistently declared inside and outside of gl_PerVertex. This breaks interface block matching. For now ignore these errors since it should be fixed with how they are declared.
#2578


mbechard:
Sorry realized this included another commit, will fix

greg-lunarg (changes requested):
The code changes look good.

Would you please add the cross.vert / cross.geom shader example from #2578 as a regression test?

mbechard:
Sure, pushed

PR #2587, opened 2021-03-22
Don't check precision qualifiers on cross-stage checks

### LINKED ISSUE #2586: Regression: False error message for precision mismatch for in/out variables
PR #2576 causes a regression of false error messages during GLSL linking reporting mismatched precision on corresponding in/out variables.

Given foo.vert:

```
#version 460
out mediump vec4 p_color;
in vec4 color;

void main()
{
        p_color = color;
}
```

and foo.frag:

```
#version 460
in vec4 p_color;
out vec4 color;

void main()
{
        color = p_color;
}
```

we are now seeing the new error:

```
$ /cygdrive/c/Users/Greg/dev/glslang/build/StandAlone/Debug/glslangValidator.exe -V --aml --amb --target-env vulkan1.2 cross.vert cross.frag
cross.vert
cross.frag
ERROR: Linking vertex stage: Precision qualifiers must match:
    p_color: " smooth out mediump 4-component vector of float" versus " smooth in highp 4-component vector of float"

SPIR-V is not generated for failed compile or link
```

GLSL 4.6 spec says:

For the purposes of determining if an output from one shader stage matches an input of the next
stage, the precision qualifier need not match.

ESSL says something similar.

Consequently, it appears a regression has been introduced. This has been reported as a problem for internal projects at an ISV.

@mbechard Can you PTAL? I am not able to assign this to you so I will assign to myself for the moment.

mbechard:
Fixes #2586

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -871,12 +871,13 @@ void TIntermediate::mergeErrorCheck(TInfoSink& infoSink, const TIntermSymbol& sy
         }
     }
 
-    // Qualifiers have to (almost) match
+    bool isInOut = crossStage &&
+                   ((symbol.getQualifier().storage == EvqVaryingIn && unitSymbol.getQualifier().storage == EvqVaryingOut) ||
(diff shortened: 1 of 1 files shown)


greg-lunarg (approved):
LGTM. Thanks!

mbechard:
pushed fixes for the error one of the clang builds had

PR #2588, opened 2021-03-22
Default to KHR extension

### LINKED ISSUE #2530: glslang emits deprecated SPV_EXT_physical_storage_buffer
I'm not sure if this is a glslang issue, but I think so.

This is related to an old issue: https://github.com/KhronosGroup/glslang/issues/1924

glslang emits `OpExtension "SPV_EXT_physical_storage_buffer"` that is required by deprecated vulkan extension `VK_EXT_buffer_device_address`. I guess we should now use the KHR version instead of the EXT, because the KHR version is now referenced in the Vulkan 1.1 spec and EXT is deprecated.

The problem came up while running the Vulkan CTS with Validation layers enabled. Vulkan CTS doesn't enable the EXT-version of an extension if there's KHR-version available. As a result, Validation layers complain about using `SPV_EXT_physical_storage_buffer` extension while `VK_EXT_buffer_device_address` is not enabled (I'm not sure if this is valid error, because the `VK_KHR_buffer_device_address` extension is enabled. So this might be Validator Layer issue too). 
The error emitted by Validation layers:
```
vkCreateShaderModule(): The SPIR-V Extension (SPV_EXT_physical_storage_buffer) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-04147)
```
Can be reproduced by running for example following Vulkan CTS tests: `dEQP-VK.binding_model.buffer_device_address.*
`

jeremy-lunarg:
Fix #2530.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -1493,7 +1493,7 @@ TGlslangToSpvTraverser::TGlslangToSpvTraverser(unsigned int spvVersion,
 
     if (glslangIntermediate->usingPhysicalStorageBuffer()) {
         addressingModel = spv::AddressingModelPhysicalStorageBuffer64EXT;
-        builder.addIncorporatedExtension(spv::E_SPV_EXT_physical_storage_buffer, spv::Spv_1_5);
+        builder.addIncorporatedExtension(spv::E_SPV_KHR_physical_storage_buffer, spv::Spv_1_5);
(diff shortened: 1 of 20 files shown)


greg-lunarg (approved):
LGTM.

PR #2589, opened 2021-03-23
Fix warnings about int comparisons of different signness

ben-clayton:
Warnings in public headers can trigger warnings-as-errors.


greg-lunarg (approved):
LGTM. Thanks!

Issue #2580, opened 2021-03-18
sparseTextureGatherOffsetsARB supported by GL_ARB_sparse_texture2 should only take constant integer offsets


pmistryNV:
Presently Test cases spv.float16Fetch.frag and spv.sparseTexture.frag use non constant integral offsets which result in generation of ImageSparseDrefGather with ConstOffsets as "Image opreand" bit mask but non constant integer offsets.

As per GL_ARB_sparse_texture2 it clearly states

**Do a texture gather operation as in textureGatherOffset(), but return  texture access residency information from the function and the filtered  lookup result in the out parameter <texel>**

This means these offsets cannot be non constant. Hence this needs to be fixed in TParseContext::builtInOpCheck

greg-lunarg:
Thank you for the bug report. Will you be submitting a fix?

pmistryNV:
Yes I will submit a change for the same.

greg-lunarg:
Fixed by #2590

PR #2590, opened 2021-03-24
For bug #2580: sparseTextureGatherOffsetsARB should only take constant offsets

pmistryNV:


pmistryNV:
@greg-lunarg please review

greg-lunarg (changes requested):
Generally looks good. Could you add an erroneous sparseTextureGatherOffsetsARB call with non-constant offset to Test/vulkan.frag to make sure the error message works?

pmistryNV:
> Generally looks good. Could you add an erroneous sparseTextureGatherOffsetsARB call with non-constant offset to Test/vulkan.frag to make sure the error message works?

@greg-lunarg this is a GLSL extension, so this should not be added to a vulkan test case. I can add a negative and positive test to 450.frag. Let me know if that would be fine.

greg-lunarg:
> I can add a negative and positive test to 450.frag. Let me know if that would be fine.

Yes, thanks!

pmistryNV:
> > I can add a negative and positive test to 450.frag. Let me know if that would be fine.
> 
> Yes, thanks!

Updated changes include changes now in 450.frag, please review

PR #2593, opened 2021-03-29
Update minimum SPIR-V requirement for GL_EXT_buffer_reference

jeremy-lunarg:
Require SPIR-V 1.3 when using GL_EXT_buffer_reference.
Migrate tests to SPIR-V 1.3 fixture as necessary.
Fix extension table key.
Fix whitespace.


greg-lunarg (approved):
LGTM

Issue #2346, opened 2020-07-14
Shared object ODR violations between glslang and SPIRV


ben-clayton:
## Problem

In #2271 we untangled a cyclic dependency between the `glslang` and `SPIRV` targets. This was solved by creating a new static `MachineIndependent` target which was linked by both `glslang` and `SPIRV` targets.

The `CI-ubuntu-clang-cmake-shared` test failure in https://github.com/KhronosGroup/glslang/pull/2336 has highlighted that we have duplicate public symbols found in both shared objects, which creates an ODR violation (undefined behaviour) - with real-world behaviour of double-delete crashes of global variables.

This problem is partly caused by https://github.com/KhronosGroup/glslang/issues/2308, because `SPIRV` has no public interface, and is exposing the same `MachineIndependent` symbols as `glslang` by default.

The logical solution here is to make the shared `MachineIndependent` symbols hidden. In doing this, I've noticed one further complication - the internal use of thread-local-storage allocators:

The `OGLCompilersDLL`'s [`InitThread()`](https://github.com/KhronosGroup/glslang/blob/b481744aea1ecf52ee4591afaa0f5e270b9d1636/OGLCompilersDLL/InitializeDll.cpp#L95) function [initializes the TLS allocator to `nullptr`](https://github.com/KhronosGroup/glslang/blob/b481744aea1ecf52ee4591afaa0f5e270b9d1636/OGLCompilersDLL/InitializeDll.cpp#L113), and is assigned an allocator in various places in `MachineIndependent/ShaderLang.cpp` ([1](https://github.com/KhronosGroup/glslang/blob/b481744aea1ecf52ee4591afaa0f5e270b9d1636/glslang/MachineIndependent/ShaderLang.cpp#L444), [2](https://github.com/KhronosGroup/glslang/blob/b481744aea1ecf52ee4591afaa0f5e270b9d1636/glslang/MachineIndependent/ShaderLang.cpp#L458), [3](https://github.com/KhronosGroup/glslang/blob/b481744aea1ecf52ee4591afaa0f5e270b9d1636/glslang/MachineIndependent/ShaderLang.cpp#L1468), etc).

`SPIRV` uses bits of `MachineIndependent` which require a `TPoolAllocator` to be assigned. Interestingly there are no calls to `SetThreadPoolAllocator()` in `SPIRV`. Nor are there any in `StandAlone`. This suggests to me that `SPIRV` has been depending on an allocator still being bound from earlier calls into `glslang`.

[...]

PR #2601, opened 2021-04-08
Removed condition for glslang-default-resource-limits instalation.

masterleming:
This way the limits are available to clients that link statically, which might be necessary due to the ORD problem https://github.com/KhronosGroup/glslang/issues/2346.


greg-lunarg:
Looks like the Linux builds are failing.

masterleming:
Right, my naïve change did not take into consideration the proper way of defining libraries. It worked only in my narrow and isolated setup. Should be much better now.

greg-lunarg (approved):
LGTM.

PR #2603, opened 2021-04-09
Fix table key and whitespace

jeremy-lunarg:
These got lost in the wash with my last revert but are still relevant.


greg-lunarg (approved):
LGTM

PR #2608, opened 2021-04-13
TextureOffset not support sampler2DArrayShadow sampler until 430.

ZhiqianXia:
Hi ,
glslang should report an error about textureOffset , because textureOffset not support sampler2DArrayShadow sampler until 430.

**shader** 
```
#version 300 es
precision mediump float;
in highp vec4 dEQP_Position;
uniform mediump sampler2DArrayShadow s;
void main()
{
                gl_Position = vec4(textureOffset(s, vec4(0), ivec2(0)));
                gl_Position = dEQP_Position;
}
```
**Reference**:
- https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.20.pdf
- https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.30.pdf

Signed-off-by: ZhiqianXia <xzq0528@outlook.com>

### DIFF
--- a/Test/baseResults/textureoffset_sampler2darrayshadow.vert.out
+++ b/Test/baseResults/textureoffset_sampler2darrayshadow.vert.out
@@ -1,0 +1,4 @@
+textureoffset_sampler2darrayshadow.vert
+ERROR: 0:9: 'texel offset' : TextureOffset not support the sampler2DArrayShadow :  version <= 420
+ERROR: 0:9: '' : compilation terminated
+ERROR: 2 compilation errors.  No code generated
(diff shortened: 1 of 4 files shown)


Test/baseResults/textureoffset_sampler2darrayshadow.vert.out
```diff
@@ -0,0 +1,4 @@
+textureoffset_sampler2darrayshadow.vert
+ERROR: 0:9: 'texel offset' : TextureOffset not support the sampler2DArrayShadow :  version <= 420
```
greg-lunarg, line 2: Could you please change the error message to:

ERROR: 0:9: 'sampler' : TextureOffset does not support sampler2DArrayShadow : version <= 420

and squash back to a single commit? Thanks! Otherwise good!

greg-lunarg (changes requested):
Just a request for tightening of error message. Otherwise good!

Thank you for the issue and the fix!

ZhiqianXia:
Replaced with PR#2609

PR #2609, opened 2021-04-14
TextureOffset not support sampler2DArrayShadow sampler until 430.

ZhiqianXia:
Hi ,
glslang should report an error about textureOffset , because textureOffset not support sampler2DArrayShadow sampler until 430.

shader

#version 300 es
precision mediump float;
in highp vec4 dEQP_Position;
uniform mediump sampler2DArrayShadow s;
void main()
{
                gl_Position = vec4(textureOffset(s, vec4(0), ivec2(0)));
                gl_Position = dEQP_Position;
}
Reference:

https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.20.pdf
https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.30.pdf

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -2191,6 +2191,16 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
                               "[gl_MinProgramTexelOffset, gl_MaxProgramTexelOffset]");
                 }
             }
+
+            if (callNode.getOp() == EOpTextureOffset) {
+                TSampler s = arg0->getType().getSampler();
+                if (s.is2D() && s.isArrayed() && s.isShadow()) {
+                    if (isEsProfile())
+                        error(loc, "TextureOffset does not support sampler2DArrayShadow : ", "sampler", "ES Profile");
+                    else if (version <= 420)
+                        error(loc, "TextureOffset does not support sampler2DArrayShadow : ", "sampler", "version <= 420");
+                }
+            }
         }
 
         break;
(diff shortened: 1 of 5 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2191,6 +2191,13 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
                               "[gl_MinProgramTexelOffset, gl_MaxProgramTexelOffset]");
                 }
             }
+
+            if (callNode.getOp() == EOpTextureOffset && version <= 420) {
```
greg-lunarg, line 2195: Sorry, but I just noticed this.

The logic would be cleaner and the error messaging more appropriate if the version test was removed here and done below.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2191,6 +2191,13 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
+                TSampler s = arg0->getType().getSampler();
+                if (s.is2D() && s.isArrayed() && s.isShadow())
+                    error(loc, "TextureOffset not support the sampler2DArrayShadow : ", "sampler",
```
greg-lunarg, line 2198: Replace the error call here with the following logic:

if (isEsProfile())
    error(loc, "TextureOffset does not support sampler2DArrayShadow : ", "sampler", "ES Profile");
else if (version <= 420)
    error(loc, "TextureOffset does not support sampler2DArrayShadow : ", "sampler", "version <= 420");

greg-lunarg (changes requested):
Sorry, some additional improvements requested.

You can make the changes on your current local branch, squash to one commit with rebase -i HEAD~2, and then force push the branch with git push -f <your fork> <your branch>.

glslang/MachineIndependent/ParseHelper.cpp
ZhiqianXia: Thank your commits.

greg-lunarg:
I just pushed a commit onto your Feature2 branch. Your new test needed to be added to the master test list. And the compare file needed to be updated accordingly.

With this change, if it passes CI, all my concerns will be addressed and I can merge.

greg-lunarg (approved):
With my latest commit, this PR should be good to merge.

ZhiqianXia:
@greg-lunarg Thank you for your commit!

PR #2616, opened 2021-04-20
Exposed #856 as --hlsl-sampled-textures in the StandAlone

### LINKED ISSUE #2611: Feature Request: Expose option to combine textures and samplers to the StandAlone Application
In #591 the functionality to automatically combine HLSL Texture*D and SamplerState objects was implemented. Since then, this functionality was not exposed to the StandAlone Application via any argument. Exposing it would simplify creating HLSL -> Vulkan shaders a lot for me, since I wouldn't need to write an own Application that uses the glslang lib anymore.

Rob2309:
This PR exposes the automatic Texture*D and SampleState merging introduced in #856 to the StandAlone application via --hlsl-sampled-textures. This closes #2611.

Since the new C++ API is only used in link mode, I am not sure whether exposing this functionality is possible via ShCompile. I only implemented it for link mode.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -111,7 +111,7 @@ bool NaNClamp = false;
 bool stripDebugInfo = false;
 bool beQuiet = false;
 bool VulkanRulesRelaxed = false;
-bool hlslSampledTextures = false;
+bool autoSampledTextures = false;
 
 //
 // Return codes from main/exit().
@@ -656,8 +656,8 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                         HlslEnable16BitTypes = true;
                     } else if (lowerword == "hlsl-dx9-compatible") {
                         HlslDX9compatible = true;
-                    } else if (lowerword == "hlsl-sampled-textures") { 
-                        hlslSampledTextures = true;
+                    } else if (lowerword == "auto-sampled-textures") { 
+                        autoSampledTextures = true;
                     } else if (lowerword == "invert-y" ||  // synonyms
                                lowerword == "iy") {
                         Options |= EOptionInvertY;
@@ -1192,7 +1192,7 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
         shader->setNoStorageFormat((Options & EOptionNoStorageFormat) != 0);
         shader->setResourceSetBinding(baseResourceSetBinding[compUnit.stage]);
 
-        if (hlslSampledTextures)
+        if (autoSampledTextures)
             shader->setTextureSamplerTransformMode(EShTexSampTransUpgradeTextureRemoveSampler);
 
         if (Options & EOptionAutoMapBindings)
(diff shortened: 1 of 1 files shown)


greg-lunarg:
Thanks for your contribution. Please sign the CLA so I can ultimately merge :)

Rob2309:
sorry, forgot about that :)

StandAlone/StandAlone.cpp
```diff
@@ -655,6 +656,8 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                         HlslEnable16BitTypes = true;
                     } else if (lowerword == "hlsl-dx9-compatible") {
                         HlslDX9compatible = true;
+                    } else if (lowerword == "hlsl-sampled-textures") {
```
dneto0, line 659: This doesn't seem like an HLSL-specific feature.
Suggest "auto-sampled-textures" instead?

dneto0 (commented):
Non, blocking but it seems the option name should not be HLSL specific?

StandAlone/StandAlone.cpp
dneto0: Hm. That seems a little ambiguous.  The original texture is already of sampled type. How about
"auto-combined-image-sampler".   That's seems to be in line with common phrasing you can easily find in reference material and support questions.

dneto0:
For my own benefit, Rob is also adding this to shaderc: https://github.com/google/shaderc/pull/1181

StandAlone/StandAlone.cpp
greg-lunarg: Yes. While the original issue was targeting HLSL, this can be used for GLSL shaders as well, so I agree taking hlsl- out of the name makes sense.

However, I think the original texture/image is NOT sampled and this option turns it into a sampled image, right? See Test/spv.texture.sampler.transform.frag and its corresponding Test/baseResults file. So I think that auto-sampled-texture might be most apt. Agree?

Also please add this option to usage() in StandAlone.cpp.

greg-lunarg:
I am going to push the option name change I suggested above and merge this so that it makes it into the next SDK. I am going to hold off on adding the option to usage until we agree on the name. You can add it in a separate PR.

I would also like you to add tests, one for GLSL and one for HLSL. There is already a GLSL file you can use named above. There might be an HLSL file as well. If not, I will attach some HLSL below that I used to check this option. You can look at the recent addition of --depfile for how to add the tests.

greg-lunarg:
Here is an HLSL shader if you want it:

```
Texture2D g_tColor;

SamplerState g_sAniso;

struct PS_INPUT
{
    float2 vTextureCoords : TEXCOORD2 ;
} ;

struct PS_OUTPUT
{
    float4 vColor : SV_Target0 ;
} ;

PS_OUTPUT MainPs ( PS_INPUT i )
{
    PS_OUTPUT ps_output ;

    ps_output . vColor = g_tColor . Sample ( g_sAniso , i . vTextureCoords . xy ) ;
    return ps_output ;
}
```

StandAlone/StandAlone.cpp
Rob2309: I would agree with --auto-sampled-textures being the most intuitive name.

PR #2620, opened 2021-04-22
Update CHANGES for 11.3.0

greg-lunarg:


DadSchoorse:
Is this supposed to be the proper 11.3.0 release? Because if yes it's missing a git tag.

greg-lunarg:
I added the tag an hour or so ago.

FWIW, you probably want to wait for 11.4.0. There is a CMake compatibility
problem in this release we just came across that requires CMake 3.12.0 and
the current max version is CMake 3.10.2.

11.4.0 should be available in the next couple of hours.

Greg Fischer

Senior Engineer

LunarG, Inc. - 3D Driver Innovations

***@***.*** ***@***.***>

PR #2634, opened 2021-05-08
Dont export inline TObjectReflection::getType()

### LINKED ISSUE #2460: Compiler error while building shared library in MSYS2/MINGW
* Build platform: Windows 10 MSYS2/MINGW
* Compiler: GCC 10.2.0 x86_64-w64-mingw32
* CMake version 3.18.4
* Ninja version 1.10.1
* Command used:

```sh
git clone --depth=1 https://github.com/KhronosGroup/glslang.git
cd glslang; mkdir bin; cd bin
MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
${MINGW_PREFIX}/bin/cmake.exe \
    -G "Ninja" \
    -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
    -DCMAKE_INSTALL_LIBDIR=lib \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON \
    ../
${MINGW_PREFIX}/bin/cmake.exe --build ./
```

With `-DBUILD_SHARED_LIBS=OFF` option, it is compiled fine.

* Compiler Output:

```c
[12/68] Building CXX object StandAlone/CMakeFiles/glslangValidator.dir/StandAlone.cpp.obj
FAILED: StandAlone/CMakeFiles/glslangValidator.dir/StandAlone.cpp.obj
F:\msys64\mingw64\bin\c++.exe -DENABLE_HLSL -DENABLE_OPT=0 -DGLSLANG_IS_SHARED_LIBRARY=1 -DGLSLANG_OSINCLUDE_WIN32 -I../StandAlone/../External -I../glslang/.. -Iinclude -I../SPIRV/.. -I../StandAlone -I../ -O3 -DNDEBUG -Wall -Wmaybe-uninitialized -Wuninitialized -Wunused -Wunused-local-typedefs -Wunused-parameter -Wunused-value -Wunused-variable -Wunused-but-set-parameter -Wunused-but-set-variable -fno-exceptions -Wno-reorder -fno-rtti -Werror=deprecated-copy -std=c++11 -MD -MT StandAlone/CMakeFiles/glslangValidator.dir/StandAlone.cpp.obj -MF StandAlone\CMakeFiles\glslangValidator.dir\StandAlone.cpp.obj.d -o StandAlone/CMakeFiles/glslangValidator.dir/StandAlone.cpp.obj -c ../StandAlone/StandAlone.cpp
In file included from ../StandAlone/DirStackFileIncluder.h:44,
                 from ../StandAlone/StandAlone.cpp:45:
F:/glslang/glslang/Public/ShaderLang.h:704:33: error: function 'const glslang::TType* glslang::TObjectReflection::getType() const' definition is marked dllimport
  704 |     GLSLANG_EXPORT const TType* getType() const { return type; }
      |                                 ^~~~~~~
[14/68] Building CXX object glslang/CMakeFiles/MachineIndependent.dir/MachineIndependent/glslang_tab.cpp.obj
ninja: build stopped: subcommand failed.
```

* Temporary fix:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ada2b8f..c6460ea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,7 +293,7 @@ function(glslang_only_export_explicit_symbols target)
     if(BUILD_SHARED_LIBS)
         target_compile_definitions(${target} PUBLIC "GLSLANG_IS_SHARED_LIBRARY=1")
         if(WIN32)
-            target_compile_definitions(${target} PRIVATE "GLSLANG_EXPORTING=1")
+            target_compile_definitions(${target} PUBLIC "GLSLANG_EXPORTING=1")
         else()
             target_compile_options(${target} PRIVATE "-fvisibility=hidden")
         endif()
```

I would be happy to provide any further information, if required.

xantares:
Fixes mingw build 
Closes #2460

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -722,7 +722,7 @@ class TObjectReflection {
 public:
     GLSLANG_EXPORT TObjectReflection(const std::string& pName, const TType& pType, int pOffset, int pGLDefineType, int pSize, int pIndex);
 
-    GLSLANG_EXPORT const TType* getType() const { return type; }
+    const TType* getType() const { return type; }
     GLSLANG_EXPORT int getBinding() const;
     GLSLANG_EXPORT void dump() const;
     static TObjectReflection badReflection() { return TObjectReflection(); }


greg-lunarg:
Or is the solution to split the definition out from the declaration? I.E. Should we declare and define getType() the same way we do getBinding()?

I am worried removing GLSLANG_EXPORT makes getType() inaccessible. It does, right? Are we sure it is not needed as part of the API?

xantares:
Both solutions are correct. When removing GLSLANG_EXPORT getType will be still inlined to the client code and still accessible (and it needs to be).

xantares:
It was never visible externally in the first place because it is inlined.
Either we keep it that way and just remove the GLSLANG_EXPORT for the build to pass on gcc, or we uninline it and the GLSLANG_EXPORT can stay.
Both solutions are correct and allow to use that symbol.

Issue #2645, opened 2021-05-18
mat4x2(constant scalar) constructor generates wrong matrix


johnstiles-google:
When glslang synthesizes a diagonal mat4x2 at compile time, it accidentally populates one additional field with the scalar value.

EXPECTED:
    mat4x2(7) == mat4x2(7, 0, 0, 7, 0, 0, 0, 0)  should return true
ACTUAL:
    mat4x2(7) == mat4x2(7, 0, 0, 7, 0, 0, 7, 0)  returns true

This appears to be an issue only with 4x2 matrices.
The root problem is the following logic in parseConst.cpp:

                        // If there is a single scalar parameter to a matrix
                        // constructor, it is used to initialize all the
                        // components on the matrix's diagonal, with the
                        // remaining components initialized to 0.0.
                        if (i == startIndex || (i - startIndex) % (matrixRows + 1) == 0 )

This logic is intended for square matrices, and happens to work for non-square matrices other than 4x2 sized ones.

greg-lunarg:
Fixed by #2646

PR #2646, opened 2021-05-18
Fix mat4x2(scalar) constructor.

johnstiles-google:
As described in #2645, 4x2 matrices constructed with a scalar are handled improperly and contain an additional unwanted scalar value. This PR adds a test for this issue and fixes the constant-matrix handling code.


greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2656, opened 2021-06-03
Fix OOB write in matrix constructor

ShabbyX:
In a matrix constructor that takes a number of components, as many
components as necessary must be taken, with the rest discarded, as GLSL
allows more components than necessary to be specified.  For example, the
following:

    mat4 m4 = mat4(v4, v4.yzwx, v4.zwx, v4.zwxy, v4.wxyz);

is equivalent to:

    mat4 m4 = mat4(v4, v4.yzwx, v4.zwx, v4.zwxy, v4.w);

glslang takes the components from the constructor and builds the single
components of the matrix in a 2D array before constructing the matrix
itself.  It however did not check for extra parameters and was thus
writing OOB to said 2D array.  This is fixed in this change

Signed-off-by: Shahbaz Youssefi <shabbyx@gmail.com>

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -2543,7 +2543,7 @@ Id Builder::createMatrixConstructor(Decoration precision, const std::vector<Id>&
         int row = 0;
         int col = 0;
 
-        for (int arg = 0; arg < (int)sources.size(); ++arg) {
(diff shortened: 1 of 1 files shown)


ShabbyX:
@greg-lunarg another clusterfuzz bug found

greg-lunarg (approved):
LGTM. No regression test required.

PR #2660, opened 2021-06-08
Temporarily disable unit tests to fix shaderc test failure

jaebaek:
We have an update for spirv-val
https://github.com/KhronosGroup/SPIRV-Tools/pull/4316 that causes a
few glslang unit test failures. We should change the unit tests but
changing the unit tests without the spirv-val update causes glslang
unit test failures. To submit the spirv-val change first, this commit
temporarily disables unit tests that fails with the spirv-val change.


jaebaek:
I already submitted the SPIRV-Tools PR with the shaderc failure. This PR is not needed. Let me close this PR.

PR #2661, opened 2021-06-08
Update spirv tools and unit tests

jaebaek:


jaebaek:
This will fix the shaderc unit test failures caused by https://github.com/KhronosGroup/SPIRV-Tools/pull/4316

PR #2673, opened 2021-06-26
allow textureLod for sampler2DShadow

AntonYudintsev:
it is perfectly supported in spirv tools and everywhere, it is just not parsed by glslang

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -6601,8 +6601,6 @@ void TBuiltIns::addSamplingFunctions(TSampler sampler, const TString& typeName, 
 
             if (lod && (sampler.isBuffer() || sampler.isRect() || sampler.isMultiSample() || !sampler.isCombined()))
                 continue;
-            if (lod && sampler.dim == Esd2D && sampler.arrayed && sampler.shadow)
-                continue;
             if (lod && sampler.dim == EsdCube && sampler.shadow)
                 continue;
 


greg-lunarg:
Can you please provide in this conversation a shader whose compilation is fixed by this PR? The simpler, the better. Thanks!

AntonYudintsev:
//something like this
uniform sampler2DArrayShadow shd
out vec4 outColor;

void main()
{
  float shadow = textureLod(shd, vec4(0.5,0.5,0.5,0/*slice*/), 0/*lod*/);//this line won't parse by glslang without commit.
  outColor = vec4(shadow);
}

greg-lunarg:
Please note that while your issue title and commit comment both say that there is a problem with sampler2DShadow, your shader uses sampler2DArrayShadow. Your file change also seems to be addressing sampler2DArrayShadow.

The GLSL spec does not specify a textureLod call for sampler2DArrayShadow. Just based on this, this PR cannot move forward.

My intuition tells me this absence in the spec is intentional. I suspect this not a generally supported function across graphics platforms or in Vulkan.

If you truly believe that this is a legitimate gap with the GLSL spec, please file an issue against the GLSL spec and once addressed, please re-open this PR. In the meantime, I will be closing this PR unmerged.

AntonYudintsev:
Indeed, I meant sampler2DArrayShadow. That was incorrect title, sorry for that.

It is indeed not "universally" supported by all platforms. And it is supported by:
Nintento Switch, iPhone7+, PS4/Xbox1, all Nvidia from at least GeForce 450, etc.

Basically, all DirectX 11 hardware support it. There is GLSL extension EXT_texture_shadow_lod that should enable it.
However, with a current glslang it is not possible at all to make sampler2DArrayShadow work at all.

I am totally fine with that PR not made it's way to glslang, but ofc, we will keep that changes internally in Gaijin, as "practicality beats purity", we prefer working code over "that is not pure GLSL spec".

greg-lunarg:
Thank you very much for the additional information. I was not aware of EXT_texture_shadow_lod. Unfortunately, it appears that glslang does not support this extension at all.

I would be open to accepting the above change if additional logic was added to recognize the presence of the  GL_EXT_texture_shadow_lod extension. I don't think this would be a lot of extra work. There is a fairly substantial framework you could leverage.

Adding this logic and seeing this PR through would not only be a nice contribution to the community, but it would also save you from continually rebasing your fork :)

PR #2674, opened 2021-06-28
fix error message for vertex struct input

dneto0:
Vertex shaders can have pipeline inputs that are arrays, but not structure
inputs. (GLSL 4.5 section 4.3.4)

But the error message for struct inputs says "cannot be a structure or array".
This PR removes the "or array" part.

Note: The array case is handled immediately after the check for
structure type.


dneto0:
I have no urgency for this.

greg-lunarg:
Will need a change to baseResults/300layout.vert.out as well. Could you make that change, squash it and push it?

arcady-lunarg:
I've updated the test result file and rebased the branch.

greg-lunarg (approved):
LGTM

Issue #2642, opened 2021-05-12
Atomic memory function can only be used for shader storage block member or shared variable.


ZhiqianXia:
Hi , 
The spec says :
A shader will **fail** to compile if the value passed to the mem argument of an atomic memory function does not correspond to a buffer or shared variable.  But the following shader didn't get an error from glslang. 

**shader**:
```
#version 310 es
layout(local_size_x = 1) in;
int y;
void main() {
  atomicAdd(y, 2);   // uint atomicAdd (inout uint mem, uint data)
}
```
ps: i'll submit the PR later.

ZhiqianXia:
https://github.com/KhronosGroup/glslang/pull/2643

greg-lunarg:
Fixed by #2643

PR #2643, opened 2021-05-12
Atomic memory function can only be used for shader storage block member or shared variable.

ZhiqianXia:
The glsl spec says:_A shader will fail to compile if the value passed to the mem argument of an atomic memory function does not
correspond to a buffer or shared variable._

**shader**:
```
#version 320 es
layout(local_size_x = 1) in;
buffer coherent Buffer { int x; };
int y;
void main() {
  atomicAdd(x, 1);
  atomicAdd(y, 2);
}
```
**issue:**
https://github.com/KhronosGroup/glslang/issues/2642

**ref:**
https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf (8.11. Atomic Memory Functions)

PR #2691, opened 2021-07-12
Atomic memory function can only be used for shader storage block member or shared variable.

ZhiqianXia:
Fix the Regression by https://github.com/KhronosGroup/glslang/pull/2643.


greg-lunarg:
Fixes #2642

PR #2702, opened 2021-07-20
Generate separate stores for partially swizzled memory stores

### LINKED ISSUE #2518: vector read-modify-write pattern to buffer memory
```cpp
#version 450 core

layout(binding=0)
writeonly buffer Foo {
  vec4 v;
};

void main() {
  v.xy = vec2(1, 2);
}
```

I noticed that storing to a vector swizzle causes a read-modify-write pattern, even when the lhs is off-chip memory, even when it's `writeonly`. I think the load to `writeonly` is UB.

This load even occurs when all components of the lhs are in the swizzle, and only the order of components is permuted:
```cpp
  v.xywz = vec4(1, 2, 3, 4);
```
```
         %20 = OpLoad %v4float %19
         %21 = OpVectorShuffle %v4float %20 %17 4 5 7 6     ; %20 never gets used!
               OpStore %19 %21
```

spirv-opt does manage to eliminate the load in this latter scenario, but it keeps the pernicious load in the former one.

The logic of this maneuver is clear. My own SPIR-V compiler emits the OpLoad, OpVectorShuffle, OpStore sequence. However I started writing a DXIL compiler, and I noticed that dxc does a better job:
* If the full lhs side is named by the swizzle in permuted order, one vectorized `@dx.op.rawBufferStore.f32` is emitted with all vector source components in inverse-permuted order, so that the destination is written in one transaction. There's no load.
* If some components of the lhs are not named in the swizzle, the transaction is broken into scalar stores. No load. I think dxc could go still better and exploit the mask capability in `@dx.op.rawBufferStore.f32` and write them all in one instruction. But what it does now is not bad.

The fact that spirv-opt doesn't fix this makes me wonder if this is a performance bug. I don't know what all the optimizers in drivers are doing, but I could imagine them missing quite a bit.

jeremy-lunarg:
Full vector and fully specified vector swizzle stores are not affected by this change, only partial swizzles ie swizzles with fewer components than the vector being stored to.

Previously the vector being stored to is loaded and any components not specified in the swizzle were used to create a full store to the vector.

While this change generates more SPIR-V instructions, it is necessary for correctness.

Fixes #2518.


greg-lunarg (changes requested):
LGTM.

Could you please amend the commit to use the title and description from the PR? Thanks!

jeremy-lunarg:
> LGTM.
> 
> Could you please amend the commit to use the title and description from the PR? Thanks!

Done. Thanks Greg.

PR #2704, opened 2021-07-26
Fix per-set argument parsing in standalone

YunHsiao:
Currently usages like this don't work at all:
```bash
glslangValidator -V --shift-UBO-binding 16 1 # add offset 16 to set 1 only
```

Seems to be a simple parsing problem.


greg-lunarg:
LGTM. Thanks for the fix.

PR #2705, opened 2021-07-26
Fix some instances of -Wunused-but-set-variable.

pkasting:
Bug: chromium:1203071


greg-lunarg:
LGTM. Thanks for the fix.

greg-lunarg:
Gah! Meant to merge :( Will do so when CI finishes.

Issue #2486, opened 2020-12-15
spir-v:  ImageRead code generation generates result type other than 4-elem vector, for HLSL compilation


dneto0:
SPIRV-Tools has added validation of OpImageRead result type.
https://github.com/KhronosGroup/SPIRV-Tools/pull/4072 

For Vulkan, the result type must be a 4-element vector.
Glslang code generation for HLSL does not always follow this rule.

Failures in the test suite are:

[ FAILED ] 4 tests, listed below:
[ FAILED ] ToSpirv/HlslCompileTest.FromFile/hlsl_rw_register_frag, where GetParam() = 16-byte object <CC-16 01-9A 2D-56 00-00 23-EF 02-9A 2D-56 00-00>
[ FAILED ] ToSpirv/HlslCompileTest.FromFile/hlsl_rw_scalar_bracket_frag, where GetParam() = 16-byte object <E2-16 01-9A 2D-56 00-00 23-EF 02-9A 2D-56 00-00>
[ FAILED ] ToSpirv/HlslCompileTest.FromFile/hlsl_rw_vec2_bracket_frag, where GetParam() = 16-byte object <13-17 01-9A 2D-56 00-00 23-EF 02-9A 2D-56 00-00>
[ FAILED ] Hlsl/HlslIoMap.FromFile/spv_rw_autoassign_frag, where GetParam() = 40-byte object <07-3B 01-9A 2D-56 00-00 23-EF 02-9A 2D-56 00-00 05-00 00-00 0A-00 00-00 14-00 00-00 0F-00 00-00 1E-00 00-00 01-01 00-00>

greg-lunarg:
Fixed by #2706

Issue #1340, opened 2018-04-09
HLSL: Invalid SPIR-V when accessing < 4 component images or buffers


gwihlidal:
I am running into a significant number of issues with HLSL -> SPIR-V using latest glslang (optimizer enabled, as well) - VK SDK 1.70.1.

There are many places in our code that define < 4 component images or buffers, and use subscript operator to access/write the values (translating to OpImageFetch or store).

A very common case if reading from depth (float):
```
Texture2D<float>  g_depth : register(t1, space0);
...
int2 pixelPos;
float depth = g_depth[pixelPos];
```
or from x/y velocity (float2):
```
RWTexture2D<float2> g_output : register(u0, space0);
g_output[pixelPos] = float2(blah1, blah2);
```

Compilation succeeds, but runtime validation fails with:
```
SPIR-V module not valid: Expected Result Type to be int or float vector type: ImageFetch
```

The current workaround in a bunch of trivial cases (but very far from ideal) is to make everything a float4, and then use appropriate swizzle masks:
```
Texture2D<float4>  g_depth : register(t1, space0);
...
int2 pixelPos;
float depth = g_depth[pixelPos].x;
```
```
RWTexture2D<float4> g_output : register(u0, space0);
g_output[pixelPos] = float4(blah1, blah2, 0, 0);
```

More complex cases are unable to compile.
Before:
```
RWTexture3D<uint> g_cellSizeTexture : register(u0, space0);
...
InterlockedAdd(g_cellSizeTexture[coord], 1);
```
After:
```
RWTexture3D<uint4> g_cellSizeTexture : register(u0, space0);
...
InterlockedAdd(g_cellSizeTexture[coord].x, 1);
```

FXC obviously disallows interlocked add on vector types, so using uint4 as a workaround won't work, and the following error will be returned:
```
error X3681: cs_5_1 only supports interlocked operations on scalar int or uint data
```

The SPIR-V spec for `OpImageFetch` seems awfully restrictive:
```
Result Type must be a vector of four components of floating-point type or integer type. Its components must be the same as Sampled Type of the underlying OpTypeImage (unless that underlying Sampled Type is OpTypeVoid).
```

Spec aside, glslang needs to syntactically support the examples I mentioned, even if some translation to valid SPIR-V needs to happen under the hood.
Effectively, glslang needs to emit something like:
```
%x = OpImageFetch %v4float ...
%y = OpCompositeExtract %float %x 0
```

[...]

PR #2706, opened 2021-07-26
Fix OpImageRead result type when compiling HLSL

Reedbeta:
Per the Vulkan spec, OpImageRead must return a 4-component vector always. When compiling HLSL, loads from a RWTexture of a template type with < 4 components would incorrectly generate an OpImageRead with a < 4 component result, resulting in validation errors.

The same issue existed with OpImageFetch (see #1340, https://github.com/KhronosGroup/SPIRV-Tools/issues/1465) and was fixed in commit 4425f24. This PR does the same thing for OpImageRead, and should address #2486.

I also added associated tests, and expanded the existing HLSL image fetch tests to check all the different types of textures, in both float and int incarnations, for completeness. That may be overkill, but it costs little to have extra tests.

Thanks for your attention!


Reedbeta:
Looks like the Appveyor build had some test failures, but I'm not seeing those locally when I run the glslangtests. Is there something else I need to do to run all the tests?

greg-lunarg:
LGTM. Thanks for the fix!

Would you mind squashing these two commits into one commit and force-push your resulting branch? It just makes a little cleaner for people to do a bisect on the repo.

Reedbeta:
Thanks! Sure, I can do that a little later - away from my desk just now.
But I think you can select "squash and merge" in GitHub as well - there
should be a dropdown on the merge button on the PR.

PR #2712, opened 2021-07-29
glslang should report a error for Feature: last case/default label not followed by statements'. 

ZhiqianXia:
the ol430 spec says:
**_Fall through labels are allowed, but it is a compile-time error to have no statement between a label and the end of the switch statement._** 

So if the glsl version <=430 , glslang should report a error for Feature: last case/default label not followed by statements'. 

**shader:**
```
#version 430
precision mediump float;
in highp vec4 dEQP_Position;

void main ()
{
	switch (1)
	{
		case 0: break;
		case 1:
	}
	gl_Position = dEQP_Position;
}
```

**specs:**

[1] https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.30.pdf
[2] https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.40.pdf

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -9188,10 +9188,13 @@ TIntermNode* TParseContext::addSwitch(const TSourceLoc& loc, TIntermTyped* expre
         // "it is an error to have no statement between a label and the end of the switch statement."
         // The specifications were updated to remove this (being ill-defined what a "statement" was),
         // so, this became a warning.  However, 3.0 tests still check for the error.
-        if (isEsProfile() && version <= 300 && ! relaxedErrors())
+        if (isEsProfile() && (version <= 300 || version >= 320) && ! relaxedErrors())
+            error(loc, "last case/default label not followed by statements", "switch", "");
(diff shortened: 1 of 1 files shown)


greg-lunarg:
Looks like you will need to fix test spv.switch.frag.

greg-lunarg:
Here is the current situation as I can see:

If no statement after last label {
    If ESProfile and (version <= 300 or version >= 320)
        error
    else if !ESProfile and (version <= 430 or version >= 460)
        error
    else
        warn
}

As far as I can tell, generation of SPIR-V does not have an effect.

Is this consistent with your reading of the specs?

If so, please change the logic here to fit this understanding of the specs.

Thanks!

ZhiqianXia:
Spec4.6 says: Fall through labels are allowed.
Spec4.3 says: Fall through labels are allowed, but it is a compile-time error to have no statement between a label and the end of the switch statement.

So , I think there is a warn for 4.6 , not an error.

greg-lunarg:
This is from the 4.60.7 GLSL spec:

Fall through labels are allowed, but it is a compile-time error to have no statement between a label and the end of the switch statement.

greg-lunarg:
Please change your code to match the logic as I have outlined above.

PR #2726, opened 2021-08-08
Update CMakeLists.txt

TheMindVirus:
Checked REGEX REPLACE for consistency failure


greg-lunarg:
Can you please describe the problem that the original source was causing?

TheMindVirus:
Due to the inclusion of CMake, the building process was prone to many failures. In practice this turned out to be a missing set of quotes around some of the variables which meant that CMake disregarded the parameters passed to it and caused the whole build to fail.

Issue #2740, opened 2021-08-26
Precision is incorrectly propagated to function call arguments


ShabbyX:
glslang has the following code in `TIntermTyped::propagatePrecision`:

```c
    TIntermAggregate* aggregateNode = getAsAggregate();
    if (aggregateNode) {
        TIntermSequence operands = aggregateNode->getSequence();
        for (unsigned int i = 0; i < operands.size(); ++i) {
            TIntermTyped* typedNode = operands[i]->getAsTyped();
            if (! typedNode)
                break;
            typedNode->propagatePrecision(newPrecision);
        }

        return;
    }
```

So given `f(a, b, c)` with the result precision being `mediump`, it would apply `mediump` to every argument of this function (that doesn't already have a precision). This includes any argument that is a constructor (because per spec constructors don't have a precision). This is not correct. For example, in ANGLE I discovered that given `texture(mediump_sampler, vec3(coords));`, `coords_expression` was being evaluated at `mediump` even though it contained only `highp` variables (inside constructors). The same expression given a `highp` sampler would evaluate the coordinates in `highp`!

greg-lunarg:
It would really help to have a full, compilable shader that exhibits the problems you describe. While the general approach of the glslang code I think is sound, there appear to be holes in the logic that need to be filled and so it would help to have a real shader so I can be sure that I am addressing your problem.

Specifically, there is logic right at the top of propagatePrecision that returns if the node already has a precision assigned. This tells me that it is not propagatePrecision that is overwriting the precision of the coords above, but that the precision is not being assigned correctly for some operations to begin with. It appears that precision is being assigned for unary and binary ops (see the calls to updatePrecision()), but other ops appear to be getting missed (builtins, constructors).

ShabbyX:
Here's a minimal repro:

```c
#version 320 es

layout(set = 0, binding = 0) uniform mediump sampler2DArray s;
layout(location = 0) out mediump float o;

highp float f(in highp ivec2 v){
    return texture(s, vec3(ivec2((v.x & 3), (v.y & 3)), 0)).x;
}

void main()
{
    o = f(ivec2(0));
}
```

glslang produces:

```swift
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 52
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
[...]

PR #2741, opened 2021-08-26
Fix precision propagation through function calls

ShabbyX:
The TIntermTyped::propagatePrecision was given a precision and it
propagated it to its operands.  For aggregates, it was applying the
precision to all its operands.  This is not correct for function calls
(including built-ins), and this change limits this to only constructors.

For example, glslang was mistranslating the following code:

    mediump sampler2D s;
    highp ivec2 coords = ...;
    mediump result = texture(s, vec3(int(coords.x) & 3, int(coords.y) & y, 3));

where parameter 2 of the texture call is a constructor (which doesn't
have a precision per spec), and the expressions inside the constructor
don't have a precision either (because they are either constructors or
literals).  glslang was incorrectly applying the precision of the result
(mediump) to all the parameters, including the coordinate calculation,
causing visual artifacts in ANGLE traces.

Fixes: #2740


glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -51,6 +51,26 @@
...
+namespace {
+// Is an aggregate operator precision-propagating.
+bool IsPrecisionPropagatingAggregate(const TIntermAggregate& node)
+{
+    if (node.isConstructor())
+    {
+        return true;
+    }
+
+    if (node.isTexture() || node.isImage())
```
greg-lunarg, line 63: This line of code is not correct. For the purposes of precision "propagation", all built-in functions (including texture()) are considered eligible. See GLSL 4.7.2:

For this paragraph, “operation” includes operators, built-in functions, and constructors, and
“operand” includes function arguments and constructor arguments.

See my general review comments for more thoughts.

greg-lunarg (changes requested):
You are right that the mediump should not propagate into the coordinate computation, but your reason is not correct and your PR is not in agreement with the GLSL spec.

In general, I believe this problem needs to be addressed in a different way. See my other comments for more details.

glslang/MachineIndependent/Intermediate.cpp
ShabbyX: > You are right that the mediump should not propagate into the coordinate computation

Ok, thanks for pointing that piece of the spec out. What makes you say that my claim is correct then?

greg-lunarg:
In your example, the variable coords is defined as highp, so anything consuming coords, such as the texture() call, should not change the computations computing the value for coords.

In general, if a precision is ever overridden, it should only increase, it should never decrease.

In general, the precision of an operation should be determined by the operands. If there is a mixture of precisions in an operation, the highest precision should be used. If no precision can be determined from the operands, the precision should come from consuming computations. If that still does not give a precision, default precision should be used.

So, in your example, the vec3 constructor should be considered highp by the rule that operands decide.

I agree that the compiler should not change the computation of coords to mediump. I suspect there is a bug in the propagation code somewhere, but I don't think the current proposed code change is the solution.

ShabbyX:
Thanks for the explanation. I went ahead and read the relevant section in the spec more closely and you are right on all accounts. In particular, I had also missed this:

> The precision used to internally evaluate an operation, and the precision qualification subsequently associated with any resulting intermediate values, must be at least as high as the highest precision qualification of the operands consumed by the operation.

which seems somewhat at odds with:

> Literal constants do not have precision qualifiers. Neither do Boolean variables. Neither do constructors.

Because constructors then must necessarily have a precision at least as large as their operands.

I'll close this PR then, and leave it to you for a fix. One possible solution could be to have `TParseContext::handleConstructorCall` actually assign a "tentative" precision instead of `EpqNone`, and have `TIntermTyped::propagatePrecision` increase precision for constructors, but nothing else (i.e. instead of early out in case of `getQualifier().precision != EpqNone`, it would special-case constructors and allow them to have their precision increased). How does that sound?

PR #2747, opened 2021-09-02
Perform update_precision on constructors and converts

### LINKED ISSUE #2740: Precision is incorrectly propagated to function call arguments

greg-lunarg:
Fixes #2740


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7691,7 +7691,13 @@ TIntermTyped* TParseContext::addConstructor(const TSourceLoc& loc, TIntermNode*
             return nullptr;
     }
 
-    return intermediate.setAggregateOperator(aggrNode, op, type, loc);
+    TIntermTyped *ret_node = intermediate.setAggregateOperator(aggrNode, op, type, loc);
+
+    TIntermAggregate *agg_node = ret_node->getAsAggregate();
+    if (agg_node && agg_node->isVector())
```
ShabbyX, line 7697: Why only vectors?

greg-lunarg: Yes, matrix and array should also be handled. I just did vector for the moment to more quickly fix your problem and get on to a couple other more pressing issues. I will open an issue to track this.

PR #2749, opened 2021-09-04
GL_EXT_spirv_intrinsics - Port extensions

AaronHaganAMD:
Add mechanism to use GL_EXT_spirv_intrinsics headers in glslang.
Ported GL_EXT_shader_realtime_clock as an example.


greg-lunarg:
Re testing failures, it looks like you will need to inform the other builds of the new files.

greg-lunarg (changes requested):
Some general comments:
- I am generally fine with moving ahead with this new framework for extension intrinsic functions.
- I would like to see the extension-specific headers generated into the "build" directory rather than into the "source" tree.
- If I am understanding this PR correctly, the header file glslang/ExtensionHeaders/GL_EXT_shader_realtime_clock.h is a generated file and should not be part of the source controlled repo, right?

AaronHaganAMD:
> * should not be part of the source controlled repo, rig

I re-worked this to take the header files from glslang\ExtensionHeaders and put them in build\include\glslang\glsl_intrinsic_header.h

greg-lunarg:
I think you are going to have to update the bazel and gn builds, which will likely mean changes to BUILD.bazel and BUILD.gn in the root directory, and possibly other supporting scripts or files.

greg-lunarg:
I am still seeing that you are attempting to add glslang/ExtensionHeaders/GL_EXT_shader_realtime_clock.h to the repo (see my comment above) which I believe is incorrect. This is the reason for the license failure above. I think if you fix this as I suggest above the license failure will go away.

greg-lunarg:
Heads up: this PR was partially reverted by #2817, specifically the port of GL_EXT_shader_realtime_clock. This was due to regressions caused for users of the glslang library interface. We were building for the SDK and did not have time to fix this, hence the revert. Please address this problem and re-submit.

PR #2756, opened 2021-09-16
GL_EXT_vulkan_glsl_relaxed - avoid growing the global uniform block with duplicates

mbechard:
When using GL_EXT_vulkan_glsl_relaxed, check to see if a
uniform already exists in the block uniform block, and if so,
ensure they are the same type. If they are the same, avoid growing the block
with a duplicate. If they are different, throw an error.


greg-lunarg (approved):
LGTM.

PR #2759, opened 2021-09-16
GL_EXT_vulkan_glsl_relaxed - retarget gl_VertexID to gl_VertexIndex

mbechard:
I'm not sure if this is the best way to do this. In the original implementation GL_EXT_vulkan_glsl_relaxed, it was incorrectly creating two references to the VertexIndex and InstanceIndex SPIR-V variables when both the *ID and *Index variants of the variables were used.
This instead adds a retargeting mechanism that allows the symbolTable entries for the *ID variants to instead just point to the same symbol as the *Index ones.
However it's possible this should be done elsewhere, possibly where the parser finds the symbol that is passed into handleVariable during parsing. It could just retarget the symbol when it encounters the *ID tokens. What do you think?


greg-lunarg (approved):
The results here are certainly better than they were.

We certainly need to do matching at a point when all needed semantics are available. Doing something too early (say, at parse token time), and you risk doing an unintended substitution.

This seems like it is well designed in this regard, so I approve moving ahead with this change.

PR #2770, opened 2021-09-28
cmake: Remove "conditions" from endif

ncesario-lunarg:
See https://cmake.org/cmake/help/latest/command/endif.html.

If the endif condition does not match the if condition verbatim, an
error is produced on some versions of cmake.

FYI @greg-lunarg @jeremy-lunarg. I ran into this and noticed there were a lot of "endif conditions" which, apparently, can be problematic for some versions of cmake.


greg-lunarg:
Is "else (<condition>)" a problem as well?

ncesario-lunarg:
> Is "else ()" a problem as well?

Looks like it is (not sure why `endif` has it's own call out): https://cmake.org/cmake/help/latest/command/if.html#command:if
> Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command.

I can add cleaning those up to this PR.

ncesario-lunarg:
Think I got most of (if not all) of these. It's building on my linux box (which is where it was failing), so it's at least working for that path.

jeremy-lunarg (approved):
Looks good to me. Apparently, this was required prior to cmake version 2.6. We now rely on a version  2.8.12 or newer.

PR #2777, opened 2021-10-13
Make modifications of GL_EXT_spirv_intrinsics

amdrexu:
1. spirv_execution_mode_id and spirv_decorate_id could support
   specialization constants. The original implementation always
   assume only normal frontend constants are valid. It is not true.

2. spirv_type donesn't support type_specifier as an option of
   spirv_type_parameter. At present, only constant_expression is the
   valid option.


amdrexu:
Change is updated.

Issue #2751, opened 2021-09-08
Incomplete cmake find_package support


rhabacker:
According to https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#libraries-providing-config-file-packages, libraries that want to support cmake find_package() in configuration mode must install a set of files as shown for the VulkanSceneGraph package.  

```
/usr/lib64/cmake/vsg/vsgConfig.cmake
/usr/lib64/cmake/vsg/vsgConfigVersion.cmake
/usr/lib64/cmake/vsg/vsgTargets-relwithdebinfo.cmake
/usr/lib64/cmake/vsg/vsgTargets.cmake
```
Once these files are installed, the vsg library can be added as a dependency to a package with cmake support by using 
```
find_package(vsg)
``` 
to a CMakeLists.txt file.

A look into the installed files of the glslang package (see for example at [openSUSE](https://build.opensuse.org/package/show/openSUSE%3AFactory/glslang)
```
/usr/lib64/cmake/HLSLTargets-relwithdebinfo.cmake
/usr/lib64/cmake/HLSLTargets.cmake
/usr/lib64/cmake/OGLCompilerTargets-relwithdebinfo.cmake
/usr/lib64/cmake/OGLCompilerTargets.cmake
/usr/lib64/cmake/OSDependentTargets-relwithdebinfo.cmake
/usr/lib64/cmake/OSDependentTargets.cmake
/usr/lib64/cmake/SPIRVTargets-relwithdebinfo.cmake
/usr/lib64/cmake/SPIRVTargets.cmake
/usr/lib64/cmake/SPVRemapperTargets-relwithdebinfo.cmake
/usr/lib64/cmake/SPVRemapperTargets.cmake
/usr/lib64/cmake/glslang-default-resource-limitsTargets-relwithdebinfo.cmake
/usr/lib64/cmake/glslang-default-resource-limitsTargets.cmake
/usr/lib64/cmake/glslangTargets-relwithdebinfo.cmake
/usr/lib64/cmake/glslangTargets.cmake
/usr/lib64/cmake/glslangValidatorTargets-relwithdebinfo.cmake
/usr/lib64/cmake/glslangValidatorTargets.cmake
/usr/lib64/cmake/spirv-remapTargets-relwithdebinfo.cmake
/usr/lib64/cmake/spirv-remapTargets.cmake
```
shows that there are no `*Config.cmake` and *Version.cmake` files. 

For example, trying to use the `glslang` package with 
```
[...]

PR #2778, opened 2021-10-13
Fix support to generate cmake config files

rhabacker:
See https://github.com/KhronosGroup/glslang/issues/2751

These commits are used to create and install cmake support files. The extent to which the generated `*Config.cmake` files are usable and complete would need to be tested.

This can be done by appropriately modifying the cmake build system of projects that use glslang, or by adding a cmake test to this repo based on ctest [`--build-and-test`](https://cmake.org/cmake/help/latest/manual/ctest.1.html#build-and-test-mode) mode.

The new macro `add_cmake_support_files` was derived from https://github.com/vsg-dev/VulkanSceneGraph/blob/ad93a77c0e7e3d0ff73552323a1568fecd52239d/cmake/vsgMacros.cmake#L97.

Since this macro is already used in a similar form in more than two repos (glslang, VulkanSceneGraph and sibling repos), after a stabilization period in this repo it might be useful to provide it together with other useful macros in a separate git repo, e.g. `vulkan-cmake-macros` to avoid cmake code duplication.


rhabacker:
This pull request is probably outdated by https://github.com/KhronosGroup/glslang/pull/2989.

rhabacker:
Rebased on https://github.com/KhronosGroup/glslang/pull/2989.

Because the SPIRV-Tools-opt development package is now considered a third-party package that must be pre-installed, the client package test fails at msvc.

arcady-lunarg:
Closing due to inactivity. Feel free to open a new PR if you want to pick this back up.

PR #2781, opened 2021-10-13
Fix incorrect link time validation for unused gl_PerVertex members

kevin-mccullough:
Whenever a built in block such as gl_PerVertex is explicitly redeclared glslang has slightly different behavior depending on whether or not the block is anonymous. For blocks that have an instance name any members absent from the redeclaration are simply deleted, while for anonymous blocks any members absent from the redeclaration are instead "hidden". Most of the codebase handles hidden members correctly but the sameStructType() function does not. As a result of this whenever identical redeclarations of gl_PerVertex with one or more members omitted appear in both the vertex and geometry stages sameStructType() will incorrectly report that the types do not match. This change fixes the logic in sameStructType() to correctly skip hidden members when comparing structures.

### DIFF
--- a/Test/baseResults/link.redeclareBuiltin.vert.out
+++ b/Test/baseResults/link.redeclareBuiltin.vert.out
@@ -1,0 +1,154 @@
+link.redeclareBuiltin.vert
+Shader version: 410
+0:? Sequence
+0:8  Function Definition: main( ( global void)
+0:8    Function Parameters: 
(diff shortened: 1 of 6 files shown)


kevin-mccullough:
I have submitted an update with tests.

greg-lunarg (approved):
LGTM. Thanks for the fix and the test!

PR #2786, opened 2021-10-19
Fix Cmake rule for extension header generation

dneto0:
Allow glslang to be embedded in an enclosing project (such as Shaderc)


dneto0:
Cc: @greg-lunarg Please take a look at your leisure.

greg-lunarg:
@dneto0 Thanks for the fix!

PR #2800, opened 2021-11-08
Fix encoding/decoding of string literals for big-endian systems

### LINKED ISSUE #202: [SPIR-V] Incorrect encoding for string literals on big-endian machines
The SPIR-V spec says, in the definition of "literal string," "the UTF-8 octets (8-bit bytes) are packed four per word, following the little-endian convention (i.e., the first octet is in the lowest-order 8 bits of the word)." But when emitting literal strings [here](https://github.com/KhronosGroup/glslang/blob/master/SPIRV/spvIR.h#L104), we just reinterpret an integer as a character array, which won't work on big-endian systems, since we'll end up packing them according to the big-endian convention instead.

mhillenbrand:
Per SPIR-V spec, a string literal's UTF-8 octets are encoded packed into
words with little-endian convention. Explicitly perform that encoding
instead of assuming that the host system is little-endian.

Note that this change requires corresponding fixes in SPIRV-Tools (proposed in PR https://github.com/KhronosGroup/SPIRV-Tools/pull/4622)

Fixes #202

Tested on x86-64 and s390x, no regression in google test-backed tests or in ./runTests-backed tests.

### DIFF
--- a/SPIRV/SPVRemapper.cpp
+++ b/SPIRV/SPVRemapper.cpp
@@ -297,15 +297,21 @@ namespace spv {
     std::string spirvbin_t::literalString(unsigned word) const
     {
         std::string literal;
+        const spirword_t * pos = spv.data() + word;
 
         literal.reserve(16);
 
-        const char* bytes = reinterpret_cast<const char*>(spv.data() + word);
-
-        while (bytes && *bytes)
-            literal += *bytes++;
-
-        return literal;
+        do {
+            spirword_t word = *pos;
+            for (int i = 0; i < 4; i++) {
+                char c = word & 0xff;
+                if (c == '\0')
+                    return literal;
+                literal += c;
+                word >>= 8;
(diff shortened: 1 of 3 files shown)


greg-lunarg:
Thinking it would be "best" to wait and merge this code at the same time as spirv-tools merges their fix and make that spirv-tools the known-good for glslang. Do you concur @mhillenbrand ?

mhillenbrand:
@greg-lunarg I agree. Tying the two fixes together via the known-good version will help avoid inconsistent combinations on s390x.

mhillenbrand:
The corresponding fix in SPIRV-Tools has been merged https://github.com/KhronosGroup/SPIRV-Tools/pull/4622

greg-lunarg:
I will updating our dependency to SPIRV-Tools early next week and will merge this PR at that time.

PR #2801, opened 2021-11-08
Thread safety fixes

haasn:
I noticed that the current process initializer is extremely thread-unsafe, in multiple ways, despite ostensibly taking a mutex.


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2803, opened 2021-11-09
Use intermOut.cpp's IsNan and IsInfinity for parse-time constant folding

### LINKED ISSUE #2802: isNan and isInf in Constant.cpp produce wrong results on big-endian systems
Both attempt to decode a `double` into high and low 32-bit `int`s via a union:

https://github.com/KhronosGroup/glslang/blob/8f56d738000ba2b39595160b2d345821ce03d18f/glslang/MachineIndependent/Constant.cpp#L49-L52

On little-endian systems that works as intended. On big-endian systems, such as s390x, the 32-bit ints end up in reversed order and thus the two functions fail to detect Infs or NaNs.

Reproduction: `glslangValidator -C -i Test/constFold.frag` and compare `out10` and `out11` to expected values -- on s390x, they are all 0 since none of the NaN or Inf conditions are detected correctly.

mhillenbrand:
There were two implementations of isInf() and isNan(), in Constant.cpp and in intermOut.cpp. The former only works on little-endian systems, the latter is a wrapper for library functions and works regardless of endianness. Move the second version into Common.h and adopt it in both places. Thereby avoid the duplication and fix for big-endian systems.

Using host library functions for identifying Inf and NaNs is consistent with the constant-folding in Constant.cpp using the host's native arithmetic anyway.

On s390x, this commit fixes the test case Glsl/CompileToAstTest.FromFile/constFold_frag.

Tested with Linux on s390x and x86-64. No regressions.

Fixes #2802

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -39,6 +39,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 #include <cstdio>
 #include <cstdlib>
 #include <list>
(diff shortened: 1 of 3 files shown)


greg-lunarg (approved):
LGTM. Thanks for the fix!

Issue #2815, opened 2021-11-11
Undefined behavior in Test/constantUnaryConversion.comp causes test failures on s390x (Linux) and aarch64 (MacOS)


mhillenbrand:
The testcase Test/constantUnaryConversion.comp converts a negative float64 to several unsigned integer data types. The GLSlang spec calls these conversions undefined (in Section 5.4.1 Conversion and Scalar Constructors). The corresponding casts in glslang (in TIntermediate::promoteConstantUnion()) are undefined behavior in C++.

The observed results differ between x86 (two's complement representation) and aarch64 and s390x (both convert negative floats to 0). Since the test case's expected output matches the behavior on x86, the test case fails on aarch64 (tested in MacOS) and s390x (in Linux) even though the behavior is aligned with the spec.

That is, the test case is stricter than the spec and expects x86-style behavior for what the spec leaves as undefined. Is that intentional (I may miss the reason, or misinterpret undefined behavior in the context of GLSlang)?

If not, I propose to remove the undefined cases from the test and, while at it, adding conversions from negative float16 and negative float32 to signed integers.

Reproduction to run with `glslangValidator -i` (reduced from constantUnaryConversion.comp):
```
#version 450
#extension GL_EXT_shader_explicit_arithmetic_types : require

const float64_t f = float64_t(- 4.0);

const uint8_t float64_t_to_uint8_t = uint8_t(f);
const uint16_t float64_t_to_uint16_t = uint16_t(f);
const uint32_t float64_t_to_uint32_t = uint32_t(f);
const uint64_t float64_t_to_uint64_t = uint64_t(f);

const uint float64_t_to_uint = uint(f);

void main(){ }
```

mhillenbrand:
Similarly, the test case `ToSpirv/HlslCompileTest.FromFile/hlsl_inf_vert` fails on aarch64 and s390x because platform-dependent behavior differs from the reference output.

The test case compiles a float constant _NaN_ into the SPIR-V binary. On x86-64, that NaN happens to be negative and the reference output expects that. aarch64 and s390x produce positive NaNs, though.

IEEE 754 allows both negative and positive NaNs and does not prescribe the sign bit when NaN is the result of an operation; I have not found more restrictice terms in the specs for SPIR-V or GLSlang. Thus, I conclude that a negative NaN should be an acceptable result and the test case would be too strict.

Short reproducer (reduced from `Test/hlsl.inf.vert`
```
float main() : SV_Position
{
[...]

PR #2820, opened 2021-11-15
[RFC] add more negative float conversions to test constantUnaryConversion.comp and remove the undefined ones

mhillenbrand:
Conversions of negative float values to unsigned integers are undefined behavior (in both GLSlang and as implemented in glslang in C++) and produce different platform-specific  results. As a result, the test case fails on aarch64 and s390x because the golden output was captured on x86, as described in #2815 . 

One option to avoid those false positive test failures is dropping the undefined conversions from the test case, as in this PR. In addition, this PR adds (defined) conversions for negative values in float16_t and float32_t.

To be clear, the only problem this is solving is noise in the test suite on !x86.

What do you think?


greg-lunarg:
Agree that conversion of negative float value to uint type is undefined and any tests with this behavior should be removed.

Re broken bazel build, maybe rebase your branch and repush?

greg-lunarg:
Closed in deference to #2821. No functional difference, just a CI rerun.

PR #2823, opened 2021-11-16
Initialize member TSymbol::uniqueId.

kda:
Using non-default flags for clang for memory sanitizer disclosed uses of uniqueId where it was uninitialized.


greg-lunarg (approved):
LGTM. Thanks for the fix!

Issue #2804, opened 2021-11-09
I/O variables declared as locals with GL_EXT_spirv_intrinsics


greg-lunarg:
A recent fix of spirv-val revealed bad SPIR-V generation for test spv.intrinsicsSpirvLiteral.vert. The variables vec4In and vec4Out are declared in the generated SPIR-V as storage class Function rather than Input and Output, respectively. This seems related to the GL_EXT_spirv_intrinsics extension since replacing the extension trappings in the GLSL source with a simple assignment gives correct declarations.

@amdrexu Can you please fix this test or find someone who can?

amdrexu:
Yes, I will take a look.

greg-lunarg:
Fixed by #2831

PR #2831, opened 2021-11-23
Fix an issue of spirv_by_reference

amdrexu:
When using this qualifier for a parameter, we make it as a pointer.
However, the function TranslateStorageClass() is therefore called and
the storage class should only be set to Function when it is invoked
to translate parameter types rather than actual argument types.


amdrexu:
This is to address the issue: https://github.com/KhronosGroup/glslang/issues/2804

jeremy-lunarg (approved):

greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2832, opened 2021-11-25
Fix validation failures of test cases regarding GL_EXT_spirv_intrinsics

amdrexu:
1. The test cases of OpReadClockKHR is invalid. The return type is
   unsigned integer rather than signed integer.
2. When SPIR-V decorate or SPIR-V type is specified, we should avoid
   auto location mapping because the semantics are totally decided by
   SPIR-V tokens.


jeremy-lunarg (approved):

jeremy-lunarg (approved):
LGTM

greg-lunarg (approved):
LGTM

PR #2834, opened 2021-11-26
Support for automap options and resource binding shift

chaoticbob:
- Added glslang_shader_options_t with enums for auto map bindings and
  auto map locations.
- Added options param to glslang_shader_parse.
- Modified glslang_shader_parse to call appropriate auto map function if
  auto map bits are set.


jeremy-lunarg (approved):
LGTM

greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2836, opened 2021-11-30
Vulkan rules relaxed for glslang C interface

chaoticbob:
Hi,

This PR adds Vulkan rules relaxed to the glslang C interface to make it more useful for Vulkan. I wasn't able to find a precedent for how this behavior should be added so I added a member variable to glslang_input_t. If this doesn't feel like a natural usage to the C interface I'm happy to make any adjustments. PTAL. 

Cheers,
Hai


jeremy-lunarg (approved):
LGTM

greg-lunarg (changes requested):
While this would work, I think even better would be to add to the interface you created with your last PR (glslang_shader_set_options) and add this option along with the auto location and auto binding options.

The reason I would prefer that approach is that it would avoid changing the externally visible structure glslang_input_t which can "break" existing code using this interface.

Does this sound ok?

chaoticbob:
@greg-lunarg Yes, that makes sense. Will make the appropriate changes.

chaoticbob:
@greg-lunarg I made the changes per your suggestion. PTAL.

glslang/CInterface/glslang_c_interface.cpp
```diff
@@ -367,6 +367,11 @@ GLSLANG_EXPORT void glslang_shader_set_options(glslang_shader_t* shader, int opt
     if (options & GLSLANG_SHADER_AUTO_MAP_LOCATIONS) {
         shader->shader->setAutoMapLocations(true);
     }
+
+    if (options & GLSLANG_SHADER_AUTO_MAP_LOCATIONS) {
```
greg-lunarg, line 371: Cut and paste error?

greg-lunarg (changes requested):
I think just one problem, otherwise good.

glslang/CInterface/glslang_c_interface.cpp
chaoticbob: Totally. This what I get for coding at 4AM. Will fix. Sorry.

chaoticbob:
@greg-lunarg Fixed the copy/paste error.

greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2838, opened 2021-12-01
Remove unnecessary badge

jeremy-lunarg:
I had copy/pasted this badge without really thinking about it's
usefulness. It is unnecessary.


greg-lunarg (approved):
LGTM. Thanks!

PR #2840, opened 2021-12-06
Do not output location decoration for certain variables in ray tracing storage classes.

### LINKED ISSUE #2736: Stop assigning Location decorations to ray payloads and callable data
The SPIR-V spec says:
> **Location**
>
> Apply only to a variable or a structure-type member. *Location* is an unsigned 32-bit integer that forms the main linkage for Storage Class **Input** and **Output** variables:
> - between the client API and vertex-stage inputs,
> - between consecutive programmable stages, or
> - between fragment-stage outputs and the client API.
>
> It can also tag variables or structure-type members in the **UniformConstant** Storage Class for linkage with the client API.
>
> Only valid for the **Input**, **Output**, and **UniformConstant** Storage Classes.

The SPV_KHR_ray_tracing spec does not modify this language.  However, GLSLang (at least the version in the Vulkan CTS) currently passes through the location from GLSL which is used for matching ray payloads and callable data with the corresponding `traceRayEXT()` or `executeCallableEXT()` instruction.  This is a left-over from the old SPV_NV_ray_tracing extension which also did the matching via location numbers.  The KHR extension does the matching using actual pointers in **OpTraceRayKHR** or **OpExecuteCallableKHR**.

alelenv:
Fixes #2736 
@greg-lunarg to review

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -8716,8 +8716,18 @@ spv::Id TGlslangToSpvTraverser::getSymbolId(const glslang::TIntermSymbol* symbol
             builder.addDecoration(id, spv::DecorationOffset, symbol->getQualifier().layoutOffset);
     }
 
-    if (symbol->getQualifier().hasLocation())
-        builder.addDecoration(id, spv::DecorationLocation, symbol->getQualifier().layoutLocation);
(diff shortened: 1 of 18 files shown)


jeremy-lunarg (commented):
LGTM

greg-lunarg (approved):
LGTM. Thanks for the fix!

Issue #2846, opened 2021-12-20
[question] Can we remove an assert for unsized array use to xfb.


ZhiqianXia:
Hi,
   I have a shader which have unsized array use to xfb and glslang code had supported the feature. But There is a assert at 
https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/linkValidate.cpp + line 1813.
According the comments at line 1759 , i think the assert seems is for unsized array case and it is  out of date.  becasue the feature had supported at ParseHelper.cpp + line 6220. So I think  we should remove the assert.

**glslang code :** 
```
    int numComponents;
    if (type.isScalar())
        numComponents = 1;
    else if (type.isVector())
        numComponents = type.getVectorSize();
    else if (type.isMatrix())
        numComponents = type.getMatrixCols() * type.getMatrixRows();
    else {
        assert(0);  // line 1813
        numComponents = 1;
    }
```

**tes shader : **
```
#version 430 core
#extension GL_ARB_enhanced_layouts : require
layout(isolines, point_mode) in;
layout (xfb_offset = 9) out double gohan[];

in  vec4 tcs_tes[];
out vec4 tes_gs;
void main()
{
    vec4 result = tcs_tes[0];
    gohan[0] = double(0);
    if (vec4(0) == result)
    {
        gohan[0] = double(1);
    }
    tes_gs += result;
}
```

jeremy-lunarg:
Agreed. This behavior is not ideal.

greg-lunarg:
Fixed by #2847

PR #2847, opened 2021-12-23
Unblock unsized error from assert

jeremy-lunarg:
Fix issue #2846.


ZhiqianXia (approved):
LGTM.

greg-lunarg (changes requested):
Let's add a test the confirms the fix. Thanks!

jeremy-lunarg:
> Let's add a test the confirms the fix. Thanks!

Test added. Thanks Greg.

PR #2849, opened 2021-12-29
Fix seg fault

### LINKED ISSUE #2848: Crash when passing a primitive type to a function that expects a struct
When there is a function that has a `struct` as one of the arguments, and the calling code incorrectly passes a primitive type (e.g. `float`) instead of an instance of that structure, the compiler crashes instead of generating an error message.

Example code:
```
#version 450

struct S { float a; };
float func(S s) { return s.a; }

layout(location = 0) out vec4 o_color;

void main()
{
	float b = 0;
	S s;
	s.a = 1;
	float c = func(b); // <-- this should generate an error but crashes instead
	o_color = vec4(c);
}
```
Compile as a fragment shader with `--target-env vulkan1.2`

Relevant part of the compiler call stack:
```
glslang::TType::sameStructType(const glslang::TType & right) Line 2453
glslang::TType::sameElementShape(const glslang::TType & right) Line 2555
glslang::TParseContext::findFunction400::__l2::<lambda>(const glslang::TType & from, const glslang::TType & to, glslang::TOperator __formal, int __formal) Line 6808
[External Code]	
glslang::TParseContextBase::selectFunction(const glslang::TVector<glslang::TFunction const *> candidateList, const glslang::TFunction & call, std::function<bool __cdecl(glslang::TType const &,glslang::TType const &,enum glslang::TOperator,int)> convertible, std::function<bool __cdecl(glslang::TType const &,glslang::TType const &,glslang::TType const &)> better, bool & tie) Line 432
glslang::TParseContext::findFunction400(const glslang::TSourceLoc & loc, const glslang::TFunction & call, bool & builtIn) Line 6839
glslang::TParseContext::findFunction(const glslang::TSourceLoc & loc, const glslang::TFunction & call, bool & builtIn) Line 6655
glslang::TParseContext::handleFunctionCall(const glslang::TSourceLoc & loc, glslang::TFunction * function, TIntermNode * arguments) Line 1325
```

The `sameStructType` method is called on an object representing `float` that has `typeName == NULL`.

jeremy-lunarg:
Check types before accessing typeName.

Fix #2848.

### DIFF
--- a/Test/baseResults/noMatchingFunction.frag.out
+++ b/Test/baseResults/noMatchingFunction.frag.out
@@ -1,0 +1,52 @@
+noMatchingFunction.frag
+ERROR: 0:17: 'func' : no matching overloaded function found 
+ERROR: 1 compilation errors.  No code generated.
+
+
+Shader version: 330
+ERROR: node is still EOpNull!
+0:8  Function Definition: func(struct-S-f11; ( global float)
+0:8    Function Parameters: 
+0:8      's' ( in structure{ global float a})
+0:10    Sequence
+0:10      Branch: Return with expression
+0:10        a: direct index for structure ( global float)
+0:10          's' ( in structure{ global float a})
+0:10          Constant:
+0:10            0 (const int)
+0:15  Function Definition: main( ( global void)
+0:15    Function Parameters: 
+0:17    Sequence
+0:17      Sequence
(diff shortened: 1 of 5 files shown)


greg-lunarg (approved):
LGTM. Thanks!

PR #2851, opened 2022-01-01
Added GLSL version override functionality and CLI

chaoticbob:
PR to address question assekd in ##2850.

This change list allows a user to override the GLSL version from the
command line or through the C and C++ interfaces. This will override the
override happens in ProcessDeferred() before DeduceVersionProfile() so
the process should still error out if the version is insufficient for
the shader code.

- Added --glsl-version <version> to CLI.
- Added parameter to route glslVersion as override version to
  preprocessor and parse functions to C++ interface.
- Updated C interface with function to override GLSL version.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -1261,6 +1261,8 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
             shader->setSourceEntryPoint(sourceEntryPointName);
         }
 
+        shader->setOverrideVersion(GlslVersion);
+
         std::string intrinsicString = getIntrinsic(compUnit.text, compUnit.count);
 
         PreambleString = "";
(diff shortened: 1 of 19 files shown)


greg-lunarg (changes requested):
The code looks fine.

It is common practice to add at least one test which indicates the capability is working to detect regressions. Functionality that involves command line options is often added into the Test/runtests script. Besides updating the script, you will just need to add a glsl shader to Test/ and an output file in Test/baseResults.

chaoticbob:
Hi Greg, thanks for the guidance. I will find a spot of time to make these changes.

greg-lunarg (changes requested):
Thanks for the tests!

Sorry to ask for one more change, but could you change the original versions to 110 so that they will cause a failure if your new option is not enabled. Otherwise looks good.

chaoticbob:
Hi Greg, do you mean the versions in the shaders or in the StandAlone.cpp?

greg-lunarg:
> Hi Greg, do you mean the versions in the shaders or in the StandAlone.cpp?

In the shaders.

greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2852, opened 2022-01-01
build: Make action targets hermetic

gnoliyil:
Fuchsia GN build requires all "action" targets to be hermetic, and they should correctly and fully state their inputs and ouptuts in "sources" and "outputs" fields (see https://fuchsia.dev/fuchsia-src/development/build/hermetic_actions
for details).

This change adds "sources" field to "glslang_extension_headers" build target so that it states all its input files in its GN target build rule.

TEST=fx set workstation.x64 --args=build_should_trace_actions=true
     fx build

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -74,6 +74,15 @@ action("glslang_extension_headers") {
 
   out_file = "${target_gen_dir}/include/glslang/glsl_intrinsic_header.h"
 
+  # Fuchsia GN build rules require all GN actions to be hermetic and they
+  # should correctly and fully state their inputs and outpus (see
+  # https://fuchsia.dev/fuchsia-src/development/build/hermetic_actions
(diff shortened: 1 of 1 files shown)


gnoliyil:
@jbauman42 @rosasco-wk FYI

gnoliyil:
@greg-lunarg Would you please take a look? Thanks!

greg-lunarg:
Sorry for the delay. Reviewing now...

jeremy-lunarg (approved):
LGTM. Were you seeing a build error?

greg-lunarg (approved):
LGTM. Thanks for the code!

gnoliyil:
> LGTM. Were you seeing a build error?

Yes. We saw build errors on Fuchsia where hermetics action targets are enforced.

PR #2862, opened 2022-01-08
#2861 macos gcc build failures

### LINKED ISSUE #2861: build failures on Mac OS X with gcc
There are two failures I'm running into when trying to build on Mac OS X with gcc (11.2.0). My CMake command is:

```
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=/opt/homebrew/Cellar/gcc/11.2.0_3/bin/g++-11 -DCMAKE_C_COMPILER=/opt/homebrew/Cellar/gcc/11.2.0_3/bin/gcc-11 ..
```

The first is invalid compiler arguments being generated (e.g. `-Xarch_arm64`) by precompiled header generation. A workaround for this is to set `CMAKE_DISABLE_PRECOMPILE_HEADERS` to ON during configure time. This seems to me to be a CMake issue proper rather than an issue with glslang.

The second is a linker error; glslang's CMakeLists.txt unconditionally adds `--no-undefined` to the linker command line if the compiler is gcc. However on Mac OS, the default linker does not understand this argument and fails with:

```
ld: unknown option: --no-undefined
collect2: error: ld returned 1 exit status
```

Since it seems to be a bit difficult to detect whether ld is gnu ld or the one shipped with mac os, a workaround could be to check the cmake platform in CMakeLists.txt in addition to the version check before adding this flag:

```
if(NOT CMAKE_VERSION VERSION_LESS "3.13" AND NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```

I have a patch with this minor change that I can create a PR from if the devs are interested.

n-morales:
Closes #2861

This PR adds a few minor changes for building with gcc on Mac OS X. It changes the default for `ENABLE_PCH` to `OFF` for that particular configuration and avoids adding the --no-undefined linker flag since it isn't valid for the default MacOS linker.

I changed the default for ENABLE_PCH rather than disabling entirely so that it would still be configurable and existing builds shouldn't break. However, it should now work out-of-the-box after cloning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,7 +106,14 @@ CMAKE_DEPENDENT_OPTION(ENABLE_HLSL
 option(ENABLE_RTTI "Enables RTTI" OFF)
 option(ENABLE_EXCEPTIONS "Enables Exceptions" OFF)
 option(ENABLE_OPT "Enables spirv-opt capability if present" ON)
-option(ENABLE_PCH "Enables Precompiled header" ON)
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND ${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
+    # Workaround for CMake behavior on Mac OS with gcc, cmake generates -Xarch_* arguments
+    # which gcc rejects
+    option(ENABLE_PCH "Enables Precompiled header" OFF)
+else()
+    option(ENABLE_PCH "Enables Precompiled header" ON)
+endif()
 option(ENABLE_CTEST "Enables testing" ON)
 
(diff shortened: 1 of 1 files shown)


greg-lunarg (approved):
LGTM. Thanks for the code!

greg-lunarg:
Please sign CLA so I can merge.

n-morales:
> Please sign CLA so I can merge.

I'm not sure why, but I have signed the CLA twice now and it still seems to be pending. I also tried the recheck link from the bot. Going to the details of the workflow, I'm informed that I have already signed the CLA.

n-morales:
Apparently I had an old email in my git config. Anyway, CLA issues should be fixed and I also rebased on current master.

Issue #1015, opened 2017-08-13
CMakeLists.txt sets CMAKE_INSTALL_PREFIX


Richard-W:
Setting CMAKE_INSTALL_PREFIX is very bad practice.

Especially since it sets it to a path relative to the build directory it breaks most packaging systems entirely since the result of `make DESTDIR=<dir> install` starts to depend on the absolute path of the build directory.

Removing this shouldn't break (sensible) workflows at all.

Zangetsu38:
I have this error also, absolutely need remove this, never touch defaut CMAKE_INSTALL_PREFIX
and set this code like this is very bad and broken all time compilling on few program on cmake, exemple with yaml-cpp it gèet error caused by set this "install"
>if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND WIN32)
    set(CMAKE_INSTALL_PREFIX "install" CACHE STRING "..." FORCE)
endif()

Zangetsu38:
no, remove this part broken nothing, is particularly set this part broken app

johnkslang:
I understand your input very clearly. I only wanted to small chance for anyone else to say or realize this will effect them, if their workflow is different than yours. I understand it is a low chance. It is just nice to ask.

Zangetsu38:
I see, but i don't think, because "CMAKE_INSTALL_PREFIX " with value "install" no have sense, this value return folder, so here big fail

Zangetsu38:
i have also see other value added no idea why
>if(WIN32)
set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Adds a postfix for debug-built libraries.")

i get so strangely d alone like that, no idea what is it

johnkslang:
I believe you are now discussing issue #1540.  Let's keep the discussions separate.

Zangetsu38:
oh ok this have already issue, but so yes, need remove this "CMAKE_INSTALL_PREFIX" re set value, and leave it normal by default, for me only it broken build with error, depending project i think

PR #2864, opened 2022-01-09
Override CMAKE_INSTALL_PREFIX only if ENABLE_GLSLANG_INSTALL is ON

scribam:
`CMAKE_INSTALL_PREFIX` variable is only useful when you want to run the CMake "install" target so this variable should only be overridden when it is necessary.

It will prevent issue https://github.com/KhronosGroup/glslang/issues/1015 on projects which include glslang but don't need to install it (using SKIP_GLSLANG_INSTALL) .


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2865, opened 2022-01-10
Add missing license to LICENSE.txt

alan-baker:
* This license is used by:
  * SPIRV/GLSL.ext.AMD.h
  * SPIRV/GLSL.ext.EXT.h
  * SPIRV/GLSL.ext.KHR.h
  * SPIRV/GLSL.ext.NV.h

This was flagged by our internal tooling as a missing license from LICENSE.txt.


greg-lunarg (approved):
Thanks for the fix!

PR #2866, opened 2022-01-11
Temporarily disable spirv1.6 tests

### LINKED ISSUE #2858: Unit test non-determinism
When SPIR-V 1.6 support was add, a test non-determinism was introduced. If you run all of the tests in Spv.FromFile.cpp together they pass, but if you filter and just run the 1.6 tests then `spv.1.6.conditionalDiscard.frag` fails with a diff in the expected SPIR-V.

```
With diff:
@@ +18,5 @@
                               Decorate 13(tex) Binding 0
                               Decorate 17(coord) Location 0
+                              Decorate 34(gl_FragColor) Location 0
                2:             TypeVoid
                3:             TypeFunction 2
```

To reproduce, run:
```
# all tests
./gtests/glslangtests

# just 1.6
./gtests/glslangtests --gtest_filter=*spv_1_6_*
```

I haven't tracked it down, but I assume there is some global state that is not properly reset (presumably related to the client environment).

greg-lunarg:
These tests will be re-enabled when a Vulkan version is released that
supports spirv1.6.

Fixes #2858

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -162,8 +162,11 @@ typedef enum {
     LAST_ELEMENT_MARKER(EShTargetCount),
 } EShTargetLanguage;
 
+// TODO(greg-lunarg): Fix non-determinism problem with Universal
+// https://github.com/KhronosGroup/glslang/issues/2858
+
 typedef enum {
-    EShTargetUniversal = 0,                           // Universal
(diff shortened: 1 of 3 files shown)


jeremy-lunarg (approved):
LGTM!

ben-clayton:
Many thanks!

PR #2868, opened 2022-01-13
Update LICENSE.txt

ben-clayton:
Expand the GPL 3 with special bison exception to the fully expanded license.

Required for Google's updated license checker.


greg-lunarg (approved):
As this is only a change in the text and not the effect of the license, this is fine.

ben-clayton:
Many thanks.

PR #2873, opened 2022-01-25
Fix comment for setEnvInput()

### LINKED ISSUE #2872: Improve comment for setEnvInput()
I'm trying to use glslang 11.7.1 with the renderdoc project. Renderdoc bundles glslang 8.13.3743.

Renderdoc calls `setEnvInput()` with a version parameter of 100, and uses it to pre-process GLSL (not GLSL+SPIR-V) sources.

Since glslang 6274ec5c (glslang 11.2.0), this appears to cause the glslang to define `GL_SPIRV` during pre-processing. `GL_SPIRV` ends up causing issues for the GLSL source output from the glslang pre-processor due to `#ifdef GL_SPIRV` usage by renderdoc.

Looking at the glslang README.md, it seems that `0` is recommended for the `version` parameter of `setEnvInput()` "For just validating (not generating code)".

What would be the recommended `version` value for using to glslang to pre-process GLSL (but, not GLSL+SPIR-V) sources? Is there anything in the glslang documentation that I'm missing that would clear this up? Why does setting it to a non-zero value cause glslang to define `GL_SPIRV`?

greg-lunarg:
Specifically, add text to clarify dialect and dialectVersion

Fixes #2872

### DIFF
--- a/glslang/Public/ShaderLang.h
+++ b/glslang/Public/ShaderLang.h
@@ -150,8 +150,8 @@ typedef enum {
 
 typedef enum {
     EShClientNone,               // use when there is no client, e.g. for validation
-    EShClientVulkan,
-    EShClientOpenGL,
+    EShClientVulkan,             // as GLSL dialect, specifies KHR_vulkan_glsl extension
+    EShClientOpenGL,             // as GLSL dialect, specifies ARB_gl_spirv extension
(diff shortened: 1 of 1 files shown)


dneto0 (commented):
Seems like an improvement to me.  :-)

PR #2874, opened 2022-01-26
fix cases where symbols in the tree didn't get updated during block merging

mbechard:
For GL_EXT_vulkan_glsl_relaxed. When merging the default uniform block,
there were cases where symbols in the tree wern't updated to match the
new block structure after merging blocks together.

This change traverses the symbol tree and updates any references to the
merged block.

### DIFF
--- a/Test/baseResults/vk.relaxed.stagelink.0.0.vert.out
+++ b/Test/baseResults/vk.relaxed.stagelink.0.0.vert.out
@@ -1,0 +1,10697 @@
+vk.relaxed.stagelink.0.0.vert
+Shader version: 460
+0:? Sequence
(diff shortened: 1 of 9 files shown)


greg-lunarg:
Could you please supply shader code which demonstrates this fix? I would like to have a test that locks this behavior in. I can create the test or you can, whichever you would prefer. Thanks!

mbechard:
re-pushed with test case included. Without this fixed the link fails with the error:
INTERNAL ERROR: Invalid Uniform variable type : gl_DefaultUniformBlock

greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2883, opened 2022-02-01
Enhance readability of error messages for GLSL

greg-lunarg:
Specifically, make GLSL link error messages more specific and output
only information relevant to the error.

Also change type printing to more closely reflect GLSL syntax. This
is the default for link error messages, but must me enabled with the
new option --enhanced-msgs for compilation error messages.

Also with --enhanced-msgs, only emit one error message per source
line.


jeremy-lunarg (approved):
LGTM

PR #2891, opened 2022-02-16
Fix Test/hlsl.namespace.frag test case

dneto0:
Before this change, the example is rejected by DXC:

$ dxc -T ps_6_0 hlsl.namespace.frag
hlsl.namespace.frag:22:73: error: call to non-static member function
without an object argument
    return N1::getVec() + N2::getVec() + N2::N3::getVec() + N2::N3::C1::getVec() * N2::gf;
                                                                ~~~~~~~~~~~~^~~~~~

The call to the class member function requires an object, or we ned to
make the function static.  This update makes the function static.

This also fixes SPIR-V validation: without this change the call
to that getVec does not have enough arguments:

error: line 69: OpFunctionCall Function <id>'s parameter count does not
match the argument count.
  %43 = OpFunctionCall %v4float %N2__N3__C1__getVec_


greg-lunarg (approved):
Thanks for the fix!

Issue #2898, opened 2022-03-01
redeclare issue for gl_FragCoord.


ZhiqianXia:
Hi ,
     I got a shader from glsl cts and the compile result was expected for pass. but I got a **error**  for the shader. 
ogl 150 spec says : 
_Within any shader, the first redeclarations of gl_FragCoord must appear before any use of gl_FragCoord._ So I think
glslang shouldn't report the error for it.

**glslang output**

_ERROR: D:\local\Temp\1244202e-a8fa-4b0c-9a67-63bb6878bc9e.tmp:10: 'gl_FragCoord' : cannot redeclare after use 

**shader code** 
```
#version 150 core
#extension GL_ARB_explicit_attrib_location : enable

#ifdef GL_ES
precision mediump float;
#endif

layout (origin_upper_left,pixel_center_integer) in vec4 gl_FragCoord;
float myGlobalVar = gl_FragCoord.x;
layout (origin_upper_left,pixel_center_integer) in vec4 gl_FragCoord;

in vec4 i;
layout (location = 0) out vec4 myColor;
const float eps=0.001;

void main() {
    myColor = vec4(0.2);
    if (gl_FragCoord.y >= 10) {
        myColor.b = 0.8;
    }
    if (gl_FragCoord.y == trunc(gl_FragCoord.y)) {
        myColor.g = 0.8;
    }
    if (gl_FragCoord.x == trunc(gl_FragCoord.x)) {
        myColor.r = 0.8;
    }

    vec4 diff = gl_FragCoord - i;
    if (abs(diff.z)>eps) myColor.b = 0.5;
    if (abs(diff.w)>eps) myColor.a = 0.5;
}
```

ZhiqianXia:
fix the issue at https://github.com/KhronosGroup/glslang/pull/2902

greg-lunarg:
Yes. Since the *first* redeclaration is before the use and all other redeclarations are identical (per the spec), this shader should be legal.

greg-lunarg:
Fixed by #2902

PR #2899, opened 2022-03-01
the first redeclarations of gl_FragCoord must appear before any use of gl_FragCoord.

ZhiqianXia:
fix the https://github.com/KhronosGroup/glslang/issues/2898.


ZhiqianXia:
replace with https://github.com/KhronosGroup/glslang/pull/2902.

PR #2904, opened 2022-03-10
Prevent Push Constant blocks being an array

sfricke-samsung:
The [WG agreed](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5080) that the following shader makes no sense

```glsl
layout(push_constant) uniform PushData {
  float a;
} data[2];

void main() {
  float x = data[1].a;
}
```

So I have a [spirv-val](https://github.com/KhronosGroup/SPIRV-Tools/pull/4742) patch ready as well when the VUID appears in a future release. The Vulkan spec change is just waiting on deciding how to best word this restriction

### DIFF
--- a/Test/baseResults/vulkan.frag.out
+++ b/Test/baseResults/vulkan.frag.out
@@ -25,37 +25,38 @@ ERROR: 0:43: 'non-opaque uniforms outside a block' : not allowed when using GLSL
 ERROR: 0:43: 'push_constant' : can only be used with a block 
 ERROR: 0:45: 'push_constant' : cannot declare a default, can only be used on a block 
 ERROR: 0:46: 'binding' : cannot be used with push_constant 
-ERROR: 0:54: 'binding' : sampler/texture/image requires layout(binding=X) 
-ERROR: 0:55: 'binding' : sampler/texture/image requires layout(binding=X) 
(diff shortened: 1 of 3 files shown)


greg-lunarg:
Thanks for the code! Please post here when the spec change goes public.

sfricke-samsung:
@greg-lunarg so the spec already states this in the [Push Constant Interface](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-resources-pushconst) section

> typed as OpTypeStruct,

as opposed to below for `Uniform` and `StorageBuffer` it says

> typed as OpTypeStruct, or an array of this type,

So the spec language is already there to justify this glslang change (since we agreed in last week's call this was not a spec bug), the internal spec fix is to add an explicit VUID in the [Standalone SPIR-V Validation](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv-module-validation-standalone) as that is the VUID used in `spirv-val`

sfricke-samsung:
@greg-lunarg the change for `spirv-val` landed https://github.com/KhronosGroup/SPIRV-Tools/pull/4742

I don't see value waiting for the spec change as it is just there to more clarify what is already stated as invalid

greg-lunarg (approved):
LGTM. Thanks for the code.

PR #2906, opened 2022-03-12
Make depfiles compatible with Windows paths

apanteleev:
Depfiles are basically Makefiles, with some special character rules. In particular, the `:` and `\` characters are special, and they will be present in any well-formed path on Windows.

This change adds character escaping to make depfiles work correctly on Windows. Verified with the Meson depfile parser, which can be found here: https://github.com/mesonbuild/meson/blob/master/mesonbuild/depfile.py 

In addition, the duplicate `it` variables were renamed for clarity.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -1168,6 +1168,27 @@ struct ShaderCompUnit {
     }
 };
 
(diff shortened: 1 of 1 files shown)


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2908, opened 2022-03-15
cmake: Install SPVRemapper runtime library

Biswa96:
This installs SPVRemapper.dll file which was missing previously.


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2916, opened 2022-03-26
fix crash in HLSL frontend

Try:
fix crash, when converting HLSL return of hull shader into spirv/glsl like arrayed output.

#fixed 2914

### DIFF
--- a/Test/baseResults/hlsl.emptystructreturn.tesc.out
+++ b/Test/baseResults/hlsl.emptystructreturn.tesc.out
@@ -1,0 +1,606 @@
+hlsl.emptystructreturn.tesc
+Shader version: 500
+vertices = 3
+vertex spacing = equal_spacing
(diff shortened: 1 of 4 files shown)


greg-lunarg:
I think this change is fine and we just need to fix the failing test hlsl.emptystructreturn.frag. This can be done by building glslangtests.exe and running it with --update-mode enabled. Then commit the new *.out to your branch.

Try:
Hi!
It seems patch requires more work: with slightly updated example, now it's crashes in `HlslParseContext::flatten`, when processing `gl_InvocationID` member of `stage_input`.

arcady-lunarg:
I've rebased your PR on main and made a small change, with which the tests appear to pass and the crash no longer reproduces on your example.

greg-lunarg (approved):
LGTM

Issue #2922, opened 2022-04-06
Can't build with mingw on windows


AndreyMlashkin:
originally reported on https://github.com/conan-io/conan-center-index/issues/10134

When compiling with mingw on windows, there is a compiler error occurs
[ 27%] Built target SPVRemapper
[ 28%] Building CXX object source_subfolder/glslang/CMakeFiles/MachineIndependent.dir/MachineIndependent/iomapper.cpp.obj
[ 30%] Building CXX object source_subfolder/glslang/CMakeFiles/MachineIndependent.dir/MachineIndependent/InfoSink.cpp.obj
[ 31%] Building CXX object source_subfolder/glslang/CMakeFiles/MachineIndependent.dir/MachineIndependent/Initialize.cpp.obj
internal error in mingw32_gt_pch_use_address, at config/i386/host-mingw32.c:192: MapViewOfFileEx: Attempt to access invalid address.

greg-lunarg:
I am sorry to hear about your build troubles, but it is beyond my charter to address this problem or guarantee compilation with all possible compilers. The current repo builds utilizing MSVS C++11, so the troubles you are seeing would seem to be with the compiler you are using. I would suggest filing an issue with that compiler.

SpaceIm:
@greg-lunarg It's fixed by setting `ENABLE_PCH` to `OFF` by default if `MINGW`.

greg-lunarg:
@SpaceIm Thanks for the fix!

Issue #2882, opened 2022-01-30
build  it for tvOS


azureplus:
[ 19%] Building CXX object glslang/CMakeFiles/MachineIndependent.dir/cmake_pch.hxx.pch
/var/folders/_l/20f4004537g0n0k2g6vwblqm0000gn/T/cmake_pch-2853c7.bc:1:1: error: expected top-level entity

make[3]: *** [glslang/CMakeFiles/MachineIndependent.dir/cmake_pch.hxx.pch] Error 1
make[2]: *** [glslang/CMakeFiles/MachineIndependent.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....

greg-lunarg:
I am not resourced to fix the build for this platform. If someone wishes to submit a fix, I would be willing to consider it.

azureplus:
> 

m1

SpaceIm:
Add `-DENABLE_PCH=0` while calling cmake. I believe that glslang should disable PCH by default for all compilers.

greg-lunarg:
@azureplus Please see the workaround or submit a PR disabling PCH similar to #2925.

@SpaceIm I don't completely disagree with your suggestion to disable PCH by default for all compilers, but I am not sure what the overall implications might be. For the moment, I am going to choose to disable selectively as problems arise. I do thank you for recognizing the problem, helping me to recognize the problem and submitting the PR.

greg-lunarg:
@azureplus I am asking you to submit the PR because you are in the best position to test if the fix works.

You will essentially need to add to the test modified by #2925. You will "OR" in the following to the test:

("${CMAKE_SYSTEM_NAME}" STREQUAL "tvOS")

If the fix does not work, let me know and I will re-open the issue.

PR #2925, opened 2022-04-07
CMake: fix MinGW build

SpaceIm:
Pre compiled headers lead to `internal error in mingw32_gt_pch_use_address, at config/i386/host-mingw32.c:192: MapViewOfFileEx: Attempt to access invalid address`.

closes https://github.com/KhronosGroup/glslang/issues/2922


SpaceIm:
As I said in https://github.com/KhronosGroup/glslang/issues/2882#issuecomment-1092162300, I believe that `ENABLE_PCH` should be `OFF` by default, and enabled if you want to improve incremental build time, because it's not porbable.

greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2928, opened 2022-04-17
Add macOS hidden files to gitignore

tellowkrinkle:
So that looking at the folder in Finder doesn't make git think there's uncommitted changes


greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2932, opened 2022-04-21
Fix WavePrefixCountBits() being off by one.

### LINKED ISSUE #2929: Using `WavePrefixCountBits` returns value off by 1
Hello,
I'm hitting this behavior where I have to offset by 1 the values I get from `WavePrefixCountBits` on my intel machine where I run my HLSL+Vulkan code - whereas on my DirectX codebase this isn't necessary.

From the documentation of [WavePrefixCountBits](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/waveprefixcountbytes), the current lane is NOT counted for the sum. It seems like `glslang` ignores this, so I had a look at the generated code.
Here I find this `%214 = OpGroupNonUniformBallotBitCount %uint %uint_3 InclusiveScan %213`, which sounds suspicious.
Looking at the compiler code, `ExclusiveScan` does indeed exist. Can you confirm that this is a bug?

Ryp:
It was counting bits up to the current lane included, whereas the
documentation says it should be excluded. This now matches dxc's behavior
as well.

Fix #2929

### DIFF
--- a/Test/baseResults/hlsl.waveprefix.comp.out
+++ b/Test/baseResults/hlsl.waveprefix.comp.out
@@ -1126,7 +1126,7 @@ local_size = (32, 16, 1)
 0:54              0 (const int)
 0:54          Constant:
 0:54            0 (const int)
-0:54        subgroupBallotInclusiveBitCount ( temp uint)
+0:54        subgroupBallotExclusiveBitCount ( temp uint)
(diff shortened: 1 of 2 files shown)


greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2933, opened 2022-04-27
Add whitelist filtering for debug comments in SPIRV-Remap.

AaronHaganAMD:
We are currently using glslang and spirv-remap for a project we would like to open source and as part of the compilation step, we need the ability to not strip out some "unused" functions in the SPIRV. We currently do this by passing in a whitelist file to the remapper. It would greatly simplify things if we could get this also added to the glslang codebase.

### DIFF
--- a/SPIRV/SPVRemapper.cpp
+++ b/SPIRV/SPVRemapper.cpp
@@ -160,15 +160,30 @@ namespace spv {
     }
 
     // Is this an opcode we should remove when using --strip?
-    bool spirvbin_t::isStripOp(spv::Op opCode) const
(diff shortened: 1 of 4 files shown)


greg-lunarg (approved):
LGTM.

mnetsch:
Note that this causes a regression in VK-GL-CTS when updating to latest glslang since the remap function signature is a breaking change with existing code that uses this function. It's trivial to fix, but heads up that others using this repo may see this issue as well. It might be worth exposing the original prototype as a wrapper or making the whitelist an optional arg after OPTS.

PR #2940, opened 2022-05-06
Disable layout error check for RT ops in presence of EXT_spirv_intrinsics

### LINKED ISSUE #2935: Raytracing Intrinsics do not work in presence of GL_EXT_spirv_intrinsics extension
Following legal shader fails to compile

#version 460
#extension GL_EXT_ray_tracing : enable
#extension GL_EXT_spirv_intrinsics: enable

layout(binding = 0) buffer block {
        vec2 op;
};
layout ( binding =  1) uniform accelerationStructureEXT as;
layout ( location = 1) rayPayloadEXT vec2 pay;

void main()
{
        traceRayEXT(as , 0xff , 0xff , 1U , 1, 1, vec3(0.5), 0.0, vec3(1.5),1.0f, 1);
        op += pay;

}

This is because in layoutTypeCheck : ParseHelper.cpp, all error checks are disabled when extension is enabled (which seems a bit too prohibitive in first place)

The issue is because raytracing intrinsics like traceRayEXT, executeCallableEXT have last argument has location id which is mapped to actual variable during spirv translation and this map is populated in above function

We need to move this code to some non error check function
I plan to have a fix

alelenv:
Fixes #2935

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -2327,7 +2327,7 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
             error(loc, "argument must be compile-time constant", "payload number", "a");
         else {
             unsigned int location = (*argp)[10]->getAsConstantUnion()->getAsConstantUnion()->getConstArray()[0].getUConst();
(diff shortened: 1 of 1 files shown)


alelenv:
@greg-lunarg : Can you review? I posted updated thoughts on the issue

greg-lunarg (approved):
LGTM.

PR #2943, opened 2022-05-17
fix structure indexing reassignment during block merging

mbechard:
For EOpIndexDirectStruct binaries, we want to visit the left symbol (the
structure) before we visit the binary, so it gets updated first.
That way we are comparing the updated structure against the target
'unitType', not the original structure.


greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2945, opened 2022-05-19
Remove unused variable

dneto0:


dneto0:
Fixes builds that have -Werror and -Wunused-variable

greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #2950, opened 2022-05-24
Generate OpLine Before OpFunction

qingyuanzNV:
This is a PR succeeding https://github.com/KhronosGroup/glslang/pull/2321 as the author of the original PR doesn't have time to finalize it. The goal is to emit OpLine before OpFunction so that downstream user of spir-v module could have a more accurate debug line information.

There's following design decisions might need to be discussed:
1. Due to different parsing behavior, for a function "void foo() { ... }", glsl sets the OpLine to the location of '(' while hlsl sets it to '{'
2. Sometimes, compiling hlsl will generate a dummy main function. Currently, the dummy function will have the same line info with the original main function. But this could be removed if wanted.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2783,6 +2783,10 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
             } else {
                 handleFunctionEntry(node);
             }
+            if (options.generateDebugInfo) {
+                const auto& loc = node->getLoc();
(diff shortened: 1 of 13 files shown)


qingyuanzNV:
> Generally looks good. Please fix build errors.

It seems spv::Instruction has a protected copy constructor. Just in case I break anything, do you know the intention of that? @greg-lunarg

qingyuanzNV:
Sorry to respond late. I have updated the change to use std::unique_ptr so the copy assignment of spv::Instruction is no longer used.

qingyuanzNV:
Have updated to get rid of std::make_unique. MSVC seems to have back ported that into C++11 so my compiler didn't catch that.

PR #2955, opened 2022-05-30
Avoid duplicate BuiltIn variables for ray tracing matrices (fix #2921)

### LINKED ISSUE #2921: Invalid SPIR-V produced when gl_ObjectToWorldEXT and gl_ObjectToWorld3x4EXT used in the same shader
Consider the following intersection shader:

```glsl
#version 460
#extension GL_EXT_ray_tracing : require

void main()
{
    mat4x3 foo = gl_ObjectToWorldEXT;
    mat3x4 bar = gl_ObjectToWorld3x4EXT;
}
```

There is no "3x4" version of the SPIR-V built-in, so glslang maps both of these built-ins to `ObjectToWorldKHR` and inserts a transpose.

However, if you use both of these in the same shader, you will get two variables decorated with the same `BuiltIn`:

```
                              Decorate 12(gl_ObjectToWorldEXT) BuiltIn ObjectToWorldKHR
                              Decorate 18(gl_ObjectToWorld3x4EXT) BuiltIn ObjectToWorldKHR
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeFloat 32
               7:             TypeVector 6(float) 3
               8:             TypeMatrix 7(fvec3) 4
               9:             TypePointer Function 8
              11:             TypePointer Input 8
12(gl_ObjectToWorldEXT):     11(ptr) Variable Input
              14:             TypeVector 6(float) 4
              15:             TypeMatrix 14(fvec4) 3
              16:             TypePointer Function 15
18(gl_ObjectToWorld3x4EXT):     11(ptr) Variable Input
               5:             Label
         10(foo):      9(ptr) Variable Function
         17(bar):     16(ptr) Variable Function
              13:           8 Load 12(gl_ObjectToWorldEXT)
                              Store 10(foo) 13
              19:           8 Load 18(gl_ObjectToWorld3x4EXT)
              20:          15 Transpose 19
```

This is not permitted by [the SPIR-V specification](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_builtin):

> For all the declarations of all the global variables and constants statically referenced by the entry-point’s call tree, within any specific storage class it is invalid to decorate with a specific BuiltIn more than once.

The same issue exists for the pair of `gl_WorldToObjectEXT` and `gl_WorldToObject3x4EXT`.

This issue is similar to https://github.com/KhronosGroup/glslang/pull/2910 but is not as easy to fix because the built-ins are not perfect aliases.

andfau-arm:
Fixes an issue where invalid SPIR-V was generated when gl_ObjectToWorldEXT and gl_ObjectToWorld3x4EXT, or gl_WorldToObjectEXT and gl_WorldToObject3x4EXT, were used in the same shader. The SPIR-V specification requires that there be at most one OpVariable decorated with a given BuiltIn value.


andfau-arm:
The CI failurse don't seem like they would be caused by my change.

andfau-arm:
I think the CI failure is because there isn't a hashing function for `spv::BuiltIn`. I've changed it to cast to `uint32_t` first, hopefully that fixes it.

PR #2962, opened 2022-06-06
Add OpSource support to the C interface.

James2022-rgb:
The C interface was lacking support for adding `OpSource` source code information instruction to the resulting SPIR-V.

This PR adds C interface counterparts of `TIntermediate::addSourceText`, `TIntermediate::setSourceFile` and `SpvOptions`.

I opted not to change the signature of `glslang_program_SPIRV_generate` so as not to break existing code,
instead adding a variant `glslang_program_SPIRV_generate_with_options` which additionally takes the `SpvOptions` counterpart,
and restructured `glslang_program_SPIRV_generate` to be implemented in terms of the latter to avoid code duplication.


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #2979, opened 2022-07-25
Fix #2658: Properly include all headers in deployments

### LINKED ISSUE #2658: The builds are broken
I tried to use the automatic CI builds, but they are broken on both Linux and Mac:
- The include folder is empty
- In debug builds the static lib binaries are not included at all.

Windows on the other hand, does include headers and binaries in both Debug and Release.

Could you please provide proper ZIP packages for Debug and Release configurations under both Linux and Mac?

spnda:
The workflow used to only zip files in `include/glslang/*`, which does not respect subdirectories. Also including `include/glslang/**/*` fixes this issue. If you look at the current builds, you can precisely observe this behaviour as `build_info.h` in the root directory is actually zipped. The `${SUFFIX}` seems to also have been blindly copied from the old CI which has no usage anymore. This workflow never worked. I fully tested this on Ubuntu 20.04 in WSL and on my macOS 13 Mac.

Also, I've rewritten the GitHub workflow to also build for Windows, but is there any reason why VS 2015 is strictly *required*? GitHub workflow only offers VS Enterprise 2017, VS Enterprise 2019, or VS Enterprise 2022.

Fixes #2658.

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -77,17 +77,18 @@ jobs:
                   zip ${ARCHIVE} \
                       bin/glslangValidator \
                       include/glslang/* \
-                      lib/libGenericCodeGen${SUFFIX}.a \
-                      lib/libglslang${SUFFIX}.a \
-                      lib/libglslang-default-resource-limits${SUFFIX}.a \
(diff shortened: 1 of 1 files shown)


jeremy-lunarg (approved):
LGTM. Thanks for the fix!

PR #2980, opened 2022-07-25
Fix: Migrate Windows CI to GitHub Actions

spnda:
This removes the need for AppVeyor in this repository. This workflow however runs on windows-2016, meaning that by default these are being built with Visual Studio Enterprise 2017. If that's not acceptable, this can wait until VS 2015 is deprecated by you. Also includes some nice things that weren't ported in the original migration.

Also, note that the AppVeyor auth token was always public in the .appveyor.yml config :) https://github.com/KhronosGroup/glslang/blob/607771c3622dc5c01e2e2ec088324383db1964fc/.appveyor.yml#L99

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -21,6 +21,17 @@ on:
     push:
         branches:
             - main
+        paths-ignore:
+            - 'README.md'
+            - 'README-spirv-remap.txt'
+            - 'LICENSE.txt'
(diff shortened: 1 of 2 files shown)


jeremy-lunarg (approved):
This is good to go now. I tested in my fork. I had to fix up some problems with the windows configuration. I removed the deletion of the appveyor file. I will follow up with another patch to remove that after this safely lands on main.

arcady-lunarg (approved):
Lgtm

PR #2985, opened 2022-07-27
Implement NonSemantic.Shader.DebugInfo.100

jeremy-lunarg:
See https://github.com/KhronosGroup/SPIRV-Registry.


greg-lunarg (approved):
LGTM.

PR #2987, opened 2022-07-29
Avoid double-free in functions cloned for vulkan relaxed mode

dneto0:
When rewriting function calls atomicCounterIncrement and
atoicCounterDecrement, clone the parameters so that the TParameter
'type' field is cloned.  This avoids double-free when both the original
and transformed functions are deleted by the parser.

Fixes a ubsan failure.


greg-lunarg (commented):
Generally a good change.

There is one previous use of copyParam in SymbolTable.cpp. Should we stick a (void) before that call to signal we are not using the return value there?

dneto0:
> There is one previous use of copyParam in SymbolTable.cpp. Should we stick a (void) before that call to signal we are not using the return value there?

I've added a commit which does that.

PR #2988, opened 2022-07-30
Fix: Build arm64 binaries for macOS

spnda:
x86_64 binaries cannot be linked when building solely for arm64. This uses the matrix to build separately for x86_64 and arm64. I specifically opted for this instead of building a universal binary as there seems to be no guarantee that universal binary 2 files can be read and understood by older linkers, and the fact that a universal binary will effectively have two copies of the library/executable wasting a considerable amount of space. Also, upgrading the runner to `macOS-11` in 457d11ebd86c089f00a934e579f9c4dc76921e71 was required for this to work.

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -106,6 +106,7 @@ jobs:
             matrix:
                 os: [{genus: macos-11, family: osx}]
                 compiler: [{cc: clang, cxx: clang++}]
+                arch: [x86_64, arm64]
                 cmake_build_type: [Debug, Release]
         steps:
             - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


greg-lunarg (approved):
LGTM. Thanks for the code!

greg-lunarg:
CI tests failed, The glslang-testsuite failed with multiple such failures:

/Users/runner/work/glslang/glslang/build/StandAlone/glslangValidator: Bad CPU type in executable

The glslang-testsuites test failed with BAD_COMMAND

I will have to revert until we have fixes for this.

spnda:
Wait before reverting. This PR makes it build arm64 binaries. However, the macos-11 and macos-12 GH runners are running on x86_64. They simply cannot run arm64 binaries. The only option currently would be to either drop tests when running arm64 (easily done by an if condition in the yaml), or wait until GitHub provides arm64-based runners. I, personally, don't think dropping tests in this case is bad, as they still run and do their job on ubuntu, macos (x86_64), and windows and I don't think there's any platform exclusive code in the glslang source.

greg-lunarg:
@spnda I am going to revert until I have come to a decision about how to handle this situation.

PR #2992, opened 2022-08-03
Revert "Revert "Fix: Build arm64 binaries for macOS""

greg-lunarg:
Reverts KhronosGroup/glslang#2991

Placeholder to reapply #2988. See this previous PR for current status.

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -106,6 +106,7 @@ jobs:
             matrix:
                 os: [{genus: macos-11, family: osx}]
                 compiler: [{cc: clang, cxx: clang++}]
+                arch: [x86_64, arm64]
                 cmake_build_type: [Debug, Release]
         steps:
             - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


spnda:
Interesting to double-revert a merge. However, I just noticed something: These are continuous deployments, not the continuous integrations. Why are the tests being run at all? The workflows defined in `continuous_integration.yml` should define everything necessary to test PRs.

greg-lunarg:
I am just polling a few more folks to decide if it is reasonable to release something that has not been tested. @spnda Can you give any other examples where this is done?

jeremy-lunarg:
I think we should wait for runner support (https://github.com/github/roadmap/issues/528). It sounds like it might not be that far off.

@spnda We retest in continuous deployment to verify that a merge didn't have unexpected results.

oscarbg:
Any update?

arcady-lunarg:
@oscarbg A MacOS arm64 build of glslang is available in the Vulkan SDK. Is there any specific reason you are interested in having this configuration tested in CI?

oscarbg:
@arcady-lunarg not really.. misunderstood this..

PR #2994, opened 2022-08-04
Capture left/rightStorage by value

aeubanks:
Capturing a bitfield by reference is not allowed, and in clang this is
enforced as of https://reviews.llvm.org/D122768.


aeubanks:
actually, let me double check that this isn't a clang bug

aeubanks:
clang bug, apologies for the noise

PR #2997, opened 2022-08-11
Remove the unused `OS_CleanupThreadData`

RandomShaper:
I was building glslang for a platform that doesn't support `PTHREAD_CANCEL_ENABLE` and during the investigation about what it was being used for I realized that its only use belongs in an unused function — `OS_CleanupThreadData`.

### DIFF
--- a/glslang/OSDependent/Unix/ossource.cpp
+++ b/glslang/OSDependent/Unix/ossource.cpp
@@ -57,52 +57,6 @@ namespace glslang {
 //
 
 //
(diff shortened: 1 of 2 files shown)


greg-lunarg (changes requested):
You will also need to remove DetachThreadLinux() in Unix/ossource.cpp. That is the reason for the one CI failure.

Please also remove the Thread cleanup comment above it.

RandomShaper:
Done.

jeremy-lunarg (approved):
LGTM. I don't think these functions are publicly supported.

greg-lunarg (approved):

PR #3011, opened 2022-08-26
Fix strict aliasing violation

hbatagelo:
Dereferencing a type-punned pointer breaks type aliasing rules, leading to undefined behavior as in this statement:
```
unsigned literal = *reinterpret_cast<unsigned*>(&floatValue);
```
This can be fixed with `std::bit_cast`, but I use `memcpy` for compatibility with C++11 (and `static_assert` for safety):
```
unsigned literal;
static_assert(sizeof(literal) == sizeof(floatValue), "sizeof(unsigned) != sizeof(float)");
memcpy(&literal, &floatValue, sizeof(literal));
```


jeremy-lunarg (approved):
LGTM

greg-lunarg (approved):
LGTM

PR #3012, opened 2022-08-29
Fix -Wundef warnings for MINGW_HAS_SECURE_API

JohannesKauffmann:
Since 12e27e17deb3102f1f6f08ec19caf5e32becb3f8, it was assumed that the
MINGW_HAS_SECURE_API macro was unconditionally defined. This is not
always the case, and GCC produces -Wundef warnings when that happens.

Fix this, by first checking if MINGW_HAS_SECURE_API is defined, and if
so, also check that it does not evaluate to zero.

As a drive-by, place parentheses around _MSC_VER for consistency.

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -66,7 +66,7 @@ std::string to_string(const T& val) {
 }
 #endif
 
-#if (defined(_MSC_VER) && _MSC_VER < 1900 /*vs2015*/) || MINGW_HAS_SECURE_API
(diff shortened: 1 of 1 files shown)


JohannesKauffmann:
Update: added braces around the whole MinGW define statement, after the `||`. This is done to prevent ambiguity because of the `||` and `&&` used together without any braces.

jeremy-lunarg (approved):
LGTM

PR #3015, opened 2022-09-01
Fix: Remove NV suffix from glslang C interface

spnda:
This removes the NV suffixes from the glslang stage and stage mask enum from the C interface. I've aliased them properly to maintain backwards compatibility.


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #3021, opened 2022-09-07
Fix OpLine prepending OpFunction reports wrong file when #line is pre…

qingyuanzNV:
This commits intend to fix the file name in OpLine prepending the OpFunction is incorrect when preprocessors changed the active file:
```
#line 123 "filename.h"
```


greg-lunarg (approved):
LGTM. Thanks for the fix!

Issue #1540, opened 2018-10-21
Setting CMAKE_DEBUG_POSTFIX to "d" breaks all CMake projects included after glslang


DiligentGraphics:
glslang's main CMake file sets global debug postfix to "d":
`set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Adds a postfix for debug-built libraries.")`
This affects every single project processed after glslang in an unexpected and undesired way. If glslang wants to use this postfix, the variable should be local to glslang's source subtree and should not be global and visible to other projects.

madwax:
I've been hit by this same problem in the past.  I ended up editing the CMakefiles but the proper way should be I think is add set_target_properties(glslang PROPERTIES DEBUG_POSTFIX "d") to glslang/CMakeLists.txt

johnkslang:
PR #1676 removes it from the cache, which seems sufficient. Sorry for the delay.

PR #3022, opened 2022-09-08
Don't force CMAKE_DEBUG_POSTFIX if forced by the user

robUx4:
This allows building in debug mode and use the library as if it was a release build.

(the line at the end was added by the GitHub editor)


jeremy-lunarg (changes requested):
We used to cache this variable, but it caused another [issue](https://github.com/KhronosGroup/glslang/issues/1540).

robUx4:
Right now I have the opposite issue. I set `CMAKE_DEBUG_POSTFIX` globally but glslang sets it back to its own. The other way would be to wrap around a `if(NOT DEFINED CMAKE_DEBUG_POSTFIX)` block.

jeremy-lunarg:
One suggestion was to use `set_target_properties`. Would that work?

jeremy-lunarg:
Closing this for now. Please feel to reopen if you find a better solution. Thanks.

PR #3027, opened 2022-09-13
cmake: Use common installation directory for cmake support files.

rhabacker:
At least on openSUSE, the usual installation location for cmake support files is ${CMAKE_INSTALL_LIBDIR}/cmake/<name>

   $ find /usr/lib64 -name '*.cmake' | grep /cmake/ | wc -l
   834
   $ find /usr/lib64 -name '*.cmake' | grep -v /cmake/ | wc -l
   8

which is also used by glslang with these changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,13 +375,13 @@ if(ENABLE_GLSLANG_INSTALL)
         include("@PACKAGE_PATH_EXPORT_TARGETS@")
     ]=])
     
-    set(PATH_EXPORT_TARGETS "${CMAKE_INSTALL_LIBDIR}/${PROJECT_NAME}/glslang-targets.cmake")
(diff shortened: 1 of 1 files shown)


jeremy-lunarg:
@Tachi107 @theblackunknown 
Do you have any feedback on this change?

theblackunknown:
I don't know the specific of Unix system lib path conventions so I will let @Tachi107 and @rhabacker decide on this. 
For me as long as this results in a installation path which is portable - which seems to be - I am ok with this change.

Tachi107:
Yes, CMake files are usually installed in lib/cmake/projname

jeremy-lunarg (approved):
LGTM. Thanks for all the patches here @rhabacker @Tachi107 and @theblackunknown.

PR #3028, opened 2022-09-13
cmake: Do not install static libraries in shared build mode

rhabacker:
The glslang cmake build system installs static libraries in addition to the dynamic glslang library, which are required when used in a package that uses glslang when calling find_package().

Distributions such as openSUSE (and perhaps others) use a "shared only" strategy, which conflicts with the current state of the glslang build system.
The static libraries mentioned are already all included in glslang and thus not needed. With this commit, these will no longer install the associated cmake support files when glslang is built shared.

See https://build.opensuse.org/request/show/1001407 for a related issue description.

### DIFF
--- a/OGLCompilersDLL/CMakeLists.txt
+++ b/OGLCompilersDLL/CMakeLists.txt
@@ -41,7 +41,7 @@ if(WIN32)
     source_group("Source" FILES ${SOURCES})
 endif(WIN32)
 
-if(ENABLE_GLSLANG_INSTALL)
+if(ENABLE_GLSLANG_INSTALL AND NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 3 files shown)


jeremy-lunarg (approved):
LGTM. Thanks for the patch!

jengelh:
This got broken again by 6f22e41e0de13aff75807cca9662730d8b3e54b6

rhabacker:
> This got broken again by [6f22e41](https://github.com/KhronosGroup/glslang/commit/6f22e41e0de13aff75807cca9662730d8b3e54b6)

You need to add https://github.com/KhronosGroup/glslang/pull/3317/commits/3c0a08992d60aac782c769a7d6838f2bd206a903 at https://build.opensuse.org/package/show/X11:Wayland/glslang to solve the problem.

PR #3032, opened 2022-09-19
Improve naming in link validation

mbechard:
since we are changing the type before the visitBinary now, we need to be comparing against the new type.
Previous test cases worked since the 'unitType' was a shallow copy and ended up getting updated in some cases to match the new type, but not all.


greg-lunarg (approved):
LGTM. Thanks for the code!

PR #3034, opened 2022-09-23
Use DebugDeclare for local variables

jeremy-lunarg:
Previously we had decided to issue DebugValue directly in glslang. However, this was incorrect and causing issues with RenderDoc.


greg-lunarg (approved):
LGTM.

Issue #3019, opened 2022-09-05
Regression in location assignment for barycentrics.


HansKristian-Work:
After merging mesh shaders, there's a regression in barycentrics support:

```
#version 450
#extension GL_EXT_fragment_shader_barycentric : require

layout(location = 0) out vec2 value;
layout(location = 0) pervertexEXT in vec2 vUV[3];
layout(location = 1) pervertexEXT in vec2 vUV2[3];
      
void main () {
    value = gl_BaryCoordEXT.x * vUV[0] + gl_BaryCoordEXT.y * vUV[1] + gl_BaryCoordEXT.z * vUV[2];
    value += gl_BaryCoordNoPerspEXT.x * vUV2[0] + gl_BaryCoordNoPerspEXT.y * vUV2[1] + gl_BaryCoordNoPerspEXT.z * vUV2[2];
}
```

now reports with `glslangValidator -H -V -o /tmp/test.spv /tmp/barycentric-khr.frag`.

```
/tmp/barycentric-khr.frag
ERROR: /tmp/barycentric-khr.frag:6: 'location' : overlapping use of location 1
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
```

`pervertexEXT` should only consume one location here.

greg-lunarg:
@pmistryNV Can you please fix? Can you also add a test to the glslang regression tests which covers this? There are two for GL_NV_fragment_shader_barycentric but none for GL_EXT_fragment_shader_barycentric.

pmistryNV:
Sure I can do that

pmistryNV:
Submitted https://github.com/KhronosGroup/glslang/pull/3036 to fix the bug

jeremy-lunarg:
Fixed by #3036.

PR #3036, opened 2022-09-26
Fix for bug #3019 : That incorrectly removed perVertexEXT qualifier as…

pmistryNV:
… arrayed IO

This bug got introduced as part of EXT_mesh_shader changes 228c672 Also updated barycentric test cases to add coverage for multiple pervertexEXT array inputs


pmistryNV:
@greg-lunarg  please review

jeremy-lunarg (approved):
LGTM! Thanks for the patch.

PR #3037, opened 2022-09-28
only use dead input elimination on vertex shaders

mbechard:
It can't be safely used unilaterally on other stages, since that will result in output/input mismatches between stages. They arn't having their output variables eliminated accordingly.


greg-lunarg (approved):
Correct. Thanks for the fix!

PR #3043, opened 2022-10-04
Change Volatile generation for HelperInvocation

### LINKED ISSUE #3042: Invalid SPIR-V generated for gl_HelperInvocation under Vulkan memory model
When using the demote_to_helper functionality under SPIR-V 1.6, glslang decorates the `gl_HelperInvocation` with `Volatile`, but this decoration is not permitted when using the Vulkan memory model. The load should use the `Volatile` memory operand instead.

For example, compiling the following shader:
```
#version 310 es

#pragma use_vulkan_memory_model

#extension GL_EXT_demote_to_helper_invocation : require

precision highp float;

layout (set=0, binding=0) buffer B {
   float o;
};

void main() {
   demote;
   o = gl_HelperInvocation ? 1.0 : 0.0;
}
```
with the command-line `glslang -V --target-env vulkan1.3 test.frag` then running `spirv-val` on the output gives the error
```
error: line 32: Volatile decoration targeting 14[%gl_HelperInvocation] is banned when using the Vulkan memory model.
  %gl_HelperInvocation = OpVariable %_ptr_Input_bool Input
```

gnl21:
For SPIR-V 1.6 HelperInvocation accesses need to be volatile to avoid undefined values when shaders execute 'demote'. Previously this was always decorated on the gl_HelperInvocation variable, but this is not valid when the Vulkan memory model is in use.

When the memory model is enabled, stop decorating the variable declaration and apply the memory semantic to access chain loads instead.

Fixes #3042


gnl21:
@greg-lunarg Although this implementation works for me, I'm not sure that adding the flag where I have is really the correct place. Can you offer any guidance?

gnl21:
Tests added and rebased to pick up the recent change to generator number.

greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #3046, opened 2022-10-11
Fix incorrect parse message of mesh shader

amdrexu:
When GL_EXT_mesh_shader is enabled, the check of layout qualifiers 'max_vertices' and 'max_primitives' should use
gl_MaxMeshOutputVerticesEXT and gl_MaxMeshOutputPrimitivesEXT.


amdrexu:
Ping for review. Thanks.

jeremy-lunarg (approved):
LGTM. Thanks for the fix!

PR #3047, opened 2022-10-11
Convert spirv.hpp line endings to LF (Unix style)

akien-mga:
This was the only header using CRLF, the rest already uses LF.


greg-lunarg (approved):
LGTM. Thanks for the fix!

PR #3050, opened 2022-10-16
mesh shader: fix implicit index-array size calculation for lines and triangles

### LINKED ISSUE #3041: PrimitiveTriangleIndicesEXT array size not matching OutputPrimitivesEXT
@pmistryNV for the VUID

> VUID-PrimitiveTriangleIndicesEXT-PrimitiveTriangleIndicesEXT-07058
The size of the array decorated with PrimitiveTriangleIndicesEXT must match the value specified by OutputPrimitivesEXT

If I have a GLSL shader such as 

```glsl
#version 460
#extension GL_EXT_mesh_shader : enable
layout(max_vertices = 1) out;
layout(max_primitives = 2) out;
layout(triangles) out;
void main()
{
    gl_PrimitiveTriangleIndicesEXT[0] = uvec3(1, 1, 1);
}
```

I get 

```swift
OpExecutionMode %main OutputPrimitivesEXT 2
OpDecorate %var BuiltIn PrimitiveTriangleIndicesEXT
%uint_6 = OpConstant %uint 6
%array = OpTypeArray %v3uint %uint_6
%ptr = OpTypePointer Output %array
%var = OpVariable %ptr Output
```

So not sure if this is a Spec ambiguity or Glslang bug, but for `array size` I would expect `OpTypeArray` to be only `2` and not `6`

Try:
Apparently compiler behavior wasn't updated to match EXT and still behaves like NV-version of mesh-shader. 

#fixed #3041


Try:
I've run `--update-mode` on a failing test-cases. Yet, not 100% sure that this will work. Something is not quite right with my testing setup: locally I have way more test failures, mostly in unrelated cases

jeremy-lunarg (approved):
LGTM. Thanks for the fix!

PR #3055, opened 2022-11-01
Improve ResourceLimits interface to be more forward compatible

### LINKED ISSUE #2822: Please move resource_limits_c.h from private to public header paths
The Vulkan SDK is now shipping with the static library libglslang-default-resource-limits.a 

To use this library, the user currently has to manually bring in the prototypes from:
/Glslang/StandAlone/resource_limits_c.h

This needs to be relocated to make it public in:
/Glslang/glslang/Include

greg-lunarg:
New interface allows users to generate ResourceLimits for interface so that additions to TBuiltInResource do not break the ABI.

Users should use the glslang-default-resource-limits library and the Public/ResourceLimits.h header. Similar changes have been made to the C interface. Use Public/resource_limits_c.h.

Fixes #2822

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -210,7 +210,7 @@ cc_library(
 cc_library(
     name = "glslang-default-resource-limits",
     srcs = ["StandAlone/ResourceLimits.cpp"],
-    hdrs = ["StandAlone/ResourceLimits.h"],
+    hdrs = ["glslang/Public/ResourceLimits.h"],
     copts = COMMON_COPTS,
     linkstatic = 1,
     deps = [":glslang"],
--- a/StandAlone/ResourceLimits.cpp
+++ b/StandAlone/ResourceLimits.cpp
@@ -37,9 +37,9 @@
 #include <sstream>
 #include <cctype>
 
-#include "ResourceLimits.h"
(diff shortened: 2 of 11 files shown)


jeremy-lunarg (approved):
LGTM

moritz-h:
@greg-lunarg Thanks for that change, having the default resources in the public interface is great. But maybe as follow up feature, it would be very useful to have this decoupling from the StandAlone library also for the implementation and CMake targets.

With this PR the headers are made public and added to the glslang library CMake target, but the implementation is still in its own target `glslang-default-resource-limits`. When using glslang as library and integrating into an other application you must use the `glslang` target for the include paths, and also to the `glslang-default-resource-limits` to link to the actual implementation. Additionally, the `glslang-default-resource-limits` target is only available when enabling `ENABLE_GLSLANG_BINARIES`, which then unnecessarily builds the stand alone executable and also brings the python depedency, even when just the default resources implementation is needed.
Maybe it would be useful to either move the default resources implementation to the `glslang` target that the implementation is at the same target as the now public headers are, or as alternative to move the `glslang-default-resource-limits` target to its own directory together with its public headers and provide it as independant library.

PR #3063, opened 2022-11-14
Fix locations of cmake files in side compat shims

svenstaro:
In https://github.com/KhronosGroup/glslang/pull/3027, the installed cmake files were stuck into a /cmake subdir but this isn't reflected in these compatibility shims. The shims were added in https://github.com/KhronosGroup/glslang/pull/2989.


jeremy-lunarg (approved):
LGTM. Thanks for fixing this.

PR #3066, opened 2022-11-15
[glslang][extension] Add support for ARB_bindless_texture.

ShchchowAMD:
Add support for ARB_bindless_texture.

There's still some features related to non_uniform. Would update soon.


greg-lunarg:
Can you please add tests for this functionality?

This can be done by adding shaders that use ARB_bindless_texture to the Test/ directory. You will also need to add the files to AST.FromFile.cpp. See switch.frag as an example.

You will also need to create result files in Test/baseResults/, but this can be done by building and running build/gtests/Release/glslangtests.exe --update-mode.

ShchchowAMD:
> 

Thanks for reminding. I'll update test and out file here.

glslang/MachineIndependent/Versions.cpp
```diff
@@ -227,6 +227,7 @@ void TParseVersions::initializeExtensionBehavior()
     extensionBehavior[E_GL_ARB_texture_query_lod]            = EBhDisable;
     extensionBehavior[E_GL_ARB_vertex_attrib_64bit]          = EBhDisable;
     extensionBehavior[E_GL_ARB_draw_instanced]               = EBhDisable;
+    extensionBehavior[E_GL_ARB_bindless_texture]          = EBhDisable;
```
greg-lunarg, line 230: Indentation

glslang/MachineIndependent/Versions.h
```diff
@@ -163,6 +163,7 @@ const char* const E_GL_ARB_texture_query_lod            = "GL_ARB_texture_query_
 const char* const E_GL_ARB_vertex_attrib_64bit          = "GL_ARB_vertex_attrib_64bit";
 const char* const E_GL_ARB_draw_instanced               = "GL_ARB_draw_instanced";
 const char* const E_GL_ARB_fragment_coord_conventions   = "GL_ARB_fragment_coord_conventions";
+const char* const E_GL_ARB_bindless_texture          = "GL_ARB_bindless_texture";
```
greg-lunarg, line 166: Indentation

greg-lunarg (changes requested):
Thanks for the test. Does your code support the full extension? If so, could we get tests for the other salient shader changes? Eg, from the extension spec:

`In particular, sampler and image types may be used as shader inputs/outputs, temporary variables, and uniform block members, and may be assigned to by shader code.`

Also, could we get an error message when this extension is used when targeting SPIR-V. For example, when compiled with this command line:

`glslangValidator.exe -V --target-env vulkan1.2 --amb --aml -o t.frag.spv t.frag`

The error would just say that this extension is not currently supported when targeting SPIR-V.

ShchchowAMD:
Thanks, I'll make another PR today.

glslang/MachineIndependent/Versions.cpp
ShchchowAMD: Fixed.

glslang/MachineIndependent/Versions.h
ShchchowAMD: Fixed.

ShchchowAMD:
> Thanks for the test. Does your code support the full extension? If so, could we get tests for the other salient shader changes? Eg, from the extension spec:
> 
> `In particular, sampler and image types may be used as shader inputs/outputs, temporary variables, and uniform block members, and may be assigned to by shader code.`
> 
> Also, could we get an error message when this extension is used when targeting SPIR-V. For example, when compiled with this command line:
> 
> `glslangValidator.exe -V --target-env vulkan1.2 --amb --aml -o t.frag.spv t.frag`
> 
> The error would just say that this extension is not currently supported when targeting SPIR-V.

Hi greg, thanks for reply. I've made another PR according to your reviews.

For tip2, I don't make sure whether I understand right, so just double check:

SPIR-V already has disassembled extension keyword:

`OpSourceExtension "GL_ARB_bindless_texture"`

So what we need to do is to disable usage of extension when generate SPV targeting vulkan?

Current glslang have checks to require spv version and vulkan, I add another check to allow some extensions reserved for vk in the latest PR.

Thanks for your review again.

greg-lunarg:
Thank you for the error check, although I would like one additional small change :)

Can you please disallow for ALL SPIR-V, generating for OpenGL as well as Vulkan?

Specifically, could you change the name of the variable vkReservedExt to spvUnsupportedExt?

Also check instead spvVersion.spv and change Vulkan to SPIR-V in the error message?

ShchchowAMD:
> 

Thanks for detailed suggestion. I've disabled it for spv and changed codes according to your comment.

greg-lunarg (approved):
LGTM. Thanks!

greg-lunarg (approved):
LGTM. Thanks!

PR #3067, opened 2022-11-19
Use CMAKE_INSTALL_FULL_LIBDIR in compat cmake files

chuangzhu:
According to
https://cmake.org/cmake/help/v3.25/module/GNUInstallDirs.html, CMAKE_INSTALL_LIBDIR can be an absolute path. For instance, Nixpkgs [defined it to an absolute path in /nix/store](https://github.com/NixOS/nixpkgs/blob/3d17b4c305cefef284109fa9d426b00f3e5072c6/pkgs/development/tools/build-managers/cmake/setup-hook.sh#L101). The output in this case is:

	# result-glslang/lib/cmake/glslangTargets.cmake:5
	include("${CMAKE_CURRENT_LIST_DIR}/../..//nix/store/3mif2zibig0cilk5dbz334278n0vlq9s-glslang-1.3.231.0/lib/glslang/glslang-targets.cmake")

Signed-off-by: Chuang Zhu <git@chuang.cz>


arcady-lunarg (approved):
LGTM, thanks for the PR.

PR #3071, opened 2022-11-23
Include: PoolAlloc: do not rely on CMake define

JohannesKauffmann:
On Windows, _DEBUG is defined by CMake if CMAKE_BUILD_TYPE is Debug. But on other platforms, this is not the case and thus in debug mode, the guard checks are not enabled.

Instead, rely on the NDEBUG define, which is always defined in release mode (Release, RelWithDebInfo and MinSizeRel). This works reliably on all platforms: It is also used to enable or disable assertions.


jeremy-lunarg (commented):
I believe this implies that this functionality was _not_ widely used outside of VS.

PR #3074, opened 2022-11-25
[glslang] Refine implicit array size interfaces.

ShchchowAMD:
Help to check builtin and other variables if across stages.

### DIFF
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -757,7 +757,7 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
                 }
                 else if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isSizedArray()) {
                     if (symbol->getWritableType().getImplicitArraySize() > unitSymbol->getType().getOuterArraySize())
-                        error(infoSink, "Implicit size of unsized array aren't match same symbol among multiple shaders.");
+                        error(infoSink, "Implicit size of unsized array doesn't match same symbol among multiple shaders.");
                 }
                 else if (unitSymbol->getType().isImplicitlySizedArray() && symbol->getWritableType().isSizedArray()) {
                     if (unitSymbol->getType().getImplicitArraySize() > symbol->getWritableType().getOuterArraySize())
--- a/glslang/MachineIndependent/linkValidate.cpp
+++ b/glslang/MachineIndependent/linkValidate.cpp
@@ -750,6 +750,21 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
                 }
 
                 // Update implicit array sizes
+                if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isImplicitlySizedArray()) {
+                    if (unitSymbol->getType().getImplicitArraySize() > symbol->getType().getImplicitArraySize()){
+                        symbol->getWritableType().updateImplicitArraySize(unitSymbol->getType().getImplicitArraySize());
+                    }
+                }
+                else if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isSizedArray()) {
+                    if (symbol->getWritableType().getImplicitArraySize() > unitSymbol->getType().getOuterArraySize())
+                        error(infoSink, "Implicit size of unsized array aren't match same symbol among multiple shaders.");
+                }
+                else if (unitSymbol->getType().isImplicitlySizedArray() && symbol->getWritableType().isSizedArray()) {
+                    if (unitSymbol->getType().getImplicitArraySize() > symbol->getWritableType().getOuterArraySize())
+                        error(infoSink, "Implicit size of unsized array doesn't match same symbol among multiple shaders.");
+                }
+
+                // Update implicit array sizes
                 mergeImplicitArraySizes(symbol->getWritableType(), unitSymbol->getType());
 
                 // Check for consistent types/qualification/initializers etc.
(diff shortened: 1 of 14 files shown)


greg-lunarg:
Can you please add tests that demonstrate the new functionality? Similar to #3066, although I think you will instead register in Link.FromFile.cpp.

ShchchowAMD:
Thanks for reminding. I'll add test for this one and also future PRs.

ShchchowAMD:
> Can you please add tests that demonstrate the new functionality? Similar to #3066, although I think you will instead register in Link.FromFile.cpp.

Add tests for link time error check.

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -744,9 +744,19 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
...
+                        symbol->getWritableType().updateImplicitArraySize(unitSymbol->getType().getImplicitArraySize());
+                    }
+                }
+                else if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isSizedArray()) {
+                    if (symbol->getWritableType().getImplicitArraySize() > unitSymbol->getType().getOuterArraySize())
+                        error(infoSink, "Implicit size of unsized array aren't match same symbol among multiple shaders.");
+                }
+                else if (unitSymbol->getType().isImplicitlySizedArray() && symbol->getWritableType().isSizedArray()) {
+                    if (unitSymbol->getType().getImplicitArraySize() > symbol->getWritableType().getOuterArraySize())
+                        error(infoSink, "Implicit size of unsized array aren't match same symbol among multiple shaders.");
```
greg-lunarg, line 759: "aren't" -> "doesn't"

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -744,9 +744,19 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
                 if (! symbol->getQualifier().hasBinding() && unitSymbol->getQualifier().hasBinding())
                     symbol->getQualifier().layoutBinding = unitSymbol->getQualifier().layoutBinding;
 
-                // Similarly for location
-                if (!symbol->getQualifier().hasLocation() && unitSymbol->getQualifier().hasLocation()) {
-                    symbol->getQualifier().layoutLocation = unitSymbol->getQualifier().layoutLocation;
```
greg-lunarg, line 749: Did you mean to remove this? If so, why?

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -759,6 +769,19 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
...
+        // Check conflicts between preset primitives and sizes of I/O variables among multiple geometry shaders
+        if (language == EShLangGeometry && unitStage == EShLangGeometry)
+        {
+            TIntermSymbol* unitSymbol = unitLinkerObjects[unitLinkObj]->getAsSymbolNode();
+            if (unitSymbol->isArray() && unitSymbol->getQualifier().storage == EvqVaryingIn && unitSymbol->getQualifier().builtIn == EbvNone)
+                if ((unitSymbol->getArraySizes()->isImplicitlySized() &&
+                        unitSymbol->getArraySizes()->getImplicitSize() != TQualifier::mapGeometryToSize(getInputPrimitive())) ||
+                    (! unitSymbol->getArraySizes()->isImplicitlySized() &&
+                        unitSymbol->getArraySizes()->getDimSize(0) != TQualifier::mapGeometryToSize(getInputPrimitive())))
+                    error(infoSink, "Not all provided sizes match across all geometry shaders in the program");
```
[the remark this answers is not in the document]
ShchchowAMD: Yes, I'll fix it.

glslang/MachineIndependent/linkValidate.cpp
ShchchowAMD: Thanks, will fix in next commit.

glslang/MachineIndependent/linkValidate.cpp
ShchchowAMD: Sorry, this is a mistake via merging codes, I'll add this block back.

ShchchowAMD:
Add a fix for corner case:
when a builtin variable is used across stages, set initialized implicit size to 0.

If stage A uses it by implicitly indexing but not in stage B, it should not report an error.

This is especially for vars in gl_PerVertex and other built-in blocks available across stages.

The main idea is to use implicitSize == 0 to determine whether a declared unsized array has been visited with an index.

For example:

VS:

...
void c(vec4 p, vec4 c[8], uint x)
{
	gl_ClipDistance[7] = mix(1.0, dot(p, c[7]), x(c, 7u));
}
...
void main() {
  ...
  c(...)
}


GS (CLIP not defined):

...
#if defined(CLIP)
void o(uint i)
{
	gl_ClipDistance[7] = gl_in[i].gl_ClipDistance[7];
}
#endif
...
void main(){
  ...
  #if defined(CLIP)
	  o(0u);
  #endif
  ...
}

ShchchowAMD:
Test info should be changed.
Latest added commit for default set implicit sized check needs a test for it.
I'll add a new commit today.

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -749,6 +749,21 @@ void TIntermediate::mergeLinkerObjects(TInfoSink& infoSink, TIntermSequence& lin
...
 
+                // Update implicit array sizes
+                if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isImplicitlySizedArray()) {
+                    if (unitSymbol->getType().getImplicitArraySize() > symbol->getType().getImplicitArraySize()){
+                        symbol->getWritableType().updateImplicitArraySize(unitSymbol->getType().getImplicitArraySize());
+                    }
+                }
+                else if (symbol->getWritableType().isImplicitlySizedArray() && unitSymbol->getType().isSizedArray()) {
+                    if (symbol->getWritableType().getImplicitArraySize() > unitSymbol->getType().getOuterArraySize())
+                        error(infoSink, "Implicit size of unsized array aren't match same symbol among multiple shaders.");
```
greg-lunarg, line 760: aren't -> doesn't :)

Also in the .out file. Thanks!

greg-lunarg (changes requested):
Just one small change that got missed. Thanks for your persistence!

glslang/MachineIndependent/linkValidate.cpp
ShchchowAMD: Sorry, I forgot this one :)
New commit has fixed this error message.

greg-lunarg (approved):
LGTM. Thanks for the code!

PR #3075, opened 2022-11-28
[glslang][extension] GL_ARB_geometry_shader4.

ShchchowAMD:
Support extension GL_ARB_geometry_shader4.
1. primitive type and vertices number would be determined in link time, hence set flags.
2. Two changes in ParseHelper have dependency on implicit array size set.
3. For some special var like gl_VerticesIn, recompute would be later than first-time compile. Would be depend on api set topology and extra info across stages.


ShchchowAMD:
Dependency PR: https://github.com/KhronosGroup/glslang/pull/3074

arcady-lunarg:
@ShchchowAMD do you plan to move this forward at some point?

PR #3082, opened 2022-12-07
[glslang] Add impl for several matrix and constant op.

ShchchowAMD:
Add impl for some ops, mainly for matrix and float / conv ops.

Finished op List:
(all of them have already been added to symbol table and related to a built-in op/function before, but not actually be implemented yet.)

```
EOpMatrixInverse
EOpTranspose
EOpPackSnorm2x16
EOpPackUnorm2x16
EOpPackHalf2x16
EOpUnpackSnorm2x16
EOpUnpackUnorm2x16
EOpUnpackHalf2x16
EOpDeterminant
EOpSinh
EOpCosh
EOpTanh
EOpAsinh
EOpAcosh
EOpAtanh
EOpFloatBitsToInt
EOpFloatBitsToUint
EOpIntBitsToFloat
EOpUintBitsToFloat
EOpDoubleBitsToInt64
EOpDoubleBitsToUint64
EOpInt64BitsToDouble
EOpUint64BitsToDouble
EOpMul (matrixCompMult)
EOpComma
```

### DIFF
--- a/glslang/MachineIndependent/Constant.cpp
+++ b/glslang/MachineIndependent/Constant.cpp
@@ -37,6 +37,7 @@
 //
 
 #include "localintermediate.h"
+#include "../../SPIRV/hex_float.h"
 #include <cmath>
 #include <cfloat>
 #include <cstdlib>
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/Constant.cpp
```diff
@@ -37,6 +37,7 @@
 //
 
 #include "localintermediate.h"
+#include "../../SPIRV/hex_float.h"
```
greg-lunarg, line 40: This is a little ugly. The MachineIndependent module should be independent of any target module such as SPIRV.

There really is nothing about hex_float.h (and bitutils.h) that is related to SPIRV. They were brought in to support SPIRV and that is why they were placed there. They do have a spvutils namespace, but that is it.

How much work would it be to move these files into MachineIndependent and remove any reference of SPIRV from them?

greg-lunarg (changes requested):

glslang/MachineIndependent/Constant.cpp
ShchchowAMD: Yes, you are right.

The reason to include hex_float here is to keep those float/bit ops following same rules/implementation.

I'd have a try to refine the folder structure and namespace locally first.

arcady-lunarg:
@ShchchowAMD do you intend to get back to this at some point?

ShchchowAMD:
> 

Thanks for reminding. I'll make a commit recently to fix issue discussed above.

dnovillo:
@ShchchowAMD we've gotten no response to this in more than a year. I'll close it for now. Please re-open if you are still interested in this.

PR #3086, opened 2022-12-08
Guard AppleClang linker options

### LINKED ISSUE #3073: Misuse of -Wl,-undefined
`CMakeLists.txt` contains some code to pass `-undefined error` to the linker as shown below.

```
    if(NOT CMAKE_VERSION VERSION_LESS "3.13")
        # Error if there's symbols that are not found at link time.
        # add_link_options() was added in CMake 3.13 - if using an earlier
        # version don't set this - it should be caught by presubmits anyway.
        if (WIN32)
            add_link_options("-Wl,--no-undefined")
        else()
            add_link_options("-Wl,-undefined,error")
        endif()
    endif()
```

The problem is that `-undefined` isn't an option to force the linker to report an undefined symbol error. See https://man7.org/linux/man-pages/man1/ld.1.html. Did you confuse it with `--unresolved-symbols`?

jeremy-lunarg:
Fix #3073


greg-lunarg (approved):
LGTM. Thanks!

PR #3100, opened 2022-12-22
Fix issues with MaxDualSourceDrawBuffersEXT

daniel-story:
This PR fixes two issues introduced with the addition of support for MaxDualSourceDrawBuffersEXT.

First, glslangValidator's ResourceLimits.cpp was never updated to support the newly defined MaxDualSourceDrawBuffersEXT limit, causing the following warning to be output whenever glslangValidator is invoked with a default configuration file generated by glslangValidator's -c argument:

> Warning: unrecognized limit (MaxDualSourceDrawBuffersEXT) in configuration file.

This PR adds the appropriate check to recognize and apply this limit.

Second, the option for MaxDualSourceDrawBuffersEXT was added to the glslang_resource_s struct with inconsistent capitalization compared to all other members of the same struct (and indeed the entire C interface). This PR updates the struct definition with a union of the old name and the new name in order to maintain backward compatibility. The next time the major version of glslang is incremented this union could be replaced with just the new (consistent) name. Alternatively, it could be permanently left as a union.

### DIFF
--- a/StandAlone/ResourceLimits.cpp
+++ b/StandAlone/ResourceLimits.cpp
@@ -505,6 +505,8 @@ void DecodeResourceLimits(TBuiltInResource* resources, char* config)
             resources->maxTaskWorkGroupSizeZ_EXT = value;
         else if (tokenStr == "MaxMeshViewCountEXT")
             resources->maxMeshViewCountEXT = value;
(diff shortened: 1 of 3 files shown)


greg-lunarg (approved):
The union is ugly, but I think it is really the best solution for the moment.

Otherwise fine. Thanks for the fix!

PR #3104, opened 2023-01-04
build: Remove AMD_EXTENSIONS / NV_EXTENSIONS defines

juan-lunarg:
These defines aren't needed anymore


jeremy-lunarg (approved):
LGTM. These appear to be leftovers from some refactoring work JohnK did back in 2019. Thanks!

PR #3108, opened 2023-01-11
Fix crash on bad structure member reference

### LINKED ISSUE #3105: Crash when trying to access a non-existing member of a struct returned by a function
If you have a function returning a struct and you directly try to access a member from this struct from the function call, but this member does not actually exist, glslang, as a library, and glslangValidator will crash.

Shader example :

```glsl
#version 460

struct A {
	float x;
};

A test() {
	return A(1.0);
}

void main() {
	test().z; // A.z does not exist, causes a crash
}
```

Crash happens during the parsing stage, in ParseHelper.cpp, in the ``handleDotDereference`` function : https://github.com/KhronosGroup/glslang/blob/06a7078ce74ab5c7801a165b8145859678831fb8/glslang/MachineIndependent/ParseHelper.cpp#L1037-L1042

The while loop is checking if ``baseSymbol`` is a ``TIntermSymbol``, and if it is not, tries to dereference it as a binary node, which is ``nullptr`` in this case, causing a crash.

The difference between 
```glsl
test().z; // A.z does not exist, causes a crash
```
and

```glsl
A c = A(1.0);
float b = c.z; // A.z does not exist, does not crash
```

is that ``baseSymbol`` is of type ``TIntermAggregate`` in the first case and ``TIntermSymbol`` in the second.

``TIntermAggregate`` does not have a ``getAsSymbolNode`` nor a ``getAsBinaryNode`` member function and use the base class ``TIntermNode`` implementations instead, which both return ``nullptr``.

``getAsBinaryNode`` is implemented by ``TIntermBinary``, so 

```cpp
auto baseSymbol = base;
while (baseSymbol->getAsSymbolNode() == nullptr) 
    baseSymbol = baseSymbol->getAsBinaryNode()->getLeft(); 
```
is trying to get a ``TIntermSymbol`` and estimates that if ``baseSymbol`` is not a ``TIntermSymbol``, then it is a ``TIntermBinary``, which is not the case here, as it is a ``TIntermAggregate``.

One possible fix is to isolate both cases, one where ``baseSymbol`` is a ``TIntermBinary`` and one where ``baseSymbol`` is a ``TIntermAggregate``, then getting the name of either the ``TIntermSymbol`` or the ``TIntermAggregate`` and write a different message when the error is from a ``TIntermAggregate`` :

```cpp
auto baseSymbol = base;
while (baseSymbol->getAsSymbolNode() == nullptr) {
    if (baseSymbol->getAsBinaryNode() == nullptr) {
        break;
    }
    baseSymbol = baseSymbol->getAsBinaryNode()->getLeft();
}
TString structName;
if (baseSymbol->getAsSymbolNode() != nullptr) {
    structName.append("\'").append(baseSymbol->getAsSymbolNode()->getName().c_str()).append("\'");
    error(loc, "no such field in structure", field.c_str(), structName.c_str());
} else if (baseSymbol->getAsAggregate() != nullptr) {
    structName.append("\'").append(baseSymbol->getAsAggregate()->getName().c_str()).append("\'");
    structName.erase(structName.end() - 2);
    error(loc, "no such field in structure returned by function", field.c_str(), structName.c_str());
}
```

I don't know if any more types can reach this code path so I am not sure if it actually works in all cases.

With the example code :

```cpp
ERROR: 0:15: 'z' : no such field in structure returned by function 'test'
ERROR: 0:15: '' : compilation terminated
```

greg-lunarg:
Fixes #3105

### DIFF
--- a/Test/baseResults/struct.error.frag.out
+++ b/Test/baseResults/struct.error.frag.out
@@ -1,0 +1,37 @@
+struct.error.frag
+ERROR: 0:12: 'z' : no such field in structure 
+ERROR: 1 compilation errors.  No code generated.
+
+
+Shader version: 460
+ERROR: node is still EOpNull!
+0:7  Function Definition: test( ( global structure{ global float x})
+0:7    Function Parameters: 
+0:8    Sequence
+0:8      Branch: Return with expression
+0:8        Constant:
+0:8          1.000000
+0:11  Function Definition: main( ( global void)
+0:11    Function Parameters: 
+0:12    Sequence
+0:12      Function Call: test( ( global structure{ global float x})
+0:?   Linker Objects
+
+
+Linked fragment stage:
+
+
+Shader version: 460
+ERROR: node is still EOpNull!
+0:7  Function Definition: test( ( global structure{ global float x})
+0:7    Function Parameters: 
+0:8    Sequence
+0:8      Branch: Return with expression
+0:8        Constant:
+0:8          1.000000
+0:11  Function Definition: main( ( global void)
+0:11    Function Parameters: 
+0:12    Sequence
+0:12      Function Call: test( ( global structure{ global float x})
+0:?   Linker Objects
+
--- a/Test/struct.error.frag
+++ b/Test/struct.error.frag
@@ -1,0 +1,14 @@
+#version 460
+
+struct A {
+	float x;
+};
+
+A test() {
+	return A(1.0);
+}
+
+void main() {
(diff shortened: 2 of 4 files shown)


jeremy-lunarg (changes requested):
LGTM but needs test case.

greg-lunarg:
@jeremy-lunarg Test case added. PTAL.

jeremy-lunarg (approved):
LGTM

PR #3114, opened 2023-01-17
Fix double expansion of macro parameters

### LINKED ISSUE #3112: macro expanded twice, leading to unexpected generated code.
This was originally filed as https://github.com/google/shaderc/issues/1288
The original submitter is @jeroenbakker-atmind, a Blender developer.

It seems macros that are used as parameters to another macro call can be expanded twice.
See the expression `m(pPosition)`.

This is a minimal reproducer:

```
#version 450

layout(push_constant) uniform constants {
  vec4 pPosition;
}
PushConstants;
#define pPosition (PushConstants.pPosition)

layout(location=0) out vec3 o;

#define m(p)  vec4(p,1.0);

void main() {
  gl_Position = m(pPosition);
}
```

glslangValidator -E y.vert shows:

```
$ glslangValidator -E y.vert
#version 450

layout(push_constant)uniform constants {
  vec4 pPosition;
}
PushConstants;


layout(location = 0)out vec3 o;



void main(){
  gl_Position = vec4((PushConstants .(PushConstants . pPosition)), 1.0);;
}
```

I'm not enough of an expert preprocessor macro expansion.  The Glsl 4.5 spec says

> #define and #undef functionality are defined as is standard for C++ preprocessors for macro definitions
both with and without macro parameters.

When I use `clang -E` to run the C++ preprocessor I get this main function out of it:
```
void main() {
  gl_Position = vec4((PushConstants.pPosition),1.0);;
}
```

arcady-lunarg:
Macro arguments were getting expanded twice, causing issues when a macro's name appeared in its definition.

Fixes #3112

I was a bit skeptical that this is the correct fix, but it does fix the issue and doesn't cause any problems in any of the testing I have done.

### DIFF
--- a/Test/baseResults/preprocessor.function_macro.vert.out
+++ b/Test/baseResults/preprocessor.function_macro.vert.out
@@ -13,9 +13,12 @@
 
 
 
+
+
 int main(){
   gl_Position = vec4(3 + 1, 3 + 4, 3 + 1);
   gl_Position = vec4(1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 + 11 + 12);
   gl_Position = vec4(4 + 3 + 3);
+  gl_Position = 4 + 3 + F . a;
 }
 
--- a/Test/preprocessor.function_macro.vert
+++ b/Test/preprocessor.function_macro.vert
@@ -13,8 +13,11 @@
   )\
 4 + 3 + Y
 
+#define F F.a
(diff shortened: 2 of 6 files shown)


arcady-lunarg:
Hm, never mind. Apparently I wasn't testing it correctly.

jeremy-lunarg (commented):
Can you please add a test?

PR #3116, opened 2023-01-18
Update appveyor environment

jeremy-lunarg:
SPIRV-Tools now requires 3.17.2 or higher. The version provided by Visual Studio 2015 is 3.16.2.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,7 +4,7 @@
 # build version format
 version: "{build}"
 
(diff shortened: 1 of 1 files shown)


arcady-lunarg (commented):
Looks good to me. For posterity, is the reason for going straight from 2015 to 2019 just future-proofing against future minimum version bumps?

arcady-lunarg (approved):

Issue #3111, opened 2023-01-14
Validator validates integer attributes in legacy GLSL


rdb:
This code should be invalid according to the specification:
```
#version 120

attribute ivec4 invalid;

void main() {
}
```

However, it passes validation.  The GLSL 1.20 spec seems very clear:
> The attribute qualifier can be used only with **float**, floating-point vectors, and matrices.

PR #3120, opened 2023-01-20
Reject non-float inputs/outputs with version < 120

arcady-lunarg:
GLSL 1.20 and prior stated that "the attribute qualifier can be used only with float, floating-point vectors, and matrices" and likewise for varying.

Fixes: #3111


jeremy-lunarg (approved):
LGTM

Issue #2411, opened 2020-10-03
Invalid BufferBlock operand gets emitted when compiling HLSL code to SPIR-V 1.5


Ryp:
I have this shader ([shader playground mirror](http://shader-playground.timjones.io/aec8aac1081244b99a3a98952f14a236))
```hlsl
#include "lib/base.hlsl"
#include "lib/indirect_command.hlsl"

#include "share/compaction.hlsl"

VK_BINDING(0, 0) ByteAddressBuffer DrawCount;

VK_BINDING(1, 0) RWByteAddressBuffer IndirectDispatchOut;
VK_BINDING(2, 0) RWByteAddressBuffer DrawCountOut;

[numthreads(1, 1, 1)]
void main()
{
    const uint command_count = DrawCount.Load(0);

    IndirectDispatchCommand dispatchCommand;
    dispatchCommand.x = div_round_up(command_count, CompactionGroupSize);
    dispatchCommand.y = 1;
    dispatchCommand.z = 1;

    // Write indirect dispatch command
    IndirectDispatchOut.Store(0, dispatchCommand.x);
    IndirectDispatchOut.Store(4, dispatchCommand.y);
    IndirectDispatchOut.Store(8, dispatchCommand.z);

    // Reset draw count for the next pass
    DrawCountOut.Store(0, 0);
}
```
That compiles with this command line:
```
glslangValidator --target-env vulkan1.2 -g -e main -V -D compaction_prepare_indirect_dispatch.comp.hlsl -o compaction_prepare_indirect_dispatch.comp.spv
```
I get this error when I try to legalize the HLSL:
```
error: line 50: 2nd operand of Decorate: operand BufferBlock(3) requires SPIR-V version 1.3 or earlier
    OpDecorate %DrawCount BufferBlock
```
I'm only seeing this error now, as I'm trying to compile all my shaders for SPIR-V 1.5. I was using SPIR'V 1.3 previously.

StarsX:
Hit the same error for HLSL StructuredBuffer: 2nd operand of Decorate: operand BufferBlock(3) requires SPIR-V version 1.3 or earlier.

FoggyMist:
Also hit the same error, submitted a fix here: https://github.com/KhronosGroup/glslang/pull/3121

Ryp:
I'm not seeing the compile error anymore

arcady-lunarg:
This was fixed by #3121

PR #3121, opened 2023-01-23
Move check if useStorageBuffer needs to be set

FoggyMist:
...From TParseContext used only by GLSL, to TParseContextBase inherited by both GLSL and HLSL paths. It caused compilations from HLSL to SPIR-V 1.3+ to use BufferBlock decoration which is no longer valid.

### DIFF
--- a/Test/baseResults/hlsl.wavebroadcast.comp.out
+++ b/Test/baseResults/hlsl.wavebroadcast.comp.out
@@ -2331,7 +2331,7 @@ local_size = (32, 16, 1)
                               MemberDecorate 20(Types) 3 Offset 64
                               Decorate 21 ArrayStride 96
                               MemberDecorate 22(data) 0 Offset 0
-                              Decorate 22(data) BufferBlock
+                              Decorate 22(data) Block
                               Decorate 24(data) DescriptorSet 0
                               Decorate 24(data) Binding 0
                               Decorate 388(dti) BuiltIn GlobalInvocationId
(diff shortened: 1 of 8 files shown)


FoggyMist:
As mentioned in https://github.com/KhronosGroup/glslang/issues/2411 some HLSL inputs are producing invalid SPIR-V output. When it happens it can be easily validated that output SPIR-V is using outdated operand of Decorate. BufferBlock has been deprecated in SPIR-V 1.3 and removed in newer versions.

My debug session went like this:

1. I found that there is only one part of the code that produces BufferBlock token during translation to SPIR-V. It is `TranslateBlockDecoration` function from `glslang/SPIRV/GlslangToSpv.cpp:378`.
2. This function is being called in only one place `glslang/SPIRV/GlslangToSpv.cpp:4842`. It is being called when compiling both HLSL and GLSL, but with different `useStorageBuffer`.
3. That flag comes from `glslangIntermediate` member that is being initialized by ParseContext class.
4. There is switch with `#ifdef` that picks right concrete implementation of ParseContext, depending on shader input (HLSL vs GLSL).
5. It turns out that `TParseContext` (which is part of GLSL path) checks if `--target-env` is SPIR-V 1.3 or newer inside constructor. But there is no such check in `HlslParseContext`'s constructor (HLSL path).
6. I moved check to common base class - `TParseContextBase`. It worked, so I regenerated test files and submitted this PR.

arcady-lunarg (approved):
Looks good.

PR #3127, opened 2023-01-30
HLSL: Add missing relaxed-precision float/int matrix expansions

rdb:
Straightforward PR that adds matrix expansions for min16float/min10float/min16int/min12int/min16uint types.  These are all valid HLSL as mentioned [here](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-keywords).

Also supported in the `matrix<type, n, n>` syntax, which also had `half` missing as a supported type.

Thanks in advance!


arcady-lunarg (changes requested):
Thanks for the PR. Could you add tests for the newly supported types? I think `Test/hlsl.mintypes.frag` looks like the place to add the new tests.

rdb:
Can do, but do you also want me to add that file to runtests then?  It doesn't seem to currently be part of the list of tests that is run.

rdb:
Updated test and baseResults file.  Haven't touched runtests.

arcady-lunarg (approved):
Lgtm, thanks for adding the tests.

rdb:
It seems that glslang has a bunch of issues relating to 1-col or 1-row matrices that caused the validation to fail (SPIR-V does not support matrices with such a size).  I have removed the 1D matrices from the test file for now, tests pass now.

jeremy-lunarg:
@rdb Thank you for fixing the test. We're investing an unrelated failure with Kokoro external CI and will merge as soon as that is resolved.

jeremy-lunarg:
Merging this despite the CI failure. Google is investigating the Kokoro job that is failing.

Issue #2439, opened 2020-10-28
RuntimeArray within a RuntimeArray generated from HLSL


ruiling:
I am investigating an issue when running vulkancts.
https://github.com/KhronosGroup/VK-GL-CTS/blob/master/external/vulkancts/modules/vulkan/tessellation/vktTessellationFractionalSpacingTests.cpp#L498
The tessellation evaluation shader is written in hlsl:

```
struct OutputStruct
{
    int numInvocations;
    float tessCoord[];
};
globallycoherent RWStructuredBuffer <OutputStruct> Output : register(b1);

void main(float2 tessCoords : SV_DOMAINLOCATION)
{
    int index;
    InterlockedAdd(Output[0].numInvocations, 1, index);
    Output[0].tessCoord[index] = tessCoords.x;
}
```

I compiled with ` ./glslangValidator -e main -V 1.tese.hlsl -o 1.spv; spirv-dis --raw-id 1.spv -o -`
The compiled spv is below:

```
               OpCapability Tessellation
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint TessellationEvaluation %4 "main" %39
               OpSource HLSL 500
               OpName %4 "main"
               OpName %11 "@main(vf2;"
               OpName %10 "tessCoords"
               OpName %15 "index"
               OpName %17 "OutputStruct"
               OpMemberName %17 0 "numInvocations"
               OpMemberName %17 1 "tessCoord"
               OpName %19 "Output"
               OpMemberName %19 0 "@data"
               OpName %21 "Output"
               OpName %36 "tessCoords"
               OpName %39 "tessCoords"
               OpName %44 "param"
               OpDecorate %16 ArrayStride 4
               OpMemberDecorate %17 0 Offset 0
               OpMemberDecorate %17 1 Offset 4
               OpDecorate %18 ArrayStride 8
               OpMemberDecorate %19 0 Coherent
               OpMemberDecorate %19 0 Offset 0
               OpDecorate %19 BufferBlock
               OpDecorate %21 DescriptorSet 0
               OpDecorate %21 Binding 1
               OpDecorate %39 Patch
               OpDecorate %39 BuiltIn TessCoord
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
          %7 = OpTypeVector %6 2
          %8 = OpTypePointer Function %7
[...]

PR #3130, opened 2023-02-01
Block-decorate Vulkan Structs with RuntimeArrays

FoggyMist:
Fixes KhronosGroup/glslang#2439

When decorating a struct for Vulkan, add [Buffer]Block decoration if the struct has a RuntimeArray member, as required by VUID-...-04680.


arcady-lunarg:
Looks good, but would it be possible to add a test case for this? Probably just using the code in the original issue would work fine as a test.

greg-lunarg:
Tests can be added by adding shaders that use the relevant functionality to the Test/ directory. You will also need to add the files to Hlsl.FromFile.cpp. See hlsl.switch.frag as an example.

You will also need to create result files in Test/baseResults/, but this can be done by building and running build/gtests/Release/glslangtests.exe --update-mode.

arcady-lunarg (approved):
Thanks for adding the test, looks good to me now.

PR #3133, opened 2023-02-02
kokoro: avoid git permissions issue that GN fails on

dneto0:
In the docker build script that Kokoro runs, the directories are owned by a different user.  Git complains about that and in the GN flow GN will error out.
In this docker flow we don't care about that warning, so within the docker config set a git global option to ignore the issue.

### DIFF
--- a/kokoro/linux-clang-gn/build-docker.sh
+++ b/kokoro/linux-clang-gn/build-docker.sh
@@ -34,7 +34,12 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 set -e # Fail on any error.
+
+. /bin/using.sh # Declare the bash `using` function for configuring toolchains.
+
 set -x # Display commands being run.
+
+using ninja-1.10.0
 
 # Disable git's "detected dubious ownership" error - kokoro checks out the repo
 # with a different user, and we don't care about this warning.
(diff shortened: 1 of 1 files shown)


dj2 (approved):

jeremy-lunarg:
@dneto0 Should we expect the CI-ubuntu-clang-gn job to pass in this PR? Or will that job start passing once this PR lands on main?

dneto0:
> @dneto0 Should we expect the CI-ubuntu-clang-gn job to pass in this PR? Or will that job start passing once this PR lands on main?

Sigh. Now it's failing for a different reason.

The ninja build tool was removed from depot_tools.  I'll patch this PR.

dneto0:
> @dneto0 Should we expect the CI-ubuntu-clang-gn job to pass in this PR? Or will that job start passing once this PR lands on main?

The intent is to land only after that bot works in this PR.
thanks

kokoro/linux-clang-gn/build-docker.sh
```diff
@@ -36,6 +36,12 @@
 set -e # Fail on any error.
 set -x # Display commands being run.
 
+using ninja-1.10.0
```
dneto0, line 39: well that didn't work.

jeremy-lunarg: The job is currently passing. Is there something else that is broken?

jeremy-lunarg: Ah, I see. I believe you were referring to the missing "using" declaration. Everything appears to be okay now. Can I proceed with the merge?

jeremy-lunarg (approved):
LGTM

PR #3134, opened 2023-02-04
change HLSL/hlslParseables.cpp to support InterlockedAdd on F=float types

rcoreilly:
This enables use of `InterlockedAdd` in HLSL on float types per this extension: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_shader_atomic_float.html

It has been tested and produces the expected results, generating an `atomicAdd` opcode.


arcady-lunarg (changes requested):
Thanks for the patch. Could you add some tests to verify that the expected results are produced? It should just be a matter of adding tests for InterlockedAdd with floats to `Test/hlsl.intrinsics.comp` and then updating the base results using `build/gtests/Release/glslangtests.exe --update-mode`.

rcoreilly:
ok did that..

arcady-lunarg (approved):
LGTM, thanks for adding the test.

PR #3135, opened 2023-02-07
Remove unused type parameter methods; NFC

svenvh:


arcady-lunarg:
Since these functions are in a publicly exported header, removing them would break the public library interface and require a major version bump.

svenvh:
A major version bump just for removing a few methods sounds like an overkill; I guess this will have to wait until there is a more compelling reason for a major version bump.

Closing now, but please let me know whenever a major version increase is imminent.

arcady-lunarg:
It's a public header file, so any 3rd party programs that compile against it could be using those methods and would no longer compile as a result. We might even want to issue a deprecation notice for these methods before we remove them. May I ask, what is the motivation for removing these methods?

svenvh:
I completely missed the fact that this is a public header with stability guarantees.  My plan was to push out some refactorings ahead of actual functional changes that are still being defined, but it seems better to defer those refactorings until the new functionality is ready so we only need a single version bump.

Issue #3139, opened 2023-02-16
Build failing with GCC13 on master commit


DeadSix27:
`glslang/glslang/MachineIndependent/SymbolTable.h:70:1: note: 'uint32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?`

Somewhere it forgets to include cstdint, I'm using latest MinGW commit and GCC13.0.1

JohannesKauffmann:
This is not MinGW specific, but GCC 13. See https://gcc.gnu.org/gcc-13/porting_to.html.

DeadSix27:
> This is not MinGW specific, but GCC 13. See https://gcc.gnu.org/gcc-13/porting_to.html.

Ah thanks for clarifying, makes sense.

hydra3333:
Hi.

Just a gentle prompt to please apply the patch asap.

Without it some mingw based builds systems, including MABS, cannot build your product ... which is a great deal less than ideal :)

PR #3144, opened 2023-02-21
Include <cstdint> header in Common.h

arcady-lunarg:
This change also cleans up some ifdef'd code for no longer supported versions of MSVC.

Fixes: #3139

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -44,6 +44,7 @@
 #else
 #include <cmath>
 #endif
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <list>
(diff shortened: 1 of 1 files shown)


jeremy-lunarg (commented):
This looks good to me. Can we clean up lines 69-83 as well? We still need the safe_vsprintf definitions. Can we remove the basetsd.h? Will that affect mingw? I think we can remove the int_types.h include. I believe that's a very old solaris header.

JohannesKauffmann:
Unfortunately, excluding `basetsd.h` does hurt mingw in this case:

```
/glslang/glslang/MachineIndependent/PoolAlloc.cpp:280:57: error: 'UINT_PTR' does not name a type; did you mean 'UINT_MAX'?
         return reinterpret_cast<void*>(reinterpret_cast<UINT_PTR>(memory) + headerSkip);
                                                         ^~~~~~~~
                                                         UINT_MAX
```

jeremy-lunarg (approved):
LGTM

PR #1621, opened 2018-12-10
CMake: Allow linking against system-installed SPIRV-Tools

akien-mga:
Fixes #1585.

@kyrios123 Can you confirm that it works for your use case too?

Output on my system:
```
+ CFLAGS='-O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -fasynchronous-unwind-tables'
+ export CFLAGS
+ CXXFLAGS='-O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -fasynchronous-unwind-tables'
+ export CXXFLAGS
+ FFLAGS='-O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -fasynchronous-unwind-tables'
+ export FFLAGS
+ LDFLAGS=' -Wl,--as-needed -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags'
+ export LDFLAGS
+ mkdir -p build
+ cd build
[...]

PR #3152, opened 2023-03-03
Look for external SPIR-V Tools build, if not building in-tree

kasper93:
This allows to build with optimizer enabled, if external SPIR-V tools libraries are available in the system. It is quite common in *nix world to package spirv-tools and glslang separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,8 +319,15 @@ if(BUILD_EXTERNAL AND IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/External)
     add_subdirectory(External)
 endif()
 
+option(ALLOW_EXTERNAL_SPIRV_TOOLS "Allows to build against installed SPIRV-Tools-opt" OFF)
 if(NOT TARGET SPIRV-Tools-opt)
-    set(ENABLE_OPT OFF)
+    if(ALLOW_EXTERNAL_SPIRV_TOOLS)
+        # Look for external SPIR-V Tools build, if not building in-tree
+        find_package(SPIRV-Tools-opt)
+    endif()
+    if(NOT TARGET SPIRV-Tools-opt)
+        set(ENABLE_OPT OFF)
+    endif()
 endif()
 
 if(ENABLE_OPT)


kasper93:
Any question or comments? I'd prefer to have it upstreamed instead of sideloading such small patch, thanks.

arcady-lunarg:
My main concern is that this may end up causing problems for someone who forgets to run update_glslang_sources.py and ends up picking up whatever spirv-tools might be installed on their system, which may lead to unexpected results. I think putting this behind a flag would be fine though.

kasper93:
> My main concern is that this may end up causing problems for someone who forgets to run update_glslang_sources.py and ends up picking up whatever spirv-tools might be installed on their system

This is exactly the case I wanted to enforce on users. Arguably it is little bit messy, because most distors splits the packages and don't build spriv-tools in-tree. While Fedora has patches for that, some others like Arch does not and silently the optimizer is disabled.

Take a look at https://gitlab.archlinux.org/archlinux/packaging/packages/glslang/-/blob/3bf6cd36b4703af66f04cb8d81dcb754a765e5fe/PKGBUILD where they explicitly state the issue about ABI compatibility between glslang and spriv-tools. In the same time, they use own packages, and because cmake won't search for it it will set ENABLE_OPT=0.

This is basically broken after https://gitlab.archlinux.org/archlinux/packaging/packages/glslang/-/commit/9ad20da7712f9ad431f3cfadd43c748117a0d9e3 where they split the build.

And this is just an example, same thing happens to MSYS2 and others.

That's why I would love for it to be enabled by default and force people to resolve dependency issue on their own.

But to move forward I updated the patch with an option and disabled by default. This is the safest option and it will be easier for everyone to enable the system package linking. Still would like to be ON tho :)

arcady-lunarg:
I think it might not be a bad idea to also add cmake warnings if ENABLE_OPT was requested and SPIRV-Tools was not found, that way it would be clear that something isn't quite as expected.

arcady-lunarg:
Fair enough, let me review your change and I'll make followup change.

arcady-lunarg (approved):
LGTM, thanks.

akien-mga:
Thanks for working on this!

I don't see changes to https://github.com/KhronosGroup/glslang/blob/main/SPIRV/CMakeLists.txt#L101-L109 unlike what I had in #1621, won't that lead to still using the vendored SPIRV-Tools headers instead of the system-packaged ones?

kasper93:
> won't that lead to still using the vendored SPIRV-Tools headers instead of the system-packaged ones?

Unlike if your patch, this does not use pkg-config. `find_package(SPIRV-Tools-opt)` will find the package and initialize fully qualified cmake target (based on installed .cmake files). And then when you use it as `target_link_libraries` it will append `INTERFACE_INCLUDE_DIRECTORIES` as defined in the target.

If anything this `target_include_directories` are only needed for in-tree build or redundant even.

You can validate that proper include dir is defined in the target by using something like this
```
get_target_property(incdir SPIRV-Tools-opt INTERFACE_INCLUDE_DIRECTORIES)
message(${incdir})
```

I think everything is correct, but if any issues will occur with this, feel free to ping me.

PR #3158, opened 2023-03-14
Add option to glslangValidator to inject preamble

moritz-h:
Add option `-P` / `--preamble-text` / `--p` to glslangValidator to add custom text to the preamble. There are already the parameters `-D` and `-U` allowing to set #define and #undef, but this options makes it possible to inject fully customizable preamble text via a command line parameter.

My intended usecase is to allow enabling of extensions, but might be useful for other things as well to have the flexibility of setting arbitrary preambles.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -258,6 +258,17 @@ public:
         text.append("\n");
     }
 
+    void addText(std::string preambleText)
+    {
+        fixLine(preambleText);
+
+        Processes.push_back("preamble-text");
+        Processes.back().append(preambleText);
+
+        text.append(preambleText);
+        text.append("\n");
+    }
+
 protected:
     void fixLine(std::string& line)
     {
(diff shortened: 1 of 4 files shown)


arcady-lunarg:
I think this feature is a bit too general and potentially confusing. Do you have a complete example of where you would be using this? If you only need to enable extensions, I think it would make sense to keep it limited to that, and I think it would also make sense to actually test that in the test cases you are adding.

moritz-h:
I would use this to validate shaders in a CI job. This would be used for a application which has glslang integrated as library to compile shaders at runtime but sets a preamble. In the CI job I would use glslangValidator to validate the shaders independently of the application, but need to inject a preamble in order to make the shaders valid.

For my specific setup, it would be sufficient to instead just mark specific extensions as enabled, but this would require much more code changes. Reading a bit through the glslangValidator sources, I saw that the preamble feature is more or less already fully implemented just not exposed as command line parameter.

But I see you point, I'm more or less requesting to bring a library API feature to the command line interface, which might is very general. But on the other side I thought it may has some value and usecases for others and only adds minimal complexity to the glslangValidator.
I'm happy to make changes if it helps to make this feature less confusing. Ideas would be to remove `-P` and `--p`. Keeping only the long name, to let it look more like a uncommon expert flag and may could write a longer description.

arcady-lunarg:
The point about this being already available through the API is a good one. I would still like to see somewhat more testing and possibly better help text. I'll give a more detailed code review shortly.

moritz-h:
Added two tests for injection of actual code and for injection of extension requirement.

For help text, would it make sense to copy the API description https://github.com/KhronosGroup/glslang/blob/12.0.0/glslang/Public/ShaderLang.h#L450-L451 ?

For the code review, there are 2 points where I am not 100% sure if its correct:
- `Processes.push_back()` I assumed that this allows replay of the commands given so the string should match the actual command line option, is this correct?
- Was not sure if line breaks should be allowed in the provided string, I copied the call to `fixLine()` from the defines feature. This will cut the string at the first line break. But maybe it makes sense to allow line breaks for arbitrary preamble strings.

PR #3159, opened 2023-03-14
Improve line info for symbol access and assignment

qingyuanzNV:
This patch intends to improve line information generation in the following cases:
```
// Because this is just a symbol access, glslang failed to generate opline for the load
if (b) {
  //...
}

// We should reset lines to += for code generated for addition and assignment
a += x > y ?
         x : y;
```

Currently, because line number during spv generation is only set during traversal of expressions, the OpLine would be missing for OpLoad if the predicate for the branch statement is just a symbol access. To fix this, added a setLine call in visit of symbol.

Also, if an assignment statement is crossing multiple lines, the additional operation and store instruction wasn't correctly mapped to where the assignment operator is.


arcady-lunarg (approved):
LGTM, thanks for contributing.

PR #3160, opened 2023-03-14
Update known_good.json

jeremy-lunarg:
Prepare for release 12.1.0.

### DIFF
--- a/known_good.json
+++ b/known_good.json
@@ -5,14 +5,14 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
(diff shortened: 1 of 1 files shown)


Keenuts:
```
In file included from [../../External/spirv-tools/tools/util/flags.cpp:15](https://cs.corp.google.com/piper///External/spirv-tools/tools/util/flags.cpp?l=15):
[../../External/spirv-tools/tools/util/flags.h:186](https://cs.corp.google.com/piper///External/spirv-tools/tools/util/flags.h?l=186):23: error: no template named 'variant' in namespace 'std'
using FlagType = std::variant<std::reference_wrapper<Flag<std::string>>,
```

Apparently spirv-tools is not build with c++17 but c++14 looking at the clang commandline.

PR #3170, opened 2023-03-21
Parameters of spirv_decorate_id should accept variables

amdrexu:
spirv_decorate_id will generate OpDecorateId. The parameter list should accept variables as part of decorations. This is because OpDecorateId allows this. The spec says:

  All such <id> Extra Operands must be constant instructions or
  OpVariable instructions.


amdrexu:
Ping for review. Thanks.

arcady-lunarg:
I'm going to take a look at this today.

arcady-lunarg (approved):
LGTM

Issue #3157, opened 2023-03-10
Support the generation of spirv NonSemantic.DebugBreak


jiaolu:
Is it possible to support the [NonSemantic.DebugBreak](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.DebugBreak.asciidoc) in glslang ?

I am thinking about there are two ways to suppor this spirv in the glslang.
First we add gdb like option to the glslang .ie.

glslangValidator -breakPoint 10 -breakFile "compute.comp". 

Or we can have new instruction like debugBreak. but this require new glsl extension ?

I prefer first option, as we don't need to change glsl source file .

arcady-lunarg:
I think the command line option approach is a reasonable one, but I am curious about the envisioned use case for setting breakpoints via glslang. Is there a debugger that would be repeatedly calling glslang to recompile GLSL source code just to set breakpoints?

jiaolu:
currently there is no such debugger.  i figure workflow is different in vendors.  might be better implemented in game engine?

arcady-lunarg:
My main concern here is that while this does seem like a potentially reasonable feature, it would add a certain inevitable amount of complexity to the code, which we would really only want to maintain if it's actually useful to someone. For example, I could just as easily imagine a debugger modifying a SPIR-V module directly by parsing the debug info to find the appropriate DebugLine and inserting the breakpoint op right after it, or doing something else that doesn't require touching glslangValidator at all.

ShchchowAMD:
https://github.com/KhronosGroup/glslang/pull/3171

arcady-lunarg:
Come to think of it, I think this would be a better fit for SPIR-V tools, and I would prefer not to implement this feature in glslang.

PR #3171, opened 2023-03-21
[SPIRV] Implement debugBreak in glslangValidator.

ShchchowAMD:
An implementation for issue #3157.

Debug Break op would be inserted before specified line, only within a function declaration body block.

If this <line> is set outside a function body block, it would be ignored (for example, behind a global variable declaration statement).

Command line would be like:

`glslangValidator.exe --breakPoint 15 --breakFile vertexShader.vert -G vertexShader.vert`

The spirv would looks like:

```
%1 = OpExtInstImport "NonSemantic.DebugBreak"
%2 = OpExtInst %void %1
```

Ref: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.DebugBreak.asciidoc

### DIFF
--- a/SPIRV/CMakeLists.txt
+++ b/SPIRV/CMakeLists.txt
@@ -64,6 +64,7 @@ set(HEADERS
     GLSL.ext.NV.h
     GLSL.ext.ARM.h
     NonSemanticDebugPrintf.h
+    NonSemanticDebugBreak.h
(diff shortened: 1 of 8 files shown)


arcady-lunarg:
I would prefer to have this done via spir-v tools rather than in glslang itself.

PR #3175, opened 2023-03-24
Fix CMake for OSDependent install

### LINKED ISSUE #3125: CMake: glslang-config.cmake is missing find_package(Threads) under Unix
The [cmake Package Configuration File](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#id17) should find all packages that the targets need, otherwise configuration fails. Under Unix the OSDependent links against Threads::Threads, so it needs to include a `find_package(Threads)`

The attached minimal example shows the issue, under Linux configuration fails:
[glslangCmakeBug.zip](https://github.com/KhronosGroup/glslang/files/10508938/glslangCmakeBug.zip)

Obviously there is an easy work-around, making the user do the `find_package(Threads)` before `find_package(glslang)`, but I'd say that is suboptimal.

moritz-h:
Fix #3125 (also reported in vcpkg: https://github.com/microsoft/vcpkg/issues/27121).

The CMake target `OSDependent` requires `Threads::Threads` on Unix, but the requirement of Threads is not exported in the installed Config.cmake file. This PR adds the necessary `find_dependency(Threads REQUIRED)` to the installed Config.cmake file.

Further the OSDependent CMake includes workarounds with seem outdated with the currently required minimum version of 3.14.0. The check if( < 3.1) could be just removed. The second mentioned issue about cross compiling was closed upstream over 5 years ago and I verified that the mentioned fix commit is contained in CMake 3.10.0. So I assume should be also save to remove if CMake 3.14 is required anyway, but I do not have a cross compilation toolchain to actually test it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,10 +356,18 @@ endif()
 if(ENABLE_GLSLANG_INSTALL)
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/glslang-config.cmake.in" [=[
         @PACKAGE_INIT@
+        @INSTALL_CONFIG_UNIX@
         include("@PACKAGE_PATH_EXPORT_TARGETS@")
     ]=])
-    
+
     set(PATH_EXPORT_TARGETS "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/glslang-targets.cmake")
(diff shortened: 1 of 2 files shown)


jeremy-lunarg (approved):
LGTM

PR #3177, opened 2023-03-31
Fix generation of conditionals with a struct result in SPIR-V

arcady-lunarg:
This PR fixes generation of conditionals with a struct result and adds some tests. Note that multiTypeStore() produces the same code as before this change in the case when the types are actually the same.


greg-lunarg (approved):
LGTM

PR #3182, opened 2023-04-06
WIP: Fix bazel CI job

jeremy-lunarg:


kokoro/windows-msvc-2015-release-bazel/build.bat
```diff
@@ -45,13 +45,13 @@ cd %SRC%
...
-wget -q https://github.com/bazelbuild/bazel/releases/download/0.29.1/bazel-0.29.1-windows-x86_64.zip
-unzip -q bazel-0.29.1-windows-x86_64.zip
+wget -q https://github.com/bazelbuild/bazel/releases/download/5.0.0/bazel-5.0.0-windows-x86_64.zip
+unzip -q bazel-5.0.0-windows-x86_64.zip
 
 :: Set up MSVC
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x64
-set BAZEL_VS=C:\Program Files (x86)\Microsoft Visual Studio 14.0
-set BAZEL_VC=C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC
+call "C:\Program Files (x86)\Microsoft Visual Studio 16.0\VC\vcvarsall.bat" x64
```
s-perron, line 52: This test is still failing because this path is not being found. I'm in the middle of updating the VM that are used to be the same as the spirv-tools tests. Once that is done, you should be use the same commands that are in the spirv-tools repo to configure the comipler:

https://github.com/KhronosGroup/SPIRV-Tools/blob/01055c60cfc0cddd2724dc674ed73d09df992360/kokoro/scripts/windows/build.bat#L34

s-perron (changes requested):
This is stilling failing because it cannot find MSVC:

```
C:\tmpfs\src\github\glslang>call "C:\Program Files (x86)\Microsoft Visual Studio 16.0\VC\vcvarsall.bat" x64
The system cannot find the path specified.
```

To fix this, I am updating the VM that is used to make sure that VS2017 and VS2019 are installed. I'll let you know when that is done.

s-perron:
The VM should be updated soon. All of the config changes on our end are through. I might take a little time to fully propagate.

s-perron (approved):

s-perron:
After this merges, I'll open a follow up PR to rename the test to windows-msvc-2019-release-bazel.

The step should be

1. Copy the directory for the current test to a directory with the new name.
1. I do an internal change to create a new CI for the new directory, and delete the old one.
1. Delete the old directory in the GitHub repo.

PR #3184, opened 2023-04-09
Improve error message of alignment offset

spencer-lunarg:
The I thought there was a bug in GLSL but once I printed the values in the error message it made sense it was my shader

error message diff

```patch
- ERROR: struct.vert:16: 'offset' : must be a multiple of the member's alignment
+ ERROR: struct.vert:16: 'offset' : must be a multiple of the member's alignment (layout offset = 8 | member alignment = 16)
```


arcady-lunarg (approved):
LGTM, more helpful error messages are always nice to have.

PR #3185, opened 2023-04-09
Use custom callbacks if they are available in 'i->callbacks'

resetius:


resetius:
Usage example: https://github.com/resetius/fdm/blob/master/compute/vulkan/shader.cpp#L24

arcady-lunarg:
I rebased your PR manually to pick up some new CI configuration, and I will merge it after the CI runs complete successfully.

PR #3193, opened 2023-04-19
Fix building on OpenBSD when building shared libs

brad0:
OpenBSD does not link shared libs against libc so it is expected that the use of --no-undefined when linking will fail.


arcady-lunarg (changes requested):
Since you're already touching this code, would you mind deleting the CMAKE_VERSION checks since we already require a minimum cmake version of 3.14?

brad0:
Sure. Will do.

arcady-lunarg (approved):
LGTM, thanks.

PR #3203, opened 2023-05-03
glslang: Do not link with pthread on QNX

K3WLBUDDY:
On QNX, pthread APIs are provided as part of libc and a separate pthread library does not exist.

This change updates the makefile of glslang to not link with pthread on QNX, thereby enabling glslang to be built for QNX platforms.


dgkoch:
cc @arcady-lunarg

arcady-lunarg (approved):
LGTM, thanks for the PR.

PR #3208, opened 2023-05-04
cmake: Don't set CMAKE_INSTALL_PREFIX

### LINKED ISSUE #1015: CMakeLists.txt sets CMAKE_INSTALL_PREFIX

juan-lunarg:
While trying to set a default for CMAKE_INSTALL_PREFIX sounds appealing it has multiple issues.

It's particularly problematic for open source projects where many users try to accomplish many different things.

It's also conflicting with the new way of installing with cmake

IE: `cmake --install build --prefix build/install`

I've already removed similar logic in various other Khronos repositories a while ago to address similar complaints.

closes #1015


arcady-lunarg (approved):
LGTM

PR #3212, opened 2023-05-05
ci: Test GCC on MacOS CI

### LINKED ISSUE #3062: Error when building under macOS X with gcc
I am trying to build glslang under macOS x arm using gcc 12.2. It gives me the following error:
`aarch64-apple-darwin22-c++-12: error: unrecognized command-line option '-Xarch_arm64'`
Of course gcc uses the flag [-march=armv8-a](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/ARM-Options.html#ARM-Options) but it still tries to use the clang flag.
My build command looks like this:
```
cmake -DCMAKE_INSTALL_PREFIX=extern/glslang/build \                      
      -DCMAKE_CXX_COMPILER=aarch64-apple-darwin22-c++-12 -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_INSTALL_PREFIX=extern/glslang/build/install \
      -Sextern/glslang \
      -Bextern/glslang/build
```
The build file inside the build directory (`build/glslang/CMakeFiles/MachineIndependent.dir/build.make`) is the only one generate with these flags. A simple search and replace with the correct flag fixes the issue.

juan-lunarg:
closes #3062

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -66,14 +66,15 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [macos-11]
-                compiler: [{cc: clang, cxx: clang++}]
+                os: [macos-11, macos-12]
+                compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
                 cmake_build_type: [Debug, Release]
         steps:
-            - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


arcady-lunarg (approved):
I suppose this won't cost us any extra maintenance effort since we already support and test gcc on other platforms. And it's good to be testing on MacOS 12 as well.

PR #3219, opened 2023-05-09
Add windows-msvc-2019-release-bazel CI config

s-perron:
The `windows-msvc-2015-release-bazel` build config name is misleading.
It now runs VS2019. We should rename it. This is the first part: create
the new config files.


arcady-lunarg (approved):
Looks like a direct copy of the old config. LGTM.

PR #3220, opened 2023-05-09
Remove the windows-msvc-2015-release-bazel config

s-perron:
The VS2015 bazel test has been stopped, so we can not remove the config
file. We have the windows-msvc-2019-release-bazel to replace it. It is
already running, and it is passing.


arcady-lunarg (approved):

PR #3221, opened 2023-05-09
Correct ctor order to make '-Wreorder-ctor' pass

zoddicus:
These lines in the ctor being out of order are causing -Wreorder-ctor to fire when trying to integrate glslang ToT into Chromium/Dawn/etc, https://logs.chromium.org/logs/dawn/buildbucket/cr-buildbucket/8781562047251466593/+/u/compile_with_ninja/stdout.


zoddicus:
@arcady-lunarg can you review this? (or send to the correct person, if it isn't you)

arcady-lunarg (approved):
The change itself looks fine, although I should note that this warning is explicitly disabled in CMakeLists.txt. I suppose a follow-up PR to change that would be in order, which I will look into.

PR #3222, opened 2023-05-09
Enable constructor reorder warnings

arcady-lunarg:
They no longer happen "all over" as the comment suggests, and downstream builds have them enabled already.


arcady-lunarg:
See https://github.com/KhronosGroup/glslang/pull/3221 for motivation.

jeremy-lunarg (approved):
LGTM

PR #3224, opened 2023-05-11
`glslangValidator`: Propagage error from `glslang::OutputSpv[Hex|Bin]`

### LINKED ISSUE #3218: glslangValidator exits successfully if it cannot write output file
glslangValidator should exit unsuccessfully if it fails to write the output file, so that it can be integrated into build systems reliably. Currently, it exits successfully:

```
$ cat shader.frag
#version 460
void main() {}
$ ls
shader.frag
$ glslangValidator -V -o non-existent/shader.frag.spv shader.frag
shader.frag
ERROR: Failed to open file: non-existent/shader.frag.spv
$ echo "$?"
0
```

allanmac:
Closes #3218.

Also it appears `KhronosGroup:main` acquired some trailing spaces from a previous commit.


arcady-lunarg (changes requested):
Hi, could you separate out the whitespace fixes into a separate commit? Also, I would suggest rewording the commit description to more clearly state what it is doing. The fix itself is good though, thanks for the PR.

allanmac:
> Hi, could you separate out the whitespace fixes into a separate commit? Also, I would suggest rewording the commit description to more clearly state what it is doing. The fix itself is good though, thanks for the PR.

#3232

arcady-lunarg (approved):
LGTM

PR #3226, opened 2023-05-16
Update known_good.json

jeremy-lunarg:


arcady-lunarg (approved):
LGTM. This picks up a spirv-opt fix for the scalar-replacement pass, as it happens.

PR #3232, opened 2023-05-22
[glslang] Strip trailing whitespace

allanmac:
There are many other files in the repo that have trailing whitespace but this PR only cleans `glslang/SPIRV` and `glslang/Standalone`.

No change in behavior.


arcady-lunarg (approved):
LGTM, thanks for the cleanup.

PR #3236, opened 2023-06-06
Fix -Wmaybe-uninitialized warnings

svenvh:
Fix some potentially uninitialized uses that are reported by GCC 13 with `-O3`.


arcady-lunarg (approved):
LGTM, thanks for the fix.

PR #3242, opened 2023-06-17
Rework how auto push_constant upgrading works a bit.

mbechard:
Ensure we traverse the entire tree and upgrade all references to the given symbol so it can be upgraded to push_constant. Without this change only one instance was upgraded, and others were left as uniform buffers.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -145,6 +145,8 @@ public:
             base->getWritableType().getQualifier().layoutComponent = at->second.newComponent;
         if (at->second.newIndex != -1)
             base->getWritableType().getQualifier().layoutIndex = at->second.newIndex;
+        if (at->second.upgradedToPushConstant)
+            base->getWritableType().getQualifier().layoutPushConstant = true;
     }
 
   private:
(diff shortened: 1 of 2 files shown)


mbechard:
Sorry to bump this, looking to upgrade my GLSLang version and would like to use the main branch.
Thanks

glslang/MachineIndependent/iomapper.cpp
```diff
@@ -1670,31 +1672,34 @@ bool TGlslIoMapper::doMap(TIoMapResolver* resolver, TInfoSink& infoSink) {
...
             if (upgraded) {
-                while (1) {
-                    auto at = std::find_if(uniformVector.begin(), uniformVector.end(),
-                                           [this](const TVarLivePair& p) { return p.first == autoPushConstantBlockName; });
-                    if (at != uniformVector.end())
-                        uniformVector.erase(at);
-                    else
-                        break;
-                }
+                auto at = std::for_each(uniformVector.begin(), uniformVector.end(),
```
ncesario-lunarg, line 1679: Is `at` still used?

ncesario-lunarg (commented):
Are there tests that could be added to verify the new behavior?

glslang/MachineIndependent/iomapper.cpp
mbechard: Thanks, removed

mbechard:
I don't really have a great testcase... this showed up in a pretty complex shader. To make the error case happen the AST needs to have multiple node that reference the same symbol, which isn't easy to make happen on-demand. You will notice that this workflow better matches all of the other code that's doing similar work in this function though

arcady-lunarg (approved):
LGTM

PR #3248, opened 2023-06-22
Implement support for GL_KHR_cooperative_matrix extension

archimedus:
Add into GLSLang support for GL_KHR_cooperative_matrix extension required for Vulkan extension VK_KHR_cooperative_matrix.


dneto0:
The support has landed in both SPIRV-Headers and SPIRV-Tools https://github.com/KhronosGroup/SPIRV-Tools/pull/5286 

Now the known-good.json can be updated with those commits.

Thanks for your patience!

archimedus:
Can someone give another review try for it please?

Thank you.

dneto0:
> Can someone give another review try for it please?

Someone has started reviewing.  Not me though.

archimedus:
> Someone has started reviewing. Not me though.

I see. Thank you for informing me.

BTW could you run checks or it is a reviewer area now?

archimedus:
BTW In "Reviewers" area I see "No reviews" on right pane.

dneto0:
I heard that the review will be delayed until Monday at the earliest.  Thanks for your patience.

archimedus:
Any progress?

arcady-lunarg:
I am going to take a look at this right now. My apologies, I had hoped to get to this sooner but I had Friday off and had some travel difficulties yesterday that meant I could not get to it until now.

archimedus:
@arcady-lunarg 

Thank you for starting review.

I see two tests failed:
 * Glsl/CompileVulkanToSpirvTest.FromFile/spv_coopmatKHR_comp
 * Glsl/CompileToAstTest.FromFile/stringToDouble_vert

The first requires dungling space (3x) at end-of-line, I will add.
```
             214:         210 CooperativeMatrixMulAddKHR 205 209 213 ASignedComponents BSignedComponents CSignedComponents ResultSignedComponents<SPACE>
```

The second seems does not relate to this change:

real:
```
0:10            1.2345678901234e+15
```
expected:
```
0:10            1.2345678901235e+15
```

Should I update this as well?

archimedus:
Now Google Tests should pass.

arcady-lunarg:
Looks like the tests now fail on other platforms. Might the change of bison version be responsible for that? Also, it looks like some of the files have line ending issues, I would prefer to stick to unix-style line endings where possibly unless an existing file already uses windows-style throughout.

archimedus:
> Looks like the tests now fail on other platforms.

I checked that issues in `stringToDouble_vert` are present in parent and thus are not introduced in this PR.

> Might the change of bison version be responsible for that?

Doubt it. I see that root problem is in code:

```
static void OutputDouble(TInfoSink& out, double value, TOutputTraverser::EExtraOutput extra)
{
        if (fabs(value) > 0.0 && (fabs(value) < 1e-5 || fabs(value) > 1e12))
            format = "%-.13e";
...
        int len = snprintf(buf, maxSize, format, value);
```

For me issue is in incorrect rounding in sprintf. The value `1.2345678901234e+15` looks like more proper than value `1.2345678901235e+15`, because it is [rounded to nearest even](https://stackoverflow.com/questions/6867693/change-floating-point-rounding-mode), though it is only IMO. I undone my changes and tried various rounding modes (see [FE_round](https://en.cppreference.com/w/c/numeric/fenv/FE_round)), though none of them resolved issue. I suppose @johnkslang is more appropriate person to ask about how to fix this.

> Also, it looks like some of the files have line ending issues, I would prefer to stick to unix-style line endings where possibly unless an existing file already uses windows-style throughout.

Fixed.

archimedus:
@arcady-lunarg 

Any progress?

archimedus:
It looks formal Checks pass now.

Waiting for reviewers opinion.

arcady-lunarg:
This is a very large change, but I am taking a look at it. In the future, to facilitate reviews it would help to write more descriptive commit messages as to what you are doing and why, and maybe also to break up the change into more manageable commits. For example, there seems to be a lot of duplicated code between the NV and KHR versions of the extension. Now, I know that they differ in some important ways, but it feels like there are opportunities for sharing more code between the two.

arcady-lunarg:
Also, it appears that the GLSL extension for this has not been published yet, so in any case I don't think this can be merged until that is published.

arcady-lunarg:
Having finally gone through all the changes, I think this is basically fine, though I do wonder about whether it might be possible to have more shared code between the NV and KHR versions. Also it would be nice to have better commit messages going forward but I'm not going to hold up this change for that. What I will hold it up for is the release of the GLSL extension spec, which I assume is coming soon.

archimedus:
@arcady-lunarg 

Thank you for review.

> and maybe also to break up the change into more manageable commits.

Creating the PR I squashed all commits that have been pushed during development. I was vague doing this, for two reasons:
1. This removes history and particular contributions
2. As far as I understand one squashed commit is more expected than a set of small ones.
I am not sure whether this can be splitted onto smaller commits as this makes sense as a whole.

> For example, there seems to be a lot of duplicated code between the NV and KHR versions of the extension. Now, I know that they differ in some important ways, but it feels like there are opportunities for sharing more code between the two.

I would avoid deduplication as looking at discussions around cooperative matrix subject I think KHR will diverge even more soon from NV version.

> Also, it appears that the GLSL extension for this has not been published yet, so in any case I don't think this can be merged until that is published.

You are right. I pinged the GLSL spec issue as well.

I updated name and description of this PR. I hope these will be used during squash/merge of the PR.

nihui:
GLSL extension: https://github.com/KhronosGroup/GLSL/pull/208

archimedus:
> FYI this is going to need to be rebased as there was another PR with a grammar change in the interim.

Should I do it now or wait for GLSL spec accepted first?

arcady-lunarg:
I suppose it's better to wait for the spec to be accepted but I don't really have any idea of when that will happen.

jeffbolznv:
FYI I've just merged the extension spec.

arcady-lunarg:
@archimedus now that the spec has landed, could you rebase this? @jeffbolznv I assume there weren't any substantial changes to the spec in the review process?

jeffbolznv:
There were no substantial changes.

archimedus:
Rebased.

arcady-lunarg (approved):
LGTM, thanks.

PR #3249, opened 2023-06-22
Force generateDebugInfo when non-semantic debug info is enabled

arcady-lunarg:
From the command line, the debug options "stack", with -gVS enabling all of generateDebugInfo, emitNonSemanticShaderDebugInfo and emitNonSemanticShaderDebugSource, however the programmatic interface allows setting the latter options without the former. In this case, the string corresponding to the source filename never gets emitted and some debuginfo instructions end up with zero ID operands, resulting in invalid SPIR-V.


jeremy-lunarg (approved):
LGTM

PR #3254, opened 2023-07-05
Spirv_intrinsics: Remove early return in layoutTypeCheck

amdrexu:
Previously, when GL_EXT_spirv_intrinsics are enabled, we disable all checks in layoutTypeCheck. This is too coarse because we can use nothing in GL_EXT_spirv_intrinsics in a shader while the necessary processing is skipped, such as addUsedLocation.

In this change, we apply fine check and more might be added if we encounter new cases in the future.


arcady-lunarg:
What use case does this allow? Are there tests you could add that would check this newly improved behavior?

amdrexu:
> What use case does this allow? Are there tests you could add that would check this newly improved behavior?

Sorry. I missed this review comment. Previously, we always use early-return when `GL_EXT_spirv_intrinsics` is enabled. This is inaccurate and will cause some necessary error check to be ignored. This change is to resolved the problem and make error check still work for types other than those defined in `GL_EXT_spirv_intrinsics`. I add a simple test case to show the problem.

Test/overlappedUniformLocation.vert
```diff
@@ -0,0 +1,10 @@
+#version 460 core
+#extension GL_EXT_spirv_intrinsics : enable
+
+// ERROR: Overlapped uniform location. Make sure it could be detected even
+// if GL_EXT_spirv_intrinsics is enabled.
+layout(location = 0) uniform vec4 v4;
```
arcady-lunarg, line 6: If you use "uniform" here, it returns a different error without your checks, I think it might be better to use "in" for the two vars.

arcady-lunarg (changes requested):
Thanks, the test makes it very clear what this is solving. Just one quick change I'd like to see on the test and this is good to go.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -6436,7 +6434,11 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
         // an array of size N, all elements of the array from binding through binding + N - 1 must be within this
         // range."
         //
-        if (! type.isOpaque() && type.getBasicType() != EbtBlock)
+#ifndef GLSLANG_WEB
```
arcady-lunarg, line 6437: Oh and we got rid of GLSLANG_WEB, so you can remove the ifdeffery here.

amdrexu: Thank you. I remove the code.

Test/overlappedUniformLocation.vert
amdrexu: I update the test.

arcady-lunarg (approved):
LGTM, thanks.

PR #3255, opened 2023-07-07
Fix C example in README

### LINKED ISSUE #3239: C interface example in README is not in C
The example in the README for how to use the functions in `glslang_c_interface.h` is not itself written in C but rather in C++ with use of things like std::vector. It should really be in C so someone could just copy and paste it into a C project.

ncesario-lunarg:
Remove the usage of C++ STL types in the C-only interface example.

Fixes #3239.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,9 +425,18 @@ warning/error and other options for controlling compilation.
 
 This interface is located `glslang_c_interface.h` and exposes functionality similar to the C++ interface. The following snippet is a complete example showing how to compile GLSL into SPIR-V 1.5 for Vulkan 1.2.
 
-```cxx
-std::vector<uint32_t> compileShaderToSPIRV_Vulkan(glslang_stage_t stage, const char* shaderSource, const char* fileName)
-{
+```c
+#include <glslang/Include/glslang_c_interface.h>
+
+// Required for use of glslang_default_resource
+#include <glslang/Public/resource_limits_c.h>
+
+typedef struct SpirVBinary {
+    uint32_t *words; // SPIR-V words
+    int size; // number of words in SPIR-V binary
+} SpirVBinary;
+
+SpirVBinary compileShaderToSPIRV_Vulkan(glslang_stage_t stage, const char* shaderSource, const char* fileName) {
     const glslang_input_t input = {
         .language = GLSLANG_SOURCE_GLSL,
         .stage = stage,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -425,9 +425,21 @@ warning/error and other options for controlling compilation.
 
 This interface is located `glslang_c_interface.h` and exposes functionality similar to the C++ interface. The following snippet is a complete example showing how to compile GLSL into SPIR-V 1.5 for Vulkan 1.2.
 
-```cxx
-std::vector<uint32_t> compileShaderToSPIRV_Vulkan(glslang_stage_t stage, const char* shaderSource, const char* fileName)
-{
+```c
+// Included for uint32_t
```
arcady-lunarg, line 429: You don't need to include this since you're already not including stdio.h and people can figure out where to get uint32_t on their own.

README.md
```diff
@@ -441,18 +453,22 @@ std::vector<uint32_t> compileShaderToSPIRV_Vulkan(glslang_stage_t stage, const c
         .force_default_version_and_profile = false,
         .forward_compatible = false,
         .messages = GLSLANG_MSG_DEFAULT_BIT,
-        .resource = reinterpret_cast<const glslang_resource_t*>(&glslang::DefaultTBuiltInResource),
+        .resource = glslang_default_resource(),
     };
 
     glslang_shader_t* shader = glslang_shader_create(&input);
 
+    SpirVBinary bin = {
```
arcady-lunarg, line 461: I like this way of doing it.

README.md
```diff
@@ -473,22 +489,24 @@ std::vector<uint32_t> compileShaderToSPIRV_Vulkan(glslang_stage_t stage, const c
...
-    glslang_program_SPIRV_get(program, outShaderModule.data());
+    bin.size = glslang_program_SPIRV_get_size(program);
+    bin.words = malloc(bin.size * sizeof(uint32_t));
+    glslang_program_SPIRV_get(program, bin.words);
 
     const char* spirv_messages = glslang_program_SPIRV_get_messages(program);
-    if (spirv_messages)
+    if (spirv_messages) {
         printf("(%s) %s\b", fileName, spirv_messages);
+    }
```
arcady-lunarg, line 504: No need to add these braces IMO.

arcady-lunarg (changes requested):
Just some minor nits, feel free to disagree with my suggestions. Thanks for taking care of this, I did test the sample code and it works.

README.md
ncesario-lunarg: I'll leave them off if that's the preferred style.

ncesario-lunarg:
> Just some minor nits, feel free to disagree with my suggestions. Thanks for taking care of this, I did test the sample code and it works.

Thanks for testing it! I tried it out on clang 16 and gcc 13, but not older compilers.

arcady-lunarg (approved):
LGTM

Issue #3092, opened 2022-12-14
HLSL: Order of const keyword can break parsing


TheJackiMonster:
While trying to compile some HLSL shaders I encountered the following issue with the `const` keyword using for local variables and function arguments.

If you use following syntax, everything is fine:
```
const float f = 1.0f;
```

But using the following syntax will throw an error by the HLSL parser:
```
float const f = 1.0f;
```

Unfortunately some shaders are written by people crosing both variants. ^^'

Another example how the current parser seems to evaluate code with functions:

GOOD:
```
void test(const float3 f) {}
```

BAD:
```
void test(float3 const f) {}
```

greg-lunarg:
Yes, this is an issue. dxc accepts `float const f = 1.0f;`

arcady-lunarg:
Fixed by #3256.

PR #3256, opened 2023-07-10
HLSL keyword declaration after declared type

Dawid-Lorenz-Mobica:
Added the ability to define a variable as const after declaring the type of the variable

Fix for Issue: https://github.com/KhronosGroup/glslang/issues/3092


arcady-lunarg (approved):
LGTM. The idea of a function returning bool that can't return false is kind of odd to me but it follows the existing code so I suppose that is fine.

PR #3257, opened 2023-07-10
cmake: Rename glslang to glslangValidator

### LINKED ISSUE #47: Change "glslangValidator" -> "glslang"
glslangValidator is becoming a SPIR-V front end, as well as a validator.  Khronos decided it should be name just "glslang".  So this change in the built executable name will likely occur at some point.

Are there any concerns about doing this?

ncesario-lunarg:
Rename glslangValidator to glslang and adds a glslangValidator symlink to the build and install directories.

Closes #47.

### DIFF
--- a/StandAlone/CMakeLists.txt
+++ b/StandAlone/CMakeLists.txt
@@ -45,13 +45,12 @@ add_custom_command(
     DEPENDS ${GLSLANG_INTRINSIC_PY}
     COMMENT "Generating ${GLSLANG_INTRINSIC_H}")
 
-#add_custom_target(glslangValidator DEPENDS ${GLSLANG_INTRINSIC_H})
-
 set(SOURCES StandAlone.cpp DirStackFileIncluder.h  ${GLSLANG_INTRINSIC_H})
 
-add_executable(glslangValidator ${SOURCES})
-set_property(TARGET glslangValidator PROPERTY FOLDER tools)
-glslang_set_link_args(glslangValidator)
+add_executable(glslang-standalone ${SOURCES})
+set_property(TARGET glslang-standalone PROPERTY FOLDER tools)
+set_property(TARGET glslang-standalone PROPERTY OUTPUT_NAME glslang)
+glslang_set_link_args(glslang-standalone)
 
 set(LIBRARIES
     glslang
@@ -70,13 +69,13 @@ elseif(UNIX)
     endif()
 endif()
 
-target_link_libraries(glslangValidator ${LIBRARIES})
-target_include_directories(glslangValidator PUBLIC
+target_link_libraries(glslang-standalone ${LIBRARIES})
+target_include_directories(glslang-standalone PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External>)
 
 if(ENABLE_OPT)
-    target_include_directories(glslangValidator
+    target_include_directories(glslang-standalone
         PRIVATE ${spirv-tools_SOURCE_DIR}/include
     )
 endif()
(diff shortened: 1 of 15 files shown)


ncesario-lunarg:
I'm not familiar with Bazel. Is it really not possible to have the output name of a target be different from the name of a target?

ncesario-lunarg:
This change depends on #3262.

StandAlone/StandAlone.cpp
```diff
@@ -1884,7 +1884,7 @@ void CompileFile(const char* fileName, ShHandle compiler)
 //
 void usage()
 {
-    printf("Usage: glslangValidator [option]... [file]...\n"
+    printf("Usage: glslang [option]... [file]...\n"
```
arcady-lunarg, line 1887: Would it make sense to print argv[0] here instead of a fixed name? In any case, I think that can be done in a followup change.

ncesario-lunarg: I had the same thought, but left it out to keep this change focused. That said, that's a pretty trivial change, so happy to add that to this PR if that's ok with you.

StandAlone/CMakeLists.txt
```diff
@@ -94,19 +95,31 @@ if(WIN32)
...
             include(\"${CMAKE_INSTALL_FULL_LIBDIR}/cmake/${PROJECT_NAME}/glslang-targets.cmake\")
         endif()
 
-        add_library(glslangValidator ALIAS glslang::glslangValidator)
+        add_library(glslang-standalone ALIAS glslang::glslang-standalone)
     ")
-    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/glslangValidatorTargets.cmake" DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/glslang-standaloneTargets.cmake" DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
+
+    # Create a symbolic link to glslang named glslang-standalone for backwards compatibility
```
arcady-lunarg, line 112: named glslangValidator

StandAlone/StandAlone.cpp
```diff
@@ -2089,8 +2089,7 @@ void usage()
            "  --variable-name <name>\n"
            "  --vn <name>                       creates a C header file that contains a\n"
            "                                    uint32_t array named <name>\n"
-           "                                    initialized with the shader binary code\n"
-           );
+           "                                    initialized with the shader binary code\n");
```
arcady-lunarg, line 2092: I think putting the closing paren on the last line was intentional to minimize spurious diffs when merging.

StandAlone/CMakeLists.txt
```diff
@@ -45,13 +45,14 @@ add_custom_command(
     DEPENDS ${GLSLANG_INTRINSIC_PY}
     COMMENT "Generating ${GLSLANG_INTRINSIC_H}")
 
-#add_custom_target(glslangValidator DEPENDS ${GLSLANG_INTRINSIC_H})
+#add_custom_target(glslang-standalone DEPENDS ${GLSLANG_INTRINSIC_H})
```
arcady-lunarg, line 48: Probably better to just delete this commented-out line rather than updating it?

arcady-lunarg (commented):
Just a couple of minor comment nits, otherwise this is good to go.

StandAlone/CMakeLists.txt
ncesario-lunarg: I would definitely prefer removing "dangling comments," I just wasn't familiar with history on this :) Happy to remove it.

StandAlone/StandAlone.cpp
ncesario-lunarg: Apparently I did leave this alone from before, but clang-format automatically changes it. I do agree having the `)` on a separate line reduces spurious diffs, but it would also be nice to be able to run clang-format on the file. I can try to find out if there's a clang-format setting that can be used to prevent this from happening.

arcady-lunarg (approved):

ioquatix:
Nice work!

StandAlone/StandAlone.cpp
ncesario-lunarg: I did play around with the clang-format settings, but wasn't able to get something working correctly.

dneto0:
I would not have made this change.  It is a breaking change to *remove* the old name.

I know there's an effort to make a symlink for the old name, but that only takes effect on Linux, if I read the CMake documentation correctly.

Issue #3089, opened 2022-12-12
Constant unsigned integer expressions in HLSL unsupported


TheJackiMonster:
I was trying to compile the shadow denoiser shaders from [here](https://github.com/GPUOpen-Effects/FidelityFX-Denoiser) with Glslang but I ran into an issue:

```
[ERROR]: Shader parsing failed {
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: ':' : Expected 
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: 'initializer' : Expected 
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: 'b10u' : unknown variable 
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: ';' : Expected 
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: 'then statement' : Expected 
ERROR: /tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h:293: 'declaration' : Expected 
/tmp/vkcv_tmp_VlNB6M/ffx_denoiser_shadows_tileclassification.h(293): error at column 47, HLSL parsing failed.
ERROR: 7 compilation errors.  No code generated.
```

For context: I compile shaders in a temporary directory to handle include paths without the need of shipping the shaders in local directories as permanent files.

Anyway the causing problem is that these HLSL shaders utilize constant expression for their bitmasks like this:
```
#define TILE_META_DATA_CLEAR_MASK 0b01u
#define TILE_META_DATA_LIGHT_MASK 0b10u
```

These constant expressions seem not be handled properly though. Because the parser tells me, it wouldn't expect such an expression or it tries to parse it as a variable which it can't find. I'm usually not working with HLSL. So I'm not sure whether these expressions are in spec or not. But I assume it should be interpreted as binary values (0b01u -> 1) and (0b10u -> 2).

greg-lunarg:
Yes, you have found an issue. The following is correctly compiled by dxc but glslang gives errors as above:

float4 MainPs(void) : SV_Target0
{
      float4 vTest = float4(0.0,0.0,0.0,0.0);
[...]

PR #3259, opened 2023-07-12
HLSL: Possibility to represent variable in binary form

Dawid-Lorenz-Mobica:
Fix for Issues: https://github.com/KhronosGroup/glslang/issues/3089


Test/hlsl.conditional.frag
```diff
@@ -18,9 +18,9 @@ float4 scalarCond()
     return ret;
 }
 
-float2 fbSelect(bool2 cnd, float2 src0, float2 src1)
-{
-    return cnd ? src0 : src1;
+float2 fbSelect(bool2 cnd, float2 src0, float2 src1)
```
arcady-lunarg, line 21: This looks like a line-ending diff, please revert.

Test/hlsl.conditional.frag
```diff
@@ -1,8 +1,8 @@
-float4 c4;
-float4 t4;
-float4 f4;
-float t;
-float f;
+float4 c4;
```
arcady-lunarg, line 1: This looks like a line-ending diff, please revert.

glslang/MachineIndependent/preprocessor/PpScanner.cpp
```diff
@@ -639,6 +640,110 @@ int TPpContext::tStringInput::scan(TPpToken* ppToken)
...
+                bool isInt16 = false;
+                ppToken->name[len++] = (char)ch;
+                ch = getch();
+
+                // Check value
+                if ((ch == '0' || ch == '1'))
+                {
+                    ival = 0;
+                    do {
+                        if (len < MaxTokenLength && ival <= 0x0fffffffffffffffull) {
```
arcady-lunarg, line 656: Why 0x0fffffffffffffffull and not 0x7fffffffffffffffull?

arcady-lunarg (changes requested):
Looks okay overall, aside from a couple issues mentioned on specific lines, and one major one: this also adds support for binary literals to GLSL, which is not actually a thing that GLSL supports. You need to add a check to make sure that we're not compiling GLSL and to bail out early in that case. I would also like to see tests for the error cases here, both the "literal too big/too long" error messages and that this doesn't work with GLSL. There should be existing "negative" test files to which you can add some new cases.

Test/hlsl.conditional.frag
Dawid-Lorenz-Mobica: Done.

Test/hlsl.conditional.frag
Dawid-Lorenz-Mobica: Done.

glslang/MachineIndependent/preprocessor/PpScanner.cpp
Dawid-Lorenz-Mobica: Yes, Right. I did copy-paste from the hexadecimal option

Test/baseResults/hlsl.numericsuffixes.frag.out
```diff
@@ -99,194 +144,136 @@ Linked fragment stage:
...
-                              Store 38(r12) 37
-              42:     24(int) Load 30(r07)
-              43:    6(float) ConvertSToF 42
-              44:    7(fvec4) CompositeConstruct 43 43 43 43
-              46:     45(ptr) AccessChain 40(ps_output) 41
-                              Store 46 44
-              47:8(PS_OUTPUT) Load 40(ps_output)
-                              ReturnValue 47
-                              FunctionEnd
+SPIR-V is not generated for failed compile or link
```
arcady-lunarg, line 279: I don't like that this changes the test to no longer generate any SPIR-V. Could you move the error cases to a separate file so that SPIR-V is still generated here?

arcady-lunarg (changes requested):
The code itself looks fine now, just one issue with your change to a test causing it to no longer generate SPIR-V code where it had previously done so. I think you need to move the negative tests to a different file, possibly a new test file.

Test/baseResults/hlsl.numericsuffixes.frag.out
Dawid-Lorenz-Mobica: Done.

arcady-lunarg (approved):
LGTM, thanks.

PR #3260, opened 2023-07-13
Use std::call_once in spv::Parameterize()

### LINKED ISSUE #342: Thread safety issue in spirvbin_t (race condition in Parameterize function)

arcady-lunarg:
There was a race condition in this function as it used a static variable to attempt to ensure global initialization was only done once, which was not thread-safe. Instead, use std::call_once, which was added to C++11 for this exact case.

Fixes #342

### DIFF
--- a/SPIRV/doc.cpp
+++ b/SPIRV/doc.cpp
@@ -45,6 +45,7 @@
 #include <cstdio>
 #include <cstring>
 #include <algorithm>
+#include <mutex>
 
 namespace spv {
     extern "C" {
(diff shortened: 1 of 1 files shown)


jeremy-lunarg (approved):
LGTM

PR #3262, opened 2023-07-14
build: Remove bazel build support

ncesario-lunarg:
This PR removes support for building with bazel. To our knowledge, the bazel build is no longer used.

FYI @dnovillo @s-perron.


ncesario-lunarg:
I'm assuming all of the google runners are bazel builds, and should no longer be triggered, but I'm not sure. I do not have permission to modify these settings.

s-perron:
Not all of the kokoro builds are bazel. You can tell from the name if it is cmake or bazel.

s-perron:
We'll take care of this on the Google side on Monday.

arcady-lunarg (approved):

PR #3264, opened 2023-07-18
cmake: Fix Android build for r25 NDK

juan-lunarg:
Currently with the build instructions provided in README.md the build will fail.

In the r25 NDK the CMake toolchain defaults to the legacy path, due to a bug in the current implementation.

https://github.com/android/ndk/issues/323


juan-lunarg:
Note as a result of this minor issue the validation layers currently have to set `ANDROID_USE_LEGACY_TOOLCHAIN_FILE=OFF` in their CMake Android builds.

arcady-lunarg (approved):

dneto0:
I'm confused. Is that the right Android NDK bug  ?

juan-lunarg:
> I'm confused. Is that the right Android NDK bug ?

Here is the CMake code in question causing the issue in `/home/juan/Android/Sdk/ndk/25.2.9519653/build/cmake/android.toolchain.cmake`

```
# CMake invokes the toolchain file twice during the first build, but only once
# during subsequent rebuilds. This was causing the various flags to be added
# twice on the first build, and on a rebuild ninja would see only one set of the
# flags and rebuild the world.
# https://github.com/android-ndk/ndk/issues/323
if(ANDROID_NDK_TOOLCHAIN_INCLUDED)
  return()
endif(ANDROID_NDK_TOOLCHAIN_INCLUDED)
set(ANDROID_NDK_TOOLCHAIN_INCLUDED true)

if(DEFINED ANDROID_USE_LEGACY_TOOLCHAIN_FILE)
  set(_USE_LEGACY_TOOLCHAIN_FILE ${ANDROID_USE_LEGACY_TOOLCHAIN_FILE})
else()
  # Default to the legacy toolchain file to avoid changing the behavior of
  # CMAKE_CXX_FLAGS. See https://github.com/android/ndk/issues/1693. <-------------
  set(_USE_LEGACY_TOOLCHAIN_FILE true)
endif()
if(_USE_LEGACY_TOOLCHAIN_FILE)
  include("${CMAKE_CURRENT_LIST_DIR}/android-legacy.toolchain.cmake")
  return()
endif()
```

PR #3267, opened 2023-07-19
Simplify PoolAlloc by using `thread_local`

### LINKED ISSUE #1429: Crash in glslang::SetThreadPoolAllocator
Hi

I try to use many thread to compile and link, then I found it's always crash in link.

I think it lost InitThread in TProgram::link(EShMessages messages)

*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'samsung/starltexx/starlte:8.0.0/R16NW/G960FXXU1BRE5:user/release-keys'
Revision: '26'
ABI: 'arm'
pid: 23338, tid: 23363, name: android.test>>> com.android.test<<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
Cause: null pointer dereference
    r0 0000001a  r1 00000000  r2 993cee9b  r3 00000030
    r4 de7fd260  r5 00000000  r6 cdff8108  r7 e5c05930
    r8 00000002  r9 e5c05930  sl cb1e9a88  fp 00000058
    ip f09b8d78  sp ce694818  lr cdf51fad  pc cdf51fac  cpsr 600f0030

backtrace:
    #00 pc 0008efac  /data/app/com.android.test-tubfI2WzdA4UcOmz32NSsQ==/lib/arm/libglslang.so

Thank you

arcady-lunarg:
This is largely a revert of the revert of #2348, which was reverted due to not working on iOS 8.

C++11 supports thread_local storage class, which means we can drop the use of OS-bespoke OS_*TLSIndex functions.

This greatly simplifies fixing #2346.

Fixes #1429


jeremy-lunarg (approved):
LGTM

Issue #3270, opened 2023-07-19
support MinGW build with Clang, not just GCC


dneto0:
A MinGW build requires extra link flags.
Currently, the MinGW build is supported with GCC as the compiler, but not Clang.
The recommendation is to check    `WIN32 AND NOT MSVC`

See https://github.com/android/ndk/issues/1464

dneto0:
I plan to post a PR to fix this.

dneto0:
Related to https://github.com/google/shaderc/issues/1343

PR #3271, opened 2023-07-19
Support MinGW build with Clang, not just GCC

dneto0:
Fixes: #3270


arcady-lunarg (approved):
LGTM, though we don't test MinGW in our CI this looks safe.

PR #3275, opened 2023-07-20
Remove obsolete files

arcady-lunarg:
WORKSPACE is related to Bazel, which is no longer supported, Test/makeDoc uses an option that glslang no longer supports.


ncesario-lunarg (approved):

jeremy-lunarg (approved):
LGTM

PR #3283, opened 2023-07-23
respect destdir for compat symlink

### LINKED ISSUE #3281: glslangValidator symlink cannot be created in a distro package build environment
Version: 12.3.1

```
CMake Error: failed to create symbolic link 'glslangValidator': Read-only file system
```

Basically, cmake tries to symlink into /usr/bin/glslangValidator -> glslang, but it completely ignores $DESTDIR. In a distro packaging environment DESTDIR will be set somewhere (in my case, to /buildstream-install/), and the rest of the system (including /usr) will be mounted read-only.

TimB87:
fixes #3281

### DIFF
--- a/StandAlone/CMakeLists.txt
+++ b/StandAlone/CMakeLists.txt
@@ -121,7 +121,7 @@ if(ENABLE_GLSLANG_INSTALL)
     # Create the same symlink at install time
     install(CODE "execute_process( \
                       COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name} \
-                      WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
+                      WORKING_DIRECTORY \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
 
     if(ENABLE_SPVREMAPPER)
         install(TARGETS spirv-remap EXPORT glslang-targets)


ncesario-lunarg (approved):

StandAlone/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ if(ENABLE_GLSLANG_INSTALL)
     # Create the same symlink at install time
     install(CODE "execute_process( \
                       COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name} \
-                      WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
+                      WORKING_DIRECTORY \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
```
ncesario-lunarg, line 124: Why does the `$` need to be escaped here?

ncesario-lunarg: Pinging @juan-lunarg in case he has a moment to provide any feedback.

TimB87: @ncesario-lunarg I have honestly no idea how cmake works but this is supposedly how it does - it works for me in my build environment and thats why I opened this pr

ncesario-lunarg: Looking at the [documentation](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html), I think this is correct (although I still don't understand the escaped `$`).

Although, looking at [$<INSTALL_PREFIX>](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:INSTALL_PREFIX) I wonder if using that is the "correct" way to fix this.

ncesario-lunarg: Nope, trying this out locally, `$<INSTALL_PREFIX>` does not include `$ENV{DESTDIR}`. So this change LGTM, just without the escaped `$` (works for me locally without the `\` prefixing `$ENV{DESTDIR}`).

TimB87: Not escaping `$ENV{DESTDIR}` leads to cmake writing it the symlink directly to `CMAKE_INSTALL_PREFIX`.

ncesario-lunarg (approved):

StandAlone/CMakeLists.txt
jbeich: Not escaping `$` leads to the variable being expanded at configure/build time. In some environments (e.g., FreeBSD Ports) `DESTDIR` may only be defined during install time.

I did a [similar fix downstream](https://github.com/freebsd/freebsd-ports/blob/cac78369bdcc/graphics/glslang/files/patch-StandAlone_CMakeLists.txt) but omitted `/` which is more common (including outside CMake) and doesn't break relative paths.

jbeich: StandAlone/cmake_install.cmake automatically created by CMake it suggests there should be no `/` to separate `DESTDIR` and `CMAKE_INSTALL_PREFIX`:
```cmake
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  if(EXISTS "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap" AND
     NOT IS_SYMLINK "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap")
    file(RPATH_CHECK
         FILE "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap"
         RPATH "")
  endif()
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/bin" TYPE EXECUTABLE FILES "/wrkdirs/usr/ports/graphics/glslang/work/.build/StandAlone/spirv-remap")
  if(EXISTS "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap" AND
     NOT IS_SYMLINK "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap")
    if(CMAKE_INSTALL_DO_STRIP)
      execute_process(COMMAND "/usr/bin/llvm-strip" "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/spirv-remap")
    endif()
  endif()
endif()
```

StandAlone/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ if(ENABLE_GLSLANG_INSTALL)
     # Create the same symlink at install time
     install(CODE "execute_process( \
                       COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name} \
-                      WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
```
juan-lunarg, line 124: 👎🏾 this logic is invalid. `CMAKE_INSTALL_BINDIR` can be an absolute path.

StandAlone/CMakeLists.txt
ncesario-lunarg: Talking with @juan-lunarg offline, it sounds like it would be better to just copy on all platforms. It also sounds like `CMAKE_INSTALL_BINDIR` can be an absolute path (a separate issue, but technically not correct in its current form). So to address both of these issues, something like the following should perhaps be done:
```diff
diff --git a/StandAlone/CMakeLists.txt b/StandAlone/CMakeLists.txt
index d1460630..2fbc533e 100644
--- a/StandAlone/CMakeLists.txt
+++ b/StandAlone/CMakeLists.txt
@@ -109,19 +109,21 @@ if(ENABLE_GLSLANG_INSTALL)

     # Create a symbolic link to glslang named glslangValidator for backwards compatibility
     set(legacy_glslang_name "glslangValidator${CMAKE_EXECUTABLE_SUFFIX}")
-    set(link_method create_symlink)
-    if (WIN32 OR MINGW)
-        set(link_method copy_if_different)
-    endif()
+    set(link_method copy_if_different)
     add_custom_command(TARGET glslang-standalone
                        POST_BUILD
                        COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name}
                        WORKING_DIRECTORY $<TARGET_FILE_DIR:glslang-standalone>)

     # Create the same symlink at install time
+    if (IS_ABSOLUTE ${CMAKE_INSTALL_BINDIR})
+        set(install_bindir ${CMAKE_INSTALL_BINDIR})
+    else()
+        set(install_bindir "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}")
+    endif()
     install(CODE "execute_process( \
                       COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name} \
-                      WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
+                      WORKING_DIRECTORY ${install_bindir})")

     if(ENABLE_SPVREMAPPER)
         install(TARGETS spirv-remap EXPORT glslang-targets)
```
Would that work for you @TimB87?

TimB87: I mean, thats a shotgun method, but it doesn't hurt and works for me as well

ncesario-lunarg: I'm ok leaving the symlinks in, it just sounded like it would be easier to test without them.

TimB87: My only feeling is that copying the file is wasteful while a symlink is the most elegant solution. But if it makes the cmake system to cumbersome just go with copying it, I wouldn't mind either way :+1:

TimB87:
> Why do we need a "compatibility" symlink? Why not just create a wrapping script named "glslangValidator" which would print a helpful string when invoked, saying that the program has been renamed to "glslang" and this compatibility script may be removed in a newer version?

Current mesa may complain if it doesn't find this: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/meson.build#L555

ncesario-lunarg:
@arcady-lunarg If you're ok with this, I think this is good to go.

@juan-lunarg pointed out [this](https://discourse.cmake.org/t/how-to-properly-install-a-symbolic-link-for-backwards-compatibility/8592/2) post, and I can do that as a follow up PR.

arcady-lunarg (approved):

PR #3286, opened 2023-07-24
ci: Test Android NDK matrix combinations

juan-lunarg:


.github/workflows/continuous_integration.yml
```diff
@@ -144,56 +144,36 @@ jobs:
...
+              run: ./update_glslang_sources.py
+            - name: Configure
               run: |
-                  ./update_glslang_sources.py
-            - name: Build
+                cmake -S . -B build/ \
+                --toolchain $ANDROID_HOME/ndk/${{ matrix.NDK }}/build/cmake/android.toolchain.cmake \
+                -D CMAKE_BUILD_TYPE=Release \
+                -D ANDROID_ABI=armeabi-v7a \
+                -D ANDROID_USE_LEGACY_TOOLCHAIN_FILE=${{ matrix.LEGACY }} \
```
juan-lunarg, line 174: My intent is to prevent build issues like PR #3264

.github/workflows/continuous_integration.yml
```diff
@@ -144,56 +144,36 @@ jobs:
...
             - name: Update Glslang Sources
+              run: ./update_glslang_sources.py
+            - name: Configure
               run: |
-                  ./update_glslang_sources.py
-            - name: Build
+                cmake -S . -B build/ \
+                --toolchain $ANDROID_HOME/ndk/${{ matrix.NDK }}/build/cmake/android.toolchain.cmake \
+                -D CMAKE_BUILD_TYPE=Release \
+                -D ANDROID_ABI=armeabi-v7a \
```
ncesario-lunarg, line 173: If we're only going to test 1 ABI, would think arm64-v8a would should be the one?

ncesario-lunarg (approved):
One question, but LGTM!

.github/workflows/continuous_integration.yml
juan-lunarg: I think so too. I just picked the ABI that was already being used. Not sure if there was an intent behind it.

\+ @dneto0 for input

.github/workflows/continuous_integration.yml
```diff
@@ -144,56 +144,36 @@ jobs:
...
         strategy:
-            fail-fast: false
-            matrix:
-                os: [ubuntu-20.04]
-                compiler: [{cc: clang, cxx: clang++}]
-                cmake_build_type: [Release]
+          matrix:
+            # Android NDK currently offers 2 different toolchains.
+            # Test both to ensure we are compatible with either approach.
+            LEGACY: [ON, OFF]
```
juan-lunarg, line 152: Example. `UNIX` is `FALSE` when `LEGACY` is true, during CMake configuration. A very minor detail which is very EASY to get wrong.

.github/workflows/continuous_integration.yml
```diff
@@ -144,56 +144,38 @@ jobs:
...
-                  export ANDROID_NDK=$HOME/android-ndk
-                  export TOOLCHAIN_PATH=$ANDROID_NDK/build/cmake/android.toolchain.cmake
-                  echo $ANDROID_NDK
-                  echo $TOOLCHAIN_PATH
-                  mkdir build && cd build
-                  cmake -DCMAKE_TOOLCHAIN_FILE=${TOOLCHAIN_PATH} -DANDROID_NATIVE_API_LEVEL=android-14 -DCMAKE_BUILD_TYPE=${{matrix.cmake_build_type}} -DANDROID_ABI="armeabi-v7a with NEON" -DBUILD_TESTING=OFF ..
-                  make -j4
+                CMAKE_GENERATOR: Ninja
+                CMAKE_C_COMPILER_LAUNCHER: ccache
+                CMAKE_CXX_COMPILER_LAUNCHER: ccache
```
juan-lunarg, line 179: Forgot to add this (which is how ccache actually gets enabled via cmake)

arcady-lunarg: Should we be enabling this for the other CI configurations as well? Presumably in a follow-up PR if so.

arcady-lunarg (approved):

.github/workflows/continuous_integration.yml
arcady-lunarg: We can make that adjustment in a follow-up PR, if it's necessary. I think an armv7 build is something we should be testing _somewhere_.

.github/workflows/continuous_integration.yml
juan-lunarg: It would help with CI time. I can do that in a follow-up PR though.

PR #3290, opened 2023-07-25
Modernize linux CI

juan-lunarg:
- Test Ubuntu-22.04
- Use ccache
- Ensure backwards compatibility with Ubuntu-20.04

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -16,22 +16,22 @@ permissions: read-all
 
 jobs:
     linux:
-        runs-on: ${{matrix.os}}
+        runs-on: ubuntu-22.04
         strategy:
             fail-fast: false
             matrix:
-                os: [ubuntu-20.04]
                 compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
                 cmake_build_type: [Debug, Release]
         steps:
-            - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


ncesario-lunarg (approved):

.github/workflows/continuous_integration.yml
```diff
@@ -16,51 +16,72 @@ permissions: read-all
...
+            - uses: actions/checkout@v3
+            - uses: lukka/get-cmake@latest  
+            - uses: actions/setup-python@v4
               with:
                   python-version: '3.7'
-            - name: Install Ubuntu Package Dependencies
-              run: |
-                  sudo apt-get -qq update
-                  sudo apt-get install -y clang-6.0
-            - name: Install GoogleTest
```
arcady-lunarg, line 35: You still need to install (check out) googletest or else you're not running most of the test suite.

arcady-lunarg (changes requested):
I think the approach to testing backwards compatibility is the right one, but removing the somewhat inaccurately named "install googletest" step gets rid of most of the test coverage we have.

.github/workflows/continuous_integration.yml
juan-lunarg: I figured that's what update_glslang_sources.py did. Sorry for the misunderstanding

arcady-lunarg (approved):
LGTM, thanks.

PR #3291, opened 2023-07-26
ci: Test CMake minimum

juan-lunarg:


juan-lunarg:
Build fails due to dependency on spirv-tools:

```
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
CMake Error at External/spirv-tools/CMakeLists.txt:15 (cmake_minimum_required):
  CMake 3.17.2 or higher is required.  You are running version 3.14.0
```

Can I raise the minimum to `3.17.2`?

arcady-lunarg:
I believe Ubuntu 20.04 only has 3.16.3 though, so I am not sure this will work either: https://packages.ubuntu.com/focal-updates/cmake. spirv-tools updated the minimum required version in https://github.com/KhronosGroup/SPIRV-Tools/pull/5041.

juan-lunarg:
> I believe Ubuntu 20.04 only has 3.16.3 though, so I am not sure this will work either: https://packages.ubuntu.com/focal-updates/cmake. spirv-tools updated the minimum required version in https://github.com/KhronosGroup/SPIRV-Tools/pull/5041.

Yes had this same situation in VVL a while ago. We just raised our CMake min to 3.17.2.

I'm not sure why 3.17.2 was picked instead of 3.16.3 but that ship sailed a while ago.

juan-lunarg:
I'm unclear on how to proceed. Effectively the real minimum for glslang is 3.17.2 due to it's dependency on spirv-tools.

While it's nice to be idiomatic and follow the convention of supporting the CMake version that comes the default package manager installations it's not required. It's mainly an idiom some projects do in order to avoid troubling users.

We broke this idiom in the vulkan validation layers and didn't receive any complaints from users.

arcady-lunarg:
I think it's fine to bump it to 3.17.2 in that case. We'll see if anyone complains.

.github/workflows/continuous_integration.yml
```diff
@@ -73,6 +73,9 @@ jobs:
           - uses: actions/setup-python@v4
             with:
                 python-version: '3.7'
+          - uses: lukka/get-cmake@latest
+            with:
+              cmakeVersion: 3.17.2
```
juan-lunarg, line 78: FWIW @arcady-lunarg in the future this will help detect these types of issues.

That's why we added this to vvl and other repos a while ago.

arcady-lunarg (approved):
LGTM, thanks!

PR #3293, opened 2023-07-28
Finish removing GLSLANG_WEB

arcady-lunarg:
This removes a stray ifdef that was missed by the rebase of the main change, and removes the whole "preprocess the grammar with m4" mechanism as that was only needed for GLSLANG_WEB.


arcady-lunarg:
cc @dj2

ncesario-lunarg (approved):
🎉

dj2 (approved):

PR #3294, opened 2023-07-31
Replace GlobalLock functions with std::mutex

arcady-lunarg:
The usage of GetGlobalLock/ReleaseGlobalLock/InitGlobalLock is replaced by std::lock_guard which is available as of c++11, and the functions are removed from the OSDependent ossource.cpp files.

std::lock_guard is already used in `StandAlone/WorkList.h` and has been there since 2017.


ncesario-lunarg (approved):
Nice!

PR #3295, opened 2023-08-01
Add emscripten build to CI

ncesario-lunarg:
Adds a build-only (no test) job for testing the emscripten (web) build on github actions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ Use the steps in [Build Steps](#build-steps), with the following notes/exception
   Bash-like environments:
   + [Instructions located here](https://emscripten.org/docs/getting_started/downloads.html#sdk-download-and-install)
 * Wrap cmake call: `emcmake cmake`
-* Set `-DBUILD_TESTING=OFF -DENABLE_OPT=OFF -DINSTALL_GTEST=OFF`.
+* Set `-DBUILD_TESTING=OFF -DENABLE_OPT=OFF`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -250,7 +250,7 @@ Use the steps in [Build Steps](#build-steps), with the following notes/exception
   Bash-like environments:
   + [Instructions located here](https://emscripten.org/docs/getting_started/downloads.html#sdk-download-and-install)
 * Wrap cmake call: `emcmake cmake`
-* Set `-DBUILD_TESTING=OFF -DENABLE_OPT=OFF -DINSTALL_GTEST=OFF`.
+* Set `-DBUILD_TESTING=OFF -DENABLE_OPT=OFF`.
```
ncesario-lunarg, line 253: This is no longer used from what I can tell.

arcady-lunarg: I think it's used by gtest.
```
External/googletest/CMakeLists.txt:
26:option(INSTALL_GTEST "Enable installation of googletest. (Projects embedding googletest may want to turn this OFF.)" ON)
```

ncesario-lunarg: Right, but `BUILD_TESTING` makes it so that gtest is not included in the build, which results in cmake printing a warning about the variable not being used. I'm not sure if gtest ever could be enabled for web (at least I don't think it can currently).

arcady-lunarg: Good point.

arcady-lunarg (approved):
LGTM.

PR #3296, opened 2023-08-03
Fix continuous deployment

ncesario-lunarg:


.github/workflows/continuous_deployment.yml
```diff
@@ -231,7 +231,7 @@ jobs:
               run: |
                   cd build/install
                   7z a ${{env.ARCHIVE}} `
-                      bin/glslangValidator.exe `
```
ncesario-lunarg, line 234: Building locally I see `glslangValidator.exe` inside `build/install/bin`, so I'm not sure why this is missing here.

StandAlone/CMakeLists.txt
```diff
@@ -122,7 +122,7 @@ if(ENABLE_GLSLANG_INSTALL)
     # Create the same symlink at install time
     install(CODE "execute_process( \
                       COMMAND ${CMAKE_COMMAND} -E ${link_method} $<TARGET_FILE_NAME:glslang-standalone> ${legacy_glslang_name} \
-                      WORKING_DIRECTORY \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
+                      WORKING_DIRECTORY \$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR})")
```
arcady-lunarg, line 125: I guess this is necessary to avoid a leading slash because the install prefix would start with a drive letter and /C:/whatever/path is not a valid Windows path?

arcady-lunarg (approved):
LGTM, I think the change makes sense.

StandAlone/CMakeLists.txt
ncesario-lunarg: Yep, that is indeed the issue on Windows.

PR #3304, opened 2023-08-17
Add GL_EXT_texture_shadow_lod support

### LINKED ISSUE #3302: Add support for GL_EXT_texture_shadow_lod for Vulkan
The following extension was added to GLSL and adds useful texture operations that are currently missing in GLSL: https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_shadow_lod.txt

The lack of this extension prevents slang from exposing some texture operations: https://github.com/shader-slang/slang/issues/3090

ncesario-lunarg:
Closes #3302.


glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8725,6 +8752,37 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
         symbolTable.setFunctionExtensions("stencilAttachmentReadEXT", 1, &E_GL_EXT_shader_tile_image);
         symbolTable.setFunctionExtensions("depthAttachmentReadEXT", 1, &E_GL_EXT_shader_tile_image);
         symbolTable.setFunctionExtensions("colorAttachmentReadEXT", 1, &E_GL_EXT_shader_tile_image);
+
+        // GL_EXT_texture_shadow_lod
+        // NOTE: These are the mangled names of individual overloads, since this extension adds overloads for
+        //       existing built-in functions.
+        if (profile == EEsProfile) {
+            if (version >= 300) {
+                symbolTable.setSingleFunctionExtensions("texture(sAS21;vf4;f1;", 1, &E_GL_EXT_texture_shadow_lod);
```
ncesario-lunarg, line 8761: I'm not a fan of this, but not sure how else to go about it.

I looked into building the mangled names at runtime (I know the mangled names are unlikely to change, but it's still a possibility), but it looks like we don't have the infrastructure for doing that in a "one-off" style, and I'm not sure it's worth adding just for this extension.

I'm happy to modify this if anyone has suggestions.

greg-lunarg: Yeah, I think splitting out and calling the code that mangles the names is the only way to make this cleaner. Is it worth it? If it doesn't add too much complexity or overhead, it might be nice. If you want to get this to Dan as soon as possible, maybe create an issue to fix it up later.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -8745,6 +8800,13 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
             symbolTable.setFunctionExtensions("textureBlockMatchSADQCOM", 1, &E_GL_QCOM_image_processing);
             symbolTable.setFunctionExtensions("textureBlockMatchSSDQCOM", 1, &E_GL_QCOM_image_processing);
         }
+
+        {
+            TempParser parser(textureShadowLodFunctions, language, version, profile, spvVersion);
```
ncesario-lunarg, line 8805: This still feels a bit hacky, but should be more "correct" in the sense that the name mangling is no longer hard-coded.

arcady-lunarg: This feels like a big hammer to use here, with things like the memory consumption. On the other hand, I think it's not actually that much extra memory, the heaviest struct is TParseContext at some 30k, and if it becomes a big issue we can do the work of pulling the function mangler out somewhat more cleanly.

glslang/MachineIndependent/SymbolTable.h
```diff
@@ -487,6 +487,19 @@ class TSymbolTableLevel {
             return (*it).second;
     }
 
+    template <typename ProcSymFn> void processSymbols(const TString& name, ProcSymFn procSym) const
+    {
+        for (auto itr = level.lower_bound(name); itr != level.cend(); ++itr) {
+            procSym(itr->second);
+        }
+    }
+
```
arcady-lunarg, line 496: Looks like dead code?

glslang/MachineIndependent/SymbolTable.h
```diff
@@ -795,6 +809,24 @@ class TSymbolTable {
         return symbol;
     }
 
+    template <typename ProcSymFn> void processSymbols(const TString& name, ProcSymFn procSym)
+    {
```
arcady-lunarg, line 813: Looks like dead code?

arcady-lunarg (changes requested):
Looks like there is some dead code that needs to be removed and I would like to see the test generate some SPIR-V, but otherwise the change looks fine.

glslang/MachineIndependent/Initialize.cpp
ncesario-lunarg: > This feels like a big hammer to use here...

I agree, I'm not a fan, but I couldn't find a better way, at least with the existing code. If this shows up again, pulling it out makes sense to me, though I'm not sure how to do that without reimplementing the parser to some degree.

glslang/MachineIndependent/SymbolTable.h
ncesario-lunarg: Yep, I forgot to get rid of this. Thanks for the catch!

arcady-lunarg (approved):
LGTM, thanks.

PR #3310, opened 2023-08-21
Fix ODR violations

silverclaw:
On a shared build, these symbols exist in both libglslang.so and libSPIRV.so, leading to an ODR violation at runtime.
Detected through ASAN.

### DIFF
--- a/glslang/Include/PoolAlloc.h
+++ b/glslang/Include/PoolAlloc.h
@@ -118,11 +118,16 @@ private:
     unsigned char* mem;           // beginning of our allocation (pts to header)
     TAllocation* prevAlloc;       // prior allocation in the chain
 
-    const static unsigned char guardBlockBeginVal;
-    const static unsigned char guardBlockEndVal;
-    const static unsigned char userDataFill;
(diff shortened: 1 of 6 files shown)


glslang/Include/PoolAlloc.h
```diff
@@ -118,11 +118,16 @@ class TAllocation {
     unsigned char* mem;           // beginning of our allocation (pts to header)
     TAllocation* prevAlloc;       // prior allocation in the chain
 
-    const static unsigned char guardBlockBeginVal;
-    const static unsigned char guardBlockEndVal;
-    const static unsigned char userDataFill;
+    static constexpr unsigned char guardBlockBeginVal = 0xfb;
```
arcady-lunarg, line 121: Would `inline constexpr` be the more correct thing to use here? Apparently it was introduced in c++17, but we already require that as a minimum.

silverclaw: I preserved the existing visibility keyword.
There are numerous static constexpr - should they all have been inline all along?

arcady-lunarg: I don't think so, I assume they produce the expected result.

PR #3314, opened 2023-08-22
Fix ByteAddressBuffer as function parameter

### LINKED ISSUE #3297: Can't pass `ByteAddressBuffer` as function parameter in HLSL with `--target-env vulkan1.2`
The following HLSL compute shader compiles but fails validation for  `--target-env vulkan1.2`
```hlsl
[[vk::binding(0, 0)]] ByteAddressBuffer buffer_position_ms;
[[vk::binding(1, 0)]] RWByteAddressBuffer r;

float3 pull_position(ByteAddressBuffer buffer_position, uint vertex_id)
{
    return asfloat(buffer_position.Load3(vertex_id * 3 * 4));
}

[numthreads(64, 1, 1)]
void main(uint gi : SV_GroupIndex)
{
    float3 position_ms = pull_position(buffer_position_ms, gi);

    r.Store3(0, asuint(position_ms));
}
```
`spirv-val` Error:
```
StorageBuffer pointer operand '54[%buffer_position_ms]' requires a variable pointers capability
  %57 = OpFunctionCall %v3float %pull_position_block__u1_0_1_u1_ %buffer_position_ms %param
```
`spirv-opt` Error if I take the resulting `.spv` file
```
Interfaces passed to OpEntryPoint must be of type OpTypeVariable. Found OpFunctionParameter.
```
This is the invocation for glslang:
```
glslangValidator --target-env vulkan1.2 -g -e main -V -D bug.comp.hlsl -o bug.spv
```

Linking #2264 since they seem related

ncesario-lunarg:
Make sure that an id represents a variable before adding it to an entry point's interface.

Fixes #3297.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -2014,7 +2014,7 @@ void TGlslangToSpvTraverser::visitSymbol(glslang::TIntermSymbol* symbol)
                 spv::StorageClass sc = builder.getStorageClass(id);
                 // Before SPIR-V 1.4, we only want to include Input and Output.
                 // Starting with SPIR-V 1.4, we want all globals.
-                if ((glslangIntermediate->getSpv().spv >= glslang::EShTargetSpv_1_4 && builder.isGlobalStorage(id)) ||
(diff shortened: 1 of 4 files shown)


Ryp:
Bumping so that it gets merged faster :)

arcady-lunarg (approved):

PR #3319, opened 2023-08-24
Add --no-link option to only compile (not link) individual shaders and export all functions

ncesario-lunarg:
After internal discussion, we will I'm changing this slightly such that `--no-link` will export _all_ functions, and the ability to export/import individual functions/variables with an attribute will be delayed until such an attribute is standardized.

<details>
<summary>[[export]]/[[import]] attribute proposal (on hold)</summary>
This change experiments with adding a "compile-only" option to glslang as well as an `[[export]]` function attribute allowing functions to be exported and then later used by spirv-link.

Proposed usage:
```bash
glslang -V --no-link myfuncs.comp -o myfuncs.spv
```

Where myfuncs.comp looks something like
```glsl
#version 450

[[export]]
int foo(...) { ... }

[[export]]
float bar(...) { ... }
```

And where "foo" and "bar" will get the "Export" linkage decoration in the resulting spir-v.

To be more "complete" this change should also:
- Add an `[[import]]` attribute for importing functions from another spir-v module
- Allow variables to be exported and imported in addition to functions (ideally using the same `[[export]]`/`[[import]]` attributes)
- Specify additional spir-v binaries at link time

However, in its current (functional) form, this change could benefit the ongoing improvements to the GPU-AV workflow in Vulkan-ValidationLayers. Before spending anymore time on this, I'm hoping to get some feedback on the overall approach here. I know `[[export]]`/`[[import]]` are not official attributes and that may not be the best mechanism for achieving the end goal here (e.g., just exporting all functions when in compile-only mode would work for the purposes of GPU-AV).
</details>


ncesario-lunarg:
FYI @jeremyg-lunarg.

jeremyg-lunarg:
I'll update my code this morning and make sure it works. I don't anticipate this change being a problem. Thanks!

jeremyg-lunarg:
Works for me locally! I pushed my updated VVL branch to depend on the newest commit.

SPIRV/GlslangToSpv.cpp
```diff
@@ -4329,6 +4340,15 @@ spv::Id TGlslangToSpvTraverser::convertGlslangToSpvType(const glslang::TType& ty
     return convertGlslangToSpvType(type, getExplicitLayout(type), type.getQualifier(), false, forwardReferenceOnly);
 }
 
+spv::LinkageType TGlslangToSpvTraverser::convertGlslangToSpv(glslang::TLinkType linkType) {
```
arcady-lunarg, line 4343: Could you give this function a better name? Given we have GlslangToSpv, convertGlslangToSpv sounds very confusing. Maybe something like glslangToSpvLinkageType?

SPIRV/SpvBuilder.cpp
```diff
@@ -2223,6 +2245,12 @@ void Builder::enterFunction(Function const* function)
         defInst->addIdOperand(funcId);
         buildPoint->addInstruction(std::unique_ptr<Instruction>(defInst));
     }
+
+    if (auto linkType = function->getLinkType(); linkType != LinkageTypeMax) {
```
arcady-lunarg, line 2249: I don't know that we have a rule against multiple statements in an if condition, but this is the only place in the codebase such a construct appears. I would prefer you hoist the linkType declaration out.

Test/baseResults/spv.exportFunctions.comp.out
```diff
@@ -0,0 +1,36 @@
...
+                              Capability Shader
+                              Capability Linkage
+               1:             ExtInstImport  "GLSL.std.450"
+                              MemoryModel Logical GLSL450
+                              Source GLSL 450
+                              Name 7  "add(f1;f1;"
+                              Name 5  "a"
+                              Name 6  "b"
+                              Name 11  "foo("
+                              Decorate 7(add(f1;f1;) Linkage Attributes 6579297 0
```
arcady-lunarg, line 15: That random constant in there is apparently the name of the exported symbol, unfortunate that it gets printed in such an unreadable way but I don't think it's worth fixing in this change.

arcady-lunarg (changes requested):
Just a few minor questions and comments, but on the whole this looks good to me.

SPIRV/GlslangToSpv.cpp
ncesario-lunarg: Yes, my original intent was to merge this with one of the existing `convertGlslangToSpvType`, but hadn't found a good way to do that and forgot to circle back. I'll rename it for now.

ncesario-lunarg: Is `convertGlslangLinkageToSpv` sound ok?

Test/baseResults/spv.exportFunctions.comp.out
ncesario-lunarg: Yeah, this is really unreadable. The `spirv-dis` output makes a lot more sense:
```asm
...
OpName %7 "add(f1;f1;"
...
OpDecorate %7 LinkageAttributes "add" Export
...
%7 = OpFunction %2 None %4
...
```
I'm not sure how easy this output is to fix for the tests (or the ROI for such a change).

SPIRV/GlslangToSpv.cpp
arcady-lunarg: Yeah, that seems like a cromulent name.

glslang/HLSL/hlslAttributes.cpp
```diff
@@ -142,6 +142,8 @@ namespace glslang {
             return EatUnroll;
         else if (name == "loop")
             return EatLoop;
+        else if (name == "export")
```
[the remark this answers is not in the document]
arcady-lunarg: What I meant is that you should put it under the `if (nameSpace == "spv")` block so that in the HLSL, the attribute is spelled `[spv::export]`.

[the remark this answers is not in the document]
ncesario-lunarg: ah, yes, I should have put it there.

arcady-lunarg (approved):
LGTM, thanks.

PR #3322, opened 2023-08-29
Hash pin and auto-update workflow dependencies

### LINKED ISSUE #3321: Hash pin GitHub Actions on workflows
Hi again, I'd like to suggest another security practice related to make the workflows as safe as possible.

Since GitHub does not provide immutable releases, hash pinning is a way of:

- prevent tag renaming attacks
- increase stability - prevent CI from break due to breaking changes on new releases blindly upgraded.
- avoid blind updates to malicious releases

And to help on keeping this versions up to date, I'd suggest adopting dependabot. Dependabot identifies and notifies security releases and it also tries to identify breaking changes and flag them on the respective PRs.

Let me know if you want to discuss more about this practice and if you have any concerns. 

Otherwise, let me know if a PR is welcome and I'll be glad to work on configuring dependabot and hash pinning the actions.

### Additional Context:

A [tag renaming attack][tag-renaming] is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

Github also recommends to **always** hash pin third party actions (see [Security Hardening for GitHub Actions][github-hashpin]). I believe it is worth it to hash pin **all** actions, not only third parties: github owned actions are still open sourced and can be compromised, even if they are significantly safer than any other action.

[tag-renaming]: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash
[github-hashpin]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

joycebrum:
Closes #3321 

I've used step security tool to help on getting the hash commits, but also I've manually validate them 

https://github.com/lukka/get-cmake/commit/4dcd3eb73017c61159eb130746fbca35d78a3d5f - v3.27.4
https://github.com/actions/checkout/commit/f43a0e5ff2bd294095638e18286ca9a3d1956744 -  v3.6.0
https://github.com/actions/setup-python/commit/61a6322f88396a6271a6ee3565807d608ecaddd1 - v4.7.0
https://github.com/actions/github-script/commit/d7906e4ad0b1822421a7e6a35d5ca353c962f410 - v6.4.1
https://github.com/hendrikmuhs/ccache-action/commit/ca3acd2731eef11f1572ccb126356c2f9298d35e - v1.2.10
https://github.com/mymindstorm/setup-emsdk/commit/29ba4851d6da084ffdc1e0fc390efadbd75df9d1 - v12

I've configured the dependabot with a maximum of 3 PRs per week. Though, it is also possible to configure it to group dependencies now https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/.  Let me know if you rather group the PRs instead.

Example of dependabot PR - https://github.com/joycebrum/glslang/pull/6

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # Necessary to update action hashes	
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Allow up to 3 opened pull requests for github-actions versions
+    open-pull-requests-limit: 3
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -47,9 +47,9 @@ jobs:
                 compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
                 cmake_build_type: [Debug, Release]
         steps:
-            - uses: actions/checkout@v3
-            - uses: lukka/get-cmake@latest
-            - uses: actions/setup-python@v4
+            - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+            - uses: lukka/get-cmake@4dcd3eb73017c61159eb130746fbca35d78a3d5f # v3.27.4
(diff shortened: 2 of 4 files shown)


arcady-lunarg (approved):
LGTM, thanks.

PR #3323, opened 2023-08-30
Emit correct name in debug info for scalar types

sajjadmirzanv:
Previously SpvBuilder would name all integer types "int/uint" and all floating-point types "float". With this change it names 64-bit types to "double" to match the GLSL language spec, and other scalar types in accordance with GL_EXT_shader_explicit_arithmetic_types (e.g. "uint8_t").

This change only affects debug info when NonSemantic.Shader.DebugInfo.100 is requested.


arcady-lunarg:
Looks like automated testing is failing, could you please fix that and then I can review your change?

sajjadmirzanv:
> Looks like automated testing is failing, could you please fix that and then I can review your change?

Sorry about that, I have corrected the new test. 

In that job the Windows build **should have failed** (it has the same failing test as the Linux and MacOS builds), but instead it passed. Maybe there's a problem in the test runner?

arcady-lunarg (approved):
LGTM, thanks.

PR #3327, opened 2023-09-08
Fix textureOffset overload

### LINKED ISSUE #3325: textureOffset doesn't need GL_EXT_texture_shadow_lod in GLSL 440 and beyond shaders
https://github.com/KhronosGroup/glslang/commit/ffefbcd9f36efcec0a940aa96eb99f0c6055cd2a introduced a regression by requiring GL_EXT_texture_shadow_lod to be requested for GLSL '450 core' shaders that contain 'float textureOffset (sampler2DArrayShadow sampler, vec4 P, ivec2 offset )'. This function is defined in at least GLSL 440, hence no other extensions should be needed.

Reporducer"

```
#version 450 core
layout(location = 0) out mediump vec4 o_color;
layout(location = 0) in highp vec4 v_texCoord;
layout(set = 0, binding = 0) uniform highp sampler2DArrayShadow u_sampler;
layout(set = 0, binding = 1) uniform buf0 { highp vec4 u_scale; };
layout(set = 0, binding = 2) uniform buf1 { highp vec4 u_bias; };

void main()
{
	o_color = vec4(textureOffset(u_sampler, v_texCoord, ivec2(7, -8)), 0.0, 0.0, 1.0);
}
```

Error log:

```
---
fragment compile log:
---

ERROR: 0:10: 'textureOffset' : required extension not requested: GL_EXT_texture_shadow_lod

ERROR: 0:10: '' : compilation terminated 

```

ncesario-lunarg:
float textureOffset(sampler2DArrayShadow sampler, vec4 P, ivec2 offset) was incorrectly requiring the GL_EXT_texture_shadow_lod extension.

NOTE: Prior to GLSL 440, this prototype was defined as

float textureOffset(sampler2DArrayShadow sampler, vec4 P, vec2 offset)

i.e., the type of 'offset' was specified as 'vec2' rather than 'ivec2'. This is believed to be a typo.

Fixes #3325.


ncesario-lunarg:
@joycebrum it looks like https://github.com/KhronosGroup/glslang/blob/main/.github/dependabot.yml is [missing a license](https://source.cloud.google.com/results/invocations/882a8393-adfb-4da5-a445-abd50407570c/targets/graphics_shader_compiler%2Fglslang%2Flicense-check%2Fpresubmit/log)?

arcady-lunarg:
Should there have been a test that would catch this? Should we add one? There seems to be comprehensive testing of `textureOffset` and friends `spv.float16Fetch.frag` but apparently only with 16-bit return values, so that didn't cover this case.

ncesario-lunarg:
> Should there have been a test that would catch this? Should we add one?

I don't think so. The code appeared to be working as intended before, I had just written the code against a typo in the spec.

> There seems to be comprehensive testing of textureOffset and friends spv.float16Fetch.frag but apparently only with 16-bit return values, so that didn't cover this case.

Maybe it would make sense to have a 32b versionof `spv.float16Fetch.frag`? I could add that as a follow up PR, or as part of this one.

ncesario-lunarg:
I added a 32b version of `spv.float16Fetch.frag` and verified that it would have given an error before my original PR went in.

ncesario-lunarg:
> > @joycebrum it looks like https://github.com/KhronosGroup/glslang/blob/main/.github/dependabot.yml is [missing a license](https://source.cloud.google.com/results/invocations/882a8393-adfb-4da5-a445-abd50407570c/targets/graphics_shader_compiler%2Fglslang%2Flicense-check%2Fpresubmit/log)?
> 
> Oh sorry about it. Will fix it now!

Thanks!

arcady-lunarg (approved):
LGTM, thanks.

PR #3331, opened 2023-09-11
Add license to dependabot.yml

joycebrum:
I've added the apache-2.0 license header. Let me know if the file should be actually under another license


ncesario-lunarg:
It might be simplest to use the existing BSD license in glslang. It might be fine to use the apache 2.0 license, but I'm not a open source license lawyer 😬. glslang was created before Khronos standardized on Apache 2.0.

arcady-lunarg (approved):
Glslang has a bit of a mix of licenses but Apache 2.0 is one of them so this is fine.

PR #3333, opened 2023-09-15
Fixed memleaks using delete operator for dynamically allocated objects

GermanAizek:
Minor PR changes that affect small memory leaks in different situations.

### DIFF
--- a/glslang/HLSL/hlslGrammar.cpp
+++ b/glslang/HLSL/hlslGrammar.cpp
@@ -2623,6 +2623,7 @@ bool HlslGrammar::acceptStructBufferType(TType& type)
     if (hasTemplateType) {
         if (! acceptTokenClass(EHTokLeftAngle)) {
             expected("left angle bracket");
+            delete templateType;
             return false;
         }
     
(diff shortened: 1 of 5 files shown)


ncesario-lunarg (commented):
I would vote for making all of these `std::unique_ptr`s or `std::optional`s, but it's up to @arcady-lunarg.

ncesario-lunarg:
@GermanAizek Did you encounter memory leaks that led to this change? It looks like these allocations are cleaned up by the [TPoolAllocator](https://github.com/KhronosGroup/glslang/blob/4c57db1595462c51d5080893f65c42fc1f318146/glslang/Include/PoolAlloc.h#L152), and running some tests with valgrind locally, I don't encounter any leaks with the current code.

Ideally, I personally would rather have the lifetimes of these variables spelled out using types such as `unique_ptr`, but changing that would be a much larger ordeal.

ncesario-lunarg:
I'm going to close this as there are currently no leaks nor other memory issues in glslang (at least as reported by running valgrind on the tests and looking at ASAN/TSAN). I'm in the process of adding ASAN/TSAN jobs to CI and I'm also looking at making the use of the pool allocator a little more explicit so that it's clear these "raw new"s are not using the standard allocator.

PR #3334, opened 2023-09-15
Optimize push_back() std::unique_ptr using C++17 std::make_unique() and emplace

GermanAizek:
You have C++17 standard in Cmakelists, so it would be good to update code.


ncesario-lunarg (changes requested):
I think this looks good with a couple minor adjustments.

arcady-lunarg:
Given this doesn't pass CI and hasn't been touched in a while, I'm going to close this.

PR #3335, opened 2023-09-15
Fix interpolant ES error

### LINKED ISSUE #3277: interpolate struct member
Acoording es32 spec:
For all of the interpolation functions, interpolant must be an input variable or an element of an input
variable declared as an array.
glcts.exe--deqp-case=dEQP-GLES31.functional.shaders.multisample_interpolation.interpolate_at_sample.negative.interpolate_struct_member

Glslang need report error for following shader
#version 320 es
struct S
{
	highp float a;
	highp float b;
};
in S v_var;
layout(location = 0) out mediump vec4 fragColor;
void main (void)
{
	fragColor = vec4(interpolateAtSample(v_var.a, 0)); // <interpolant> is not an input variable (just a member of)
}

ncesario-lunarg:
The restriction of no swizzling and no struct fields as an interpolant were not being checked when using the ES profile.

Fixes #3277.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -2571,7 +2571,7 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
             requireExtensions(loc, 1, &E_GL_EXT_shader_atomic_float2, fnCandidate.getName().c_str());
         }
 
-        const TIntermTyped* base = TIntermediate::findLValueBase(arg0, true , true);
+        const TIntermTyped* base = TIntermediate::traverseLValueBase(arg0, true, true);
         const char* errMsg = "Only l-values corresponding to shader block storage or shared variables can be used with "
                              "atomic memory functions.";
         if (base) {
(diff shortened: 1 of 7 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2536,10 +2536,53 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
+            const bool esProfile = isEsProfile();
+            const bool swizzleOkay = !esProfile && (version >= 440);
+
+            std::string interpolantErrorMsg = "first argument must be an interpolant, or interpolant-array element";
+            bool isValid = true; // Assume that the interpolant is valid until we find a condition making it invalid
+            bool isIn = false;   // Checks whether or not the interpolant is a shader input
+            bool structAccessOp = false; // Whether or not the previous node in the chain is a struct accessor
+            TIntermediate::traverseLValueBase(
+                arg0, swizzleOkay, false,
+                [&isValid, &isIn, &interpolantErrorMsg, esProfile, &structAccessOp](const TIntermNode& n) {
```
ncesario-lunarg, line 2548: @arcady-lunarg @jeremy-lunarg I went back and forth with creating a dedicated function to fix this or a more generic function to traverse. Let me know what you prefer style-wise.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -43,11 +43,12 @@
...
 
-#include <string>
-#include <vector>
 #include <algorithm>
-#include <set>
 #include <array>
+#include <functional>
+#include <set>
+#include <string>
+#include <vector>
```
ncesario-lunarg, line 51: The only actual addition here is `<functional>`. `clang-format` put them in alphabetical order.

glslang/MachineIndependent/ParseHelper.cpp
arcady-lunarg: I think it's better to keep it inline but I also would prefer a brief comment explaining what this is doing.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -573,6 +574,8 @@ class TIntermediate {
 
     // Tree ops
     static const TIntermTyped* findLValueBase(const TIntermTyped*, bool swizzleOkay , bool BufferReferenceOk = false);
+    static const TIntermTyped* traverseLValueBase(const TIntermTyped*, bool swizzleOkay, bool bufferReferenceOk = false,
```
arcady-lunarg, line 577: I'm a bit confused as to why a) this returns a value since your one use of it doesn't consume the return value at all and b) why proc has a default value. Were you trying to combine `findLValueBase` and `traverseLValueBase` and just didn't finish the job? It might not be a bad idea to combine those two btw since the traversal code is annoying and would be best written only once.

arcady-lunarg (changes requested):
A couple comments on some details but the overall approach is good. Thanks for adding some pretty comprehensive tests.

glslang/MachineIndependent/localintermediate.h
ncesario-lunarg: Yes, that was my original intent, but felt it was a little messy and so I wanted to get some feedback first. I would also prefer these to be combined, but wasn't sure what the preference would be for the generic traversal code. If you're ok with the generic approach, I'll combine them.

arcady-lunarg: I think it's cleaner this way, the complicated traversal logic exists only once.

glslang/MachineIndependent/localintermediate.h
```diff
@@ -573,6 +574,8 @@ class TIntermediate {
 
     // Tree ops
     static const TIntermTyped* findLValueBase(const TIntermTyped*, bool swizzleOkay , bool BufferReferenceOk = false);
```
arcady-lunarg, line 576: I think you need to delete this because the function doesn't exist.

arcady-lunarg (changes requested):
Thanks, I think it's cleaner like this, one small nit left to address and this can be merged.

glslang/MachineIndependent/localintermediate.h
ncesario-lunarg: Ah, yep, thanks for catching that!

arcady-lunarg (approved):

PR #3336, opened 2023-09-15
Fix segfault with atomic arg check

### LINKED ISSUE #3332: Segfault when r-value used in first argument to atomic*()
`glslang[Validator]` will fail with a segmentation fault on the following input:
```glsl
#version 440 // Can be any version number, even gibberish

// Any shader stage
void main() {
    uint a = 5;
    // Can be any atomic function, and any expressions as long as
    // first argument is an r-value
    // Notably, doesn't occur with user-defined functions with `inout` parameters.
    atomicAdd(a * 2, 0);
}
```

Output:
```
$ glslangValidator -v test.comp
Glslang Version: 11:13.0.0
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 13.0.0
GLSL Version: 4.60 glslang Khronos. 13.0.0
SPIR-V Version 0x00010600, Revision 1
Khronos Tool ID 8
SPIR-V Generator Version 11
ARB_GL_gl_spirv version 100
Segmentation fault (core dumped)
```

ncesario-lunarg:
Makes sure that we have an l-value before checking the storage type of the mem argument passed to an atomic memory operation.

Fixes #3332.

### DIFF
--- a/Test/baseResults/atomicAdd.comp.out
+++ b/Test/baseResults/atomicAdd.comp.out
@@ -1,5 +1,5 @@
 atomicAdd.comp
-ERROR: 0:18: 'atomicAdd' : Atomic memory function can only be used for shader storage block member or shared variable. 
+ERROR: 0:18: 'atomicAdd' : Only l-values corresponding to shader block storage or shared variables can be used with atomic memory functions. 
 ERROR: 1 compilation errors.  No code generated.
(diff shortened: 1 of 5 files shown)


arcady-lunarg (approved):
LGTM, thanks.

PR #3339, opened 2023-09-22
Improve preprocessor ouput format

chirsz-ever:
There are still some imperfections, for example `++i` would be convert to `++ i`. I think the current state is good enough.


ncesario-lunarg (approved):
LGTM pending CI.

Thanks for the contribution @chirsz-ever. Was there an issue associated with this?

chirsz-ever:
> Was there an issue associated with this?

No, I didn't find related issues.

arcady-lunarg:
Is this a purely aesthetic change to the appearance of the preprocessor output? Seems like an improvement on that front at least.

chirsz-ever:
> Is this a purely aesthetic change to the appearance of the preprocessor output?

Yes. Most users of glslang do not care about the preprocess output, and glslang's error message only give the line numbers. But I rencently just need the preprocess output for human reading.

In fact IMO the most right scheme is, not refomart code, just keeping the code as what it looks like when input it to the preprocessor. Let's fix it in the future when someone really needs it.

PR #3343, opened 2023-09-25
Add ASAN and TSAN integration testing

ncesario-lunarg:
Add jobs for testing with -fsanitize=address and -fsanitize=thread.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -51,15 +51,16 @@
 #include "../SPIRV/doc.h"
 #include "../SPIRV/disassemble.h"
 
-#include <cstring>
-#include <cstdlib>
+#include <array>
+#include <atomic>
 #include <cctype>
 #include <cmath>
-#include <array>
+#include <cstdlib>
+#include <cstring>
 #include <map>
 #include <memory>
-#include <thread>
 #include <set>
+#include <thread>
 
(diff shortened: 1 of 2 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -144,8 +145,9 @@ void FreeFileData(char* data);
 void InfoLogMsg(const char* msg, const char* name, const int num);
 
 // Globally track if any compile or link failure.
-bool CompileFailed = false;
-bool LinkFailed = false;
+std::atomic<int8_t> CompileFailed{0};
```
arcady-lunarg, line 148: Would it be better to use `std::atomic<bool>` here, or is there some reason this has to be int8_t?

arcady-lunarg (commented):
I just have one question about the atomics, otherwise this looks good, thanks for getting it done.

StandAlone/StandAlone.cpp
ncesario-lunarg: `std::atomic<bool>` would have been my first choice, but it doesn't seem to work with `fetch_or`.

arcady-lunarg: Ah, okay. That's annoying but I suppose it's not a big deal to use int8_t.

arcady-lunarg (approved):
LGTM, thanks.

PR #3359, opened 2023-10-13
spirv: don't emit invalid debuginfo for buffer references

arcady-lunarg:
Currently no debug info is emitted for buffer reference types, which resulted in the SPIR-V backend code asserting when trying to emit the debug info for struct member that had such a type. Instead, the code now skips such struct members. Full debug info for buffer references may require forward references in the debug info instructions, which is currently prohibited by the spec.


SPIRV/SpvBuilder.cpp
```diff
@@ -1029,7 +1029,10 @@ Id Builder::makeCompositeDebugType(std::vector<Id> const& memberTypes, char cons
     for(auto const memberType : memberTypes) {
         assert(debugTypeLocs.find(memberType) != debugTypeLocs.end());
 
-        memberDebugTypes.emplace_back(makeMemberDebugType(memberType, debugTypeLocs[memberType]));
+        // There _should_ be debug types for all the member types but currently buffer references
+        // do not have member debug info generated.
+        if (debugId[memberType])
```
jeremy-lunarg, line 1034: This could filter a larger set than just buffer references, but I think it's okay for now.

jeremy-lunarg (approved):
Okay for now. Hopefully this gets better when we get the new extensions.

SPIRV/SpvBuilder.cpp
arcady-lunarg: If we've gotten this far, the options are emit invalid debug info (with zero Ids) or not emit any debuginfo at all, and I think the latter is preferable.

ncesario-lunarg (approved):

SPIRV/SpvBuilder.cpp
jeremy-lunarg: I agree, but it would be nice to know where the other holes might be.

PR #3364, opened 2023-10-16
Add missing initialization of `compile_only` field for SPIRV C interface

ajmd17:
Looks like the `compile_only` struct field of `glslang_spv_options_t` is not being set, so it's just holding garbage memory and causing unwanted behaviour by implicitly resolving to `true` most times.


jeremy-lunarg (approved):
Good catch! Thanks!

PR #3369, opened 2023-10-19
Initialize spv_options in glslang_program_SPIRV_generate

arcady-lunarg:
This will prevent any recurrence of issues like the one addressed by PR #3364.


jeremy-lunarg (approved):
LGTM

PR #3370, opened 2023-10-21
Remove debugOptions from internal classes

### LINKED ISSUE #3348: Need to propagate 64 bit options correctly in glslang standalone
https://github.com/KhronosGroup/glslang/commit/4c57db1595462c51d5080893f65c42fc1f318146 introduced a 64 bit TOptions in StandAlone.cpp that get's passed down as a 32 bit option in ShConstructCompiler and ShCompile. Causing loss of precision.

These functions need to be corrected to take 64 bit flags. This has a wider implac since ShConstructCompiler and ShCompile are exposed API's of the library.

ncesario-lunarg:
The debug options passed down from the public ShConstruct* functions to internal code is unused. This change removes the internal debugOptions fields and attempts to make it more obvious these fields are not used.

This change does not change the public-facing interface.

This change also adds the -Wshorten-64-to-32 warning to the StandAlone build in order to avoid unwanted 64-to-32 bit conversions in the future.

Closes #3348.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -516,7 +516,7 @@ void ProcessGlobalBlockSettings(int& argc, char**& argv, std::string* name, unsi
 
     if (set) {
         errno = 0;
-        int setVal = ::strtol(argv[curArg], nullptr, 10);
+        int setVal = static_cast<int>(::strtol(argv[curArg], nullptr, 10));
         if (errno || setVal < 0) {
             printf("%s: invalid set\n", argv[curArg]);
             usage();
@@ -528,7 +528,7 @@ void ProcessGlobalBlockSettings(int& argc, char**& argv, std::string* name, unsi
 
     if (binding) {
         errno = 0;
-        int bindingVal = ::strtol(argv[curArg], nullptr, 10);
+        int bindingVal = static_cast<int>(::strtol(argv[curArg], nullptr, 10));
         if (errno || bindingVal < 0) {
             printf("%s: invalid binding\n", argv[curArg]);
             usage();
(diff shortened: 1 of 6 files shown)


StandAlone/StandAlone.cpp
```diff
@@ -516,7 +516,7 @@ void ProcessGlobalBlockSettings(int& argc, char**& argv, std::string* name, unsi
 
     if (set) {
         errno = 0;
-        int setVal = ::strtol(argv[curArg], nullptr, 10);
+        int setVal = static_cast<int>(::strtol(argv[curArg], nullptr, 10));
```
ncesario-lunarg, line 519: These casts were necessary due to the addition of `-Wshorten-64-to-32`.

glslang/Public/ShaderLang.h
```diff
@@ -330,18 +330,13 @@ GLSLANG_EXPORT void ShDestruct(ShHandle);
...
-    bool forwardCompatible = false,      // give errors for use of deprecated features
-    EShMessages messages = EShMsgDefault // warnings and errors
-    );
+GLSLANG_EXPORT int ShCompile(const ShHandle, const char* const shaderStrings[], const int numStrings,
+                             const int* lengths, const EShOptimizationLevel, const TBuiltInResource* resources,
+                             int,                      // debugOptions unused
+                             int defaultVersion = 110, // use 100 for ES environment, overridden by #version in shader
+                             bool forwardCompatible = false,      // give errors for use of deprecated features
+                             EShMessages messages = EShMsgDefault // warnings and errors
+);
```
ncesario-lunarg, line 339: These whitespace changes are coming from clang-format.

What is our policy on clang-format? It looks like we don't enforce clang-format changes, yet we have a .clang-format file checked into the repo. Should we either enforce clang-format (perhaps modifying the .clang-format file first), or just get rid of the .clang-format file if we don't plan on enforcing it?

arcady-lunarg (approved):

Issue #3320, opened 2023-08-25
RFC: Changes to the glslang API/ABI


arcady-lunarg:
Given the difficulty of maintaining ABI compatibility when all the internals of the compiler are exposed as a public API, I propose the following changes to the way that glslang is built.
1. The final result of the build of the compiler is one library, libglslang.so or libglslang.a as appropriate for the type of build, and the 'glslang' binary. Likewise, for spirv-remap there would be one library, libSPVRemapper.so (or libSPVRemapper.a) and the spirv-remap binary.
2. The only header files installed by the build process would be:
```
Public/ResourceLimits.h
Public/ShaderLang.h
Public/resource_limits_c.h
Include/glslang_c_interface.h
Include/glslang_c_shader_types.h
Include/ResourceLimits.h
MachineIndependent/Versions.h
SPIRV/GlslangToSpv.h
SPIRV/disassemble.h
SPIRV/SPVRemapper.h (for the remapper)
```
3. The only entrypoints exposed by libglslang will be the ones declared in the above header files (except the last one, which declares the entry points for libSPVRemapper). The API and ABI for these will be kept stable, with the SOVERSION being incremented anytime there is a backward-incompatible change. We will add a CI hook to use libabigail to ensure that no backwards-incompatible changes are made.
4. If for some reason you want to use the internals of glslang (for example, just the AST, or just the SPIR-V emitting backend), you are welcome to a) include glslang as a subproject in your project, and b) link against it statically. These internals are subject to binary-incompatible changes at any time, though we will try to maintain source compatibility.

arcady-lunarg:
CC @rhabacker @jengelh @ben-clayton @haasn @SpaceIm as people whose names have appeared in issues related to shared library builds, I would be curious to hear your opinions on this.

haasn:
This would work for my use case. I only need TShader and GlslangToSpv, pretty much.

jengelh:
Not (actively) maintaining compatibility like right now is a possibility too; the file just needs to be marked such that it appears to be "different" often enough so the ELF loader can reject _possibly-incompatible_ library<>program combinations.

<details><summary><i>exempli gratia</i></summary>
<pre>
# can't rely on .git/ dir existing, so no git commit id available
cs=$(find * -type f ( -name "*.[ch]" -o -name "*.cc" ) -exec md5sum {} + | cut -b 1-32)
echo "$cs { global: *; };" >x.ver
[...]

PR #3371, opened 2023-10-21
Generate pkg-config file during install builds

### LINKED ISSUE #1715: Add pkgconfig file in the sources
Hi

is possible add a pkgconfig file in the sources, like vulkan-loader and spirv-tools?

greetings

jasonliu--:
This commit causes CMake to generate a pkg-config file when `ENABLE_GLSLANG_INSTALL` is enabled. This allows software projects that use pkg-config (and not CMake) to locate external dependencies (e.g., Godot 4.x), to find and properly link to a pre-built glslang package.

Closes #1715.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,4 +370,35 @@ if(ENABLE_GLSLANG_INSTALL)
         DESTINATION
             "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
     )
+
+    # Generate a pkg-config file, so that software projects which use
+    # pkg-config to locate dependencies can find glslang
+
+    # This template is filled-in by CMake's `configure_file(... @ONLY)`.
+    # The `@...@` are substituted by CMake's configure_file(), either
+    # from variables set in CMakeLists.txt or by CMake itself.
+    # (Based on: https://www.scivision.dev/cmake-generate-pkg-config/)
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc.in" [=[
+prefix="@CMAKE_INSTALL_PREFIX@"
+exec_prefix="${prefix}"
+libdir="${prefix}/lib"
+includedir="${prefix}/include"
+
+Name: @PROJECT_NAME@
+Description: official reference compiler front end for the OpenGL ES and OpenGL shading languages
+Version: @PROJECT_VERSION@
+Cflags: -I"${includedir}"
+Libs: -L"${libdir}" -l@PROJECT_NAME@
+    ]=])
+    configure_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc.in"
+        "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc"
+        @ONLY
+    )
+    install(
+        FILES
+            "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc"
+        DESTINATION
+            "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
+    )
 endif()


arcady-lunarg:
Unfortunately the pkg-config file that this produces doesn't actually work, when I try to build with pkg-config I get undefined errors because libglslang.a isn't actually self-contained and you generally will need to link against at least libSPIRV.a as well. Same goes for shared library builds I think. We're working on unifying the two into one single library file though, please see #3320.

jasonliu--:
Hi @arcady-lunarg, so are you saying that the pkg-config file doesn't work in general as a concept, or are you saying that it's missing some pieces, i.e. also needing to link against `libSPIRV.{a,dylib}` as well? Because if it's the latter, could you let me know what are the minimum libraries that are needed? Just `libglslang` and `libSPIRV`? Or what about the other libs, `libHLSL` and `libSPVRemapper`? I could easily add them to the `Libs:` line in the pkg-config file, that's no problem.

arcady-lunarg:
The pkg-config file is fine as a concept, but your implementation of it is broken. I suggest waiting a little while for me to finish the work of rearranging the libraries a bit and then your pkg-config file will actually be correct.

juan-lunarg:
- This doesn't properly handle `BUILD_SHARED_LIBS`.

`Libs.private` is only needed when `BUILD_SHARED_LIBS` is OFF.

- Doesn't handle absolute installation locations which can be provided to CMake:

See: https://github.com/KhronosGroup/SPIRV-Headers/issues/282

- Testing?

Both vulkan-loader and spirv-headers currently test the generated .pc file to some extent. If the testing doesn't happen here it should happen during SDK testing since adding this will mean verifying another deliverable.

jasonliu--:
@juan-lunarg , do you believe that the same technique that is used in the issue you referenced would also work here for glslang?

> Both vulkan-loader and spirv-headers currently test the generated .pc file to some extent. If the testing doesn't happen here it should happen during SDK testing since adding this will mean verifying another deliverable.

I'm willing to add similar tests to this PR, I wasn't aware that other projects were doing these tests.

juan-lunarg:
> @juan-lunarg , do you believe that the same technique that is used in the issue you referenced would also work here for glslang?

Yes but keep in mind that is a header only library. You can look at the loader code as well for reference.

NOTE: The loader's package config is simpler since the loader can only be installed as a SHARED library so there isn't a perfect example for you to look at.

> I'm willing to add similar tests to this PR, I wasn't aware that other projects were doing these tests.

👍🏾

juan-lunarg:
> @juan-lunarg , do you believe that the same technique that is used in the issue you referenced would also work here for glslang?

This repo actually summarizes the issue nicely
https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path

arcady-lunarg:
Now that glslang has been mostly consolidated to one library file, I think it might be time to pick this back up.

jpr42:
> Now that glslang has been mostly consolidated to one library file, I think it might be time to pick this back up.

https://github.com/KhronosGroup/glslang/pull/3898

CMakeLists.txt
```diff
@@ -370,4 +370,35 @@ if(ENABLE_GLSLANG_INSTALL)
...
+    # (Based on: https://www.scivision.dev/cmake-generate-pkg-config/)
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc.in" [=[
+prefix="@CMAKE_INSTALL_PREFIX@"
+exec_prefix="${prefix}"
+libdir="${prefix}/lib"
+includedir="${prefix}/include"
+
+Name: @PROJECT_NAME@
+Description: official reference compiler front end for the OpenGL ES and OpenGL shading languages
+Version: @PROJECT_VERSION@
```
jpr42, line 389: btw this doesn't work since glslang doesn't set the project version.

GLSLANG_VERSION is

rossburton:
Note that many distros are doing this themselves with variations of the patch.

- Fedora https://src.fedoraproject.org/rpms/glslang/blob/main/f/0001-pkg-config-compatibility.patch
- OpenEmbedded https://git.openembedded.org/openembedded-core/tree/meta/recipes-graphics/glslang/glslang/0001-generate-glslang-pkg-config.patch
- Debian https://salsa.debian.org/xorg-team/vulkan/glslang/-/blob/debian-unstable/debian/patches/0001-pkg-config-compatibility.patch?ref_type=heads

jpr42:
> Note that many distros are doing this themselves with variations of the patch.

I did discover that when I was working on this. As I mentioned in my PR attempting this generating the pkg-config files correctly in all circumstances is not possible. See my PR attempt/explanation when I worked on this.

IMO the correct path forward at this point to support CPS (Common Package Specification):
https://www.kitware.com/common-package-specification-is-out-the-gate/

PR #3375, opened 2023-10-24
Fix debug info file and source strings

jeremy-lunarg:
While reviewing another PR, I noticed the DebugSource file and text operands were not being set correctly in the test output. I discussed this with @ncesario-lunarg and he noticed that the "generate debug information" option is implicitly set when generating nonsemantic shader debug information from the command-line. This change makes the test fixture consistent with the command-line behavior (i.e. "-gVS"), which was my original intent when I added these tests a while ago.


Test/baseResults/spv.debuginfo.bufferref.glsl.frag.out
```diff
@@ -13,7 +13,7 @@ spv.debuginfo.bufferref.glsl.frag
                               MemoryModel PhysicalStorageBuffer64EXT GLSL450
                               EntryPoint Fragment 14  "main" 69 122
                               ExecutionMode 14 OriginUpperLeft
-               1:             String  ""
+               1:             String  "spv.debuginfo.bufferref.glsl.frag"
                8:             String  "uint"
               15:             String  "main"
               18:             String  "// OpModuleProcessed auto-map-locations
```
jeremy-lunarg, line 19: Ugh. I just noticed this stuff is still being prepended. Investigating...

ncesario-lunarg (approved):

arcady-lunarg:
Ping? I think this needs to be rebased as well.

arcady-lunarg:
I rebased this but the line ending problem persists and git makes it very hard to figure out what is wrong or fix it.

arcady-lunarg (approved):

PR #3377, opened 2023-10-24
Support NonSemantic DebugValue and generate it for const arg

qingyuanzNV:
For passing-by-value semantic, DebugDeclare cannot be used for parameters because there's no pointer.

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -1182,13 +1182,26 @@ Id Builder::makeDebugExpression()
     return debugExpression;
 }
 
-Id Builder::makeDebugDeclare(Id const debugLocalVariable, Id const localVariable)
+Id Builder::makeDebugDeclare(Id const debugLocalVariable, Id const pointer)
 {
     Instruction* inst = new Instruction(getUniqueId(), makeVoidType(), OpExtInst);
     inst->addIdOperand(nonSemanticShaderDebugInfo);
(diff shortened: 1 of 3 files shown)


jeremy-lunarg:
Thanks @qingyuanzNV. Can you provide a test case that demonstrates what this fix is addressing?

jeremy-lunarg:
@qingyuanzNV Is this validation missing from SPIRV-Tools?

qingyuanzNV:
> Thanks @qingyuanzNV. Can you provide a test case that demonstrates what this fix is addressing?

"spv.debuginfo.const_params.glsl.comp" in glslang's test is a good example. Basicaly, for a function definition like shown below:
```
int foo(const int x) {
    return x;
}
```
glslang will generate the parameter instruction "OpFunctionParameter %x %float", noting the type is actually a plain float (pass-by-value). However, the OpDebugDeclare actually requires its second argument provided to be a pointer to its original value. So we have to either using a dummy variable or using OpDebugValue instead.

Technically, what I said above is a little different than the real spec. OpDebugDeclare actually requires the second argument to be id of an OpVariable. But that's a different topic. Though SPIR-V Tools allows OpFunctionParameter as well.

> @qingyuanzNV Is this validation missing from SPIRV-Tools?

No, it looks like no pointer type check is in SPIRV-Tools.

jeremy-lunarg (approved):
Sorry for the delay. This LGTM

PR #3378, opened 2023-10-25
Fix Windows GHA Tests

ncesario-lunarg:
The Windows tests job in Github Actions was incorrectly reporting 'success' on failed runs. This change ensures a single command is run in each test 'step' to help remove any guesswork w.r.t. the shell behavior on each platform.

This change also takes the oppportunity to make the formatting more consistent in continuous_integration.yml.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -7,301 +7,296 @@
 name: Continuous Integration
 
 on:
-    workflow_dispatch:
-    pull_request:
-        branches:
-            - main
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
 
 permissions: read-all
 
 jobs:
-    linux:
-        runs-on: ubuntu-22.04
-        strategy:
-            fail-fast: false
-            matrix:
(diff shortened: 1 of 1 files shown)


ncesario-lunarg:
It's probably best to make sure "hide whitespace" is checked when looking at the diff, as pretty much the whole file got formatted.

.github/workflows/continuous_integration.yml
```diff
@@ -7,301 +7,296 @@
...
+          CMAKE_C_COMPILER_LAUNCHER: ccache
+          CMAKE_CXX_COMPILER_LAUNCHER: ccache
+      - name: Build
+        run: cmake --build build
+      - name: Install
+        run: cmake --install build --prefix build/install
+      - name: Test
+        run: ctest --output-on-failure --test-dir build
+      - name: Test (standalone)
+        run: cd Test && ./runtests
```
jeremy-lunarg, line 64: If I understand the original problem this patch is addressing, then is it possible that the Test directory does not exist and the runtests script can silently fail?

jeremy-lunarg: Oops... this is linux... sorry. Disregard.

jeremy-lunarg (approved):
LGTM

Issue #3376, opened 2023-10-24
bool debug type not generated for non semantic debugging


chaocNV:
It appears that when we make bool type in SPIRV builder via` Id Builder::makeBoolType(bool const compilerGenerated)`, it has a hidden argument of `bool const compilerGenerated = true`. And generating debug type for bool is always gated on that argument being false: 
```
if (emitNonSemanticShaderDebugInfo && !compilerGenerated)
{
        auto const debugResultId = makeBoolDebugType(32);
        debugId[type->getResultId()] = debugResultId;
}
```
I believe this is redundant as there is no harm in creating a debug bool type regardless. It's a primitive type like int/float and those do not have this check. Currently all the usage of makeBoolType do not pass in a false argument and therefore bool debug type is never generated. 

My suggested fix would be remove `compilerGenerated ` from `makeBoolType` entirely. If you agree, I can create a PR for it. 

To illustrate the problem, following shader would crash in glslang currently:
```

#version 460

uniform UBO {
    bvec2 uniform_bvec2;
};

buffer SSBO {
    bvec2 buffer_bvec2;
};



void main() {
    buffer_bvec2 = uniform_bvec2;
}

```

arcady-lunarg:
Yeah, I think the correct thing is to just always emit the boolean debug type and I would welcome a PR. Before you create the PR however, it would be a good idea to check that this works fine in renderdoc, since that is the ultimate consumer of this debug info.

chaocNV:
PR created 

> Yeah, I think the correct thing is to just always emit the boolean debug type and I would welcome a PR. Before you create the PR however, it would be a good idea to check that this works fine in renderdoc, since that is the ultimate consumer of this debug info.

PR created. I'm not very familiar with renderDoc. Could you help give this a quick run on it if you have the setup?

arcady-lunarg:
This has been fixed by the linked PR.

PR #3379, opened 2023-10-25
Always enable the generation of OpDebugBasicType for bool type

chaocNV:
Fixing issue https://github.com/KhronosGroup/glslang/issues/3376


arcady-lunarg (changes requested):
This doesn't seem right, if you look at the changes in the test results it's generating the debug type multiple times and massively increasing the number of IDs used. You need to make sure the bool debug type only gets emitted once.

chaocNV:
> This doesn't seem right, if you look at the changes in the test results it's generating the debug type multiple times and massively increasing the number of IDs used. You need to make sure the bool debug type only gets emitted once.

Fixed. Could you help merge?

chaocNV:
A friendly pain on review/merge @ncesario-lunarg @arcady-lunarg

arcady-lunarg (approved):
What was putting me off was the fact that your change makes the SPIR-V smaller, but it is actually because the current version of the code apparently repeatedly generates OpDebugTypeBasic for bool, whereas after your change it is only generated once. So your change is making things more correct.

PR #3380, opened 2023-10-25
Add debug source continued

chaocNV:


arcady-lunarg:
Is this still going to evolve into something that should be merged, or can we close this?

PR #3384, opened 2023-10-30
tweak error behavior for redeclared uniforms for vulkan-relaxed

mbechard:
avoids nullptr crash from reading memberType's name, and more user friendly error message.


arcady-lunarg (approved):
LGTM

PR #3385, opened 2023-10-31
Use googletest version 1.14.0

juan-lunarg:
https://github.com/google/googletest/issues/3128

Seems like at one point googletest had an issue with an older version of Xcode. I wasn't able to repro the build failure with v1.14.0

This adds googletest to the known_good.json to simplify CI / developer workflow


arcady-lunarg (approved):
LGTM, this certainly does make things a lot simpler.

PR #3387, opened 2023-11-01
Make pool allocations more explicit

ncesario-lunarg:
This removes the ability for some classes to be implicitly allocated using the pool allocator. Now NewPoolObject should be used to explicitly allocate any object using the pool allocator.

Some other things this change does:
- Remove some of the NewPool* non-generic functions, replacing them with a single "make_unique" style function.
- Existing pointers allocated with the standard allocator are replaced with smart pointers.


arcady-lunarg:
I think we're going to take a different approach here, plus this isn't really a priority at the moment.

PR #3388, opened 2023-11-06
Out-of-range floats should overflow to infinity

jimihem:
…40 and 1.0e-50 are normal values.

Shader1：
precision highp float;
out vec4 my_FragColor;
void main()
{
// Out-of-range floats should overflow to infinity // GLSL ES 3.00.6 section 4.1.4 Floats:
// "If the value of the floating point number is too large (small) to be stored as a single precision value, it is converted to positive (negative) infinity" float correct = isinf(1.0e40) ? 1.0 : 0.0;
my_FragColor = vec4(0.0, correct, 0.0, 1.0);
}

precision highp float;
out vec4 my_FragColor;
void main()
{
// GLSL ES 3.00.6 section 4.1.4 Floats:
// "A value with a magnitude too small to be represented as a mantissa and exponent is converted to zero." // 1.0e-50 is small enough that it can't even be stored as subnormal. float correct = (1.0e-50 == 0.0) ? 1.0 : 0.0;
my_FragColor = vec4(0.0, correct, 0.0, 1.0);
}
KHR-GLES3.number_parsing.float_out_of_range_as_infinity

### DIFF
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -2590,6 +2590,18 @@ TIntermConstantUnion* TIntermediate::addConstantUnion(double d, TBasicType baseT
 {
     assert(baseType == EbtFloat || baseType == EbtDouble || baseType == EbtFloat16);
 
+    if (isEsProfile() && (baseType == EbtFloat || baseType == EbtFloat16)) {
+        int exponent = 0;
+        frexp(d, &exponent);
+        int minExp = baseType == EbtFloat ? -126 : -14;
+        int maxExp = baseType == EbtFloat ? 127 : 15;
+        if (exponent > maxExp) { //overflow, d = inf
+            d = std::numeric_limits<double>::infinity();
+        } else if (exponent < minExp) { //underflow, d = 0.0;
+            d = 0.0;
+        }
+    }
+
     TConstUnionArray unionArray(1);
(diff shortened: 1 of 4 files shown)


arcady-lunarg:
Hi, thanks for the patch. I think I see what you're doing here, and it makes sense. However, could you update the expected test results in `Test/baseResults`? There are instructions for how to do this in the README. Also, could you rewrite the commit description to explain a bit more clearly what exactly the change is meant to do? Seems like you're trying to handle values that are out of range for half-precision types?

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -2589,7 +2589,17 @@ TIntermConstantUnion* TIntermediate::addConstantUnion(bool b, const TSourceLoc&
 TIntermConstantUnion* TIntermediate::addConstantUnion(double d, TBasicType baseType, const TSourceLoc& loc, bool literal) const
 {
     assert(baseType == EbtFloat || baseType == EbtDouble || baseType == EbtFloat16);
-
+    if (baseType == EbtFloat)
+        d = (float)d;
```
ncesario-lunarg, line 2593: Mixing no `{}` with `{}` a single if/else is a bit unreadable IMHO (same comment below).

arcady-lunarg (approved):
LGTM, thanks.

PR #3390, opened 2023-11-07
Add a helper applySpirvDecorate to handle spirv_decorate_xxx directives

amdrexu:
Some code paths could be shared.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -66,6 +66,7 @@ namespace spv {
 #include <iomanip>
 #include <list>
 #include <map>
+#include <optional>
 #include <stack>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


SPIRV/GlslangToSpv.cpp
```diff
@@ -4705,6 +4706,64 @@ spv::Id TGlslangToSpvTraverser::convertGlslangToSpvType(const glslang::TType& ty
     return spvType;
 }
 
+// Apply SPIR-V decorations to the SPIR-V object (provided by SPIR-V ID). If member index is valid (>= 0), the
```
arcady-lunarg, line 4709: I am not a fan of using this magical -1 value here. Maybe it would be cleaner to use something like `std::optional`?

arcady-lunarg (changes requested):

SPIRV/GlslangToSpv.cpp
ncesario-lunarg: ➕

amdrexu: Thank you. It is a good idea. I revise the change.

ncesario-lunarg (approved):

arcady-lunarg (approved):

PR #3396, opened 2023-11-07
Fix some GCC warnings

ncesario-lunarg:
This fixes some -Wconversion warnings from gcc.

~~There are still several GCC conversion warnings in Types.h due to the use of bit fields that will be resolved in a future change.~~


glslang/Include/ConstantUnion.h
```diff
@@ -503,7 +503,7 @@ class TConstUnion {
         case EbtInt:    returnValue.setIConst(iConst % constant.iConst); break;
         case EbtUint:   returnValue.setUConst(uConst % constant.uConst); break;
         case EbtInt8:   returnValue.setI8Const(i8Const % constant.i8Const); break;
-        case EbtInt16:  returnValue.setI8Const(i8Const % constant.i16Const); break;
+        case EbtInt16:  returnValue.setI8Const(static_cast<int8_t>(i8Const % constant.i16Const)); break;
```
arcady-lunarg, line 506: I think this is an actual bug and this should be setI16Const. Adding the warning turned out to be helpful after all!

arcady-lunarg (changes requested):
I think the warning actually found us a genuine bug, we should probably fix that.

glslang/CMakeLists.txt
```diff
@@ -134,6 +134,9 @@ add_library(MachineIndependent STATIC ${MACHINEINDEPENDENT_SOURCES} ${MACHINEIND
 set_property(TARGET MachineIndependent PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_property(TARGET MachineIndependent PROPERTY FOLDER glslang)
 
+# -Wunused-but-set-variable is triggered in code generated by bison that we do not control. Turn this warning off, but only for the generated.
+set_source_files_properties(MachineIndependent/glslang_tab.cpp PROPERTIES COMPILE_FLAGS -Wno-unused-but-set-variable)
```
ncesario-lunarg, line 138: @arcady-lunarg this will get rid of the `warning: variable 'yynerrs' set but not used [-Wunused-but-set-variable]
` warning. I think  it's ok to turn this off for a generated file like this.

arcady-lunarg (changes requested):
Still one issue left and this can be merged I think.

arcady-lunarg (approved):
LGTM, thanks

PR #3398, opened 2023-11-14
Each number sign (#) can be preceded in its line only by spaces or horizontal tabs

jimihem:
Following case need report error.
 #define VEC4_STRING_PARAM(a, b, c, d) vec4(#a, #b, c, d)
According spec
Each number sign (#) can be preceded in its line only by spaces or horizontal tabs. It may also be followed by spaces and horizontal tabs, preceding the directive.

glcts.exe --deqp-case=KHR-GL45.shaders.preprocessor.basic.stringification_vertex 
glcts.exe --deqp-case=KHR-GL45.shaders.preprocessor.basic.stringification_fragment


ncesario-lunarg:
@jimihem Looks like these  changes effect a lot of tests. Would you be able to fix those as well as add a test that catches this error inside `Test/badMacroArgs.frag`?

glslang/MachineIndependent/preprocessor/PpContext.h
```diff
@@ -527,7 +553,7 @@ class TPpContext {
     public:
         tStringInput(TPpContext* pp, TInputScanner& i) : tInput(pp), input(&i) { }
         virtual int scan(TPpToken*) override;
-
+        bool isStringInput() { return true; }
```
arcady-lunarg, line 556: You need to put "override" here to avoid a compiler warning.

arcady-lunarg (changes requested):
One minor nit here, but I think this looks good. If you want, you can fix the tests in tokenLength.vert to not trigger this new error, but I can also do that in a followup change.

arcady-lunarg:
I've applied the fix myself as well as fixed up the test to generate fewer errors.

PR #3401, opened 2023-11-16
Improve non-semantic debug info for variables

qingyuanzNV:
This patch fixes two things:
1. Correctly set the source location of global variable correctly. Currently, glslang will set the location of the debug variable to the site of its first use, which is definitely undesirable. This patch forwards the site of definition from `TSymbol` to `TIntermSymbol` and finally to `createDebugGlobalVariable`. This is technically inefficient usage of space, but we don't have AST node for decls in glslang. 
```
int x; // <-- now we point here

int foo() {
    return x; // <-- we used to point here in DebugGlobalVariable
}
```
2. Set FlagTypePassByReference/FlagTypePassByValue properly to parameter so the consumer could identify a wrapping pointer and a real pointer in parameter type.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -4311,7 +4311,8 @@ spv::Id TGlslangToSpvTraverser::createSpvVariable(const glslang::TIntermSymbol* 
         initializer = builder.makeNullConstant(spvType);
     }
 
-    return builder.createVariable(spv::NoPrecision, storageClass, spvType, name, initializer, false);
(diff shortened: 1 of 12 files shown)


qingyuanzNV:
This PR is still a draft.

@jeremy-lunarg Could you please take a quick look on the structure of the change? The change is somewhat wasteful in the use of storage, but I couldn't come up with a better solution without being hacky.

arcady-lunarg:
I wouldn't be too concerned about the resource usage of storing extra `TSourceLoc`s, but you can try to quantify the impact on memory usage. glslang even has some rudimentary ability to do this itself via the OS_DumpMemoryCounters function, you can try to use that though it may require making some source code changes to use it.

PR #3406, opened 2023-11-20
Fix CMake spirv-tools dependency

moritz-h:
The build is currently broken when using external SPIRV-Tools. This happens when `find_package(SPIRV-Tools-opt)` is used here: https://github.com/KhronosGroup/glslang/blob/589431af5c86b0be3cc23175258451be1f5d25b3/CMakeLists.txt#L271

The problem is using `${spirv-tools_SOURCE_DIR}`. This variable is defined by CMake on [a project() command](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html). But a spirv-tools project() command is only present when using spirv-tools via `add_subdirectory`. Therefore, when using external spirv-tools via find_package, the variable is just empty and no include dirs are set.

This PR is trying to fix this by linking the spirv-tools-opt target instead, which should set the include directories automatically and will work with using spirv-tools via add_subdirectory and find_package. I have no full overview of the glslang cmake sources, so I'm not aware of side effects, but here is what I intended to do:
- Do not include ${spirv-tools_SOURCE_DIR}/source. I hope there is no dependency on internal sources of the spirv-tools repo, but in that case external spiv tools would probably not work at all.
- It seems headers of the SPIRV target include spirv-tools headers, so SPPIRV-Tools-opt seems to be a public dependency of the SPIRV target
- remove spirv-tools include from StandAlone. From a quick search through the sources it seems no spirv-tool headers are used. If this was present because of transitive dependency of the SPIRV target this should now be solved by the SPIRV target having a public dependency on SPIRV-Tools-opt

### DIFF
--- a/SPIRV/CMakeLists.txt
+++ b/SPIRV/CMakeLists.txt
@@ -106,11 +106,7 @@ if(WIN32 AND BUILD_SHARED_LIBS)
 endif()
 
 if(ENABLE_OPT)
-    target_include_directories(SPIRV
-        PRIVATE ${spirv-tools_SOURCE_DIR}/include
-        PRIVATE ${spirv-tools_SOURCE_DIR}/source
-    )
-    target_link_libraries(SPIRV PRIVATE MachineIndependent SPIRV-Tools-opt)
+    target_link_libraries(SPIRV PRIVATE MachineIndependent PUBLIC SPIRV-Tools-opt)
     target_include_directories(SPIRV PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../External>
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/External>)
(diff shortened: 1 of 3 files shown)


ncesario-lunarg (approved):
This LGTM and like "more correct" cmake, but I'll defer to @arcady-lunarg (and @juan-lunarg if he can review).

ncesario-lunarg:
Looks like the Android build is not happy with this change.

juan-lunarg (approved):
💯

juan-lunarg:
> Looks like the Android build is not happy with this change.

Can you try re-running Android CI? @ncesario-lunarg it looks completely unrelated.

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
```

moritz-h:
> Looks like the Android build is not happy with this change.
```
CMake Error at /home/runner/work/_temp/1493929610/cmake-3.27.7-linux-x86_64/share/cmake-3.27/Modules/CMakeDetermineSystem.cmake:154 (message):
  Could not find toolchain file:
  /usr/local/lib/android/sdk/ndk/23.2.8568313/build/cmake/android.toolchain.cmake
```
I think this looks more like a general problem of the CI job. I.e. #3405 has the exact same problem.

juan-lunarg:
> I think this looks more like a general problem of the CI job. I.e. #3405 has the exact same problem.

I see the problem. I'll make a PR.

juan-lunarg:
Android failure is unrelated:
https://github.com/KhronosGroup/glslang/pull/3407/files

PR #3407, opened 2023-11-20
Fix Android CI

juan-lunarg:
NDK 23.2.8568313 was removed from the latest runner


juan-lunarg:
https://github.com/actions/runner-images/issues/8383

ncesario-lunarg (approved):
Thanks @juan-lunarg!

PR #3409, opened 2023-11-21
Add iOS build to CI

juan-lunarg:
Fix iOS build as well as minor Android cleanup since the problems for both platforms are so similar

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,22 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
     set(CMAKE_BUILD_TYPE "Debug")
 endif()
 
+# Currently iOS and Android are very similar.
+# They both have their own packaging (APP/APK).
+# Which makes regular executables/testing problematic.
+#
+# Currently the only deliverables for these platforms are
+# libraries (either STATIC or SHARED).
+#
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):
LGTM! Thanks again Juan.

juan-lunarg:
FYI @richard-lunarg

CMakeLists.txt
```diff
@@ -59,6 +59,22 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
...
+# They both have their own packaging (APP/APK).
+# Which makes regular executables/testing problematic.
+#
+# Currently the only deliverables for these platforms are
+# libraries (either STATIC or SHARED).
+#
+# Furthermore testing is equally problematic.
+if (IOS OR ANDROID)
+    set(ENABLE_GLSLANG_BINARIES OFF)
+    set(SPIRV_SKIP_EXECUTABLES ON)
```
juan-lunarg, line 72: For extra context the ndk-builds currently in glslang and spirv-tools also don't create executables. For the aforementioned reasons.

juan-lunarg: So this should also speed up our Android / iOS CI in downstream repos since we can avoid building extra executables.

arcady-lunarg (approved):

PR #3410, opened 2023-11-21
Cleanup MacOS CI

juan-lunarg:
Along with minor commits addressing comments in this file.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Test (standalone)
         run: cd Test && ./runtests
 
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -125,29 +119,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, macos-12]
+        os: [macos-12, macos-13]
```
juan-lunarg, line 122: Developer survey shows little to no users still use MacOS 11 or earlier.
Screenshot 2023-11-21 at 12 46 52 PM

ncesario-lunarg (approved):

PR #3411, opened 2023-11-21
cmake: Cleanup usage of option

juan-lunarg:
By default option is OFF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ include(CMakePackageConfigHelpers)
 # Needed for CMAKE_DEPENDENT_OPTION macro
 include(CMakeDependentOption)
 
-option(BUILD_SHARED_LIBS "Build Shared Libraries" OFF)
+option(BUILD_SHARED_LIBS "Build Shared Libraries")
 option(BUILD_EXTERNAL "Build external dependencies in /External" ON)
 
 set(LIB_TYPE STATIC)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -59,16 +59,15 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
     set(CMAKE_BUILD_TYPE "Debug")
 endif()
 
-option(SKIP_GLSLANG_INSTALL "Skip installation" ${SKIP_GLSLANG_INSTALL})
```
juan-lunarg, line 62: This is actively confusing and hard to read.

ncesario-lunarg (commented):
Seems fine to me, but at the same time is there any downside to making the initial value explicit?

juan-lunarg:
> Seems fine to me, but at the same time is there any downside to making the initial value explicit?

It's less idiomatic.

CMakeLists.txt
juan-lunarg, line 62: This is the main reason I made this PR. It's difficult to tell by reading the code that SKIP_GLSLANG_INSTALL is OFF by default.

ncesario-lunarg: Yeah, I agree this was confusing and should change. I'm not sure I agree that the proposed change is clearer and/or more idiomatic than `option(SKIP_GLSLANG_INSTALL "Skip installation" OFF)`, but I'll leave that up to @arcady-lunarg :)

juan-lunarg: I don't mind either way. I just use this approach in every other repo I work on.

arcady-lunarg: Judging by the cmake book, not having "OFF" is more idiomatic.

PR #3412, opened 2023-11-21
Fix Xcode 15 linker warning

juan-lunarg:
XCode 15 includes a completely new linker implementation. Hence some flags are deprecated.

The default is now error so we can simply remove the linker option.


juan-lunarg:
This fixes the following linker warning:

```
[build] ld: warning: -undefined error is deprecated
```

ncesario-lunarg (approved):

PR #3413, opened 2023-11-22
fix: Parse SV_ViewID keywords for hlsl

FrostyLeaves:
I tried using glslang to parse hlsl and found that SV_ ViewID will be treated as part of the vertex structure. It will cause my vertex data to be incorrectly parsed into irregular mesh.

Add "SV_VIEWID" to SemanticMap for generate spv files with "VK_KHR_multiview" OpExtension.


arcady-lunarg (approved):
Thanks for the PR, and thanks for adding a test. LGTM.

PR #3415, opened 2023-11-22
Linker option cleanup

juan-lunarg:
1. Move -Wl,--no-undefined into CI script

Removes complexity around setting this linker option from the
CMakeLists.txt while still providing error checking.

Furthermore given there are now 4 linkers it's simpler to just have this linker option in CI than worry about individual users setups:
- gnu
- gold
- llvm
- mold

Not to mention how this apparently doesn't work for OpenBSD and
Emscripten it's much simpler to set this via LDFLAGS instead.

2. Set linker warnings as errors for Windows builds

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -35,6 +35,7 @@ jobs:
           CMAKE_GENERATOR: Ninja
           CMAKE_C_COMPILER_LAUNCHER: ccache
           CMAKE_CXX_COMPILER_LAUNCHER: ccache
+          LDFLAGS: -Wl,--no-undefined
       - name: Build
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):

juan-lunarg:
> Furthermore given there are now 4 linkers it's simpler to just have this linker option in CI than worry about individual users setups:

Technically there is also the Apple linker. So I guess that's 5 actually.

juan-lunarg:
Does this PR look good to you @arcady-lunarg ?

arcady-lunarg:
The one thing I would be concerned about is developers not seeing the potential errors this warning is meant to catch, or at least not catching them until they try to run the CI.

PR #3417, opened 2023-11-22
Remove OVERRIDE_MSVCCRT

juan-lunarg:
CMake 3.15 removes the need for all of this custom code.


CMakeLists.txt
```diff
@@ -118,10 +118,6 @@ endif()
 
 if(WIN32)
     set(CMAKE_DEBUG_POSTFIX "d")
-    option(OVERRIDE_MSVCCRT "Overrides runtime of MSVC " ON)
-    if(MSVC AND OVERRIDE_MSVCCRT)
-        include(ChooseMSVCCRT.cmake)
-    endif()
```
juan-lunarg, line 124: Maintaining a custom approach is worse for package managers and users of glslang long term.

ncesario-lunarg (approved):
Always nice to see special case code like this removed :)

arcady-lunarg (approved):

PR #3419, opened 2023-11-22
Cleanup ENABLE_PCH

juan-lunarg:
Min is 3.17 checks for 3.16 are no longer needed.


CMakeLists.txt
```diff
@@ -102,10 +102,11 @@ option(ENABLE_OPT "Enables spirv-opt capability if present" ON)
 if(MINGW OR (APPLE AND ${CMAKE_CXX_COMPILER_ID} MATCHES "GNU"))
```
juan-lunarg, line 102: Part me wonders if this problem still exists but that's another PR

arcady-lunarg (approved):

PR #3420, opened 2023-11-23
No external install include dir

moritz-h:
It seems there is no `External` directory installed. This leads to include dir not found CMake errors when glslang is used from other tools with find_package().

### DIFF
--- a/SPIRV/CMakeLists.txt
+++ b/SPIRV/CMakeLists.txt
@@ -108,8 +108,7 @@ endif()
 if(ENABLE_OPT)
     target_link_libraries(SPIRV PRIVATE MachineIndependent PUBLIC SPIRV-Tools-opt)
     target_include_directories(SPIRV PUBLIC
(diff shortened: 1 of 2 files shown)


arcady-lunarg:
cc @juan-lunarg for review

juan-lunarg:
@moritz-h is correct. External isn't used.

```
cmake --install build --prefix /tmp | grep External
warning: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive library: /tmp/lib/libOGLCompiler.a the table of contents is empty (no object file members in the library define global symbols)
warning: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive library: /tmp/lib/libHLSL.a the table of contents is empty (no object file members in the library define global symbols)
```

juan-lunarg (approved):

PR #3425, opened 2023-11-28
Remove known_good_khr.json

juan-lunarg:
Looking at the git history this file isn't used at all.
https://github.com/KhronosGroup/glslang/commits/main/known_good_khr.json

Only changes I can see come from 2018


ncesario-lunarg:
It would be nice to see this go, but is it possible this is used for extension development in gitlab?

juan-lunarg:
> It would be nice to see this go, but is it possible this is used for extension development in gitlab?

Not sure. This PR is also a question. If it is still used I'd like to make that more apparent.

arcady-lunarg:
The gitlab version hasn't been updated in a while either but it is still potentially useful for doing development in gitlab on something that a) isn't public and b) involves changes to headers/tools as well as glslang, with those changes being staged in gitlab. While that workflow hasn't seen much use of late, I don't think it's one we want to preclude.

ncesario-lunarg:
I'm not sure how helpful the "site map" is in `update_glslang_sources.py`. I'd instead vote for removing that map and just putting `https://github.com` or `git@gitlab.khronos.org` in the `site` field for each entry in `known_good.json`, and then remove `known_good_khr.json`. This would be more aligned with what VVL does, which I prefer (but maybe I'm just biased).

juan-lunarg:
That would provide consistency and QoL features

ncesario-lunarg:
I believe the change I'm suggesting would just be a simplification, but happens to make the update script less restrictive. I can put up a PR.

PR #3427, opened 2023-11-28
Simplify update_glslang_sources.py

ncesario-lunarg:
Store the full URL to a repo in known_good.json rather than splitting up the URL into "site" and "subrepo." This separation appears to be for use with the "--site" flag in update_glslang_sources.py.

The "--site" flag has been removed. This appears to have been added as a convenience for switching between github and gitlab. However, this would not work currently as the organizations for glslang dependencies do match between github and gitlab, and "subrepo" contains both the organization and the repo name.

FYI @juan-lunarg.


juan-lunarg (approved):
👍🏾

arcady-lunarg:
I would prefer to keep the existing workflow for now.

PR #3430, opened 2023-11-29
support GL_ARB_texture_multisample extension.

jimihem:
This extension allows the use of "texelFetch" and "textureSize" with 2DMS sampler.
#version 140
out float result;

uniform sampler2DMS data;
uniform sampler2DMSArray data1;
void main()
{
result = texelFetch(data, ivec2(0), 3).r;
ivec2 temp = textureSize(data);
result = texelFetch(data1, ivec3(0), 3).r;
ivec3 temp1 = textureSize(data1);
}

### DIFF
--- a/Test/GL_ARB_texture_multisample.vert
+++ b/Test/GL_ARB_texture_multisample.vert
@@ -1,0 +1,14 @@
+#version 140
+#extension GL_ARB_texture_multisample : enable
+
+out float result;
+
+uniform sampler2DMS data;
+uniform sampler2DMSArray data1;
+void main()
+{
+result = texelFetch(data, ivec2(0), 3).r;
+ivec2 temp = textureSize(data);
+result = texelFetch(data1, ivec3(0), 3).r;
+ivec3 temp1 = textureSize(data1);
+}
--- a/Test/GL_ARB_texture_multisample.vert
+++ b/Test/GL_ARB_texture_multisample.vert
@@ -2,13 +2,27 @@
 #extension GL_ARB_texture_multisample : enable
 
 out float result;
+out int result1;
+out uint result2;
 
 uniform sampler2DMS data;
 uniform sampler2DMSArray data1;
+uniform isampler2DMS data2;
+uniform usampler2DMSArray data3;
+uniform usampler2DMS data4;
+uniform isampler2DMSArray data5;
 void main()
 {
     result = texelFetch(data, ivec2(0), 3).r;
     ivec2 temp = textureSize(data);
     result = texelFetch(data1, ivec3(0), 3).r;
     ivec3 temp1 = textureSize(data1);
+	result1 = texelFetch(data2, ivec2(0), 3).r;
+    temp = textureSize(data2);
+	result2 = texelFetch(data3, ivec3(0), 3).r;
+    temp1 = textureSize(data3);
+	result2 = texelFetch(data4, ivec2(0), 3).r;
+    temp = textureSize(data4);
+	result1 = texelFetch(data5, ivec3(0), 3).r;
+    temp1 = textureSize(data5);
 }
--- a/Test/GL_ARB_texture_multisample.vert
+++ b/Test/GL_ARB_texture_multisample.vert
@@ -7,8 +7,8 @@ uniform sampler2DMS data;
 uniform sampler2DMSArray data1;
 void main()
 {
-result = texelFetch(data, ivec2(0), 3).r;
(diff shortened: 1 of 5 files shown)


Test/GL_ARB_texture_multisample.vert
ncesario-lunarg, line 1: I'm guessing this is not supposed to be empty?

Test/baseResults/GL_ARB_texture_multisample.vert.out
ncesario-lunarg, line 1: Probably not the desired output?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1352,7 +1352,11 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
                 requireInt16Arithmetic(loc, "built-in function", "(u)int16 types can only be in uniform block or buffer storage");
             if (builtIn && fnCandidate->getType().contains8BitInt())
                 requireInt8Arithmetic(loc, "built-in function", "(u)int8 types can only be in uniform block or buffer storage");
-
+            if (builtIn && (fnCandidate->getBuiltInOp() == EOpTextureFetch || fnCandidate->getBuiltInOp() == EOpTextureQuerySize)) {
+                TString typeName = (*fnCandidate)[0].type->getSampler().getString();
+                if ((typeName == "sampler2DMS" || typeName == "sampler2DMSArray") && version == 140)
```
ncesario-lunarg, line 1357: I think this makes sense after looking at the spec, but just to clarify, this extension is _only_ required for version 140? i.e., it is not possible to use this extension prior to 140?

ncesario-lunarg (changes requested):
I think this looks good apart from the test file.

glslang/MachineIndependent/ParseHelper.cpp
jimihem: Thanks for your suggesstion, I have made modifications to this PR.

Test/GL_ARB_texture_multisample.vert
```diff
@@ -0,0 +1,14 @@
...
+out float result;
+
+uniform sampler2DMS data;
+uniform sampler2DMSArray data1;
+void main()
+{
+result = texelFetch(data, ivec2(0), 3).r;
+ivec2 temp = textureSize(data);
+result = texelFetch(data1, ivec3(0), 3).r;
+ivec3 temp1 = textureSize(data1);
```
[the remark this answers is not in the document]
jimihem: I have organized the format @ncesario-lunarg

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1352,7 +1352,11 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
                 requireInt16Arithmetic(loc, "built-in function", "(u)int16 types can only be in uniform block or buffer storage");
             if (builtIn && fnCandidate->getType().contains8BitInt())
                 requireInt8Arithmetic(loc, "built-in function", "(u)int8 types can only be in uniform block or buffer storage");
-
+            if (builtIn && (fnCandidate->getBuiltInOp() == EOpTextureFetch || fnCandidate->getBuiltInOp() == EOpTextureQuerySize)) {
+                TString typeName = (*fnCandidate)[0].type->getSampler().getString();
+                if ((typeName == "sampler2DMS" || typeName == "sampler2DMSArray") && version <= 140)
```
ncesario-lunarg, line 1357: Do the integer versions of these functions need to be handled?

ncesario-lunarg (commented):
One question, but LGTM.

Looks like there are really 2 distinct changes here; I'm not sure how they are related?

glslang/MachineIndependent/ParseHelper.cpp
arcady-lunarg: I think rather than calling `getString()` here it would be better to directly check the relevant property, which I guess is `getSampler()->isMultiSample()`. And that would handle the usampler/isampler case transparently as well.

arcady-lunarg (changes requested):
Could you add tests for usampler2DMS and isampler2DMS just for completeness? Also, I think it would be better not to construct the sampler type string and just check the TSampler's properties directly via the appropriate accessors.

arcady-lunarg (approved):
Thanks, LGTM

PR #3432, opened 2023-11-29
Add BUILD_WERROR option

ncesario-lunarg:
Adds a cmake BUILD_WERROR option to enable warnings as errors. This option is off by default, but is enabled for CI runs.

This change also cleans up a few more compiler warnings.


arcady-lunarg (approved):

arcady-lunarg:
Newer versions of cmake have a built-in mechanism to do this but until we require cmake 3.24 this is fine.

ncesario-lunarg:
> Newer versions of cmake have a built-in mechanism to do this but until we require cmake 3.24 this is fine.

Yes, I left a comment to remind that this can be cleaned up if we bump the cmake version.

PR #3433, opened 2023-11-29
Emit DebugTypePointer when non-semantic debug info is enabled.

sajjadmirzanv:


ncesario-lunarg (approved):
This LGTM, but should probably have a look by @jeremy-lunarg.

arcady-lunarg (approved):
LGTM, thanks.

PR #3434, opened 2023-11-30
Improve overflow_underflow_toinf_0 test somewhat

arcady-lunarg:
Add test cases that will make explicit ±0.0 and ±INF appear in the AST output to make sure those cases are handled correctly.
Also, move overflow_underflow_toinf_0 from runtests to gtest. gtest is the preferred framework for simple tests that don't require    passing special command-line options to glslang.


arcady-lunarg:
cc: @jimihem this is a follow-up change with some minor improvements to your earlier PR.

ncesario-lunarg (approved):

PR #3438, opened 2023-12-01
location aliasing

jimihem:
when location aliasing, the aliases sharing the location must have the same underlying numerical type (floating-point or integer) and the same auxiliary storage and interpolation qualification. The following case, glslang need report error.

layout(vertices = 1) out;

layout (location = 1, component = 0)  in  double gohan[]; layout (location = 1, component = 2)  in  float goten[];

in  vec4 vs_tcs[];
out vec4 tcs_tes[];

void main()
{
}


arcady-lunarg (approved):
LGTM, thanks for adding tests.

jimihem:
@arcady-lunarg Can you help me check why this test item crash because there is no linux-asan runtime environment on my computer.
[linux-asan (gcc, g++, Debug, -fsanitize=address)]

arcady-lunarg:
I think the problem is that a `vector<TIORange>` is being turned into a `vector<TRange>` somewhere and then deallocated, and this passed ASAN because the structs were the same size somehow, and your adding new struct members changed that? But that seems like a pretty bizarre error in any case, which needs to be investigated, and I will try to look into. In any case, the information produced by ASAN is in the CI output.

jimihem:
@arcady-lunarg Has this test crash problem been fixed ?

arcady-lunarg:
This is a real bug, not related to your change. Apparently when `qualifier.storage` is equal to `EvqTileImageEXT`, it overruns the `usedIO` array and corrupts whatever is next in the `TIntermediate` struct. I'm going to make a patch.

svenvh:
@jimihem after your commit, the following fragment shader results in `ERROR: 0:4: 'location' : the aliases sharing the location 0 must be the same basic type and interpolation qualification`:

```
#version 450
  
layout(location=0, component=0) flat in uint u;
layout(location=0, component=1) flat in int i;

void main() {
}
```

Is that intentional?  My reading of the GLSL specification is that differences in signedness shouldn't matter:

> Two (or more) variables that share the same location must satisfy the following conditions:
>  Share the same base data type. So you can have alias locations for floating-point types, or integer types (signed and unsigned does not matter for this), but you cannot have a float sit alongside an integer.

pmistryNV:
This commit will fail tests from GL CTS https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/openglcts/modules/gl/gl4cEnhancedLayoutsTests.cpp

It will error out following aliases based on  mismatch in auxilary qualifiers. I think the check is not correct and must let auxilary qualifiers mismatch.

layout (location = 1, component = 0) centroid flat in int gohan;
layout (location = 1, component = 1)  flat in ivec3 goten;

pmistryNV:
> layout (location = 1, component = 0) centroid flat in int gohan;
> layout (location = 1, component = 1) flat in ivec3 goten;

Filed a bug to track this bug https://github.com/KhronosGroup/glslang/issues/3598

PR #3439, opened 2023-12-01
Only install/test if PROJECT_IS_TOP_LEVEL

juan-lunarg:
Further remove installing glslangtests. There isn't a need to do that. glslangtests isn't a deliverable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,11 @@
 cmake_minimum_required(VERSION 3.17.2)
 project(glslang)
 
+if (CMAKE_VERSION VERSION_LESS "3.21")
+    # https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
+    string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
+endif()
+
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # Adhere to GNU filesystem layout conventions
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -302,33 +300,38 @@ if(ENABLE_GLSLANG_BINARIES)
...
-if(GLSLANG_TESTS)
-    enable_testing()
-    add_subdirectory(gtests)
+# Testing / installation only makes sense when the project is top level.
+#
+# Otherwise add_subdirectory users have a harder time consuming the library.
+# Since glslang will pollute the installation and add undesirable testing.
+if(PROJECT_IS_TOP_LEVEL)
+    option(GLSLANG_TESTS "Enable glslang testing")
+    if(GLSLANG_TESTS)
```
juan-lunarg, line 309: For precedent we have been doing this in vulkan-headers and spirv-headers and it's been working without issue. In fact it fixed multiple workflow issues for more than a few developers.

gtests/CMakeLists.txt
```diff
@@ -68,21 +68,6 @@ if(GLSLANG_TESTS)
         glslang_pch(glslangtests ${CMAKE_CURRENT_SOURCE_DIR}/pch.h)
         set_property(TARGET glslangtests PROPERTY FOLDER tests)
         glslang_set_link_args(glslangtests)
-        if(ENABLE_GLSLANG_INSTALL)
-            install(TARGETS glslangtests EXPORT glslang-targets)
```
juan-lunarg, line 72: There is no reason to install tests

arcady-lunarg (approved):
Seems reasonable to me, we definitely don't need to be installing the tests.

PR #3443, opened 2023-12-04
Fix continous_deployment

juan-lunarg:
libSPIRV-Tools.a and libSPIRV-Tools-opt.a aren't being installed.

### DIFF
--- a/External/CMakeLists.txt
+++ b/External/CMakeLists.txt
@@ -71,8 +71,7 @@ endif()
 if(ENABLE_OPT AND NOT TARGET SPIRV-Tools-opt)
     if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/spirv-tools)
         set(SPIRV_SKIP_TESTS ON CACHE BOOL "Skip building SPIRV-Tools tests")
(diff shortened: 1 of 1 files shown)


External/CMakeLists.txt
```diff
@@ -71,8 +71,7 @@ endif()
 if(ENABLE_OPT AND NOT TARGET SPIRV-Tools-opt)
     if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/spirv-tools)
         set(SPIRV_SKIP_TESTS ON CACHE BOOL "Skip building SPIRV-Tools tests")
-        # EXCLUDE_FROM_ALL will prevent unneccessary build/installation from spirv-tools
-        add_subdirectory(spirv-tools EXCLUDE_FROM_ALL)
```
juan-lunarg, line 75: Currently continuous deployment ships static libraries. As a result all relevant static libraries need to be shipped.

This is generally speaking why package managers prefer SHARED libraries for shipping.

juan-lunarg: TLDR: SHARED libraries result in less deliverables / implementation details being exposed to the user

arcady-lunarg (approved):

Issue #3223, opened 2023-05-11
sampler or image can be used as return type only when the extension GL_ARB_bindless_texture enabled


vbachet:
The following code generates an error since https://github.com/KhronosGroup/glslang/pull/3066

```
layout(set = 0, binding = 0) uniform texture2D textures[];
layout(set = 0, binding = 1) uniform sampler samplers[];

sampler2D getTextureSampler(int texID, int smpID)
{
    return sampler2D(textures[texID], samplers[smpID]);
}
```

The error message is `sampler or image can be used as return type only when the extension GL_ARB_bindless_texture enabled`

If we try to enable this extension, we get the following error: `GL_ARB_bindless_texture not allowed when using generating SPIR-V codes`

PR #3444, opened 2023-12-06
Advertise GL_ARB_bindless_texture as unavailable for SPIR-V outputs

sbourasse:
As mentioned in issue https://github.com/KhronosGroup/glslang/issues/3223, the support for `GL_ARB_bindless_texture` results in confusing error messages, where a user is asked to enable an extension that is actually unavailable to them.

This PR amends one such error message to be clearer. It also moves the unconditionnal definition of the `GL_ARB_bindless_texture` macro so it isn't defined when targeting SPIR-V, since defining it anyway could break shader code that relied on it to test for the extension's availability.


arcady-lunarg (approved):
LGTM, thanks.

PR #3450, opened 2023-12-14
Fixed check uint64 before set bool

GermanAizek:
@arcady-lunarg U64 value type was incorrectly checked, commit fixed this.


arcady-lunarg (approved):
This doesn't have much practical effect since the uint64_t or int64_t gets converted to a bool anyway, but it's better to be meticulous with our types.

PR #3451, opened 2023-12-14
Update version of macOS deployment GHA runner

ncesario-lunarg:
@juan-lunarg Should we be setting `MACOSX_DEPLOYMENT_TARGET` in our builds as well?

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -101,7 +101,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [{genus: macos-11, family: osx}]
+                os: [{genus: macos-12, family: osx}]
                 compiler: [{cc: clang, cxx: clang++}]
                 cmake_build_type: [Debug, Release]
         steps:


juan-lunarg:
> @juan-lunarg Should we be setting `MACOSX_DEPLOYMENT_TARGET` in our builds as well?

I found no one was keeping it up to date. So I'm hesitant to recommend it.

That's why I added the usage of the macos-11 runners.

You can also wait until this issue gets fixed and continue using macos-11 runners with the latest version of CMake:
https://github.com/lukka/get-cmake/issues/98

Or remove usage of get-cmake for macos continuous deployment since the cmake version for macos-11 that comes with the docker container is good enough:
https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md

ncesario-lunarg:
I don't have a strong preference. @arcady-lunarg?

juan-lunarg:
> You can also wait until this issue gets fixed and continue using macos-11 runners with the latest version of CMake:
https://github.com/lukka/get-cmake/issues/98

Looks like lukka just fixed this. Just a matter of time before the bots update the dependency now.

ncesario-lunarg:
In that case I'll wait to see if the next deployment is successful before merging this.

juan-lunarg:
Looks like dependabot is set to only run 3 times a week. Meaning get-cmake won't be updated till next week with the fix. Unless you just wanna update it manually.

ncesario-lunarg:
This was resolved by https://github.com/KhronosGroup/glslang/pull/3456.

PR #3453, opened 2023-12-15
Add support for GL_ARB_shading_language_include

### LINKED ISSUE #249: GL_ARB_shading_language_include not supported but GL_GOOGLE_include_directive is
code like this

```
#extension GL_ARB_shading_language_include : require

#include </foo/bar.glsl>
```

gives me these errors

```
ERROR: 0:2: '#extension' : extension not supported: GL_ARB_shading_language_include
ERROR: 0:4: '#include' : required extension not requested: GL_GOOGLE_include_directive
```

may i ask why the ARB extension is not supported? #37 talks about this but the reasoning was not clear to me.

ncesario-lunarg:
Add support for GL_ARB_shading_language_include. Usage is identical to the way GL_GOOGLE_include_directive currently works glslang (since GL_ARB_shading_language_include is inherently a runtime feature and glslang is an offline compiler).

Users can simulate their runtime environment by using a custom glslang::TShader::Includer or using filenames that match their GL runtime names.

Closes #249.

### DIFF
--- a/Test/baseResults/preprocess.arb_shading_language_include.vert.err
+++ b/Test/baseResults/preprocess.arb_shading_language_include.vert.err
@@ -1,0 +1,4 @@
+ERROR: 0:7: '#include' : Could not process include directive for header name: bar.h
+ERROR: 1 compilation errors.  No code generated.
+
+
--- a/glslang/MachineIndependent/preprocessor/Pp.cpp
+++ b/glslang/MachineIndependent/preprocessor/Pp.cpp
@@ -973,7 +973,8 @@ int TPpContext::readCPPline(TPpToken* ppToken)
             break;
         case PpAtomInclude:
             if(!parseContext.isReadingHLSL()) {
-                parseContext.ppRequireExtensions(ppToken->loc, 1, &E_GL_GOOGLE_include_directive, "#include");
+                const std::array exts = { E_GL_GOOGLE_include_directive, E_GL_ARB_shading_language_include };
+                parseContext.ppRequireExtensions(ppToken->loc, exts, "#include");
             }
             token = CPPinclude(ppToken);
             break;
--- a/Test/baseResults/preprocess.include_directive_missing_extension.vert.err
+++ b/Test/baseResults/preprocess.include_directive_missing_extension.vert.err
@@ -1,0 +1,7 @@
+ERROR: 0:3: '#include' : required extension not requested: Possible extensions include:
+GL_GOOGLE_include_directive
+GL_ARB_shading_language_include
+ERROR: 0:3: '#include' : Could not process include directive for header name: bar.h
+ERROR: 2 compilation errors.  No code generated.
+
+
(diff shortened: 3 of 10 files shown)


glslang/MachineIndependent/preprocessor/Pp.cpp
```diff
@@ -973,7 +973,11 @@ int TPpContext::readCPPline(TPpToken* ppToken)
             break;
         case PpAtomInclude:
             if(!parseContext.isReadingHLSL()) {
-                parseContext.ppRequireExtensions(ppToken->loc, 1, &E_GL_GOOGLE_include_directive, "#include");
+                const std::vector<const char*> exts = [this]() -> std::vector<const char*> {
+                    if (parseContext.version >= 150) return { E_GL_GOOGLE_include_directive, E_GL_ARB_shading_language_include };
```
arcady-lunarg, line 977: This isn't actually correct, if you read the extension spec, it's _written against_ 1.50 but _requires_ 1.10, so the extension can be used for any version.

arcady-lunarg (changes requested):
My reading of the [spec](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_shading_language_include.txt) seems to indicate that the extension can be used with any version of GLSL, the code should be changed to reflect that, and it also obviates the need for one of the tests.

glslang/MachineIndependent/preprocessor/Pp.cpp
ncesario-lunarg: Yep, this should be fixed now.

Test/baseResults/preprocess.arb_shading_language_include.vert.err
```diff
@@ -0,0 +1,4 @@
+ERROR: 0:7: '#include' : Could not process include directive for header name: bar.h
```
arcady-lunarg, line 1: Is this failure expected? The test looks like it's supposed to pass.

ncesario-lunarg: I believe this is correct. This is effectively checking that the output is _not_

```
ERROR: 0:7: '#include' : required extension not requested: GL_GOOGLE_include_directive
```

That said, all the pre-processor test output/logic is a bit confusing to me.

All the preprocessor tests use the default "ForbidIncluder": https://github.com/KhronosGroup/glslang/blob/d84255296c3d19642515237c8d016bb685bd0bed/gtests/TestFixture.h#L658 which means that no include files can be resolved.

e.g., https://github.com/KhronosGroup/glslang/blob/main/Test/preprocessor.include.enabled.vert is a "positive" test for `GL_GOOGLE_include_directive`, yet it produces errors because none of the include paths can be resolved: https://github.com/KhronosGroup/glslang/blob/main/Test/baseResults/preprocessor.include.enabled.vert.err.

arcady-lunarg: Ah, that explains it. We can leave it as is for now, and maybe I will do a followup change to fix the preprocessor testing.

arcady-lunarg (approved):

PR #3463, opened 2023-12-24
Fix MinGW build

ncesario-lunarg:


ncesario-lunarg:
@arcady-lunarg I'm going to merge this as it should be low risk and should fix a build issue.

arcady-lunarg:
Thanks for fixing this.

Issue #925, opened 2017-06-08
Not all built-in functions can portably return a const.


johnkslang:
Khronos made an explicit whitelist of which built-in functions, when given constant arguments, must be semantically constant when parsing.  This list is below.  Glslang can help portability by

1. ensuring all of these can be semantically constant
2. giving warnings on any others that get used in a context requiring a semantic constant

### Angle and Trigonometric Functions
radians
degrees
sin
cos
asin
acos

### Exponential Functions
pow
exp
log
exp2
log2
sqrt
inversesqrt

### Common Functions
abs
sign
floor
trunc
round
ceil
mod
min
max
clamp

### Geometric Functions
length
dot
normalize

de2nis:
Does Matrix Function, matrixCompMult(), belong to that list?
OpenGL ES Shading Language version 1.00 defines this function and, in addition, Validator requires constant initializers in ES profile.

johnkslang:
I think the list is accurate WRT matrixCompMult(). Glslang should give a warning if it is attempted with all constants.

de2nis:
Thanks.
Is this task under development? mod() implementation seems to be missing. If it's true I could give some help with a PR.

johnkslang:
Sure, if you want to help. I imagine best to somehow flag/hook all the folding operations that glslang does, but are not on the whitelist, and then issue a warning for those.

PR #3465, opened 2023-12-27
Support constant expression calculated by matrixCompMult.

jimihem:
glslang report error for  "const mat2 cval = matrixCompMult(mat2(1.0), mat2(1.0));".
Spec says:
The value returned by certain built-in function calls whose arguments are all constant expressions is also constant expressions.

### DIFF
--- a/Test/baseResults/matrixCompMult.vert.out
+++ b/Test/baseResults/matrixCompMult.vert.out
@@ -1,0 +1,33 @@
+matrixCompMult.vert
+Shader version: 460
+0:? Sequence
+0:5  Function Definition: main( ( global void)
+0:5    Function Parameters: 
+0:?     Sequence
+0:26      move second child to first child ( temp float)
+0:26        'o1' ( smooth out float)
+0:26        Constant:
+0:26          18.000000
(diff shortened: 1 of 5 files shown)


jimihem:
@ncesario-lunarg Please review this PR.

jimihem:
@arcady-lunarg  Please help review this PR.

arcady-lunarg (commented):
The idea of this change is sound, I do wonder if this is the best place to do this though. Why not put it in `fold` directly?

arcady-lunarg:
Actually, have you seen the discussion in #925? From reading it, it sounds like matrixCompMult is not supposed to be constant-folded.

jimihem:
@arcady-lunarg glslcompiler in mesa support matrixCompMult constant fold.
I have made some modifications based on your feedback.

arcady-lunarg (approved):
Thanks, I think this version of the code looks better. I'm going to merge this for now, I don't think anyone would complain about this not technically complying with the spec.

arcady-lunarg:
I've rebased your branch, as soon as it clears CI I will do a squash and merge.

Issue #2717, opened 2021-08-01
glslangValidator formatting of warnings/errors not clickable from IDE


alecazam:
I tried switching our codebase over to use glslangValidator from shaderc, and that broke all of our warning/error clickthrough from IDE's such as Xcode.  I then noticed that shaderc reformats the warning/errors so that they are IDE friendly.  Fixing this would make working on large shader codebases much more pleasant.

greg-lunarg:
Could you give an example of a good shaderc message and its unfriendly glslang counterpart?

alecazam:
This is how shaderc reformats the errors that it gets back.   These are clickable when using absolute paths.  So now I have Xcode run our make file, that calls shaderc, and shaderc dumps these out to the Xcode IDE.  Then I can clickthrough to errors and warnings.   We're only using shaderc now, so I don't have a glslangValidator example.  

/Users/me/folder/Shaders/Sand.frag:8: error: '' :  syntax error, unexpected IDENTIFIER
1 error generated.

And this is a warning.  I think Xcode is producing the line/location off the warning
/Users/me/folder/Shaders/Intermediate/Sand.macos.vs.metal:130:7: warning: unused function 'Foo' [-Wunused-function]
float Foo(thread const float& v, thread const float& a, thread const float& b)

arcady-lunarg:
I think there should be enough between #3467 and #3614 that you can pass the right options to get your clickable errors with the full path and the column number easily enough with the right options.

PR #3466, opened 2023-12-28
Added flag to output absolute paths for errors & warning

Sharlock93:
Uses std::filesystem to generate an absolute path when outputting errors and warnings.
I needed this myself for easy jumps in vim.
This was also requested in this issue https://github.com/KhronosGroup/glslang/issues/2717

Please let me know if this is a "viable" solution or if I'm missing something in terms of contribution etiquette. It seemed the most straight forward way to add this given the C++17 requirement.


arcady-lunarg:
Yes, using `std::filesystem` is fine, indeed there is already some precedent for it in spirv-remap.

Sharlock93:
Sorry I closed the PR because the commit had a different email compared to my github one and it triggered a bunch of stuff, I'm preparing a different PR with correct github user.
Thanks for the confirmation.

PR #3467, opened 2023-12-28
Adding flags for outputting absolute paths for messages

Sharlock93:
Uses std::filesystem to create absolute paths when outputting messages.
I needed this myself for easy jumps in vim/visual studio error lists.
This was also requested in this issue https://github.com/KhronosGroup/glslang/issues/2717

Please let me know if I'm missing anything major or if I'm missing something in terms of contribution etiquette.


arcady-lunarg (changes requested):
You will need to add this new flag to the usage message. Also, when I test this, it does something like
```
~/glslang$ build/StandAlone/glslang --absolute-path Test/cppBad2.vert 
Test/cppBad2.vert
ERROR: /home/arcady/glslang/0:3: 'macro expansion' : End of input in macro b
ERROR: /home/arcady/glslang/0:3: '' : compilation terminated 
```
Which does not seem like the desired result since the filename is missing.

Sharlock93:
> You will need to add this new flag to the usage message. Also, when I test this, it does something like
> 
> ```
> ~/glslang$ build/StandAlone/glslang --absolute-path Test/cppBad2.vert 
> Test/cppBad2.vert
> ERROR: /home/arcady/glslang/0:3: 'macro expansion' : End of input in macro b
> ERROR: /home/arcady/glslang/0:3: '' : compilation terminated 
> ERROR: 2 compilation errors.  No code generated.
> ```
> 
> Which does not seem like the desired result since the filename is missing.

Yeah that definitely was something I didn't fully consider when the filename is missing. 
Latest commit should address that, I noticed that the file name was never sort of "lowered" to the compiler only the contents were (at least from just stepping into the code) so I had to store the file name somewhere, I'm not sure if that is the best place for it so any feedback on it would be appreciated.

I can also imagine some optimization around having a big shader with a lot of errors and caching the absolute path once its been outputted the first time.

arcady-lunarg (approved):
LGTM, thanks

PR #3489, opened 2024-01-25
Fix 'maximally_reconverges' token to match the GLSL spec

jeffbolznv:
See discussion at https://github.com/KhronosGroup/GLSL/pull/227#discussion_r1467007040.


arcady-lunarg (approved):
LGTM

PR #3490, opened 2024-01-26
SPV: Complete OpAssumeTrueKHR, OpExpectKHR

younggwan-arm:


arcady-lunarg (approved):
This is fine, but I am wondering why you need it given that glslang doesn't emit these opcodes. Are you using glslang's built-in disassembly after emitting them via intrinsics?

janharaldfredriksen-arm:
>Are you using glslang's built-in disassembly after emitting them via intrinsics? 

Exactly.
The CTS for this (VK_KHR_shader_expect_assume) was written with intrinsics. That worked fine, except we needed the changes here.

We are working on an actual GLSL extension (and glslang support) for this as well, but it will take a little bit longer for that PR to land upstream.

arcady-lunarg:
It would be nice if the doc.cpp stuff could be auto-generated because the current process of updating it manually wasn't really designed with intrinsics-based extensions in mind, but that's a much bigger change.

Issue #3403, opened 2023-11-18
Add Declaration Nodes to AST for Better Debug Info


qingyuanzNV:
Glslang is having trouble to generate good debug information for variables because we don't have AST representation for variable definitions. The variables are generated at the site of its first use, where some of the original declaration information is already lost.

In https://github.com/KhronosGroup/glslang/pull/3401, I attempted to correct the global variable line information by adding a TSourceLoc to track the original definition location in TIntermSource. But this approach doesn't fix a broader issue that some variables may be stripped away because of constant folding or unused symbol.

We could probably do better with an **optional toggle** to allow glslang to build AST nodes for variable declaration. A proof-of-concept implementation is at [here](https://github.com/KhronosGroup/glslang/compare/main...qingyuanzNV:glslang:add_ast_declnode). This should be optional to 1) don't slow down standard library compilation 2) don't break existing programs that work on glslang AST

Given an example shader program:
```
out float outz;
```
and
```glsl
#version 450
#extension GL_GOOGLE_include_directive : require
#include "test.h"

// CHECK: COMMENT ONE!
const float one = 1;
in float inx, iny;
in float inz;

float bar(int x) {
    return x + one;
}

float foo(float x, const float y) {
    return x + y;
}

void main() {
    const int localOne = 1;
    outz = foo(inx, iny) + bar(localOne);
}
```

We'll create an AST
```
Shader version: 450
Requested GL_GOOGLE_cpp_style_line_directive
Requested GL_GOOGLE_include_directive
gl_FragCoord origin is upper left
0:? Sequence
0:2  Sequence
0:?     Decl: outz
0:6  Sequence
0:?     Decl: one
0:7  Sequence
0:?     Decl: inx
0:?     Decl: iny
0:8  Sequence
0:?     Decl: inz
0:10  Function Definition: bar(i1; ( global highp float)
0:10    Function Parameters:
0:10      'x' ( in highp int)
0:11    Sequence
0:11      Branch: Return with expression
0:11        add ( temp highp float)
0:11          Convert int to float ( temp highp float)
0:11            'x' ( in highp int)
0:11          Constant:
0:11            1.000000
0:14  Function Definition: foo(f1;f1; ( global highp float)
0:14    Function Parameters:
0:14      'x' ( in highp float)
0:14      'y' ( const (read only) highp float)
0:15    Sequence
0:15      Branch: Return with expression
0:15        add ( temp highp float)
0:15          'x' ( in highp float)
[...]

PR #3496, opened 2024-02-02
Clean up line tracking logic for debug info

qingyuanzNV:
This patch cleans up the line tracking logic when debug info is turned on, including:
1. Add correct file tracking. We used to have `sourceFileStringId` and `currentFileId` in the `SpvBuilder`, but both were always OpString id of the main file. Renamed them to `mainFileId` and `currentFileId` and implemented file tracking accordingly.
2. Make insertion of OpLine/OpDebugLine/OpDebugScope lazy. We used to insert OpLine right after setting the current debug source location, sometimes resulting in multiple cascading OpLines without any meaningful instructions in between. By making the process lazy, an OpLine is only inserted when an instruction is added with a dirty line tracker.
3. Added per-block line/scope tracker. So whenever an OpLine/OpDebugLine/OpScope is inserted, the state is stored in the `Block` instead of `SpvBuilder`. This is how the standard defines the lifetime of such instruction.

Also, some minor changes are:
1. Skip setting debug source location for linking object at the end of the AST
2. Rename some functions to better reflect that they are debug info related
3. No longer generate an OpLine preceding an OpFunction when Non-Semantic Debug is requested. The line information is available via the extension already.

~~As a result of this patch, we should have non-semantic debug reporting the correct file for variables, functions and .etc.~~
Hopefully, after this patch, we could have more accurate information to fix the file reporting for variables, functions and .etc for non-semantic debug extension. Also, this helps avoid a lot of cascading OpLines for a potential fix for #3403


qingyuanzNV:
@arcady-lunarg We are trying to catch an internal deadline and put this patch in. Could you help review this PR? Thanks!

arcady-lunarg:
Yes, I am going to review it today, it also fixes some issues for us. I was
at Vulkanised the past few days and didn't get a chance to look at it in
enough detail.

SPIRV/SpvBuilder.h
```diff
@@ -409,11 +417,17 @@ class Builder {
     // Also reset current last DebugScope and current source line to unknown
     void setBuildPoint(Block* bp) {
         buildPoint = bp;
-        lastDebugScopeId = NoResult;
-        currentLine = 0;
+        // TODO: Technically, change of build point should set line tracker dirty. But we'll have bad line info for
+        //       branch instructions. Commenting this for now because at least this matches the old behavior.
+        // dirtyLineTracker = true;
```
arcady-lunarg, line 422: Do you have a pending change to clean this up? I don't really like the idea of checking in TODOs or commented code. Actually, you should just delete the commented out line, but keep the TODO comment for now.

arcady-lunarg (changes requested):
Just a couple of nits, but overall I think this looks good, and the results seem fine. Also your PR description is great, and you should use some version of it as a commit message (and if not, I'll probably do "squash and merge" and do it anyway).

SPIRV/spvIR.h
```diff
@@ -292,6 +320,12 @@ class Block {
     std::vector<std::unique_ptr<Instruction> > localVariables;
     Function& parent;
 
+    // Track source location of the last source location marker instruction.
+    std::optional<DebugSourceLocation> currentSourceLoc;
+
+    // Track scope of the last debug scope instruction.
+    std::optional<spv::Id> currentDebugScope;
```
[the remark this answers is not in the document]
qingyuanzNV: Added assert.

Technically, there's an OpDebugNoScope to clear the current scope. But as we never generate the op and I don't see a use case of the op in glslang, I didn't add a way to clear the tracker to keep it simple.

SPIRV/SpvBuilder.h
qingyuanzNV: Done

qingyuanzNV:
Could you help squash and merge with the commit message below (so it's easier for you to review the new changes):
```
Clean up the debug line info tracking and generation. 
- Correctly populate the field `currentFileId` with the presence of include directive
- Support lazy OpLine/OpDebugLine generation only when a real instruction is added instead of a debug location is set
- Improve the debug location tracking to per-block instead of just per-builder
- A few bug fixes related to debug source info
```

qingyuanzNV:
By the way, one important change of this PR is that I added an `addInstruction(...)` to `SpvBuilder` to hijack all instructions added to the current build point. You guys might need be alert on this when reviewing new changes to make sure that all new instructions added to the current build point should go through that function instead of directly calling `buildPoint.addInstruction(...)`.

arcady-lunarg (approved):
LGTM, thanks, this is a huge improvement.

PR #3502, opened 2024-02-10
Emit debug info for accelerationStructure and rayQuery variables.

sajjadmirzanv:
A minor quirk is that multiple source types will get mapped to the same debug type, for example both accelerationStructureNV and accelerationStructureEXT are just named accelerationStructure. But I expect a shader author won't use both extensions in the same program, they'll just pick one, so this should be OK.


arcady-lunarg:
In the underlying SPIR-V `OpTypeAccelerationStructureNV` and `OpTypeAccelerationStructureKHR` have the same opcode, so it's kind of hard to avoid conflating them into just one type.

arcady-lunarg (changes requested):
Could you add a test for this? The change itself looks fine and isn't doing a lot, just adding an empty struct for the acceleration structure so the test should be fairly trivial to write.

sajjadmirzanv:
Added a test case (copy of an existing one, rayQuery.rgen)

arcady-lunarg (approved):
LGTM, thanks.

Issue #3507, opened 2024-02-12
cmake: allow enabling of tests and install independently of whether the project is top level


dneto0:
Glslang tests and installs only if Glslang is the top-level project directory.
See https://github.com/KhronosGroup/glslang/blob/47a02a987c4e6dc573dfca1b80e5e3eacf4e2ed8/CMakeLists.txt#L307 

This makes the project non-composable.  It's also a regression from before where enclosing projects like Shaderc want to test Glslang's tests.

Ultimately, the disabling of tests caused us to miss the following problem
https://github.com/KhronosGroup/glslang/issues/3505 

(It seems to me it's an anti-pattern to overuse PROJECT_IS_TOP_LEVEL?)

cc: @juan-lunarg @arcady-lunarg

AnyOldName3:
Looks like you posted this while I was writing up https://github.com/KhronosGroup/glslang/issues/3509, which covers this and a couple of other, smaller problems.

PR #3508, opened 2024-02-12
Allow external control of whether Glslang will be tested or installed

dneto0:
Expose GLSLANG_TESTS and GLSLANG_ENABLE_INSTALL as options that can be controlled from an enclosing project, or from the command line.

They retain the prior default behaviour. In particular, if Glslang is not the top level project, then they default to OFF.

Fixes: #3507


dneto0:
This might need work to cope with SHARED vs. STATIC variations. That is behind some of the bot failures, I think

arcady-lunarg (approved):
I think this is a reasonable compromise, preserve the current defaults but allow for installation when someone using glslang as a subproject wants it. LGTM.

PR #3513, opened 2024-02-13
Fill correct file name to OpDebugFunction and add line info for parameters' OpDebugDeclare

qingyuanzNV:
This patch fixes:
1. OpDebugFunction's file entry should be the correct file instead of always being main file
2. OpDebugDeclare and OpDebugValue of the function parameters should be now attributed with an OpDebugLine pointing to the function definition line. (They had no associated line info)

For example, say we have:
```
// test.h
void foo(int x, int y) { }

// main.frag
#include "test.h"

void main() {
    foo(1, 2);
}
```
The existing implementation would point `foo` to "main.frag" instead of "test.h" in DebugFunction instruction. Also, it doesn't generate DebugLine instruction for DebugDeclare of parameter `x` and `y`, leaving them with no line info associated. This patch should fix both.


qingyuanzNV:
It turns out it's a non-trivial effort to add a test for this. I have to change the test infrastructure to pass in a customized includer and optionally pass in the shader name for non-semantic debug.

Btw, I noticed that the source text for main file is wrong for the test result. I didn't fix it because it looks like an old issue, not a regression by this PR.

qingyuanzNV:
Change to draft. The test infra part is increasingly complicated, and I think it's worth a dedicated PR. I'd like to extract the test infra part and commit that first.

arcady-lunarg (approved):
Looks good, thanks for the change.

PR #3518, opened 2024-02-14
Remove implicit fallthrough

AnyOldName3:
This is intended so that downstream projects consuming glslang with FetchContent or similar means can use `-Wimplicit-fallthrough` without getting warning spam.

I've used my best judgement to determine whether the implicit fallthrough was desired, or was simply unreachable code. `std::unreachable` is unavailable until C++23, but I saw places where `default: assert(0);` was used, so copied that. There were a few places where some code might actually have been reachable and intended to return a value that represented an error, so someone should double check that kind of thing.

I've run the tests, and get a single failure that also fails on main, so it doesn't look like I broke anything, but I don't know how exhaustive the tests are.

This doesn't enable `-Wimplicit-fallthrough`, but that might be desirable, and can be done in another commit.


jeremy-lunarg (approved):
LGTM

arcady-lunarg (approved):

PR #3519, opened 2024-02-14
NV_shader_atomic_fp16_vector

jeffbolznv:
This implements the old OpenGL/GLSL extension https://registry.khronos.org/OpenGL/extensions/NV/NV_shader_atomic_fp16_vector.txt using the new SPIR-V extension https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/NV/SPV_NV_shader_atomic_fp16_vector.html


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2524,6 +2524,17 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
                 error(loc, "only supported on image with format r64i", fnCandidate.getName().c_str(), "");
             else if (callNode.getType().getBasicType() == EbtUint64 && imageType.getQualifier().getFormat() != ElfR64ui)
                 error(loc, "only supported on image with format r64ui", fnCandidate.getName().c_str(), "");
+        } else if(callNode.getType().getBasicType() == EbtFloat16 && 
+                ((callNode.getType().getVectorSize() == 2 && arg0->getType().getQualifier().getFormat() == ElfRg16f) ||
+                  (callNode.getType().getVectorSize() == 4 && arg0->getType().getQualifier().getFormat() == ElfRgba16f))) {
+            if ((fnCandidate.getName().compare(0, 14, "imageAtomicAdd") == 0) ||
+                 (fnCandidate.getName().compare(0, 19, "imageAtomicExchange") == 0) ||
+                 (fnCandidate.getName().compare(0, 19, "imageAtomicMin") == 0) ||
```
arcady-lunarg, line 2532: Should this be 14 rather than 19 here? Unfortunately, `std::basic_string::starts_with` was only added in c++20, but maybe a helper would be good to avoid needing to count characters.

SPIRV/GlslangToSpv.cpp
```diff
@@ -204,7 +204,8 @@ class TGlslangToSpvTraverser : public glslang::TIntermTraverser {
     spv::Id createBinaryMatrixOperation(spv::Op, OpDecorations&, spv::Id typeId, spv::Id left, spv::Id right);
     spv::Id createUnaryOperation(glslang::TOperator op, OpDecorations&, spv::Id typeId, spv::Id operand,
                                  glslang::TBasicType typeProxy,
-                                 const spv::Builder::AccessChain::CoherentFlags &lvalueCoherentFlags);
+                                 const spv::Builder::AccessChain::CoherentFlags &lvalueCoherentFlags,
+                                 const glslang::TType &opType);
```
arcady-lunarg, line 208: It seems kind of odd to be passing both the TType and TBasicType here but I kind of see why it's necessary, so I think it's fine to leave this as is.

arcady-lunarg (changes requested):
One minor nit, otherwise this looks fine. What a curious hardware capability to have.

glslang/MachineIndependent/ParseHelper.cpp
jeffbolznv: Done, thanks.

arcady-lunarg (approved):
LGTM, thanks.

AnyOldName3:
First things first, I don't know if this is how things are supposed to work, so I apologise for the noise if I'm unnecessarily raising red flags.

After this PR was merged, the new test fails unless SPIRV-Tools and SPIRV-Headers are updated to the new commit in `known_good.json`. That's fine on its own (although it briefly made me think I'd broken something on my end), but I imagine that if `ALLOW_EXTERNAL_SPIRV_TOOLS` is set and the system-installed library is used, it will probably be an older commit and the test will fail there, too. I just wanted to make sure no one had forgotten to add some conditional compilation if a feature check failed.

If instead it's expected that some tests fail with commits other than the one in `known_good.json`, it might be a good idea to document that, e.g. in the testing part of the readme and the docstring for the `ALLOW_EXTERNAL_SPIRV_TOOLS` CMake option.

arcady-lunarg:
The only supported configuration is with the versions of headers/tools in `known_good.json`. If you're using your own version of SPIRV-Tools, you're on your own. That's why we didn't even have the `ALLOW_EXTERNAL_SPIRV_TOOLS`. @AnyOldName3 if you think the documentation around this option could be improved, patches are certainly welcome.

PR #3520, opened 2024-02-15
Quote paths in runtest

AnyOldName3:
This means it won't die if the build path contains spaces.

I've got the build directory in the VS Code CMake extension set to `${workspaceFolder}/build/${buildKit}` so I can have MSVC and MSYS2 (MinGW UCRT) build systems without having to clone the repo twice and sync my changes back and forth for testing. `${buildKit}` contains spaces, and there's apparently nothing I can do to stop that, so it's preferable that the tests can handle the build directory having spaces.


arcady-lunarg (approved):
LGTM, thanks.

PR #3522, opened 2024-02-15
-Wmissing-field-initializers fixes

AnyOldName3:
Like with the last few PRs from me, it fixes a warning that's commonly enabled in downstream projects that might want to consume glslang via FetchContent or equivalent.

It doesn't actually enable the warning, but that might be desirable.

I think I found a bug in the Spv.FromFile.cpp tests that would have been prevented had this warning been enabled all along. I had to guess the value for the missing field, so went for the most common one in the list. The only test case that used a different value before had an RWTexture1D, so if baseImageBinding is meaning the same kind of image as image load store in OpenGL, it would make sense that that would be the only one to need a non-zero value for the binding.

I'm a little concerned that the test wasn't previously failing with the incorrectly-assigned fields as it implies they don't make any difference, so the test might be too permissive.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -85,7 +85,7 @@ public:
             addGlobalReference(base->getAccessName());
 
         if (target) {
-            TVarEntryInfo ent = {base->getId(), base, ! traverseAll};
+            TVarEntryInfo ent = {base->getId(), base, ! traverseAll, {}, {}, {}, {}, {}, {}, {}};
             ent.stage = intermediate.getStage();
             TVarLiveMap::iterator at = target->find(
                 ent.symbol->getAccessName()); // std::lower_bound(target->begin(), target->end(), ent, TVarEntryInfo::TOrderById());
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/iomapper.cpp
```diff
@@ -85,7 +85,7 @@ class TVarGatherTraverser : public TLiveTraverser {
             addGlobalReference(base->getAccessName());
 
         if (target) {
-            TVarEntryInfo ent = {base->getId(), base, ! traverseAll};
+            TVarEntryInfo ent = {base->getId(), base, ! traverseAll, {}, {}, {}, {}, {}, {}, {}};
```
arcady-lunarg, line 88: This is so ugly but I don't really see any better option since designated initializers are only supported in C++20. I guess once we get to that this can be refactores.

arcady-lunarg (approved):
This LGTM, you can add the warning flags to the CMakeLists once all the warning fix PRs are in.

glslang/MachineIndependent/iomapper.cpp
AnyOldName3: Yeah, I had exactly the same thoughts.

PR #3525, opened 2024-02-19
Fix implicit-fallthrough warnings and enable the warning flags in CMakelists.txt

arcady-lunarg:
This follows up on work done by @AnyOldName3 in #3518, clang is a bit pickier than gcc when it comes to fallthrough comments.


jeremy-lunarg (approved):
LGTM

PR #3528, opened 2024-02-21
Fix HLSL built-in passthrough via inout

Svenny:
Without the change this one-liner gets  `pos` output to a varying, not the position built-in.

```hlsl
void main(inout float4 pos : SV_Position, inout float2 uv : TEXCOORD0) {}
```

I'm not sure if this is the right way to fix it (I even don't really understand the meaning of `declaredBuiltIn` field)


arcady-lunarg:
Given that this doesn't break anything, I think this is fine. `declaredBuiltin` seems to be intended for this sort of use.

arcady-lunarg (approved):

PR #3531, opened 2024-02-27
GL_EXT_control_flow_attributes2 support.

wooyoungqcom:
The actual support has been available with
GL_EXT_control_flow_attributes.
This change set is to handle

  "#extension GL_EXT_control_flow_attributes2 : <val>"


arcady-lunarg (changes requested):
You also need to add `#define GL_EXT_control_flow_attributes2 1` to the preamble. But also, it seems kind of wrong to implement this extension as a no-op. What is the expected behavior if only the control_flow_attributes2 is enabled?

wooyoungqcom:
Thanks. I thought about adding it, then decided not to.  Anyways, it is added.

As for the second question, if I compile the test  spv.1.4.LoopControl.frag with no extension enabled, I get

ERROR: spv.1.4.LoopControl.frag:10: 'attribute' : required extension not requested: GL_EXT_control_flow_attributes



I am afraid adding something specific for GL_EXT_control_flow_attributes2 would break shaders out there that may have used the attributes, though I am not sure if there are any that are using those attributes.

wooyoungqcom:
updated the grammar to allow shaders that enable GL_EXT_control_flow_attributes2 only.

arcady-lunarg (changes requested):
This looks good enough for now, aside from two issues:
1. You need to regenerate the grammar using version 3.8.2 of Bison, which is the same version that was used to generate the current grammar.
2. You need to update the expected tests results, just run `build/gtests/glslangtests --update-mode` and check in the resulting change to `Test/baseResults/460.frag.out`.

wooyoungqcom:
Generated the grammar using bison 3.8.2.

For the test results, there is an error while running the test suite, and I don't see 460.frag got executed.
The same error occurred with the main branch without any of my changes.

arcady-lunarg:
I am not seeing this failure, neither is the CI. That's really odd, are you sure you're doing a clean build from latest source and not picking up any changes? Also, if you make your branch editable by maintainers I can just push the updated test as a separate commit on top of yours.

wooyoungqcom:
I did the clean build with the latest source. Let me try it again. One difference I have from the main branch is I had to add
"stdc++fs" 

target_link_libraries(spirv-remap SPVRemapper ${LIBRARIES} -lstdc++fs)

Without it, the build fails on my Linux box.

Let me also make the branch editable by maintainers.

wooyoungqcom:
Clean  re-build from the scratch resolved the issue.

I updated Test/baseResults/460.frag.out

PR #3539, opened 2024-03-05
Support for SPV_QCOM_image_processing2

wooyoungqcom:


known_good.json
arcady-lunarg, line 1: This looks like a whitespace-only change, could you revert it?

arcady-lunarg (changes requested):
The code itself looks fine, but your tests aren't being run, you need to add them to the appropriate list of tests in `gtests/Spv.FromFile.cpp` and generate base results that will end up in the `Test/baseResults` directory by running `build/gtests/glslangtests --update-mode` or similar after you have added the tests to the test list and rebuilt.
Also please revert the whitespace change to `known_good.json`.

wooyoungqcom:
Thank you so much for the review.
I hope I made all the changes you mentioned.

arcady-lunarg (approved):
LGTM, thanks.

PR #3540, opened 2024-03-06
Increase TStorageQualifier bitfield width

### LINKED ISSUE #3538: Compile warning on VS2022
I am getting this compiler warning on VS2022 about not enough bits in a bitfield interfacing with glslang b0df68c490363ac347e454f79886293f661ce4f6 from our C++ code:

```
1>H:\dev\source2\main\src\thirdparty\glslang\glslang\Include\Types.h(576,25): error C2220: the following warning is treated as an error
1>(compiling source file 'vulkan/renderdevicevulkan.cpp')
1>H:\dev\source2\main\src\thirdparty\glslang\glslang\Include\Types.h(576,25): warning C5249: 'glslang::TQualifier::storage' of type 'glslang::TStorageQualifier' has named enumerators with values that cannot be represented in the given bit field width of '6'.
1>(compiling source file 'vulkan/renderdevicevulkan.cpp')
1>H:\dev\source2\main\src\thirdparty\glslang\glslang\Include\Types.h(576,25):
1>to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
1>H:\dev\source2\main\src\thirdparty\glslang\glslang\Include\BaseTypes.h(137,5):
1>see enumerator 'glslang::EvqLast' with value '32'
```

jeremy-lunarg:
Fix #3538.

Visual Studio 2022 added a new warning to detect when enumerators can not be represented within the given bit field width. This warning is disabled by default but can be enabled using the flag /w15249. This PR increases the width by one to accommodate the signed maximum value of EvqLast, which is currently 32.

Perhaps a future improvement would be to use scoped enums; however, that is more work as the typing is more strict and would require more changes.

### DIFF
--- a/glslang/Include/Types.h
+++ b/glslang/Include/Types.h
@@ -573,7 +573,8 @@ public:
     }
 
     const char*         semanticName;
-    TStorageQualifier   storage   : 6;
+    TStorageQualifier   storage   : 7;
+    static_assert(EvqLast < 64, "need to increase size of TStorageQualifier bitfields!");
     TBuiltInVariable    builtIn   : 9;
     TBuiltInVariable    declaredBuiltIn : 9;
     static_assert(EbvLast < 256, "need to increase size of TBuiltInVariable bitfields!");


glslang/Include/Types.h
```diff
@@ -573,7 +573,8 @@ class TQualifier {
     }
 
     const char*         semanticName;
-    TStorageQualifier   storage   : 6;
+    TStorageQualifier   storage   : 7;
```
ncesario-lunarg, line 576: Would `enum TStorageQualifier : uint8_t {` work here and avoid the bitfield resize (though I hopefully nobody is relying on this bitfiield size...)?

If the size of `TQualifier` doesn't need to remain constant, could some (maybe all?) of these bitfields be replaced with fixed width integer types?

jeremy-lunarg: > Would `enum TStorageQualifier : uint8_t {` work here and avoid the bitfield resize (though I hopefully nobody is relying on this bitfiield size...)?

Yes, I mention that in the commit message. I did it this way because there was already a precedent for this change and using uint8_t would have required additional changes.

> If the size of `TQualifier` doesn't need to remain constant, could some (maybe all?) of these bitfields be replaced with fixed width integer types?

Yes, I think that would be ideal.

arcady-lunarg: They're crammed into bitfields presumably out of concern for memory consumption, we would have to check what the impact of that would be on particularly large shaders (though I think it might not be a huge impact in the end) but that's a separate project from this and we should just get this done first.

arcady-lunarg (approved):

Issue #425, opened 2016-08-01
Files with a UTF-8 BOM marker confuse the #version check


DeanoC:
The #version check looks at the first byte of a file, if the file has a UTF-8 BOM marker this fails as # is not the first byte (it is the first character though...).

Whilst it may be that it shouldn't handle BOM files, would be nice to fire a warning at least as its largely a hidden error (most editors will hide the BOM so its correct to visual inspection)

mknejp:
Seems like looking at the first byte of the file is incorrect anyways according to the spec:

> The #version directive must occur in a shader before anything else, except for comments and white space

johnkslang:
> The #version check looks at the first byte

I don't think that is true:

 - The #version finder skips initial characters to find the first #version, it does not require # as the first byte.
 - The #version semantics checking enforces ES vs. desktop rules, which differ, but neither requires # in the first byte.

> if the file has a UTF-8 BOM marker this fails 

Isn't it correct to fail?

Note there is no formal specification of processing "a file", and that is absent in the programmatic (use through API) use of glslang.  It is just a convenience provided by the standalone command-line version that turns a file into a text string. Maybe you want to provide an enhancement to the standalone tool that processes BOM files the way you think is correct?

DeanoC:
You're correct about the whitespace and comment (I just checked the code) but it doesn't skip the BOM markers. The consumeWhitespaceComment returns with the BOM lined up so the next get() returns the first BOM byte. This then fails the if(get() != '#') check, so the version declaration is skipped.

A warning (as its easy to detect) would help users, in the case where I detected it, it took a while to figure out. The text editors loaded the file correctly showing #version correct but glslang seems to get the wrong version. Yes in theory it should have been cleaned off by the loader before passing to glslang, but a warning (i.e. BOM marker detected in shader string) might be nice.

johnkslang:
> it doesn't skip the BOM markers

Correct.  There is nothing in the spec about doing that, and the "inside" of glslang, where it's looking for # should not do that.

If you want a change, I see two choices:

1. change the standalone wrapper to strip these off before handing to through glslang's API

2. change the standalone wrapper to detect and warn

I don't immediately know how to do BOM detection or proper testing of it, so a PR would help.

DeanoC:
Since 4.1 GLSL (section 3.1) is spec'ed to be UTF8, BOM handling is a part of that.

[...]

PR #3543, opened 2024-03-12
Support files with UTF8BOM character

SteveUrquhart:
This PR allows glslangvalidator to consume the byte-order mark (BOM) when it appears at the start of a text stream.  The Unicode spec, Version 9.0, Section 23.8, reads: "Systems that use the byte order mark must recognize when an initial U+FEFF signals the byte order. In those cases, it is not part of the textual content and should be removed before processing,"

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -2167,6 +2167,20 @@ char* ReadFileData(const char* fileName)
 
     fseek(in, 0, SEEK_SET);
 
+    if (count > 3) {
+        unsigned char head[3];
+        if (fread(head, 1, 3, in) == 3) {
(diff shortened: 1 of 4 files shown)


arcady-lunarg:
This would fix #425.

arcady-lunarg (changes requested):
FYI, you should generally only add a file to either the `runtests` script or to the `gtests/*.FromFile.cpp` lists, not both. Also I would strongly recommend testing on your own before creating the PR.

The code change itself looks good though, I think this is the right way to solve it and there's no point trying to argue with Windows putting a UTF-8 BOM in files.

arcady-lunarg (approved):
LGTM, thanks.

PR #3547, opened 2024-03-19
build: fix CI sanitizer failures

arcady-lunarg:
This is needed because newer linux kernels cause problems for the sanitizers.


jeremy-lunarg (approved):
Interesting bug. LGTM. Thanks @arcady-lunarg

PR #3550, opened 2024-03-21
Fix issues of the interaction between cooperative_matrix and spirv_intrinsics

amdrexu:
`coopmat<>` type definition allows type parameters. To make it accepte types defined by `spirv_type` directive, we add spirvType info to the type parameters. This change is to support this case. And a test is added to show the missing usage.


arcady-lunarg (approved):
Looks good, I assume in practice this will be used for other extension-defined types that aren't just half-floats.

amdrexu:
> Looks good, I assume in practice this will be used for other extension-defined types that aren't just half-floats.

Yes, it is used to do some experiments on other future arithmetic types.

PR #3551, opened 2024-03-21
Fix an issue of getExtBuiltins()

amdrexu:
This function is used to import SPIR-V extended instruction set. It mistakenly treated the name of SPIR-V extended instruction set as the name of SPIR-V extension. For example, when we have such code

`getExtBuiltins("NonSemantic.DebugBreak")`

'NonSemantic.DebugBreak' is added to SPIR-V extension. Rather, the SPIR-V extension name is 'SPV_KHR_non_semantics_info'. Therefore, we should avoid this since the name of SPIR-V extended instruction set is not necessarily equal to that of relevant SPIR-V extension. Adding a SPIR-V extension must be done by calling `addExtension()` explicitly.


arcady-lunarg:
Could you add a test for this? Maybe using the DebugBreak instruction, it would be a good demonstration too.

amdrexu:
> Could you add a test for this? Maybe using the DebugBreak instruction, it would be a good demonstration too.

Sure. I will add it.

arcady-lunarg (approved):
LGTM, thanks.

PR #3555, opened 2024-03-25
Out-of-range floats should overflow to infinity

jimihem:
"2147483649" of 1.0e2147483649 is out of range that positive 32bits int can store.

    precision highp float;
    out vec4 my_FragColor;
    void main()
    {
        // Out-of-range floats should overflow to infinity
        // GLSL ES 3.00.6 section 4.1.4 Floats:
        // "If the value of the floating point number is too large (small) to be stored as a single precision value, it is converted to positive (negative) infinity"
        float correct = isinf(1.0e2147483649) ? 1.0 : 0.0;
        my_FragColor = vec4(0.0, correct, 0.0, 1.0);
    }

### DIFF
--- a/Test/baseResults/positive_infinity.frag.out
+++ b/Test/baseResults/positive_infinity.frag.out
@@ -1,0 +1,51 @@
+positive_infinity.frag
+Shader version: 300
+0:? Sequence
+0:4  Function Definition: main( ( global void)
+0:4    Function Parameters: 
+0:9    Sequence
+0:9      Sequence
+0:9        move second child to first child ( temp highp float)
+0:9          'correct' ( temp highp float)
+0:9          Constant:
+0:9            1.000000
+0:10      move second child to first child ( temp highp 4-component vector of float)
+0:10        'my_FragColor' ( out highp 4-component vector of float)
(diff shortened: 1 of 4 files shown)


arcady-lunarg (changes requested):
This is a real issue, and while your change does appear to fix it in the given case, I don't think this fully solves the issue. What you probably want to do is clamp the exponent in the parsing loop to make sure that rather than overflowing it stays above the range where the float will be considered to be an infinity. The test is good though, thanks for that.

jimihem:
PR 3388 has fixed some cases about float overflow. This PR fix a new case. The max positive interger that 32bit int can represent is 2147483647, in this case it's 2147483649.

arcady-lunarg:
Won't your "fix" still break if the exponent is larger than can be represented in 64 bits? It's also not necessary, any exponent with an absolute value above 310 is too big to fit in a double anyway. So IMO it would make more sense to just check the exponent and clamp it to some arbitrary "too big" value which will be handled correctly by the rest of the code.

arcady-lunarg:
I'm going to do my own version of this PR, #3562 but I'll keep the part of the commit that adds the test (with you as the author), thanks for the test.

Issue #3536, opened 2024-03-03
Incorrect SPIR-V generated when using a fixed-size array of sampler2D with -R (relaxed vulkan rules)


slime73:
I'm using glslang's SPIR-V generator together with SPIRV-Cross to cross-compile OpenGL GLSL to MSL. I believe this bug is in glslang rather than SPIRV-Cross but I'm happy to move this report if that's not the case.

I'm using this command to generate SPIR-V: `glslang -V -R --amb --aml test.frag`
`test.frag` has the following code:
```glsl
#version 450 core

out vec4 outColor;

uniform sampler2D mainTex;
uniform sampler2D textures[5];

void main()
{
    outColor = texture(textures[1], vec2(0.5)) + texture(textures[3], vec2(0.5)) + texture(mainTex, vec2(0.5));
}
```

And then I run `spirv-cross --msl --msl-version 20100 frag.spv` to get the final MSL shader. With `-R` (as above), this is the result. it's only declaring and sampling from one of the textures in the original array, and the name of the texture is slightly mangled:
<details>
<summary>bad SPIR-V disassembly</summary>

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 31
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %outColor
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 450
               OpName %main "main"
               OpName %outColor "outColor"
               OpName %textures_1_ "textures[1]"
               OpName %mainTex "mainTex"
               OpName %textures "textures"
               OpDecorate %outColor Location 0
               OpDecorate %mainTex DescriptorSet 0
               OpDecorate %mainTex Binding 0
               OpDecorate %textures DescriptorSet 0
               OpDecorate %textures Binding 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
    %v4float = OpTypeVector %float 4
%_ptr_Output_v4float = OpTypePointer Output %v4float
   %outColor = OpVariable %_ptr_Output_v4float Output
         %10 = OpTypeImage %float 2D 0 0 0 1 Unknown
         %11 = OpTypeSampledImage %10
%_ptr_UniformConstant_11 = OpTypePointer UniformConstant %11
[...]

Issue #3526, opened 2024-02-20
Vulkan Relaxed Outputs Extra Sampler Variables


bjornbytes:
I updated glslang and noticed that in vulkan-relaxed mode it started outputting extra combined image sampler variables without set/binding decorations.  Here's a simple example that demonstrates the issue:

```glsl
#version 460

layout(set = 0, binding = 0) uniform texture2D tex;
layout(set = 0, binding = 1) uniform sampler samp;

layout(location = 0) out vec4 pixel;

void main() {
  pixel = texture(sampler2D(tex, samp), vec2(0.));
}
```

Compiling this with `glslangValidator -V -R shader.frag` and then validating with `spirv-val --target-env vulkan1.1 frag.spv` outputs the following:

```
error: line 26: UniformConstant id '13' is missing DescriptorSet decoration.
From Vulkan spec, section 14.5.2:
These variables must have DescriptorSet and Binding decorations specified
  %texsamp = OpVariable %_ptr_UniformConstant_11 UniformConstant
```

I bisected and landed on c59b876ca0f5b672d7cfeb4d591b346e97b1966c as the commit that introduced the issue.

Before the change (or when removing `-R`), the working shader looked like this:

```spirv
       %main = OpFunction %void None %3
          %5 = OpLabel
         %13 = OpLoad %10 %tex
         %17 = OpLoad %14 %samp
         %19 = OpSampledImage %18 %13 %17
         %23 = OpImageSampleImplicitLod %v4float %19 %22
               OpStore %pixel %23
               OpReturn
               OpFunctionEnd
```

After the change, the combined image sampler was hoisted to top-level scope:

```spirv
         %10 = OpTypeImage %float 2D 0 0 0 1 Unknown
         %11 = OpTypeSampledImage %10
%_ptr_UniformConstant_11 = OpTypePointer UniformConstant %11
    %texsamp = OpVariable %_ptr_UniformConstant_11 UniformConstant
...
       %main = OpFunction %void None %3
          %5 = OpLabel
         %14 = OpLoad %11 %texsamp
         %18 = OpImageSampleImplicitLod %v4float %14 %17
               OpStore %pixel %18
               OpReturn
               OpFunctionEnd
```

I don't really understand c59b876ca0f5b672d7cfeb4d591b346e97b1966c yet, but it looks like it might be overeager about converting samplers?  There aren't any samplers in structs or function parameters here, so I wouldn't expect any change to the spirv, compared to non-relaxed mode.

[...]

PR #3558, opened 2024-03-27
Branch out of relaxed rules on opaque arguments declared at top-level

sbourasse:
This change is introduced in an attempt to fix issue #3536.

Even though top-level opaque type variables are semantically valid under Vulkan semantics, previous versions would mistakenly move them around, resulting in unnecessary complex binaries.


bjornbytes:
I can confirm that this fixes #3526!

arcady-lunarg (approved):
LGTM

Issue #3544, opened 2024-03-13
error in combination with preprocessor directives can obfuscate error location.


phr34k:
I'd like to address an issue with error handeling specifically in combination with common preprocessor directives like #include #line. I'll give you a short summary of the problem and how these system interact, and the situation of the problem. 

```
layout( location = 0 ) out vec4 color1
#line 500
layout( location = 0 ) out vec4 color2:
```

Results in an "ERROR: frag.glsl:500: unexpected LAYOUT, expecting COMMA or SEMICOLON.". Strictly spoken the error message is correct. However it highlight an important issue, preprocessor directives can cause unterminated statements easily transfer accross natural file boundaries (e.g. #include directives) which makes these source of these statements near impossible to locate, especially in larger code bases where reusabiliity is imminent this class of errors are easilly obfuscated.

```
layout( location = 0 ) out vec4 color1
```

```
layout( location = 0 ) out vec4 color2:
```

```
#include "a"
#include "b"
```

In the above scenario, the origional statement occurs in "a" but the factual error is only detected in "b".  I'd propose that unterminated statements could report the location where the statement begins, and the specific token or (sub-)statement location reported as detail(s).

I have some past expierence in compilers and language(s) generation but haven't used yacc before so I'd be happy just get a few pointers.  I believe this error is presently automatically generated from the grammer (glslang.y) and explictly generated/handled inline so I suspect the location (simple_statement) needs to propegate so that this can be used during error handling procedures.

arcady-lunarg:
Yeah, I think ultimately the way to deal with this is to keep track of the start of the unterminated statement. Part of the issue is that `#include` processing happens in the tokenizer, at a level below that of the `glslang.y` bison parser. Error reporting is definitely an area where glslang could use considerable improvement, patches are definitely welcome.

phr34k:
@arcady-lunarg good to hear you concur with my line of thought. Do you have any highlevel pointers in terms of 'glslang.y', it's hard to understand thirdparty code at face value, so I'm wondering what generally would.

What I had in mind:

- I assume that you that this error statements in particulair are auto generated, so I'd have to patch up the catch all error handler.
- I assume that some kind of global context is available from bison, that is also available during the error handler(s) 
- I'm thinking some kind of inline code to identify when statements are opened and terminated (state-machine)
- The part where the auto generated errors are handled then simply use the additional (context-)information to report altered locations.

[...]

PR #3559, opened 2024-03-27
Report statement errors

phr34k:
Initial changes to improve error reporting in combination with preprocessor directives. The intent of this changeset is the modify glslang in a way  it reports location where the statement origionally begin, so that any syntax errors can report that location primarilly and/or report the type of statement.

All unit tests passed locally

This changeset modifies the glslang in the following ways:

- modified the standalone glslang  to include --report-statements feature toggle to enable functionallity optionally
- modified the function signature of parseHelper::parserError to add in the location
- modified the grammar to use bison custom syntax error reporter
- modified the parserContext to keep track of some extra meta data w.r.t. statement and scope boundaries
- modified the grammar to track the statement/scope boundaries either trough bison "subroutines" or mid-action rules

Example 

```
#line 200 
layout( .... ) fragColor1
#line 500 
layout( .... ) fragColor2:
```

See https://github.com/KhronosGroup/glslang/issues/3544

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -110,6 +110,7 @@ enum TOptions : uint64_t {
     EOptionInvertY = (1ull << 30),
     EOptionDumpBareVersion = (1ull << 31),
     EOptionCompileOnly = (1ull << 32),
+    EOptionReportStatements = (1ull << 33),
 };
 bool targetHlslFunctionality1 = false;
 bool SpvToolsDisassembler = false;
@@ -898,6 +899,8 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                         Options |= EOptionDumpVersions;
                     } else if (lowerword == "no-link") {
                         Options |= EOptionCompileOnly;
+                    } else if (lowerword == "report-statements") {
+                        Options |= EOptionReportStatements;
                     } else if (lowerword == "help") {
                         usage();
                         break;
@@ -1132,6 +1135,8 @@ void SetMessageOptions(EShMessages& messages)
 {
     if (Options & EOptionRelaxedErrors)
         messages = (EShMessages)(messages | EShMsgRelaxedErrors);
+    if (Options & EOptionReportStatements)
+        messages = (EShMessages)(messages | EShMsgReportStatements);
     if (Options & EOptionIntermediate)
         messages = (EShMessages)(messages | EShMsgAST);
     if (Options & EOptionSuppressWarnings)
@@ -2117,7 +2122,8 @@ void usage()
            "                                    initialized with the shader binary code\n"
            "  --no-link                         Only compile shader; do not link (GLSL-only)\n"
            "                                    NOTE: this option will set the export linkage\n"
-           "                                          attribute on all functions\n");
+           "                                          attribute on all functions\n"
+           "  --report-statements               Report errors on statement\n");
 
     exit(EFailUsage);
 }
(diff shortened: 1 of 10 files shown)


arcady-lunarg:
This looks interesting, but it's going to take a little while for me to give this a proper review.

phr34k:
Sounds good, surely there are some corner cases that can have been easy to miss, but I hope it's a small step in the right direction!

phr34k:
@arcady-lunarg did you ever get a chance to review this?

arcady-lunarg:
I am not sure I like the idea of mark_start/mark_end grammar rules. Maybe we need to update glslang's location type to be a span instead, i.e. having bot a start line/column and an end line/columns? I need to read bison's documentation a bit more as well, and I think the thing I would find the most helpful here is a before/after example of what effect this change has on some real use case.

phr34k:
> I am not sure I like the idea of mark_start/mark_end grammar rules. Maybe we need to update glslang's location type to be a span instead, i.e. having bot a start line/column and an end line/columns? I need to read bison's documentation a bit more as well, and I think the thing I would find the most helpful here is a before/after example of what effect this change has on some real use case.

Sure, a second pair of eyes would be great. I mean, people working with more expierence in bison have a cleaner solution that I just didn't know about, so definitely open to ideas.

Here's what I can tell you, what I remember on top of my head:

I did have a play around with `%define api.location.type {LocationType}` to define a custom location type, and bison's own built-in span location type. I believe one of the problems I had with that was that bison parser stack can reduce at abritary points and it could have already reduced half of the 'entire' statement so whatever was still on parser stack or from the last group never matched the correct locations.  

With regards to `mark_start/mark_end`  the biggest problem I ran into were bison imposed restrictions w.r.t. `mid rule actions`. Most of the grammer have parts where multiple rules are used in small variations `i.e. RETURN SEMICOLON | RETURN expression SEMICOLON`. One of the problems I had is that bison encounters such an action, according various resources bison commit to that one rule and stops evaluating other rules that could be a better fit. 

So if you have `RETURN { ... } SEMICOLON | RETURN { ... } expression SEMICOLON` despite being identical `mid rule actions`  it commits to the first rule, by seeing a `return` symbol, and will not evaluate the second rule anymore - throwing errors it's expecting `semicolon` when `expr` is used.  One of the recommended workarounds around this problem, by placing the `mid-rule action` in it's own symbol, that's how I ended up making `mark_start/mark_end`.

The first approach to me felt like I was running into a dead end given that I couldn't have any control over the parser stack so I figured the next best approach was marking statement boundaries manually, and use the look-a-head token to mark the location.

---- 

For an real-world use case example, let's agree that shader complexity can outgrow where they are kept in a single file (see for example amd's fiddelity fx cacao). Basically there might be multiple shaders, often they do have shared function definitions that are reused. 

In one of the shaders you could have a definition like this, but just forgetting an semi-colon or a curly brace:

attributes.glsl.inc:

```
#IFDEF CONDITION_1
layout( .... ) fragColor1
#ENDIF

#IFDEF CONDITION_2
void pbr() {
#ENDIF
```

material-a.glsl:

```
#define CONDITION_1
#include "attributes.glsl.inc" 
layout( .... ) fragColor2:
void main() {  ....  }
``` 
material-b.glsl:

```
#define CONDITION_2
#include "attributes.glsl.inc" 
layout( .... ) fragColor2:
void main() {  ....  }
``` 


So now you have these two materials a and b, when an grammar violation occurs bison reports you an message along the lines of epected `symbol semicolon` but received `layout` and then references the location of `layout`. However to understand where the problem really comes from, you have to understand how `functions.glsl.inc` got evaluated, and that's where the problem is. 

Sometimes you don't really know, by looking at the shader or there might be so many includes it's obfuscated from the reader, and so even if you know the source is `functions.glsl.inc` the question is 'where in functions.glsl.inc'.  In the change set that I prepared, I marked the statement boundaries, so instead of reporting the error on layout it reports the last know location of where the statement begin, so it'll tell you 'attributes.glsl.inc line 1' or 'attributes.glsl.inc line 5' which drastically helps you narrow down your search.

PR #3562, opened 2024-03-30
Check for exponent overflow in float parser

arcady-lunarg:
Even for a double precision float, the largest valid exponent is 308, so
clamp exponents to 500 when parsing to avoid overflow of the parsed
exponent value if the exponent is too big.

Also add a test for this case courtesy of @jimihem.


jeremy-lunarg (approved):
LGTM

PR #3564, opened 2024-04-01
Using reserve() in GlslangToSpv.cpp

GermanAizek:
We know the count inserted elements, we can optimize codegeneration.


jeremy-lunarg (approved):
Seems okay to me but, out of curiosity, did you encounter this as a result of a performance measurement?

GermanAizek:
> Seems okay to me but, out of curiosity, did you encounter this as a result of a performance measurement?

hi, I'm from PPSSPP project, PSP emulator runs on android and any lags/delays on weak devices are very noticeable there.

arcady-lunarg:
I would still be interested in concrete benchmark numbers if you can get them. Should this be done more broadly in the SPIR-V builder? It seems kind of odd that it's just those two opcodes that get a `reserve()` call, but it certainly doesn't hurt to do it.

PR #3565, opened 2024-04-01
Using std::make_unique C++17 in some functions

GermanAizek:
If contributors approve my changes, I can fix them for all instructions.


arcady-lunarg:
You should probably run the same tests that get run in CI before you make a pull request, for example I run `ninja -C build test` (make sure you've run `update_glslang_sources.py` and that building of tests is enabled).

PR #3566, opened 2024-04-01
Maximum optimization inserts using reserve() for operands

GermanAizek:
@jeremy-lunarg,
This PR changes very affect on performance Release and Debug configuration.

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -157,6 +157,7 @@ Id Builder::makePointer(StorageClass storageClass, Id pointee)
 
     // not found, make it
     type = new Instruction(getUniqueId(), NoType, OpTypePointer);
+    type->reserveOperands(2);
     type->addImmediateOperand(storageClass);
     type->addIdOperand(pointee);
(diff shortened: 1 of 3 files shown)


SPIRV/SpvBuilder.cpp
```diff
@@ -2808,6 +2883,7 @@ Id Builder::createLvalueSwizzle(Id typeId, Id target, Id source, const std::vect
         components[channels[i]] = numTargetComponents + i;
 
     // finish the instruction with these components selectors
+    swizzle->reserveOperands(numTargetComponents);
```
jeremy-lunarg, line 2886: +2?

jeremy-lunarg (commented):
Looks okay to me, but I'm genuinely curious in the changes you observed in performance.

arcady-lunarg (approved):

arcady-lunarg:
@GermanAizek I haven't been able to see any significant change from this in my attempts at benchmarking but also we don't really have any good performance testing for glslang at the moment. I would still love to see some actual numbers in terms of how much this improves performance in your use case.

PR #3575, opened 2024-04-17
Better follow HLSL offsets rules

Svenny:
Matrices consuming one vector are treated like vectors for alignment, and there is no "trailing padding" for matrices and arrays.


arcady-lunarg (approved):
LGTM

PR #3577, opened 2024-04-19
Keep vec1.x l-value in HLSL

Svenny:
Allows assignments like `float1 v1; v1.x = ...` to compile.

I couldn't simply do the same with mutliple selectors (EOpVectorSwizzle) - this fails assert in SpvBuilder trying to emit aggregate access into float type.

Changes in hlsl.shapeConv.frag are just renumbering, duplicated load (80) has disappeared for whatever reason.


arcady-lunarg (approved):
LGTM

PR #3578, opened 2024-04-19
A couple of fixes to the GL_QCOM_image_processing2 support

wooyoungqcom:
1) handle the extension requirements of GL_QCOM_image_processing2
   correctly
2) add QCOM image processing decorations to ids only once.


Test/baseResults/spv.tpipBlockMatchGatherSAD.frag.out
```diff
@@ -1,17 +1,19 @@
 spv.tpipBlockMatchGatherSAD.frag
-Validation failed
 // Module Version 10000
 // Generated by (magic number): 8000b
 // Id's are bound by 72
 
                               Capability Shader
+                              Capability TextureBlockMatchQCOM
                               Capability Bad
```
arcady-lunarg, line 8: I think there's an update to `SPIRV/doc.cpp` that you are missing here.

arcady-lunarg (changes requested):
Sorry, I should have reviewed the original change more thoroughly, we really shouldn't be checking in expected results with "Validation failed" messages. However, it seems like you still have a "Bad" in your expected results, which is probably something you forgot to add to the appropriate switch statement in `SPIRV/doc.cpp`.

arcady-lunarg (approved):

PR #3586, opened 2024-04-25
Increase the TIntermediate::usedIo array

arcady-lunarg:
In 0bbec2e8f, the code started using usedIo[4], however it never increased the size of the array to make that index actually valid.


arcady-lunarg:
cc: @janharaldfredriksen-arm

janharaldfredriksen-arm:
ouch.
This looks like it works as a quickfix. Is it worth adding some assertions somewhere that can trap this kind of thing? The hardcoded indices could be prone to issues like this, but the number of storage classes don't change very often so perhaps fixing that isn't worth the effort.

jeremy-lunarg (approved):
LGTM

Issue #3595, opened 2024-05-05
Miss  glslang_shader_shader_set_entry_point


metarutaiga:
I tried to use C Interface to compile HLSL.
But it miss a function to set entry point.

https://github.com/KhronosGroup/glslang/blob/e8dd0b6903b34f1879520b444634c75ea2deedf5/glslang/Public/ShaderLang.h#L471

https://github.com/KhronosGroup/glslang/blob/e8dd0b6903b34f1879520b444634c75ea2deedf5/glslang/Include/glslang_c_interface.h#L257
Maybe
GLSLANG_EXPORT void glslang_shader_set_entry_point(glslang_shader_t* shader, const char* s); 

and

GLSLANG_EXPORT void glslang_shader_set_invertY(glslang_shader_t* shader, bool invertY);

PR #3597, opened 2024-05-08
Added glslang_shader_set_entry_point and glslang_shader_set_invert_y

metarutaiga:
Added PR from Issue : https://github.com/KhronosGroup/glslang/issues/3595


arcady-lunarg:
FYI you're going to need to make your commit with the same email you use for github or else the CLA checker gets confused.

PR #3599, opened 2024-05-10
add back layoutLocation to public API

### LINKED ISSUE #3491: Access layoutLocation via public headers
I used to do 
```cpp
program->getUniform(i).getType()->getQualifier().layoutLocation
```

to get the uniform location declared as:
```
layout(location = 2) uniform ...
```

now that the `TType` definition is gone from the public headers, is there another way to get to it? Or is this something that needs to be added back to the public API?

paroj:
fixes #3491

### DIFF
--- a/glslang/MachineIndependent/ShaderLang.cpp
+++ b/glslang/MachineIndependent/ShaderLang.cpp
@@ -2105,7 +2105,7 @@ const char* TProgram::getInfoDebugLog()
 // Reflection implementation.
 //
 
-int TObjectReflection::layoutLocation() const { return type->getQualifier().layoutLocation; }
+unsigned int TObjectReflection::layoutLocation() const { return type->getQualifier().layoutLocation; }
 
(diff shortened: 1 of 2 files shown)


arcady-lunarg (changes requested):
The idea is fine, you just need to fix the function signature so it builds successfully.

glslang/MachineIndependent/ShaderLang.cpp
```diff
@@ -2105,6 +2105,8 @@ const char* TProgram::getInfoDebugLog()
 // Reflection implementation.
 //
 
+int TObjectReflection::layoutLocation() const { return type->getQualifier().layoutLocation; }
```
paroj, line 2108:
```suggestion
unsigned int TObjectReflection::layoutLocation() const { return type->getQualifier().layoutLocation; }
```

arcady-lunarg (approved):
LGTM, thanks.

PR #3613, opened 2024-05-31
SPV_EXT_replicated_composites/GL_EXT_spec_constant_composites

jeffbolznv:
Opportunistically generate SPV_EXT_replicated_composites instructions when the operands all match and when requested by pragma. Also implement GL_EXT_spec_constant_composites, which allows composite types to be spec constants.


jeffbolznv:
GLSL extension is at https://github.com/KhronosGroup/GLSL/pull/236

arcady-lunarg (approved):
LGTM.

Issue #3238, opened 2023-06-09
[glslangValidator] Column of error report always zero


FrostKiwi:
Installed the MSYS2 package `mingw-w64-x86_64-glslang` in the MinGW64 environment to allow VSCode to lint Shaders via https://github.com/hsimpson/vscode-glsllint

I created GLSL errors and ran `glslangValidator project.vert project.frag` in the Mingw64 environment of MSYS2. It outputs the column of an error as always zero. See screenshot of the Windows Terminal at the very bottom.
<details><summary>Sample output</summary>

```console
$ glslangValidator project.vert project.frag
project.vert
ERROR: 0:4: '' :  syntax error, unexpected IDENTIFIER, expecting LEFT_BRACE or COMMA or SEMICOLON
ERROR: 1 compilation errors.  No code generated.


project.frag
ERROR: 0:10: 'normze' : no matching overloaded function found
ERROR: 0:10: '=' :  cannot convert from ' const float' to ' temp highp 3-component vector of float'
ERROR: 0:10: '' : compilation terminated
ERROR: 3 compilation errors.  No code generated.
```
</details>
<details><summary>glslangValidator --version</summary>

```
Glslang Version: 11:12.1.0
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 12.1.0
GLSL Version: 4.60 glslang Khronos. 12.1.0
SPIR-V Version 0x00010600, Revision 1
GLSL.std.450 Version 100, Revision 1
Khronos Tool ID 8
SPIR-V Generator Version 11
GL_KHR_vulkan_glsl version 100
ARB_GL_gl_spirv version 100
```
</details>

This results in the linter always displaying error squiggles at the first token of the line, which is visually unpleasing and an extra workflow step of mouse-overing that squiggles to understand what's actually going on.
![image](https://github.com/KhronosGroup/glslang/assets/60887273/f11a6046-4d8a-4594-b338-fd90b0ebdbfa)
![image](https://github.com/KhronosGroup/glslang/assets/60887273/45724cfd-4724-4e58-8200-1242f11ea814)

antaalt:
I was interested by this issue, after looking a bit into this, it seems the zero is not the column but the TSourceLoc::string integer that is printed, which I don't really get what use it is for...
[...]

PR #3614, opened 2024-06-01
Add column to location logs.

antaalt:
The column is never printed to error logs. This commit simply add the column to the log after the line. This is really useful for external linting applications that will locate the error for user.

The issue here is that this might be a breaking change as some external parser might be relying on the log format to work correctly, still I think it should be pushed in order to be used as an efficient linter.

This is solving [this issue](https://github.com/KhronosGroup/glslang/issues/3238)


arcady-lunarg:
I think this is a good idea in principle, you're almost certainly going to need to update some test results though, and also check that the column numbers are actually valid in those test results, since it's likely nobody's actually checked the column numbers since they were never really visible.

antaalt:
I have checked that the column display is valid using vscode-glsllint and it seems to be working fine

arcady-lunarg (approved):
This looks good, but it definitely is a bit of a compatibility break. I think it's worth it, but I might let this sit for a bit in case anyone has other thoughts.

arcady-lunarg:
Actually, thinking about this more, is there a way to add this behind an option. Would that work with your use case?

antaalt:
It should be possible yes, I can try to look into it

antaalt:
Here I add an option for this, but right now, all test are based on this option, there should probably be some test that do not use this option

arcady-lunarg (changes requested):
Looks good, though I think I would prefer not to update every single test result and leave your one new test as the test of the error column functionality. Could you remove the passing of the new option from the test fixture (but keep your runtests test, we need that)?

antaalt:
Should be good to go, I tried to add a test for it through fixture, but thought the runtests should be enough.

arcady-lunarg (approved):
LGTM, thanks for this change and for bearing with me through the review process.

PR #3617, opened 2024-06-04
Generate debuginfo for buffer reference types

arcady-lunarg:
With the new `OpExtInstWithDebugInfo` we can finally generate recursive types in nonsemantic debuginfo, which we can use to properly provide type information for buffer reference types.


jeremy-lunarg (approved):
LGTM

PR #3618, opened 2024-06-07
Add support for the ARM extended matrix layout

silverclaw:


arcady-lunarg (commented):
Something seems off about this, looking at the expected test result, I don't see the new constants appearing, nor do I see the new capability/extension in the SPIR-V.

silverclaw:
> Something seems off about this, looking at the expected test result, I don't see the new constants appearing, nor do I see the new capability/extension in the SPIR-V.

You are of course correct - I'm comparing the SPIRV ID rather than the actual constant when adding the capability. I'll fix that shortly.

As for the constant not appearing in the SPV, I think it's there:

> 76:     58(int) Constant 4202

Note that this test is a copy of the original cooperative matrix test with the layout specifier replaced.

arcady-lunarg (approved):
Looks good, thanks.

Issue #3616, opened 2024-06-04
Incorrect codegen when there is redundant float16_t conversion


FanShupei:
For background, you may see https://github.com/google/shaderc/issues/1418

I just understand this is actually a glslang bug, so I raise the issue here.

For the compute shader, `glslang` generates ill-formed spirv binary, demonstrated by the following commands
```
$ glslang --target-env vulkan1.3 issue.comp
$ spirv-val comp.spv
error: line 65: Expected input to have different bit width from Result Type: FConvert
  %33 = OpFConvert %half %31
```

```glsl
#version 450
#extension GL_EXT_shader_16bit_storage : require
layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;

#define TYPE float16_t // or 'int64_t', 'uint16_t', also generated ill-formed spirv
// #define TYPE float, int, uint, these are fine
layout(binding = 0) readonly buffer A { TYPE data_a[]; };
layout(binding = 1) writeonly buffer D { TYPE data_d[]; };

void main() {
    const uint i = gl_GlobalInvocationID.x;
    data_d[i] = TYPE(data_a[i]);
}
```

I take a look at the disassembly, it is indeed incorrect. The spec on `OPFConvert` says that "The component width must not equal the component width in Result Type.", so the op actually should be eliminated.
```
; SPIR-V
; Version: 1.6
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 37
; Schema: 0
               OpCapability Shader
               OpCapability StorageBuffer16BitAccess
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID %_ %__0
               OpExecutionMode %main LocalSize 1 1 1
               OpSource GLSL 450
               OpSourceExtension "GL_EXT_shader_16bit_storage"
               OpName %main "main"
               OpName %i "i"
               OpName %gl_GlobalInvocationID "gl_GlobalInvocationID"
               OpName %D "D"
               OpMemberName %D 0 "data_d"
               OpName %_ ""
               OpName %A "A"
               OpMemberName %A 0 "data_a"
               OpName %__0 ""
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpDecorate %_runtimearr_half ArrayStride 2
               OpMemberDecorate %D 0 NonReadable
[...]

Issue #3607, opened 2024-05-21
Incorrect SPIRV codegen for 8bit/16bit variables in buffers


alelenv:
For the following simple shader
```
#version 460


#extension GL_EXT_shader_8bit_storage : require
#extension GL_EXT_shader_16bit_storage : require
#extension GL_EXT_shader_explicit_arithmetic_types_float16 : require


layout(binding = 1 ) uniform _16bit_storage
{
        i16vec4 i16v4;
};

// This is read back and checked on the CPU side to verify the converions
layout(binding = 2 ) writeonly buffer ConversionOutBuffer
{
        i8vec4 i16v4_to_i8v4;
} cob;

out vec4 fcolor;

void main()
{
        // Conversions
        {
                cob.i16v4_to_i8v4   = i8vec4(i16v4);
        }

        bool RED = true;
        bool GREEN = false;

        fcolor = vec4( (RED) ? 1.0f : 0.0f,
                                   (GREEN) ? 1.0f : 0.0f,
                                   0.0f, 1.0f);
}
```

We see some weird codegen when loading members from UBO and doing a convert

```
         %19 = OpAccessChain %_ptr_Uniform_v4short %_ %int_0
         %20 = OpLoad %v4short %19
         %22 = OpCompositeExtract %int %20 0
         %23 = OpCompositeExtract %int %20 1
         %24 = OpCompositeExtract %int %20 2
         %25 = OpCompositeExtract %int %20 3
         %26 = OpCompositeConstruct %v4int %22 %23 %24 %25
         %27 = OpSConvert %v4char %26
```

OpCompositeExtract tried to extract as 'int' when %20 is a vector of i16/shorts

arcady-lunarg:
Yeah that definitely isn't valid SPIR-V, I'll take a look. Interestingly this seems to be the one cast that isn't currently covered by our unit tests.

ravi688:
@alelenv, how did you compiled your shader?
Trying to execute the following command reports an error for me:
```
$ glslang -G test.frag -o test.frag.spv
test.frag
ERROR: test.frag:20: 'location' : SPIR-V requires location for user input/output 
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
```
The test.frag contains the shader you embedded into your first messaged in this issue.

arcady-lunarg:
[...]

PR #3622, opened 2024-06-16
[BugFix] Skip identity conversions for 8-bit and 16-bit types

ravi688:
Original Issue Reported: https://github.com/KhronosGroup/glslang/issues/3616

This change affects the following type of shaders (which do identity conversion for 8/16 bit types):
```
layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
#define TYPE float16_t
layout(binding = 0) readonly buffer A { TYPE data_a[]; };
layout(binding = 1) writeonly buffer D { TYPE data_d[]; };

void main() {
    const uint i = gl_GlobalInvocationID.x;
    data_d[i] = TYPE(data_a[i]);
}
```
Earlier (before this fix), it generated incorrect SPIR-V convert instructions.

**Ran `ctest` and it reports all tests are passsing:**
Test project /home/ravi/OpenSource/glslang/build
    Start 1: glslang-testsuite
1/2 Test #1: glslang-testsuite ................   Passed   30.53 sec
    Start 2: glslang-gtests
2/2 Test #2: glslang-gtests ...................   Passed   15.38 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =  45.92 sec

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -4,6 +4,7 @@
 // Copyright (C) 2015-2018 Google, Inc.
 // Copyright (C) 2017, 2019 ARM Limited.
 // Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
+// Modifications Copyright (C) 2024 Ravi Prakash Singh.
 //
 // All rights reserved.
 //
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -8540,7 +8540,8 @@ TIntermTyped* TParseContext::constructBuiltIn(const TType& type, TOperator op, T
         basicOp = EOpConstructFloat16;
         // 8/16-bit storage extensions don't support constructing composites of 8/16-bit types,
         // so construct a 32-bit type and convert
-        if (!intermediate.getArithemeticFloat16Enabled()) {
+        // and do not generate any conversion if it is an identity conversion, i.e. float16_t(<float16_t> var)
+        if (!intermediate.getArithemeticFloat16Enabled() && (node->getBasicType() != EbtFloat16)) {
             TType tempType(EbtFloat, EvqTemporary, type.getVectorSize());
             newNode = node;
             if (tempType != newNode->getType()) {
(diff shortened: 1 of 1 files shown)


ravi688:
Contribution License Aggreement instructed me to include `Copyright (C) [yyyy] [MyName]`, Shall do it as an Individual?

dneto0:
> Contribution License Aggreement instructed me to include `Copyright (C) [yyyy] [MyName]`, Shall I do it as an Individual?

If you work for a tech company they may own the copyright on your work.
If you don't know then do it as an individual.  

thanks!

arcady-lunarg (changes requested):
Thanks for the PR. I think the change itself looks reasonable. However, it would be good to add some unit tests for this (and possibly the pattern in the similar issue #3607, if your PR fixes that as well). Let me know if you have any questions about the testing framework that glslang uses or about how to add tests.

arcady-lunarg:
Actually, I'm just going to make a commit adding some test cases on top of yours and then I will merge this.

ravi688:
HI @arcady-lunarg ,

Apologies for the late response. This change doesn't fix the #3607 issue. I planned to look that into this weekend. Sorry if I will be late. You may proceed to merge this.

arcady-lunarg:
Though, now that I actually read the [GLSL_EXT_shader_16bit_storage spec](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_shader_16bit_storage.txt), it seems that these identity conversions are actually forbidden and indeed that converting directly from 8-bit to 16-bit without going through int is also forbidden. That seems unnecessarily strict and I have to wonder if that was the original intention of the spec writers to forbid this. @jeffbolznv, as the original spec writer, might you weigh in here?

arcady-lunarg:
The original GLSL spec does say "Identity constructors, like float(float) are also legal, but of little use", so IMO this should still be allowed in the sized storage extensions as well.

jeffbolznv:
> That seems unnecessarily strict and I have to wonder if that was the original intention of the spec writers to forbid this. @jeffbolznv, as the original spec writer, might you weigh in here?

The original 8/16-bit storage support in Vulkan was very limited because some IHVs didn't really support 8/16-bit types and could only load them from memory but not really operate on them as smaller types. I don't remember specifics of the GLSL extension beyond that.

arcady-lunarg (approved):
I'm going to add the tests myself, thanks, this change looks good.

PR #3625, opened 2024-06-17
Update known_good.json

rjodinchr:
Also update test that is now passing


dneto0 (approved):

arcady-lunarg (approved):

arcady-lunarg:
We really should have a check for tests with the "Validation failed" string in the results, but first we need to get rid of all the "validation failed" results we already have.

PR #3626, opened 2024-06-20
Add test cases for redundant type conversions

arcady-lunarg:
These redundant type conversions were generating illegal SPIR-V when only the 8-bit/16-bit storage extensions and not the corresponding arithmetic extensions were enabled.


jeremy-lunarg (approved):
LGTM

PR #3628, opened 2024-06-21
Generate vector constructions more efficiently when sizes match

arcady-lunarg:
When two vectors are the same size, there is no need to extract the components and construct a new vector.


ravi688:
This change leads to an assertion failure for the following shader:
```
#version 460

#extension GL_EXT_shader_8bit_storage : require
#extension GL_EXT_shader_16bit_storage : require

layout(binding = 1 ) uniform _16bit_storage
{
        int16_t i16;
	i16vec4 i16v4;
};

// This is read back and checked on the CPU side to verify the converions
layout(binding = 2 ) writeonly buffer ConversionOutBuffer
{
	int8_t i8;
	i8vec4 i8v4;
} cob;

out vec4 fcolor;

void main()
{
        // Conversions
        {
                cob.i8   = int8_t(i16);
		cob.i8v4 = i8vec4(i16v4);
        }

        bool RED = true;
        bool GREEN = false;

        fcolor = vec4( (RED) ? 1.0f : 0.0f,
                                   (GREEN) ? 1.0f : 0.0f,
                                   0.0f, 1.0f);
}
```

Assertion:
glslang: /home/ravi/OpenSource/glslang/SPIRV/SpvBuilder.cpp:3429: spv::Id spv::Builder::createConstructor(spv::Decoration, const std::vector<unsigned int>&, spv::Id): Assertion `resultTypeId == getTypeId(sources[0])' failed.

ravi688:
The assertion failure is fixed in PR #3631.

jeremy-lunarg (approved):
Looks good. We should follow up on #3631 after this lands.

arcady-lunarg:
The assertion is in place of generating invalid SPIR-V so I wouldn't count it as a regression.

PR #3630, opened 2024-06-22
Update comment to more succint for 8/16-bit types

ravi688:
... that 8/16-bit composites can't be directly constructed.

Earlier the comment looked misleading that composites of 8/16-bit types can't be constructed at all. As per the GL_EXT_shader_16bit_storage extension, they can't be constructed directly, however, can be constructed indirectly.


ravi688:
To the maintainer, I'm not sure if the comment updates in this PR are worthwhile - so feel free to reject it. 

Also, I've made a comment on #3607 for a possible solution. Suggestions are appreciated for that.

arcady-lunarg (approved):

PR #3633, opened 2024-06-24
Fix compilation for WASI target

antaalt:
I was looking into compiling glslang for [WASM WASI](https://wasi.dev/) target, and I successfully did so by removing some reference to threads as WASI does not support threads yet. 
So here is a small PR that disable thread safety in order to allow compilation for WASI, the preprocessor guarding is safe so this should not impact anything other than WASI target.

### DIFF
--- a/glslang/Include/Common.h
+++ b/glslang/Include/Common.h
@@ -94,6 +94,11 @@ std::string to_string(const T& val) {
     #pragma warning(disable : 4201) // nameless union
 #endif
 
+// Allow compilation to WASI which does not support threads yet.
+#ifdef __wasi__ 
+#define DISABLE_THREAD_SUPPORT
+#endif
+
 #include "PoolAlloc.h"
 
 //
--- a/glslang/MachineIndependent/PoolAlloc.cpp
+++ b/glslang/MachineIndependent/PoolAlloc.cpp
@@ -35,14 +35,21 @@
 #include "../Include/Common.h"
 #include "../Include/PoolAlloc.h"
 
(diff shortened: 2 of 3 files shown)


arcady-lunarg:
Is this the complete set of changes needed? I also see uses of the `<mutex>` header in `SPIRV/doc.cpp` and `StandAlone/Worklist.h`. Would you mind sharing instructions as to how you went about building for WASI?

antaalt:
On my side, it was enough to compile, the `<mutex>` header does not seems to be an issue, but I might add it to the preprocessor aswell to be sure.

For the compilation, I went with [glslang-rs](https://github.com/antaalt/glslang-rs) which is a rust wrapper for glslang, and build c++ bindings using a build.rs file. It is fairly simple, just listing all the files to build and linking them. I did not looked into CMakeList because WASI [does not look supported yet](https://gitlab.kitware.com/cmake/cmake/-/issues/19223), but maybe it could be added to the build system, which require clang with a sysroot retrieved from WASI SDK (__wasi__ preprocessor should only be defined in clang targeting wasi, which is the only compiler supporting wasi right now).

antaalt:
For the background, I am trying to compile it to WASI because vscode now support [running WASI module](https://code.visualstudio.com/blogs/2023/06/05/vscode-wasm-wasi) and I was trying to have a linter extension with glslang directly embedded into the extension, it is working nicely but require these few adjustments to compile.

arcady-lunarg (approved):

PR #3635, opened 2024-06-25
Don't emit duplicate decorations.

### LINKED ISSUE #3627: Constant generated with two RelaxedPrecision decorations
If there are to global constants with the same value, the same OpConstant will be reused for both. However, the decoration will be added once for each. This is invalid spir-v.

https://shader-playground.timjones.io/001f515dcbc39de283c41893ee3e33a4

s-perron:
It is invalid if the same decoration is applied to the same id multiple times. This adds a check before adding a decoration that the decoration is not alreay in the list. If it is, then the duplicate is not added.

Fixes #3627


SPIRV/SpvBuilder.cpp
```diff
@@ -4347,4 +4346,40 @@ void Builder::dumpModuleProcesses(std::vector<unsigned int>& out) const
     }
 }
 
+bool Builder::DecorationInstructionLessThan::operator()(const std::unique_ptr<Instruction>& lhs,
```
arcady-lunarg, line 4349: The nice thing about this is that decorations will always be emitted in a canonical order regardless of the details of the code generation in the backend.

arcady-lunarg (approved):

PR #3642, opened 2024-07-02
Fix `glslangValidator` installation

friendlyanon:
This caused the `glslang[tools]` vcpkg feature to not install properly, because the `glslangValidator` executable didn't exist for tool installation.


friendlyanon:
I have submitted the commit from this PR to vcpkg so the package works.

arcady-lunarg (approved):

friendlyanon:
Thank you. This was also tested and merged to the vcpkg package as well.

PR #3645, opened 2024-07-03
Sanitize debug source location tracking for implicit branch and return

qingyuanzNV:
This patch improves the debug line generation for implicit branch/return instructions. "IMPLICIT" means that a branch/return instruction is generated by the compiler to compose a structural construct instead of being translated from a program statement. Therefore:
1. All branches without an explicit "break/continue" are implicit.
2. All conditional branches are implicit.
3. All returns without an explicit "return" are implicit.

For implicit branch instruction, we always don't give it a debug location. Instead, we attach it to the whatever preceding instruction in the block. Typically, they should have the line number of the selection expression or the last statement in if/else body. Sometimes, we get empty if/else body. In this case, no debug location will be generated. (Ideally, we could have it point to '}' of the empty block. But glslang will set the statement node to nullptr instead of creating an empty aggregate)
For implicit return instruction, we always give it a debug location pointing to the closing '}' of the function body.

Additionally, this patch fix a bug where Phi instruction in the selection statement doesn't get a debug location.


SPIRV/SpvBuilder.h
```diff
@@ -429,6 +428,11 @@ class Builder {
     // Optionally, additional debug info instructions may also be prepended.
     void addInstruction(std::unique_ptr<Instruction> inst);
 
+    // Append an instruction to the end of the current build point without prepending any debug instructions.
```
arcady-lunarg, line 431: This helper is a good idea, much cleaner than the solution I came up with for fixing OpPhis.

glslang/Include/intermediate.h
```diff
@@ -1733,6 +1736,10 @@ class TIntermAggregate : public TIntermOperator {
     TPragmaTable* pragmaTable;
     TSpirvInstruction spirvInst;
     TLinkType linkType = ELinkNone;
+
+    // Marking the end source location of the aggregate.
+    // This is currently only set for a compound statement or a function body, pointing to '}'.
+    TSourceLoc endLoc;
```
arcady-lunarg, line 1742: You should call `endLoc.init()` in the constructor to avoid uninitialized garbage in the case where this doesn't get set.

arcady-lunarg (changes requested):
This PR looks good but it needs to be rebased, and you need to address the initialization of TIntermAggregate::endLoc.

arcady-lunarg:
@qingyuanzNV ping?
Could you rebase and address the one small piece of review feedback and then we can merge this?

qingyuanzNV:
> @qingyuanzNV ping? Could you rebase and address the one small piece of review feedback and then we can merge this?

Sorry, I missed the email. I'll address your review comments today.

qingyuanzNV:
Before we merge this, there's some problem with this patch. The SPIR-V spec says the following for OpPhi:
> Within a block, this instruction must appear before all non-OpPhi instructions (except for OpLine and OpNoLine, which can be mixed with OpPhi).
The exception doesn't include the DebugLine and DebugNoLine instruction in the Non-Semantic Debug extension, meaning that we are producing illegal modules with this patch when NS debug is on. However, I do think this is an outdated requirement in the spec. @arcady-lunarg what do you think? I suppose we should avoid emitting DebugLine/DebugNoLine for now and update the spec text first.

arcady-lunarg:
It seems like not allowing nonsemantic instructions before `OpPhi` is intentional, the [NonSemantic.Shader.DebugInfo.100 spec](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.html) explicitly calls out this as a change from the earlier OpenCL debuginfo spec. I did add some code to not emit debug instructions before an OpPhi and all further changes should comply with this restriction.

arcady-lunarg (approved):

PR #3646, opened 2024-07-04
Generate debug name matching the source language for opaque types

qingyuanzNV:
This patch improves the non-semantic debug name of opaque (sampler) types. Instead of using a generic name of "type.sampler" or something alike, we generate a name that matches the GLSL source language. So OpTypeImage would have debug name of "imageXXX" or "textureXXX". OpTypeSampler would have debug name of "sampler" or "samplerShadow". OpTypeSampledImage would have debug name of "samplerXXX".

Also, as a minor tweak, TSampler::getString() now should returns a std::string. This avoids memory for type names piling up in the pool when the function is called repetitively.

Question: need we retain the existing behavior for HLSL source? It seems other names for debug types are GLSL centric, but when a HLSL source program is compiled, it would look weird to have type names not exist in the language.


arcady-lunarg:
I think ideally for HLSL we should either keep the names as they are now (at least that won't break compatibility) or else use proper HLSL names, I don't think it makes sense to switch to GLSL-based names.

qingyuanzNV:
Thanks for comment! Let me restore the old names for HLSL path.

qingyuanzNV:
I noticed today that DXC is also generating the same name for opaque types like "type.2d.image". It would be great if this could be a written convention in the non-semantic debug spec. Then we could associate the linkage name of an opaque type with the actual SPIR-V type and we could easily identify opaque types across different languages. The SPIR-V generator like GLSLANG could only update the display name in DebugTypeComposite to the name in the source language and still use the linkage name like "@type.2d.image"

dnovillo:
> I noticed today that DXC is also generating the same name for opaque types like "type.2d.image". It would be great if this could be a written convention in the non-semantic debug spec. Then we could associate the linkage name of an opaque type with the actual SPIR-V type and we could easily identify opaque types across different languages. The SPIR-V generator like GLSLANG could only update the display name in DebugTypeComposite to the name in the source language and still use the linkage name like "@type.2d.image"

Could you add an issue on the gitlab repo? This sounds like a desirable fix to the spec.

dnovillo (commented):
> > The rebase is causing test failures. PTAL. Thanks.
> 
> Sorry, I wasn't intending to sending this rebase to review. I feel this is too hacky to be merged. I'm working on a better version. 

Sure. Thanks.

qingyuanzNV:
I end up just refined the old PR to avoid changing the behavior for HLSL target and added a few test cases. @dnovillo please help review.

PR #3648, opened 2024-07-05
kokoro: use Python 3.12 for Linux builds

dneto0:
Bug: crbug.com/350048185


dneto0:
This is a routine update of the tooling for the Kokoro builds

arcady-lunarg (approved):

PR #3649, opened 2024-07-05
Update macOS requirement from 11 to 12 in github CD workflow

arcady-lunarg:
Our github continuous delivery workflow appears to be failing for macOS, possibly because the requested version of the OS is too old.


jeremy-lunarg (approved):
LGTM

PR #3656, opened 2024-07-12
add cross-stage check for missing outputs

mbechard:
If an 'in' is present in a shader stage, make sure a matching 'out' is present in the previous stage. Only enabled when doing Vulkan.

This commit also fixes a bug where previous stage's linkerObjects got polluted with 'in' variables from the next stage when merging linker objects.

### DIFF
--- a/Test/baseResults/iomap.crossStage.vk.2.vert.out
+++ b/Test/baseResults/iomap.crossStage.vk.2.vert.out
@@ -1,0 +1,313 @@
+iomap.crossStage.vk.2.vert
+Shader version: 460
+0:? Sequence
+0:8  Function Definition: main( ( global void)
+0:8    Function Parameters: 
+0:10    Sequence
+0:10      move second child to first child ( temp highp 4-component vector of float)
+0:10        val: direct index for structure ( out highp 4-component vector of float)
+0:10          'anon@0' ( out block{ out highp 4-component vector of float val})
+0:10          Constant:
+0:10            0 (const uint)
+0:10        Constant:
+0:10          0.500000
+0:10          0.500000
+0:10          0.500000
+0:10          0.500000
+0:11      move second child to first child ( temp highp 4-component vector of float)
+0:11        'color' ( smooth out highp 4-component vector of float)
+0:11        Constant:
+0:11          1.000000
+0:11          1.000000
+0:11          1.000000
+0:11          1.000000
+0:12      move second child to first child ( temp 4-component vector of float)
+0:12        gl_Position: direct index for structure ( gl_Position 4-component vector of float Position)
+0:12          'anon@1' ( out block{ gl_Position 4-component vector of float Position gl_Position,  gl_PointSize float PointSize gl_PointSize,  out unsized 1-element array of float ClipDistance gl_ClipDistance,  out unsized 1-element array of float CullDistance gl_CullDistance})
+0:12          Constant:
+0:12            0 (const uint)
+0:12        Constant:
+0:12          1.000000
+0:12          1.000000
+0:12          1.000000
+0:12          1.000000
+0:?   Linker Objects
+0:?     'anon@0' ( out block{ out highp 4-component vector of float val})
+0:?     'color' ( smooth out highp 4-component vector of float)
+0:?     'anon@1' ( out block{ gl_Position 4-component vector of float Position gl_Position,  gl_PointSize float PointSize gl_PointSize,  out unsized 1-element array of float ClipDistance gl_ClipDistance,  out unsized 1-element array of float CullDistance gl_CullDistance})
+
+iomap.crossStage.vk.2.geom
+Shader version: 460
+invocations = -1
+max_vertices = 3
+input primitive = points
+output primitive = triangle_strip
+0:? Sequence
+0:23  Function Definition: main( ( global void)
+0:23    Function Parameters: 
+0:25    Sequence
+0:25      Sequence
+0:25        Sequence
+0:25          move second child to first child ( temp highp int)
+0:25            'i' ( temp highp int)
+0:25            Constant:
+0:25              0 (const int)
+0:25        Loop with condition tested first
+0:25          Loop Condition
+0:25          Compare Less Than ( temp bool)
+0:25            'i' ( temp highp int)
+0:25            Constant:
+0:25              3 (const int)
+0:25          Loop Body
+0:26          Sequence
+0:26            move second child to first child ( temp highp 4-component vector of float)
+0:26              'colorOut' (layout( stream=0) out highp 4-component vector of float)
+0:26              component-wise multiply ( temp highp 4-component vector of float)
+0:26                indirect index ( temp highp 4-component vector of float)
+0:26                  'color' ( in 1-element array of highp 4-component vector of float)
+0:26                  'i' ( temp highp int)
+0:26                val: direct index for structure ( in highp 4-component vector of float)
+0:26                  indirect index ( temp block{ in highp 4-component vector of float val})
+0:26                    'vv' ( in 1-element array of block{ in highp 4-component vector of float val})
+0:26                    'i' ( temp highp int)
+0:26                  Constant:
+0:26                    0 (const int)
+0:27            move second child to first child ( temp highp 4-component vector of float)
+0:27              vv2Val: direct index for structure (layout( stream=0) out highp 4-component vector of float)
+0:27                'anon@0' (layout( stream=0) out block{layout( stream=0) out highp 4-component vector of float vv2Val})
+0:27                Constant:
+0:27                  0 (const uint)
+0:27              Constant:
+0:27                1.000000
+0:27                1.000000
+0:27                1.000000
+0:27                1.000000
+0:28            EmitVertex ( global void)
+0:25          Loop Terminal Expression
+0:25          Post-Increment ( temp highp int)
+0:25            'i' ( temp highp int)
+0:30      EndPrimitive ( global void)
+0:?   Linker Objects
+0:?     'vgo1' ( in 1-element array of highp 4-component vector of float)
+0:?     'color' ( in 1-element array of highp 4-component vector of float)
+0:?     'colorOut' (layout( stream=0) out highp 4-component vector of float)
+0:?     'vv' ( in 1-element array of block{ in highp 4-component vector of float val})
+0:?     'anon@0' (layout( stream=0) out block{layout( stream=0) out highp 4-component vector of float vv2Val})
+
+iomap.crossStage.vk.2.frag
+Shader version: 460
+gl_FragCoord origin is upper left
+0:? Sequence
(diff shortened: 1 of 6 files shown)


mbechard:
Not totally confident about all the cases this can affect. Is there newer Vulkan workflows where it's legal to do cross-stage checking in GLSLang but the in/outs don't need to match?

greg-lunarg:
> Not totally confident about all the cases this can affect. Is there newer Vulkan workflows where it's legal to do cross-stage checking in GLSLang but the in/outs don't need to match?

I am not aware of Vulkan workflow where this is legal. I think this should be fine.

slime73:
Thanks for doing this, I just ran into a situation where this sort of validation would help a lot.

Would it be possible to have an opt-in setting to make it run even when vulkan isn't used? A new `EShMessages` enum flag maybe. My users write shaders and might not always be using vulkan but will expect their shaders to run when vulkan is used, so it'd be handy if I could make this validation unconditionally happen since I don't use separate shader objects in OpenGL.

That being said, if that would hold up the review I'd rather avoid that and add my own small patch in a fork for now instead. :)

mbechard:
@slime73 Seems like that would be a good second PR, just to keep things separate.

@arcady-lunarg ok I'm happy to have this reviewed then when you have a chance

arcady-lunarg (approved):
LGTM

pmistryNV:
Does the interface IN/OUT names have to match? Is this defined in any spec? 
For example in the commit vertex shader had "out vec4 color"

does the IN variable geometry shader has to have same name? Can it not have a different name? 
I think if this change is trying to introduce linking rules, then it needs first a specification in vulkan.

mbechard:
Hey, good question. For input/output blocks it seems pretty well defined that they are matched by name. The GLSL spec refers to the OpenGL for full rules for interface matching. For blocks:
```
• An output block is considered to match an input block in the subsequent
shader if the two blocks have the same block name, and the members of the
block match exactly in name, type, qualification, and declaration order.
```
However it does say:
```
Output blocks that do not match in name, but have a location and match
in every other way listed above may be considered to match by some
implementations, but not all - so this behaviour should not be relied
upon.
```
Which doesn't seem like something we should follow, since not a guaranteed behavior.

For input/output variables it is less clear. For example the GLSL spec states:
```
Tessellation control, evaluation, and geometry shader input variables get the per-vertex values written out by output variables of the same names in the previous active shader stage.
```
and for fragment shaders:
```
For this interface, the last active shader stage output variables and fragment shader input variables of the same name must match in type and qualification, with a few exceptions
```

But the OpenGL spec does state this is also valid match for in/out variables:
```
the two variables are declared with the same location and component layout qualifiers and match in type and qualification
```

That behavior seems more specific to GL_ARB_separate_shader_objects though, which this link operation is not doing, as it's trying to create a singular shader object/program.

I would think the possible change to this PR would be to allow for matching based on location for the in/out variables rather than name, but keep it name-based for the in/out blocks.

pmistryNV:
Few points.
1. All these interface matching requirements are in OpenGL spec section 7.4.1 and not Vulkan. So we should fix the check in the code that restricts this only for vulkan and make it only for OpenGL
2. And for OpenGL interface matching yes i agree that for IN/OUT interface variables matching should be based on location ,layout qualifier and type

pmistryNV:
For vulkan matching most of the requirements are listed at SPIRV level and needs to be looked into https://docs.vulkan.org/spec/latest/chapters/interfaces.html 

Interface Matching
An output variable, block, or structure member in a given shader stage has an interface match with an input variable, block, or structure member in a subsequent shader stage if they both adhere to the following conditions:

- They have equivalent decorations, other than:

   - XfbBuffer, XfbStride, Offset, and Stream

   - one is not decorated with Component and the other is declared with a Component of 0

   - Interpolation decorations

   - RelaxedPrecision if one is an input variable and the other an output variable

- Their types match as follows:

     - if the input is declared in a tessellation control or geometry shader as an OpTypeArray with an Element Type equivalent to the  OpType* declaration of the output, and neither is a structure member; or

    - if the [maintenance4 feature is enabled, they are declared as OpTypeVector variables, and the output has a Component Count value higher than that of the input but the same Component Type; or

    - if the output is declared in a mesh shader as an OpTypeArray with an Element Type equivalent to the OpType* declaration of the input, and neither is a structure member; or

    - if the input is decorated with PerVertexKHR, and is declared in a fragment shader as an OpTypeArray with an Element Type equivalent to the OpType* declaration of the output, and neither the input nor the output is a structure member; or

    - if in any other case they are declared with an equivalent OpType* declaration.

 If both are structures and every member has an interface match.

mbechard:
> 1. All these interface matching requirements are in OpenGL spec section 7.4.1 and not Vulkan. So we should fix the check in the code that restricts this only for vulkan and make it only for OpenGL

This check was specifically added for Vulkan though. I didn't enable it for OpenGL since I wanted to do a more minimal change (and I don't have OpenGL test cases).
Specifically, I need to ingest GLSL shaders written by end-user artists that use my software, not artists that work in-house with developer support. They have no notion of what a location/binding/offset/stride are. Additionally, those artists shouldn't have to know what the rendering API used underneath is. Most shaders written in GLSL should be ingestible in both OpenGL and Vulkan without the artist knowing (assuming they don't use any OpenGL or Vulkan-only GLSL features). They just want to declare variables and it's up to the rendering developer to fill in the uniforms/inputs with values and handle connecting Vulkan to their shaders.

https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt was made to aid with this.

I use the autoMapBindings/autoMapLocations features in GLSLang that allow shaders to be written without those decorations. Also autoPushConstants which allows push constants to be used without the artist having to know about them. These features are essential to making GLSL useful for these artists.

So the link phase needs to first ensure that we have valid GLSL interface matching between all shader stages, and give the artist useful errors if they make mistakes. Then the iomapper/mapIO() is used after that to fill in the bindings/locations.

Back to the spec. Some of those quotes are from the GLSL spec directly, not only the OpenGL section that is referred to. If you read **only** the GLSL spec, then all matching is done by name only, which is what I based this check on. The GLSL spec is the reference document an artist has available to them refer to when learning about GLSL. 
They do not (and should not) need to know about SPIR-V and any rules it may outline. That's up to the GLSL->SPIR-V converter to handle, but the artist needs errors at the GLSL level, which is why this check has been added here.

So I'm not really sure if matching SPIR-V interface matching checks is correct at this point, as that should be done later on when SPIR-V comes into play. In particular since locations/bindings may not have been assigned yet at this point anyways.

Having said all of that. I'm fine with making this check optional, and something that is only enabled as an option to the linker. But also augmenting this check so for in/outs, if a location is defined is used instead of the name, would be good as well.

arcady-lunarg:
This feature is already purely opt-in already, you have to compile all the shaders together and request linking, and it makes sense that these checks work on the level of GLSL semantics, so I tend to agree with @mbechard here that this is fine as is.

pmistryNV:
> > 1. All these interface matching requirements are in OpenGL spec section 7.4.1 and not Vulkan. So we should fix the check in the code that restricts this only for vulkan and make it only for OpenGL
> 
> This check was specifically added for Vulkan though. I didn't enable it for OpenGL since I wanted to do a more minimal change (and I don't have OpenGL test cases). Specifically, I need to ingest GLSL shaders written by end-user artists that use my software, not artists that work in-house with developer support. They have no notion of what a location/binding/offset/stride are. Additionally, those artists shouldn't have to know what the rendering API used underneath is. Most shaders written in GLSL should be ingestible in both OpenGL and Vulkan without the artist knowing (assuming they don't use any OpenGL or Vulkan-only GLSL features). They just want to declare variables and it's up to the rendering developer to fill in the uniforms/inputs with values and handle connecting Vulkan to their shaders.
> 
> https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt was made to aid with this.
> 
> I use the autoMapBindings/autoMapLocations features in GLSLang that allow shaders to be written without those decorations. Also autoPushConstants which allows push constants to be used without the artist having to know about them. These features are essential to making GLSL useful for these artists.
> 
> So the link phase needs to first ensure that we have valid GLSL interface matching between all shader stages, and give the artist useful errors if they make mistakes. Then the iomapper/mapIO() is used after that to fill in the bindings/locations.
> 
> Back to the spec. Some of those quotes are from the GLSL spec directly, not only the OpenGL section that is referred to. If you read **only** the GLSL spec, then all matching is done by name only, which is what I based this check on. The GLSL spec is the reference document an artist has available to them refer to when learning about GLSL. They do not (and should not) need to know about SPIR-V and any rules it may outline. That's up to the GLSL->SPIR-V converter to handle, but the artist needs errors at the GLSL level, which is why this check has been added here.
> 
> So I'm not really sure if matching SPIR-V interface matching checks is correct at this point, as that should be done later on when SPIR-V comes into play. In particular since locations/bindings may not have been assigned yet at this point anyways.
> 
> Having said all of that. I'm fine with making this check optional, and something that is only enabled as an option to the linker. But also augmenting this check so for in/outs, if a location is defined is used instead of the name, would be good as well.

again the https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt doesn't talk about interface matching. That was purely about the language compatibiltiy. Also the GLSL spec mentions OpenGL spec for interface matching. And Vulkan interface matching requirements are different. So I don't agree to this check for Vulkan

pmistryNV:
> This feature is already purely opt-in already, you have to compile all the shaders together and request linking, and it makes sense that these checks work on the level of GLSL semantics, so I tend to agree with @mbechard here that this is fine as is.

But it has to adhere to the specification, otherwise GLSLANG is not spec compliant.

pmistryNV:
Adding @jeffbolznv to comment on this fix. Jeff this patch in GLSLANG is trying enforce interafce matching rules of opengl into vulkan flow. Kindly have a look at the discussion and share your thoughts.

jeffbolznv:
I don't think the spec is clear enough about this, particularly from a "pure" GLSL perspective (i.e. without referring to API environment specs). I suggest filing an issue at https://github.com/KhronosGroup/GLSL/issues to clarify it.

I, personally, think location matching should be allowed, and this is the closest spec language I can find:
```
For outputs of all other shader stages, the location specifies a
vector number that can be used to match against inputs in a subsequent shader stage, even if that
shader is in a different program object.
```

But this sort of contradicts the quote above about "...input variables get the per-vertex values written out by output variables of the same names in the previous active shader stage." I suspect this was just some old language that didn't get updated when locations were added, or an issue of merging two extensions written at around the same time.

mbechard:
It feels like the middle ground here is to augment this check so that if locations are assigned at this stage (they may not be yet, due to the auto-assignment pass not yet being executed), they are used for matching instead of by name.
This would be for in/out variables only, not for in/out blocks.

Issue #792, opened 2017-03-23
Readonly decorations are only set on members of the struct type


gwihlidal:
Glslang only sets the readonly flags on members of the struct type. While it seems legal to do so, this is really awkward for reflection (i.e. spirv_cross) A workaround (like https://github.com/KhronosGroup/SPIRV-Cross/pull/139 - which likely won't reach mainline, since it's reporting decorations that don't exactly exist), is to check if you're querying an SSBO, and poke into the members to find the flags, which is kind of ugly.

If would be great if glslang could promote these flags up to the parent, instead.

Thanks,
Graham

johnkslang:
GLSL says

> The memory qualifiers coherent, volatile, restrict, **readonly**, and writeonly may be used in the
declaration of buffer variables (i.e., members of shader storage blocks). 

SPIR-V says

> **NonWritable**
> Apply to an object or a member of a structure type...

Shouldn't all tools be supporting that? It would seem strange to have glslang not represent GLSL correctly instead.

johnkslang:
This is all set up to support different flags on different members.

Maybe there should be a post process that 

- checks to see if for any flag, all members of a composite have the same setting for the flag, and
- if so, promote the flag out of the members up to the composite?

Is that the request? It would not fix other tools in the general sense, they would still have bugs when things vary per member.

Or, this is a request against specifications to not allow such variance? This would need be made against the issue-tracking for each involved specification (GLSL, Vulkan, SPIR-V).

Or, is this just about HLSL run-time, where it's the only member, and maybe when glslang is synthesizing the composite, it can put the flag on the composite instead? That might trigger new paths in drivers.

chaoticbob:
> - checks to see if for any flag, all members of a composite have the same setting for the flag, and
> - if so, promote the flag out of the members up to the composite?
>
> Is that the request? It would not fix other tools in the general sense, they would still have bugs when things vary per member.

Out of curiosity, what's an example in which the first of the above isn't true? i.e. where members have different flags? Maybe when you write to one member out of many?

Seems like in the case of SSBO reflection, it would be less work for tasks, like figuring out if an SSBO is a SRV or UAV, if one could determine RO/RW state from the composite.

johnkslang:
> Out of curiosity, what's an example in which the first of the above isn't true? i.e. where members have different flags?

[...]

PR #3662, opened 2024-07-17
Apply GLSL memory decorations to top-level OpVariable

mmoult:
Apply memory decorations from GLSL source to the top-level OpVariable. Previously, these decorations would only be applied to individual members. While this is correct behavior, it is more convenient for some front ends to see the decorations (specifically ReadOnly and WriteOnly) applied to the whole variable rather than individual members.

This change relates to https://github.com/KhronosGroup/glslang/issues/792.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -4372,7 +4372,14 @@ spv::Id TGlslangToSpvTraverser::createSpvVariable(const glslang::TIntermSymbol* 
         initializer = builder.makeNullConstant(spvType);
     }
 
-    return builder.createVariable(spv::NoPrecision, storageClass, spvType, name, initializer, false);
+    spv::Id var = builder.createVariable(spv::NoPrecision, storageClass, spvType, name, initializer, false);
+    std::vector<spv::Decoration> topLevelDecorations;
+    glslang::TQualifier typeQualifier = node->getType().getQualifier();
+    TranslateMemoryDecoration(typeQualifier, topLevelDecorations, glslangIntermediate->usingVulkanMemoryModel());
+    for (auto deco : topLevelDecorations) {
+        builder.addDecoration(var, deco);
+    }
+    return var;
 }
 
 // Return type Id of the sampled type.
--- a/Test/baseResults/hlsl.attributeC11.frag.out
+++ b/Test/baseResults/hlsl.attributeC11.frag.out
@@ -132,8 +132,10 @@ Validation failed
                               Decorate 43(buffer1) BufferBlock
                               MemberDecorate 43(buffer1) 0 NonWritable
                               MemberDecorate 43(buffer1) 0 Offset 0
+                              Decorate 45(buffer1) NonWritable
                               Decorate 45(buffer1) Binding 1
                               Decorate 45(buffer1) DescriptorSet 0
+                              Decorate 46(buffer3) NonWritable
                               Decorate 46(buffer3) Binding 3
                               Decorate 46(buffer3) DescriptorSet 2
                               Decorate 47(ci) SpecId 13
--- a/Test/baseResults/hlsl.buffer.frag.out
+++ b/Test/baseResults/hlsl.buffer.frag.out
@@ -205,6 +205,7 @@ Validation failed
                               Decorate 37(buf2) BufferBlock
                               MemberDecorate 37(buf2) 0 NonWritable
                               MemberDecorate 37(buf2) 0 Offset 0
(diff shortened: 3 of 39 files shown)


mmoult:
Also, to explain my reasoning:

[johnkslang](https://github.com/johnkslang) suggested:

> Maybe there should be a post process that
> - checks to see if for any flag, all members of a composite have the same setting for the flag, and
> - if so, promote the flag out of the members up to the composite?

I think that would work, but it seemed unnecessarily complex. Consider this snippet of GLSL:
```
layout(set = 3, binding = 1, std430) readonly buffer InputBuffer
{
    mediump float exposure_bloom;
} in_buf;
```
When converting the type to GLSL, the qualifiers determined from `TranslateMemoryDecoration` would be applied as decorations to each of the members (in `TGlslangToSpvTraverser::decorateStructType`). Each of the decorations identified in that function can suitably be applied to the OpVariable itself (in this case, `in_buf`). Therefore, I transferred similar logic to apply to the created variable.

I think that it would be possible to run such a post-processing step, but we would have to recursively transfer decorations, which didn't seem fun. For example, consider this:
```
layout(set = 3, binding = 1, std430) readonly buffer InputBuffer
{
    mediump float exposure_bloom;
} in_buf[6];
```

Before my change, the NonWritable decoration is applied to the members of InputBuffer. This needs to bubble up not only from the top struct (which cannot directly handle such a decoration- "Must be applied only to [memory object declarations](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#MemoryObjectDeclaration) or members of a structure type."), but through the encapsulated array too.

arcady-lunarg:
You're going to have to update the expected test results, using a command like `build/gtests/glslangtests --update-mode` and then added the changed files from `Test/baseResults/` to your PR.

jeffbolznv:
I'm not clear on what happens in SPIR-V if you decorate both the top-level variable and some member, and the two disagree.

I think this TranslateMemoryDecoration function also handles coherent/volatile sort of decorations, and these are supposed to be accumulated as you walk through the type hierarchy. Decorating the top-level variable seems like it will generate the wrong results. I think there should be test cases for this, but it might only be for when vulkan memory model is enabled which uses per-instruction modifiers.

mmoult:
@jeffbolznv Based on other SPIR-V decorations I have seen (such as how FPControls2 is handled), it seems reasonable to me that if the decorations of the top-level variable and its members disagree, the member decorations would take priority for each member. In other words, the top-level decorations should be like a default. Do you know someone specific we could ask to resolve this lack of clarity with the specification?

Yes, the `TranslateMemoryDecoration` function also handles coherent/volatile decorations. However, the description for all of `Volatile`, `Coherent`, `NonWritable` and `NonReadable` start with this snippet:

> Must be applied only to [memory object declarations](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#MemoryObjectDeclaration) or members of a structure type.

Therefore, I would assume that using them as decorations for the top-level `OpVariable` would be correct. Why do you believe that "decorating the top-level variable seems like it will generate the wrong results"? If it would generate the wrong result, should the SPIR-V spec be edited to change the wording?

This is the first time I have tried committing to this repository, so I don't know who to ask about things like this. I also figure that because an issue already exists for the enhancement I wanted, it is more likely to be legitimate and worth fixing. Thank you for your patience.

jeffbolznv:
I've opened an internal SPIR-V issue to clarify the question of whether member decorations replace or add to the top-level variable decorations. I think NVIDIA's implementation will "add", but I'm not totally sure.

I think I've convinced myself that coherent/volatile are not a problem. I was concerned about a case like:

```
layout(binding = 0) devicecoherent buffer X
{
    workgroupcoherent int y;
};
```

However, we don't actually have decorations for those, so it ends up not being an issue.

mmoult:
Did the internal SPIR-V issue give you the information you were looking for? Should we update the SPIR-V spec to include commentary on potentially conflicting decorations as discussed? I want to make sure this issue isn't forgotten.

jeffbolznv:
It's not fully resolved yet, but seems to be trending toward "unioning" the multiple decorations. So I think, at this point, don't let my concerns block this change.

mmoult:
I see. Thanks for the update.

Do you know who I should ping for formal reviews?

arcady-lunarg:
I am the one who does reviews, I'll take a look at this in the near future.

mmoult:
@arcady-lunarg May I request the current status?

arcady-lunarg (approved):
This seems reasonable to me.

PR #3661, opened 2024-07-16
location aliasing

jimihem:
when location aliasing, the aliases sharing the location must have the same underlying numerical type and bit width (floating-point or integer, 32-bit versus 64-bit, etc.) and the same auxiliary storage and interpolation qualification.

PR #3663, opened 2024-07-18
Add some comments and rename a helper function

arcady-lunarg:
This adds some documentation of the checkTypes() function to make clearer what exactly it is checking. A bit of a cleanup after #3661.


jeremy-lunarg (approved):
LGTM

PR #3671, opened 2024-07-22
Don't emit debug instructions before an OpPhi

### LINKED ISSUE #3665: Non-semantic OpExtInst is incorrectly placed before OpPhi in a block
I initially raised this issue for the validation layers, but it turns out this is actually a codegen issue. The original issue can be found [here](https://github.com/KhronosGroup/glslang/issues/3669). This was further clarified by the specification for [SPV_KHR_non_semantic_info](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_non_semantic_info.html), which describes precisely how this is invalid.

Below is mostly copy-pasted from the linked issue.

**Environment**
- glslang version: 14.3.0
- SDK version: 1.3.283.0

**The Issue**
I have a shader that has been compiled with glslang's C++ API with debug info. Debug info has been enabled through the following code:
```cpp
shader.setDebugInfo(true);
...
auto options = glslang::SpvOptions{
    .generateDebugInfo = true,
    .stripDebugInfo = false,
    .disableOptimizer = true,
    .emitNonSemanticShaderDebugInfo = true,
    .emitNonSemanticShaderDebugSource = true,
};
```
The full code for compiling shaders can be found [here](https://github.com/JuanDiegoMontoya/Frogfood/blob/main/src/Fvog/Shader2.cpp#L129-L209).

Attempting to create a shader module with the produced SPIR-V (attached at the end) results in this message from the validation layers:
> Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
OpPhi must appear within a non-entry block before all non-OpPhi instructions (except for OpLine, which can be mixed with OpPhi).
%2234 = OpPhi %bool %2187 %2160 %2233 %2189
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.3.283.0/windows/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)

The instruction that it's complaining about is located in this block:
```
[2002]  %2190 = OpLabel
[2003]  %2235 = OpExtInst %19 %1 DebugScope %388
[2004]  %2234 = OpPhi %146 (%2187 : %2160) (%2233 : %2189)
[2005]  OpSelectionMerge %2237 None
[2006]  OpBranchConditional %2234 %2236 %2237
```

**Expected Behavior**
OpExtInst should not be placed before OpPhi in a block.

**Additional Info**
The shader in question can be found here: https://github.com/JuanDiegoMontoya/Frogfood/blob/main/data/shaders/visbuffer/CullMeshlets.comp.glsl
The affected SPIR-V is attached in [bug.zip](https://github.com/user-attachments/files/16289870/bug.zip).

arcady-lunarg:
Nonsemantic instructions aren't allowed before an OpPhi, so don't emit line and debug scope instructions when the instruction being emitted is an OpPhi.

Fixes #3665


jeremy-lunarg (approved):
LGTM

PR #3672, opened 2024-07-23
Expose setGlobalUniform* methods to the C API

Akaricchi:


arcady-lunarg (approved):
Seems reasonable to me.

Issue #3668, opened 2024-07-22
HLSL struct layout with arrays broken since Apr-18


BartmanAbyss:
since 593dbafd0d4391cbc917d4246f94f870860a7cd5 (2024-04-18) @Svenny 

see this reduced test case:
```hlsl
struct GfxIterationData {
	float mIntegrationTrafo[24];
	float mWind[3];
	uint mIsTurning;
};

StructuredBuffer<GfxIterationData> bIterData : register(t2);
RWByteAddressBuffer bParticles : register(u0); //contains float4 data

[numthreads(1, 1, 1)] void main(uint blockIdx : SV_GroupID, uint threadIdx : SV_GroupThreadID) {
	bParticles.Store(0, bIterData[0].mIsTurning);
}
```

before said commit it generates this SPV for GfxIterationData
```
MemberName 30(GfxIterationData) 0  "mIntegrationTrafo"
MemberName 30(GfxIterationData) 1  "mWind"
MemberName 30(GfxIterationData) 2  "mIsTurning"
...
MemberDecorate 30(GfxIterationData) 0 Offset 0
MemberDecorate 30(GfxIterationData) 1 Offset 96
MemberDecorate 30(GfxIterationData) 2 Offset 108
```
after that commit
```
MemberDecorate 30(GfxIterationData) 0 Offset 0
MemberDecorate 30(GfxIterationData) 1 Offset 84
MemberDecorate 30(GfxIterationData) 2 Offset 84
```
which is obviously incorrect.

see SPIRV/GlslangToSpv.cpp line 5413

Svenny:
My bad, forgot to check non-cbuffer case when doing that patch. Should be fixed by #3677

PR #3677, opened 2024-07-29
Fix HLSL offsets for non-cbuffers (#3668)

Svenny:
Forgot to check that case when doing #3575


arcady-lunarg (approved):
LGTM, thanks for the quick fix.

PR #3678, opened 2024-07-29
ci: Add a list of known validation fails

arcady-lunarg:
Add a Test/baseResults/validation_fails.txt file to track test results with known validation fails. This will hopefully make it easier to spot when new test results are added that have invalid SPIR-V.


jeremy-lunarg (approved):
LGTM

PR #3682, opened 2024-07-30
add optimizerAllowExpandedIDBound to SpvOptions

mbechard:
This option increases the max_id_bound in the optimizer to 0x3FFFFFFF, and then runs a compact-id pass if the final result has a max_id greater than the standard limit of 0x3FFFFF.

### DIFF
--- a/SPIRV/GlslangToSpv.h
+++ b/SPIRV/GlslangToSpv.h
@@ -53,6 +53,7 @@ struct SpvOptions {
     bool emitNonSemanticShaderDebugInfo {false};
     bool emitNonSemanticShaderDebugSource{ false };
     bool compileOnly{false};
+    bool optimizerAllowExpandedIDBound{false};
 };
 
 void GetSpirvVersion(std::string&);
--- a/SPIRV/SpvTools.cpp
+++ b/SPIRV/SpvTools.cpp
@@ -218,9 +218,20 @@ void SpirvToolsTransform(const glslang::TIntermediate& intermediate, std::vector
     optimizer.RegisterPass(spvtools::CreateCFGCleanupPass());
 
     spvtools::OptimizerOptions spvOptOptions;
+    if (options->optimizerAllowExpandedIDBound)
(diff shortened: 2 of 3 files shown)


mbechard:
I was thinking this could alternatively have this as a number of max_id that is allowed, but I can't seem to find any way to query what is allowed on different implementations, even though the SPIR-V spec implies 0x3FFFFF is just the minimum required.
This PR implies some implementations can go higher:
https://github.com/microsoft/DirectXShaderCompiler/pull/6654

arcady-lunarg:
I think in practice, the implementation limit on ID bounds comes from when they try to allocate an array sized by the ID bound. I guess not that many people have been hitting that bound so far. I need to think a bit whether it's better to have this option as a flag or a numeric value.

mbechard:
The case that brought this change forward was a shader that is hitting the max_id_bound during the optimization passes that GLSLang applies. Specifically, it runs out of IDs during the SSARewritePass, which hits asserts. It then crashes in the IfConversion pass.
This change allows the optimizer to go beyond the ID limit while its working through its optimizations, but then brings the IDs down below the 0x3FFFFF limit so is a valid shader for all GPUs.
It would seem there are a situations at play here:
* The above one, where the optimizer can use more IDs during its passes, but needs to reduce it at the end.
* If the caller knows the implementation can handle higher IDs, just allow those to stay.
* Add a way to always apply compact-ids, since it seems like that is useful for general performance of the shaders.
* Add errors when the resulting shader is using too many IDs

arcady-lunarg:
Upon a bit more thought, I think your current approach is a reasonable one. I don't really want to always run the compact-ids pass, there's some value in manual debugging with keeping the IDs stable through optimization, but also that argument doesn't really apply when you have more than 4 million IDs. And when you have that many IDs it's probably best to actually compact them given the issues with allocation of arrays. Keeping all of this behind an option is a good idea, then only the people who really need it and understand the implications will enable it.

arcady-lunarg (approved):

PR #3683, opened 2024-07-30
Add SPIRV/SpvTools.h back to the public headers

arcady-lunarg:
Clean up the includes in SpvTools.h so that it doesn't require any transitive dependencies and add it back to the headers that get installed.


jeremy-lunarg (approved):
LGTM

PR #3685, opened 2024-08-03
Iomapper api cleanup

### LINKED ISSUE #3681: Change to CMake install rule for the include directory?
It seems like there are lots of missing headers in the install/include directory now. For example MachineIndependent seems to only come with Versions.h now, and is missing other useful headers such as iomapper.h.
SPIRV is missing SpvTools.h is another example of a useful header that's no longer copied.
Is there a new workflow to follow?

arcady-lunarg:
Clean up iomapper.h to make it suitable as a public API

Remove the dependency on LiveTraverser.h, which transitively includes localintermediate.h. As a result, some things have been moved from iomapper.h to iomapper.cpp, specifically the TVarEntry type and the constructor and destructor of TGlslIoMapper.

Add MachineIndependent/iomapper.h to the list of installed headers.

Fixes #3681

### DIFF
--- a/glslang/CInterface/glslang_c_interface.cpp
+++ b/glslang/CInterface/glslang_c_interface.cpp
@@ -37,7 +37,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "glslang/Public/ShaderLang.h"
 #include "glslang/Include/ShHandle.h"
 
+#include "glslang/Include/BaseTypes.h"
 #include "glslang/Include/ResourceLimits.h"
+#include "glslang/Include/Types.h"
 #include "glslang/MachineIndependent/iomapper.h"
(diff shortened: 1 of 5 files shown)


arcady-lunarg:
In the process of developing this, it turned out that there is a pre-existing array overrun bug that was not getting caught by the address sanitizer. I'm going to fix that before I merge this.

jeremy-lunarg (approved):
LGTM

mbechard:
This doesn't seem complete yet, as there are many types in iomapper.h that the includer doesn't have definitions for, such as TType, TString, TLayoutPacking, TStorageQualifier etc.

mbechard:
Also need access to `SpvVersion` that `TIntermediate` has an accessor for via `getSpv()`, for use in `glslang::MapToSpirvToolsEnv`. Possibly this should be held in the `TProgram`, to avoid needing to expose TIntermediate as anything more than an opaque type.

PR #3692, opened 2024-08-09
Add setSourceFile and addSourceText to TShader

### LINKED ISSUE #3691: How to set source information after the move to public headers?
In version 13, you could attach source information to the SPIRV glslang created with [setSourceFile](https://github.com/KhronosGroup/glslang/blob/7c4d91e7819a1d27213aa3499953d54ae1a00e8f/glslang/MachineIndependent/localintermediate.h#L1084) and [addSourceText](https://github.com/KhronosGroup/glslang/blob/7c4d91e7819a1d27213aa3499953d54ae1a00e8f/glslang/MachineIndependent/localintermediate.h#L1086).  This was useful when wanting to see the original source in a tool like RenderDoc.  [With the move of TIntermediate to the internal API](https://github.com/KhronosGroup/glslang/issues/3320), what is the way to handle this now in glslang and still use the public interface?

arcady-lunarg:
These are now accessible through the C++ ShaderLang.h public API as well as the C API.

Fixes #3691


jeremy-lunarg (approved):
LGTM

PR #3697, opened 2024-08-13
ci: Get rid of redundant testing

arcady-lunarg:
It's not necessary to run the runtest tests separately, as they're already being run by the ctest testing process (as glslang-testsuite).


jeremy-lunarg (approved):
LGTM

PR #3698, opened 2024-08-14
Merge ancillary libraries into main glslang library and stub originals

daniel-story:
This PR integrates the contents of the existing GenericCodeGen, MachineIndependent, OSDependent, and SPIRV libraries directly into the main glslang library. The original libraries are turned into stubs to be retained for backwards compatibility for a temporary transition period, similar to how the deprecation of the OGLCompiler and HLSL libraries was handled in the past.

The motivation for this change is similar to the motivation for the past deprecation and removal of the OGLCompiler and HLSL libraries. This change will simplify build environments for downstream users by providing them with a "one-stop-shop" glslang static library, and it also lays a foundation for future formal support for a glslang.dll/so dynamic library.


arcady-lunarg (approved):
LGTM, the failure on kokoro android is caused by a simple typo on Android.mk that I will fix in a followup change.

PR #3703, opened 2024-08-20
Move definition of GLSLANG_EXPORT to visibility.h

arcady-lunarg:
The GLSLANG_EXPORT macro is already used (and defined) in two separate places and will be used in a few more in the near future. To reduce duplication it is now in its own header.


jeremy-lunarg (approved):
LGTM

PR #3705, opened 2024-08-23
Add symbol visibility annotations to symbols in SPIRV/

### LINKED ISSUE #2308: Determine libSPIRV target 'public' interface
As [part of the work](https://github.com/KhronosGroup/glslang/issues/2283) to apply Semantic Versioning to the glslang project, we need to determine what we consider as part of the public, stable API and what is internal.

The [Semantic Versioning introduction](https://semver.org/#introduction) states:

> For this system to work, you first need to declare a public API. This may consist of documentation or be enforced by the code itself. Regardless, it is important that this API be clear and precise. Once you identify your public API, you communicate changes to it with specific increments to your version number. 

For the `glslang` target we have been able to classify [`glslang/Public/ShaderLang.h`](https://github.com/KhronosGroup/glslang/blob/master/glslang/Public/ShaderLang.h) as the public API. This is now (#2303) enforced through defaulting to `-fvisibility=hidden` and only making public the methods and types annotated with `GLSLANG_EXPORT`.

`libSPIRV` is another library that [is already found](https://www.archlinux.org/packages/extra/x86_64/glslang/) in linux distribution packages, but AFAICT, there is no clear separation between the public and private interface. As such, any changes to the [`SPIRV`](https://github.com/KhronosGroup/glslang/tree/master/SPIRV) source files will likely require a major semantic version bump, which severely reduces the benefits for semantic versioning.

I see a few options:

1. We spend the time to try classifying the public interface, and build a `SPIRV/Public/SPIRV.h` file, similar to `glslang`.

2. Remove `libSPIRV` from the 'public' interface.
It's hard to know how many people out there are using the `libSPIRV` target - but one potential solution is to make this an entirely 'private' library, and ensure this is statically linked into the `glslangValidator` executable, removing it from the install target. This would remove it from existing linux distributions, breaking anyone who is currently depending on it.

3. We simply ignore semantic versioning for `libSPIRV`. Breaking API changes to `libSPIRV` are not considered for major version bumping. Semantic versioning will only apply for the `glslang` target.

Feedback and opinions welcome.

arcady-lunarg:
Symbols in the SPIRV/ directory that are meant to be part of the public API of glslangnow have GLSLANG_EXPORT annotations that make sure they will be accessible when building glslang as a shared library if -fvisibility=hidden is enabled.

Fixes #2308

### DIFF
--- a/SPIRV/Logger.h
+++ b/SPIRV/Logger.h
@@ -37,6 +37,7 @@
 
 #include <string>
 #include <vector>
+#include "glslang/Include/visibility.h"
 
 namespace spv {
 
@@ -58,7 +59,7 @@ public:
 
     // Returns all messages accumulated in the order of:
     // TBD functionalities, missing functionalities, warnings, errors.
-    std::string getAllMessages() const;
+    GLSLANG_EXPORT std::string getAllMessages() const;
 
 private:
     SpvBuildLogger(const SpvBuildLogger&);
--- a/SPIRV/GlslangToSpv.h
+++ b/SPIRV/GlslangToSpv.h
@@ -39,6 +39,7 @@
 #include <vector>
 
 #include "Logger.h"
+#include "glslang/Include/visibility.h"
 
 namespace glslang {
 class TIntermediate;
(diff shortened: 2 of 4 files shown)


jeremy-lunarg (approved):
LGTM

SPIRV/Logger.h
```diff
@@ -58,7 +59,7 @@ class SpvBuildLogger {
 
     // Returns all messages accumulated in the order of:
     // TBD functionalities, missing functionalities, warnings, errors.
-    std::string getAllMessages() const;
+    GLSLANG_EXPORT std::string getAllMessages() const;
```
jengelh, line 62: Exporting only individual class member functions is problematic. A class has additional metadata and/or functions, some of which may be inlined, not inlined, or implicitly generated (including, but not limited to, constructors, destructors, typeinfo, virtual tables, template instantiations, etc.). Anything implicitly generated ought to be exported at all times to facilitate runtime symbol collapsing, otherwise one will run into ODR violations.

Classes themselves should be marked GLSLANG_EXPORT IMO, and if any member functions need hiding, they should be GLSLANG_HIDDEN (to counter the class-level GLSLANG_EXPORT).

arcady-lunarg: Thanks for the feedback. Since I will be doing more improvements to symbol visibility and actually enabling building with hidden visibility, I will fix this before the next release.

PR #3709, opened 2024-08-27
cmake: Fix dependencies of newly-stubbed libraries

arcady-lunarg:
Some downstream projects may only link to e.g. the glslang::SPIRV library and rely on transitive dependencies to pull in everything else. To keep this working, the SPIRV and MachineIndependent library stubs are now linked against the glslang library.


arcady-lunarg:
@jeremyg-lunarg this should fix the issue you were seeing in VVL.

jeremyg-lunarg:
> @jeremyg-lunarg this should fix the issue you were seeing in VVL.

This does indeed fix my problem. Thanks!

abouvier:
Are you sure about the `PRIVATE` visibility? Because `glslang::glslang` is still not pulled when only `glslang::SPIRV` is linked. As shown [here](https://github.com/cemu-project/Cemu/issues/1435) for example. The visibility should probably be `INTERFACE` (or `PUBLIC`?).

PR #3714, opened 2024-09-02
ci: Remove redundant testing from continuous deployment workflow

arcady-lunarg:
The runtests script is already run as part of the cmake-controlled testing, so can be removed from the CD workflow like it was from the CI ones.


jeremy-lunarg (approved):
LGTM

PR #3720, opened 2024-09-10
Convert keyword maps to static initializers

arcady-lunarg:
Rather than having a function that needs to be called from ShInitialize to initialize some global costants, convert them to use static initializers. This also ends up saving a bit of space in the binary, especially when building with GCC.

### DIFF
--- a/glslang/HLSL/hlslScanContext.cpp
+++ b/glslang/HLSL/hlslScanContext.cpp
@@ -78,414 +78,396 @@ struct str_hash
 };
 
 // A single global usable by all threads, by all versions, by all languages.
-// After a single process-level initialization, this is read only and thread safe
-std::unordered_map<const char*, glslang::EHlslTokenClass, str_hash, str_eq>* KeywordMap = nullptr;
-std::unordered_set<const char*, str_hash, str_eq>* ReservedSet = nullptr;
-std::unordered_map<const char*, glslang::TBuiltInVariable, str_hash, str_eq>* SemanticMap = nullptr;
+const std::unordered_map<const char*, glslang::EHlslTokenClass, str_hash, str_eq> KeywordMap {
+    {"static",glslang::EHTokStatic},
+    {"const",glslang::EHTokConst},
+    {"unorm",glslang::EHTokUnorm},
+    {"snorm",glslang::EHTokSNorm},
+    {"extern",glslang::EHTokExtern},
+    {"uniform",glslang::EHTokUniform},
+    {"volatile",glslang::EHTokVolatile},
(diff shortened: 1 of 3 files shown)


glslang/HLSL/hlslScanContext.cpp
```diff
@@ -494,49 +476,41 @@ void HlslScanContext::fillInKeywordMap()
...
-    (*SemanticMap)["SV_GROUPTHREADID"] =          EbvLocalInvocationId;
-    (*SemanticMap)["SV_GROUPINDEX"] =             EbvLocalInvocationIndex;
-    (*SemanticMap)["SV_GROUPID"] =                EbvWorkGroupId;
-    (*SemanticMap)["SV_DOMAINLOCATION"] =         EbvTessCoord;
-    (*SemanticMap)["SV_DEPTH"] =                  EbvFragDepth;
-    (*SemanticMap)["SV_COVERAGE"] =               EbvSampleMask;
-    (*SemanticMap)["SV_DEPTHGREATEREQUAL"] =      EbvFragDepthGreater;
-    (*SemanticMap)["SV_DEPTHLESSEQUAL"] =         EbvFragDepthLesser;
-    (*SemanticMap)["SV_STENCILREF"] =             EbvFragStencilRef;
+#if 0
```
jeremy-lunarg, line 479: Did you intend to remove this block of code?

jeremy-lunarg (approved):
I left a question about a block of dead code but otherwise looks good.

glslang/HLSL/hlslScanContext.cpp
```diff
@@ -494,49 +476,41 @@ void HlslScanContext::fillInKeywordMap()
...
     // stage, it would just be ignored as it is likely there as part of an output struct from one stage
     // to the next
-    bool bParseDX9 = false;
-    if (bParseDX9) {
-        (*SemanticMap)["PSIZE"] = EbvPointSize;
-        (*SemanticMap)["FOG"] =   EbvFogFragCoord;
-        (*SemanticMap)["DEPTH"] = EbvFragDepth;
-        (*SemanticMap)["VFACE"] = EbvFace;
-        (*SemanticMap)["VPOS"] =  EbvFragCoord;
-    }
```
arcady-lunarg, line 504: This is the block of code I am preserving with `#if 0`

glslang/HLSL/hlslScanContext.cpp
arcady-lunarg: It was de-facto commented out in the original, so I am preserving that.

PR #3721, opened 2024-09-11
Remove redundant calls to spv::Parameterize() from tests

arcady-lunarg:
This means the tests won't need to have this entrypoint exposed when building as a shared library with hidden symbols.


jeremy-lunarg:
Huh. Is this redundant because the disassembler does it?

arcady-lunarg:
Yeah. And it's the first thing that `Disassemble()` does.

jeremy-lunarg (approved):
LGTM

Issue #3716, opened 2024-09-03
Windows shared library build fails with LNK1104: cannot open file '..\glslang\Debug\glslang-default-resource-limitsd.lib'


AnyOldName3:
I just tried a DLL build of glslang (passed `-DBUILD_SHARED_LIBS` to CMake), and hit three fatal errors. Two were `fatal error LNK1104: cannot open file '..\glslang\Debug\glslang-default-resource-limitsd.lib'` and one was `fatal error LNK1104: cannot open file '..\SPIRV\Debug\SPVRemapperd.lib'`. The files are specified with the `/IMPLIB` flag for the linker commands that would be expected to generate them.

I've had a look at the DLLs with this tool https://github.com/lucasg/Dependencies and it seems that they don't actually export any symbols. MSVC doesn't bother to emit an import library when a DLL exports no symbols. You can make GCC behave like MSVC here (and shrink binary size) by passing `-fvisibility=hidden` and marking everything as `__attribute__((__visibility__("default")))` if it'd be `__declspec(dllexport)` with MSVC.

It's possible that some `GLSLANG_EXPORT` macro calls are missing and symbols are supposed to be exported but aren't - after all, GCC and Clang both export all symbols by default, so it's easy to miss. If no functions are exported, linking with these DLLs won't do anything other than running their static initialisers and destructors, and none of their functions will be callable. Looking at the headers for these libraries, it certainly *looks* like there are functions that should be callable.

When this is fixed, I'd suggest adding a Windows DLL build to the CI so it doesn't get broken again.

arcady-lunarg:
Now that we've factored out `visibility.h` into its own header, this should be pretty straightforward to add. This also needs to be done for the SPV remapper, especially because there are global symbols that clash between SPVRemapper and glslang libraries and they get linked together into the `glslangtests` binary.

arcady-lunarg:
@AnyOldName3 could you give it a try with #3725? That should fix the issues with those libraries in particular, though I'm not entirely sure that glslang.dll will be exporting the correct set of symbols but we're working on getting everything exported correctly so that we can have a fully functional shared library build not just on unix but on windows as well.

arcady-lunarg:
Thanks, I will have fixes for the other stuff shortly, just need to do a
bit more testing before I push it.

arcady-lunarg:
[...]

PR #3722, opened 2024-09-12
Add symbol visibility annotations to glslang-resource-limits library

arcady-lunarg:
Among other things, this allows using it as a DLL on Windows because unlike Linux, the default on windows is hidden visibility unless things are explicitly exported.


arcady-lunarg:
This should address #3716

jeremy-lunarg (approved):
LGTM

PR #3723, opened 2024-09-12
Allow compute derivative modes when the workgroup dimensions come from spec constants

jeffbolznv:
The spec doesn't explicitly say whether this is allowed or not, but I think it's clearly the intent that it should be possible to use derivatives with spec constant workgroup size.


arcady-lunarg (approved):

jeffbolznv:
Thanks for merging. I just noticed that since this is doing the checks as it parses the shader, whether you get an error depends on the order of setting the derivative layout vs setting the workgroup dimensions (spec constant or not). This is a pre-existing issue and not a huge deal for me, but maybe worth fixing eventually.

arcady-lunarg:
Could you create an issue about that so we don't forget about it?

PR #3725, opened 2024-09-12
Explicitly export symbols from libSPVRemapper

### LINKED ISSUE #3716: Windows shared library build fails with LNK1104: cannot open file '..\glslang\Debug\glslang-default-resource-limitsd.lib'

arcady-lunarg:
Windows requires symbols to be explicitly exported from DLLs, this change simply marks the entire spirvbin_t class as exported.

The macros from visibility.h are included inline to get around include path issues.

Fixes #3716

### DIFF
--- a/SPIRV/CMakeLists.txt
+++ b/SPIRV/CMakeLists.txt
@@ -100,6 +100,7 @@ if (ENABLE_SPVREMAPPER)
         POSITION_INDEPENDENT_CODE ON
         VERSION   "${GLSLANG_VERSION}"
         SOVERSION "${GLSLANG_VERSION_MAJOR}")
+    glslang_only_export_explicit_symbols(SPVRemapper)
 endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)
--- a/SPIRV/SPVRemapper.h
+++ b/SPIRV/SPVRemapper.h
@@ -41,6 +41,21 @@
 #include <cstdlib>
 #include <exception>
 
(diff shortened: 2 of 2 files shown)


jeremy-lunarg (approved):
LGTM

PR #3726, opened 2024-09-13
Improved iomapper api

### LINKED ISSUE #3681: Change to CMake install rule for the include directory?

arcady-lunarg:
The TProgram::mapIO method takes a TIoMapResolver and a TIoMapper, however the only way to obtain instances of these was from methods in iomapper.h. Rather than try to expose that header as part of the API, new methods are added to the public ShaderLang.h header to create a TDefaultGlslIoResolver and a TGlslIoMapper and return them as pointers to their respective base classes, which are defined in the public header.

Also update the tests to use these new public entry points, rather than relying on private ones.

Fixes #3681

### DIFF
--- a/glslang/CMakeLists.txt
+++ b/glslang/CMakeLists.txt
@@ -262,7 +262,6 @@ if(GLSLANG_ENABLE_INSTALL)
         Include/glslang_c_interface.h
         Include/glslang_c_shader_types.h
         Include/ResourceLimits.h
-        MachineIndependent/iomapper.h
         Include/visibility.h
         MachineIndependent/Versions.h)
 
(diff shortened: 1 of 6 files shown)


arcady-lunarg:
cc @mbechard This PR includes the new entry points for getting the IO mapper and resolver.

jeremy-lunarg (approved):
LGTM

PR #3729, opened 2024-09-15
Expose setResourceSetBinding() method to the C API

Akaricchi:
This adds `glslang_shader_set_resource_set_binding()` to the C API as a counterpart to `setResourceSetBinding()`. I need this to set the descriptor set for automatically mapped samplers when compiling legacy OpenGL shaders in Vulkan mode with relaxed semantics.


arcady-lunarg (approved):
LGTM

PR #3732, opened 2024-09-18
Final symbol visibility fixes

arcady-lunarg:
This should be the last of the symbol visibility fixes, allowing building and testing glslang as a shared library.


jeremy-lunarg (approved):
LGTM

PR #3734, opened 2024-09-20
Fix SPIR-V support in GN build

ShabbyX:
The cmake build recently got ENABLE_SPIRV which is enabled by default. This was missing from the GN build, accidentally removing SPIR-V support in this build.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -235,7 +235,7 @@ template("glslang_sources_common") {
       ]
     }
 
-    defines = []
+    defines = [ "ENABLE_SPIRV=1" ]
     if (invoker.enable_opt) {
       sources += [ "SPIRV/SpvTools.cpp" ]
       defines += [ "ENABLE_OPT=1" ]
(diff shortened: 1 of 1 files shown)


ShabbyX:
The worst part of this was that glslang would succeed but not generate any errors (or output). @daniel-story who added this flag in 4f01996c9df0e2edb8823a8260066467c828a341, it would be nice if you made glslang complain if glslang is asked to produce SPIR-V but ENABLE_SPIRV is not defined.

arcady-lunarg (approved):

arcady-lunarg:
@ShabbyX would it be possible to get at least one GN configuration in CI, preferably in kokoro, so that this doesn't happen again?

ShabbyX:
> @ShabbyX would it be possible to get at least one GN configuration in CI, preferably in kokoro, so that this doesn't happen again?

I hope so? I believe something similar was done for VVL, so other LunarG folks would know what to do. I don't believe I have any access to do anything myself.

FWIW, build was fine, and glslang didn't fail compilation or anything. It just silently didn't produce any output. A build test would not have caught it.

PR #3736, opened 2024-09-20
Print an error when SPIR-V is requested and ENABLE_SPIRV=0 is set

arcady-lunarg:
This case fails slightly less silently now.
Also make ENABLE_OPT depend on ENABLE_SPIRV, since when there's no SPIR-V optimization doesn't really make sense.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -1507,9 +1507,9 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
 
     std::vector<std::string> outputFiles;
 
-#ifdef ENABLE_SPIRV
     // Dump SPIR-V
     if (Options & EOptionSpv) {
+#ifdef ENABLE_SPIRV
         CompileOrLinkFailed.fetch_or(CompileFailed);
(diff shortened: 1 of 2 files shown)


arcady-lunarg:
@ShabbyX does this more or less address the issue you raised in #3734?

StandAlone/StandAlone.cpp
```diff
@@ -1569,8 +1569,10 @@ void CompileAndLinkShaderUnits(std::vector<ShaderCompUnit> compUnits)
                 }
             }
         }
-    }
+#else
+        Error("This version of glslang does not have SPIR-V support");
```
jeremy-lunarg, line 1573: s/version/configuration sounds slightly more appealing to me

jeremy-lunarg (approved):
Small nit-pick but otherwise LGTM

ShabbyX:
> @ShabbyX does this more or less address the issue you raised in #3734?

Yes thank you! And +1 to Jeremy's wording suggestion

StandAlone/StandAlone.cpp
arcady-lunarg: Good point, updated.

PR #3737, opened 2024-09-21
Use line and column from node on DebugLexicalBlock.

sajjadmirzanv:
Old implementation had every DebugLexicalBlock hardcoded to (0, 0), but IR nodes know their source location so we may as well attach it to the DebugLexicalBlock.


arcady-lunarg:
You should add a test that exercises this functionality.

sajjadmirzanv:
> You should add a test that exercises this functionality.

I discovered that there was a bug in the test runner, which is why the existing nonsemantic tests were not already triggering this code. Running glslang -gVS on those files emits DebugLexicalLine, but it was missing when run through glslangtests. 

In TestFixture.h, the code was only generating debug info for the 2nd stage of compilation, from IR -> SPIRV. I enabled the flag so it also emits debug info during Source Language -> IR compilation. The result is that we get EOpScope nodes in the tree, so the GlslangToSpv converter can emit DebugLexicalBlock instructions. 

I had to make a few other small changes to some tests as well.

gtests/TestFixture.h
```diff
@@ -262,6 +262,14 @@ class GlslangTest : public GT {
             shader.setAutoMapLocations(true);
             shader.setAutoMapBindings(true);
         }
+
+        if (enableDebug) {
+            shader.setDebugInfo(true);
+        }
+        if (enableNonSemanticShaderDebugInfo) {
```
arcady-lunarg, line 269: I like this forcing of correct usage. We might want to force `enableNonSemanticShaderDebugInfo` to set `enableDebug` at a lower level, but I will investigate that in a separate PR.

arcady-lunarg (approved):
LGTM, I think this looks reasonable.

PR #3740, opened 2024-09-23
Update CHANGES for 15.0.0

jeremy-lunarg:


CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
 
+## 15.0.0 2024-09-23
+### Breaking changes
+* Symbol visibility across all libraries have been sanitized
+
+### Other changes
+* Report error if generating SPIR-V for misconfigured glslang
+* Add setResourceSetBinding method to the API
+* Add interface to get the GLSL IO mapper and resolver
+* Explicitly export symbols from libSPVRemapper
```
arcady-lunarg, line 14: I would condense all the symbol visibility changes to one point at put it at the top.
Something like:
Explicitly export all symbols that are part of the public API. Building glslang as a DLL or shared library is now possible and supported.
Maybe we should even put something in the README?

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Add symbol visibility annotations to glslang-resource-limits library
+* Allow compute derivative modes when the workgroup dimensions are spec constants
+* Improve debug location of branch/return instructions
+* Silence preprocessor '#' error reporting for inactive #if/#ifdef/#elif/#else
```
arcady-lunarg, line 18: "error reporting in inactive #if/#else/... blocks" perhaps?

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Apply GLSL memory decorations to top-level OpVariable
+* Move definition of GLSLANG_EXPORT to visibility.h
+* Merge ancillary libraries into main glslang library and stub originals
+* Add setSourceFile and addSourceText to TShader
```
arcady-lunarg, line 22: Add *public* setSourceFile and addSourceText methods to TShader *class*

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Add type checks for hitObjectNV
+* Fix initialization of arrays in TGlslIoMapper
```
arcady-lunarg, line 24: Internal fix, I think there's no need to call this out.

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Add optimizerAllowExpandedIDBound to SpvOptions
+* Add SpvTools.h back to public headers 
+* Add cross-stage check for missing outputs
+* Fix HLSL offsets for non-buffers
+* Add types and functions for IO mapping to API
+* Add function to set preprocessed code to API
+* Add set/get version functions to API
+* Add missing calls SetThreadPoolAllocator
```
arcady-lunarg, line 32: Internal fix, not sure if this needs to be called it. If you call it out, probably best to say where.

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Expose setGlobalUniform functions to API
+* Don't emit debug instructions before an OpPhi
+* Add command-line and API option to enable column location
+* Improve location aliasing checks
+* Fix undefined behaviors caused by integer overflows
```
arcady-lunarg, line 37: Internal fix

CHANGES.md
```diff
@@ -3,6 +3,47 @@
...
+* Support constant expression calculated by matrixCompMult
+* Fix crash caused by atomicCounter() use without arguments
+* Fix undefined behavior exposed by unscoped enumerated types
```
arcady-lunarg, line 40: Internal fix

CHANGES.md
```diff
@@ -3,6 +3,47 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## 15.0.0 2024-09-23
+### Breaking changes
+* Symbol visibility across all libraries have been sanitized
+
+### Other changes
+* Report error if generating SPIR-V for misconfigured glslang
```
arcady-lunarg, line 11: I would actually call out the main thing here:
* Allow building glslang without the SPIR-V backend using the new ENABLE_SPIRV build option.

The error thing isn't a change from the previous released version (which didn't allow building without SPIR-V at all).

arcady-lunarg (changes requested):
Thanks for putting this together, I have a few suggestions.

CHANGES.md
jeremy-lunarg: `Explicitly export all symbols that are part of the public API`

I'll put that in the breaking changes section instead of what I have.

`Building glslang as a DLL or shared library is now possible and supported.`

That part sounds like it should go in the README. I'll add that to my PR as well.

CHANGES.md
jeremy-lunarg: I'll remove it.

CHANGES.md
jeremy-lunarg: I'll remove but only added to highlight improved stability.

jeremy-lunarg:
Thanks @arcady-lunarg. Please take another look.

jeremy-lunarg:
@arcady-lunarg, do you think we should remove any of the old bullets from `News` section of the README?

CHANGES.md
```diff
@@ -3,6 +3,41 @@
...
+### Breaking changes
+* Explicitly export all symbols that are part of the public API and hide other symbols by default
+
+### Other changes
+* Allow building glslang without the SPIR-V backend using the new ENABLE_SPIRV build option
+* Add setResourceSetBinding method to the API
+* Add interface to get the GLSL IO mapper and resolver
+* Allow compute derivative modes when the workgroup dimensions are spec constants
+* Improve debug location of branch/return instructions
+* Silence preprocessor '#' error reporting in inactive #if/#ifdef/#elif/#else
```
arcady-lunarg, line 16: Could you add the word "blocks" at the end?

CHANGES.md
```diff
@@ -3,6 +3,41 @@
...
+* Apply GLSL memory decorations to top-level OpVariable
+* Move definition of GLSLANG_EXPORT to visibility.h
+* Merge ancillary libraries into main glslang library and stub originals
+* Add public setSourceFile and addSourceText to TShader
```
arcady-lunarg, line 20: "methods to TShader" (or even "methods to TShader class")

arcady-lunarg (changes requested):
A couple more nits, otherwise I think this is fine.

arcady-lunarg (approved):
LGTM, thanks.

PR #3741, opened 2024-09-25
A few more tweaks for #3681

mbechard:
expose TIoMapper::setAutoPushConstantBlock()
add overload for MapToSpirvToolsEnv() that takes TIntermediate instead of SpvVersion.


arcady-lunarg (approved):
LGTM, thanks

PR #3743, opened 2024-09-25
WIP: Fix nonsemantic debuginfo line attribution for cooperative matrix

jeffbolznv:
I got a bug report that coopMatMulAdd is being attributed to the wrong source line (see https://github.com/KhronosGroup/glslang/commit/24fc8fd14fa0439db931ce0457a13a2cb7a6df6a#diff-193931d1256437fa2080564013b33c48a7a6abab57180a985a9c53edb6509fdfR518, line 518). The fix for that is pretty straightforward, but while fixing this I realized that other support for coopmat types is missing and I'm not sure how to resolve that.


SPIRV/SpvBuilder.cpp
```diff
@@ -439,6 +439,14 @@ Id Builder::makeCooperativeMatrixTypeKHR(Id component, Id scope, Id rows, Id col
     constantsTypesGlobals.push_back(std::unique_ptr<Instruction>(type));
     module.mapInstruction(type);
 
+    if (emitNonSemanticShaderDebugInfo)
+    {
+        // There's no nonsemantic debug info instruction for cooperative matrix types.
```
jeffbolznv, line 444: I'm not sure how we're supposed to deal with a type instruction for a type that nonsemantic debuginfo doesn't know about. DebugInfoNone sounds like it's supposed to be usable for "unknown" things, but spir-val complains about this. It also complains about using OpSpecConstant as an arraydim, which the spec doesn't appear to consider. If I leave the debuginfo out, glslang will assert later.

@baldurk any suggestions for how to deal with this?

baldurk: I'm not familiar with what co-operative matrices are but could you just treat them as normal matrices? ie. use `DebugTypeMatrix` and not encode any extra information about them? Then it would be a question for anyone who wanted to consume debug information about co-operative matrices if they need any of that extra information and add something to represent it.

Not allowing spec constants for things sounds like an oversight, anywhere the spec accepts a constant it should accept spec constants too because they kind of are just constants and the idea (as I understand it) is that they should always be accepted everywhere.

jeffbolznv: DebugTypeMatrix isn't a great fit because normal matrices are treated like arrays of vectors, but cooperative matrices specify a component type and two dimensions. So there's no vector type to put in the instruction.

What are your thoughts on using DebugInfoNone for types? The spec says "Other instructions can refer to this one in case the debugging information is unknown, not available, or not applicable." It doesn't explicitly say whether this is allowed for types or not.

baldurk: I don't think there's anything to stop you making a vector of whichever size you'd like (unlike base SPIR-V), but yeh maybe it's not the best fit.

DebugInfoNone is used in specific places in the spec when called out, it's not a general use thing that can be used anywhere as otherwise that makes the spec a pain to work with if any random bits of data can be NULL.

Maybe if you could represent it as an opaque object - see the bottom of `DebugTypeComposite` where it mentions having no members, similar to a `struct Foo; Foo *foo;` type thing.

arcady-lunarg: I would vote for an opaque object as well, I think that makes the most sense here.

jeffbolznv: ok, I'll give this a try. Cooperative matrices do act like composite types in some ways, so it's a reasonable thing to do.

jeffbolznv: Had some trouble with line endings in the OpSource, but I think this should pass CI now and is ready for review.

arcady-lunarg:
@jeffbolznv Looks like you're trying to check in a test result with a "Validation failed" message. Is there any way to resolve that validation failure? You can also add your new test to Test/baseResults/validation_fails.txt but I'd prefer not to do that since we really shouldn't be generating invalid SPIR-V.

arcady-lunarg:
> error: NonSemantic.Shader.DebugInfo.100 DebugTypeArray: Component Count must be OpConstant with a 32- or 64-bits integer scalar type or DebugGlobalVariable or DebugLocalVariable with a 32- or 64-bits unsigned integer scalar type
  %90 = OpExtInst %void %1 DebugTypeArray %84 %C_COLS

Where `%C_COLS` is defined as `%C_COLS = OpSpecConstantOp %uint UDiv %TILE_N %lN`. Seems like the nonsemantic debuginfo spec doesn't allow this. Should it?

jeffbolznv:
> Where `%C_COLS` is defined as `%C_COLS = OpSpecConstantOp %uint UDiv %TILE_N %lN`. Seems like the nonsemantic debuginfo spec doesn't allow this. Should it?

This is the spec issue discussed above, and Baldur says (https://github.com/KhronosGroup/glslang/pull/3743#discussion_r1775290805) it should probably be supported. But it's a real change to the extension and I'm not sure if it could cause problems.

jeffbolznv:
I think the issue I originally intended to fix I can reproduce without the spec constants, so I'll try to do that.

jeffbolznv:
@arcady-lunarg I think this is ready.

arcady-lunarg (approved):
LGTM

PR #3764, opened 2024-10-16
Fix printing of 'unknown stage' during crossStageCheck

rj123-nv:
Adds a 'stage' field to symbol nodes to remember what stage a symbol was
defined in.
Previously, 'unknown' stage was printed because linker object symbols are
merged into a TIntermediate with language=EShLangCount while cross stage
rules are checked. Storing the stage in the symbol fixes this problem.

Fixes errors such as 

WARNING: Linking unknown stage and fragment stages: Matched shader interfaces are using different instance names.
    unknown stage stage: Block: crossStageBlock2 Instance: blockName1: ""
    fragment stage: Block: crossStageBlock2 Instance: blockName2: ""


glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -59,10 +60,12 @@ namespace glslang {
 void TIntermediate::error(TInfoSink& infoSink, const char* message, EShLanguage unitStage)
```
rj123-nv, line 60: This could be changed to accept another "stage" argument, but that would require changing a bunch of callers which I opted not to do.
If getStage() is EShLangCount, as in the case of the intermediate created for merging linker objects, just opt to only print "unitStage"

jeffbolznv:
@arcady-lunarg can you please help review?

arcady-lunarg (approved):
Looks reasonable, and the messages are an improvement. I think you'll need to rebase this before it can be merged.

rj123-nv:
Thanks for reviewing. I rebased the changes

PR #3768, opened 2024-10-18
ci: Update macOS runner to macOS-latest

charles-lunarg:
That way it doesn't need to be updated manually in the future.

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -100,7 +100,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [{genus: macos-13, family: osx}]
+                os: [{genus: macos-latest, family: osx}]
                 compiler: [{cc: clang, cxx: clang++}]
                 cmake_build_type: [Debug, Release]
         steps:


arcady-lunarg:
Are there any compatibility issues with binaries built on latest running on older macos versions?

charles-lunarg:
Good question - I am not too sure.  The macos-latest is running arm64 which may produce binaries that can't be run on x64 only macOS versions. But I cannot say what the incompatibilities between macos-13, 14, and 15 would be. @richard-lunarg may have a more concrete understanding of the situation.

richard-lunarg:
The environment variable MACOSX_DEPLOYMENT_TARGET will affect which features of the shared runtime can be used. We've been setting this to 11.0 and the resulting binaries run all the way back to macOS 11. There's some environment variables Xcode will honor. This for example will build multi-arch binaries:
ARCHS="x86_64 arm64"
ONLY_ACTIVE_ARCH=NO


cmake will likewise honor:
CMAKE_OSX_ARCHITECTURES="x86_64;arm64"

arcady-lunarg:
In that case, I think we need to a) switch this to macos-latest as proposed here and b) set the deployment target, and c) enable the multiarch options as well.

jeremy-lunarg:
@charles-lunarg Do you have time to implement Arcady's suggestions on this PR?

jeremy-lunarg:
Thanks @charles-lunarg and @richard-lunarg

PR #3771, opened 2024-10-24
python 3.12: remove deprecated datetime.utcfromtimestamp

selfisekai:


build_info.py
```diff
@@ -139,7 +139,7 @@ def describe(directory):
             # clock time with environment variable SOURCE_DATE_EPOCH
             # containing a (presumably) fixed timestamp.
             timestamp = int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
-            formatted = datetime.datetime.utcfromtimestamp(timestamp).isoformat()
+            formatted = datetime.datetime.fromtimestamp(timestamp, datetime.UTC).isoformat()
```
arcady-lunarg, line 142: I think you need to use datetime.timezone.utc here, as datetime.UTC was only added in 3.11, and I'd prefer not to break compatibility with older Pythons over something so trivial.

arcady-lunarg (changes requested):
One minor issue, otherwise this is fine. Thanks for noticing this.

build_info.py
selfisekai: done

arcady-lunarg (approved):

PR #3772, opened 2024-10-24
Simplify glslang grammar for identifier lists

gnl21:
This was more complicated than it needed to be. The old grammar matched what was in the spec and GLSL#251 fixes it there as well.


gnl21:
@arcady-lunarg here is one of the grammar changes to match the GLSL PRs that you requested.

gnl21:
I've added a new commit because the error messages in the test suite have changed. They have become slightly less helpful, I think, because given
```
uniform foo
```
a LEFT_BRACE is a possible continuation (to declare a uniform block `foo`) but this isn't mentioned in the message anymore. I assume this is because bison has already reduced the identifier to an identifier_list, which rules out this completion.

Overall I think that this illustrates a short-coming with bison error handling rather than anything that should be adjusted in this change. Most of the occurrences in the test suite are where this is generated by having misunderstood some previous token anyway, so neither the old or the new messages make sense.

gnl21:
The spec PR for this change has just been merged.

arcady-lunarg (approved):
LGTM, thanks.

arcady-lunarg:
This will need to be rebased because of an intervening change to glslang.y, feel free to do that or to give maintainers access to push to your branch and I would be able to do it myself.

dnovillo:
@gnl21 are you still interested in merging this?

PR #3773, opened 2024-10-24
Implement GLSL_NV_cooperative_matrix2

jeffbolznv:


jeffbolznv:
@arcady-lunarg can you take a look? I don't know how closely you looked at the internal version.

glslang/MachineIndependent/glslang.y
```diff
@@ -3535,6 +3536,20 @@ type_specifier_nonarray
...
+    }
+    | TENSORVIEWNV {
+        parseContext.tensorLayoutViewCheck($1.loc, "tensorViewNV", parseContext.symbolTable.atBuiltInLevel());
+        $$.init($1.loc, parseContext.symbolTable.atGlobalLevel());
+        $$.basicType = EbtTensorViewNV;
+    }
+    | FUNCTION {
+        $$.init($1.loc);
+        $$.basicType = EbtFunction;
+    }
```
arcady-lunarg, line 3552: What does the FUNCTION token do? I don't see any use of EbtFunction anywhere other than here, and I don't see any tests for the "__function" keyword. The GLSL spec also has a FUNCTION token without any explanation that I can see.

jeffbolznv: `__function` is used in the prototypes for coopMatReduceNV and coopMatLoadTensorNV for the "function pointer" function parameters. That gets mapped to this grammar rule by the KeywordMap. And then EbtFunction is used in ParseHelper and GlslangToSpv to detect and handle these function pointers.

jeffbolznv: @arcady-lunarg are there any specific changes you want in this area? It's kind of a gray area for the spec because this is really an implementation detail for how glslang handles these built-in functions, so I'm not sure it makes sense to document FUNCTION there (maybe I should remove it from the GLSL spec).

I'd like to get this merged soon to unblock other dependencies.

arcady-lunarg (changes requested):
A few comments, and one actual thing I'd like changed, namely the spirv.hpp.
Also apparently GLSL has function pointers now, that's kind of cool.

SPIRV/spirv.hpp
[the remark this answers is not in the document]
jeffbolznv: ok, I ran dos2unix and added back the `: unsigned` to the masks (including the few new masks). But I kept all the other updates to the file. I get a lot of conflicts whenever I work on a new extension and need to rebase, it would be nice if we could eventually find a way to directly use one of the upstream headers.

SPIRV/disassemble.cpp
```diff
@@ -338,6 +338,18 @@ int SpirvStream::disassembleString()
     return decoderes.first;
 }
 
+static uint32_t popcount(uint32_t mask)
+{
```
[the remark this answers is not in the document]
jeffbolznv: Yeah, I don't think there's a way to get popcount without using compiler intrinsics until C++20, so I just did this for now.

arcady-lunarg (approved):
Thanks, LGTM

PR #3782, opened 2024-10-28
Add support for GL_EXT_texture_offset_non_const

### LINKED ISSUE #3765: Allow Offset in addition to ConstOffset for texture operations
As part of an upcoming Vulkan maintenance extension I'm developing CTS tests for, we want to allow using Offset in addition to ConstOffset for image operations. This was already possible for image gather operations, but not for others, and is mainly enforced by [VUID-StandaloneSpirv-Offset-04663](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/StandaloneSpirv.html#VUID-StandaloneSpirv-Offset-04663). There are other VUIDs affected by this but that's the main one. On the SPIR-V side there's no such restriction, so SPIR-V doesn't need any special capability nor a new extension to support this behavior.

While developing CTS tests I managed to get what I wanted from glslang by applying the following diff.

```diff
diff --git a/glslang/MachineIndependent/ParseHelper.cpp b/glslang/MachineIndependent/ParseHelper.cpp
index 215ccd99..b187ce8a 100644
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -2332,9 +2332,7 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
                                     arg0->getType().getSampler().shadow;
             if (f16ShadowCompare)
                 ++arg;
-            if (! (*argp)[arg]->getAsTyped()->getQualifier().isConstant())
-                error(loc, "argument must be compile-time constant", "texel offset", "");
-            else if ((*argp)[arg]->getAsConstantUnion()) {
+            else if (((*argp)[arg]->getAsTyped()->getQualifier().isConstant()) && (*argp)[arg]->getAsConstantUnion()) {
                 const TType& type = (*argp)[arg]->getAsTyped()->getType();
                 for (int c = 0; c < type.getVectorSize(); ++c) {
                     int offset = (*argp)[arg]->getAsConstantUnion()->getConstArray()[c].getIConst();
```

But I'm not used to glslang's code so I don't know if my changes are correct and complete, and I'm fairly positive I'm breaking several tests with that change. Could someone take a look at this? Ideally I'd like glslang to support this so I can just update the glslang commit reference in CTS and know this is going to work.

/cc @spencer-lunarg since this is related to validation too.

rg3igalia:
Closes #3765


rg3igalia:
I noticed the constness of the offset argument was not being checked with sparse texture functions, so I've added a commit on top with some additional code and tests.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2309,18 +2310,26 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
         int arg = -1;
         switch (callNode.getOp()) {
-        case EOpTextureOffset:          arg = 2;  break;
-        case EOpTextureFetchOffset:     arg = (arg0->getType().getSampler().isRect()) ? 2 : 3; break;
-        case EOpTextureProjOffset:      arg = 2;  break;
-        case EOpTextureLodOffset:       arg = 3;  break;
-        case EOpTextureProjLodOffset:   arg = 3;  break;
-        case EOpTextureGradOffset:      arg = 4;  break;
-        case EOpTextureProjGradOffset:  arg = 4;  break;
+        case EOpSparseTextureOffset:      // fallthrough
```
arcady-lunarg, line 2322: It probably makes sense to just reformat this to consolidate all the "arg = 2" cases together, likewise for the "arg = 3" cases, etc. And then put the "arg = whatever; break;" on a separate line from the "case".

arcady-lunarg (commented):
Should these test cases be generating SPIR-V? I feel like that might be useful here, since ultimately this is an extension for use with Vulkan and SPIR-V.

rg3igalia:
> Should these test cases be generating SPIR-V? I feel like that might be useful here, since ultimately this is an extension for use with Vulkan and SPIR-V.

Yes, it makes total sense. I wasn't aware of the distinction between the different test groups as I'm not familiar with glslang's code base. Thanks for the tip!

glslang/MachineIndependent/ParseHelper.cpp
rg3igalia: Ack. Will fix.

rg3igalia:
I see the CI failures are related to the shaders failing validation and suggesting something like this:

```diff
--- Test/baseResults/validation_fails.txt	2024-11-05 14:19:47.844071917 +0000
+++ -	2024-11-05 14:25:25.835244652 +0000
@@ -68,4 +68,6 @@
 Test/baseResults/spv.separate.frag.out
 Test/baseResults/spv.sparseTextureClamp.frag.out
 Test/baseResults/spv.sparseTexture.frag.out
+Test/baseResults/spv.sparsetextureoffset_non_const.vert.out
+Test/baseResults/spv.textureoffset_non_const.vert.out
 Test/baseResults/vk.relaxed.errorcheck.vert.out
```

Which may be solved once the [SPIR-V Tools issue](https://github.com/KhronosGroup/SPIRV-Tools/issues/5857) is solved. So what should I do in the mean time? Update the file in this MR?

arcady-lunarg:
We can wait until the SPIRV-Tools change is merged and the you can update the `known_good.json` to point to the new SPIRV-Tools commit.

rg3igalia:
Alright, feel free to merge if you want in the mean time. The CTS tests using this are still under review so I'm not in a hurry.

zmike:
What's going on here?

arcady-lunarg:
@zmike I've been waiting for the SPIRV-Tools change to land before I merge this, but at this point maybe it's not worth it to wait.

rg3igalia:
Now that the SPIRV-Tools pull request has landed, I've updated the reference commit and made sure no validation fails for the added tests. No idea what needs to be done for the "Linux Backcompat" check to pass, though.

arcady-lunarg:
The backcompat failure is because spirv-tools bumped their minimum required cmake version.

arcady-lunarg (approved):

pmistryNV:
@rg3igalia for spare texture ops, making offsets non-constant is it defined in the GL_EXT_texture_offset_non_const spec?

rg3igalia:
> @rg3igalia for spare texture ops, making offsets non-constant is it defined in the GL_EXT_texture_offset_non_const spec?

The short answer is **yes**, but this is not _directly_ related to GL_EXT_texture_offset_non_const because GLSL does not define specific operations for sparse images.

However, this need comes from VK_KHR_maintenance8 in Vulkan. The spec for the extension says that the [maintenance8 feature](https://registry.khronos.org/vulkan/specs/latest/man/html/VkPhysicalDeviceMaintenance8FeaturesKHR.html) allows us to "Support Offset (as an alternative to ConstOffset) image operand in texture sampling and fetch operations", and this is reflected in [VUID-RuntimeSpirv-Offset-10213](https://registry.khronos.org/vulkan/specs/latest/man/html/RuntimeSpirv.html):

> If the [maintenance8](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#features-maintenance8) feature is not enabled, image operand Offset must only be used with OpImage*Gather instructions

So with maintentance8, you can use Offset with a variety of other operations, including OpImageSparse* operations. In fact, the [glslang implementation](https://github.com/KhronosGroup/glslang/blob/8b822ee8ac2c3e52926820f46ad858532a895951/glslang/MachineIndependent/ParseHelper.cpp#L2556) will allow Offset with OpImageSparse* when GL_EXT_texture_offset_non_const is enabled. Is this missing CTS coverage?

PR #3783, opened 2024-11-01
ci: Refactor job matrices to use include/exclude

arcady-lunarg:
This removes some duplication from the job definitions


arcady-lunarg:
I don't think this is really worth the effort of rebasing.

PR #3785, opened 2024-11-05
ci: Add continuous_integration dependency

spencer-lunarg:
In effort to reduce the number of limit KhronosGroup minutes we share across repos, this adds dependency to the Continuous Integration CI to only expensive jobs once quicker/smaller jobs run first

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -10,6 +10,7 @@ permissions: read-all
 
 jobs:
   linux:
+    needs: linux-shared # fastest job to do sanity check
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
@@ -78,6 +79,7 @@ jobs:
         run: ctest --output-on-failure --test-dir build
 
   linux-asan:
+    needs: linux-shared # fastest job to do sanity check
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
@@ -149,13 +151,15 @@ jobs:
         run: ctest --output-on-failure --test-dir build
 
   macos:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: windows-shared
     runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-14, macos-13]
+        os: [macos-14]
         compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
-        cmake_build_type: [Debug, Release]
+        cmake_build_type: [Release]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: lukka/get-cmake@b516803a3c5fac40e2e922349d15cdebdba01e60 # v3.30.5
@@ -169,6 +173,8 @@ jobs:
       - run: ctest --output-on-failure --test-dir build
 
   macos-shared:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: windows-shared
     runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
@@ -189,6 +195,7 @@ jobs:
       - run: ctest --output-on-failure --test-dir build
 
   windows:
+    needs: linux-shared # fastest job to do sanity check
     runs-on: ${{matrix.os.genus}}
     permissions:
       contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -149,13 +151,15 @@ jobs:
...
 
   macos:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: windows-shared
     runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-14, macos-13]
+        os: [macos-14]
```
spencer-lunarg, line 160: I just realized we only run `Continuous Deployment` on Mac 13, Clang, release

so really all of Mac should be the same version

.github/workflows/continuous_integration.yml
```diff
@@ -149,13 +151,15 @@ jobs:
...
         compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
```
spencer-lunarg, line 161: if we don't release a GCC version, we should just save CI and not run it here

.github/workflows/continuous_integration.yml
spencer-lunarg:
```suggestion
        os: [macos-13]
```

.github/workflows/continuous_integration.yml
spencer-lunarg:
```suggestion
        compiler: [{cc: clang, cxx: clang++}]
```

.github/workflows/continuous_integration.yml
```diff
@@ -149,13 +151,15 @@ jobs:
...
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: windows-shared
     runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-14, macos-13]
-        compiler: [{cc: clang, cxx: clang++}, {cc: gcc, cxx: g++}]
-        cmake_build_type: [Debug, Release]
+        os: [macos-13]
```
spencer-lunarg, line 160:
```suggestion
        os: [macos-13, macos-14] # 13 has x86 and 14 has ARM
```

arcady-lunarg:
I don't think this will add very much value: CI jobs mostly succeed, and this just adds latency to every single run. I think it would be better to prune the ci matrix instead, the finding about "gcc" on macos will certainly help us a lot with that.

spencer-lunarg:
> CI jobs mostly succeed

https://github.com/KhronosGroup/glslang/actions/workflows/continuous_integration.yml?query=is%3Afailure
https://github.com/KhronosGroup/glslang/actions/workflows/continuous_integration.yml?query=is%3Asuccess

there have been 137 failure and 531 successes (so that is a 20% failure rate)

is the latency something you find that would honestly be a bottleneck,

arcady-lunarg:
@spencer-lunarg Could you rebase? I'll approve and merge it if you do.

.github/workflows/continuous_integration.yml
```diff
@@ -149,13 +152,15 @@ jobs:
...
   macos:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: windows-shared
     runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
       matrix:
         os: [macos-latest]
         compiler: [{cc: clang, cxx: clang++}]
-        cmake_build_type: [Debug, Release]
```
arcady-lunarg, line 158: We probably want to keep a macos debug build just in case.

spencer-lunarg: why? Have we ever seen it fail ONLY on Mac Debug build?

spencer-lunarg: We came to same conclusion on VVL and other repo that why do we need to worry about the Mac debug build, we want to make sure it works on release to ship, but the number of mac developers on Glslang (or just people on glslang) I can't see why to justify this IMO

arcady-lunarg: I think the combination of arm64 and mac is useful enough that I would like to run the tests with debug assertions, especially since the tests don't take that long to run and the rate of PRs is a bit lower compared to VVL.

spencer-lunarg: because `glslang` is currently not on the "top 5 offenders" of CI action minutes yet, sure will revert

PR #3788, opened 2024-11-05
Add missing <sstream> header include

HuYuxin:
This is needed to fix the below errors on some compilers:

Initialize.cpp:4476:27: error: implicit instantiation of undefined template 'std::basic_stringstream'
std::stringstream cooperativeMatrixFuncs;

### DIFF
--- a/glslang/MachineIndependent/Initialize.cpp
+++ b/glslang/MachineIndependent/Initialize.cpp
@@ -52,6 +52,7 @@
 //
 
 #include <array>
+#include <sstream>
 #include "Initialize.h"
 #include "span.h"
 


HuYuxin:
This is to address the build failure mentioned in this [comment](https://github.com/KhronosGroup/glslang/commit/d0e67c58134377f065a509845ca6b7d463f5b487#commitcomment-148708922)

arcady-lunarg (approved):

arcady-lunarg:
What build configuration causes this error? Because we should probably be testing it in CI.

HuYuxin:
> What build configuration causes this error? Because we should probably be testing it in CI.

We use http://libcxx.llvm.org/  as the C++ standard library.

PR #3789, opened 2024-11-06
Replace individual EOpConv<A>To<B> with a single EOpConvNumeric

jeffbolznv:


arcady-lunarg (approved):
LGTM, this is a very nice refactor.

PR #3796, opened 2024-11-15
Cleanup old CMake related docs in readme.md

jpr42:
The bit about CMAKE_MSVC_RUNTIME_LIBRARY has lost all context. Originally it was related to OVERRIDE_MSVCCRT. Removing it fully is more clear.

Cleanup the configure instructions. Using newer CMake cli.

Remove bit about the CMake GUI being tested. Doesn't make sense to have in the README.md

Remove section about GLSLANG_TESTS. This documentation is now outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,6 @@
 
 4. `OGLCompiler` and `HLSL` stub libraries have been fully removed from the build.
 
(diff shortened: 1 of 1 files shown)


arcady-lunarg (approved):
LGTM, thanks for cleaning things up.

PR #3797, opened 2024-11-16
Improve auto push_constants upgrading.

mbechard:
Ensure all copies of the global uniform block symbol get updated with the correct layoutPacking when it's auto upgraded to a push_constant. In some cases there are multiple copies, and only one was getting changed previously.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -65,7 +65,7 @@ struct TVarEntryInfo {
     long long id;
     TIntermSymbol* symbol;
     bool live;
-    bool upgradedToPushConstant;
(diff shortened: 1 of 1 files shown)


arcady-lunarg:
Would it make sense to add a unit test for this? In what case was the problem being triggered?

mbechard:
I've only see it occur in one huge shader. It's from an end-user so I don't think they'd want it posted to the repository. I'm not sure under what circumstances the tree ends up with duplicated symbols, but this fix follows other similar workflows at least.
I also don't think the testing infrastructure is setup to test this kind of stuff anyways. We'd need a way to apply different rules to different test cases I think.

arcady-lunarg (approved):
Looks fine.

PR #3801, opened 2024-11-26
ci: Remove gcc from macos matrix

arcady-lunarg:
Turns out gcc on macos by default is just an alias for clang, so these configurations were adding no value at all.


jeremy-lunarg (approved):
Huh. There must be a way to actually install GCC, but I don't know how important that is to MacOS users.

PR #3803, opened 2024-11-28
Fix to allow using the glslang_shader_parse from C API without preprocess

antaalt:
Hello,

I went into problem with the C API of glslang. Here is a small repro sample of the issue:

```glsl
#version 450
#extension GL_EXT_debug_printf : require
void main() {
    debugPrintfEXT("global thread id %v3u\n", gl_GlobalInvocationID);
}
```

```c
// Simple C code to run 
glslang_shader_t* shader = glslang_shader_create(&shader_input);
glslang_shader_preprocess(shader, &shader_input);
glslang_shader_parse(shader.handle.as_ptr(), &shader_input);
```
This will produce an `End of line in string` error for `\n` within string, but glslangValidator does not produce this error. After digging into it, it seems calling preprocess will replace `\n` string by a newline feed character (Code at `MachineIndependant/preprocessor/PpScanner.cpp:1136`), which will then be used by parse & produce this error. 

With this CL, I try to fix this by not forcing the parse function to have preprocess executed before. Right now the C API dont allow parsing without preprocessing while the C++ API allow it. This issue should probably be solved more deeply to allow preprocess & parsing aswell but as its a very minor issue, I dont think it should really be a problem.

### DIFF
--- a/glslang/CInterface/glslang_c_interface.cpp
+++ b/glslang/CInterface/glslang_c_interface.cpp
@@ -431,6 +431,9 @@ GLSLANG_EXPORT int glslang_shader_preprocess(glslang_shader_t* shader, const gls
 
 GLSLANG_EXPORT int glslang_shader_parse(glslang_shader_t* shader, const glslang_input_t* input)
 {
+    // Seems glslang is preprocessing files twice if we call preprocess then parse on the preprocessed string.
+    // As we are forcing the use of preprocess before parse, this could lead to some issues with line return being preprocessed twice for example.
+    // Should avoid preprocessing twice in this case ideally, but just allow using C Api without preprocess for now.
     if (shader->preprocessedGLSL.length() > 0) {
         const char* preprocessedCStr = shader->preprocessedGLSL.c_str();
         shader->shader->setStrings(&preprocessedCStr, 1);
@@ -447,6 +450,7 @@ GLSLANG_EXPORT int glslang_shader_parse(glslang_shader_t* shader, const glslang_
         glslang::TShader::Includer& Includer = (input->callbacks.include_local||input->callbacks.include_system)
             ? static_cast<glslang::TShader::Includer&>(callbackIncluder)
             : static_cast<glslang::TShader::Includer&>(dirStackFileIncluder);
+        shader->shader->setStrings(&input->code, 1);
         return shader->shader->parse(
             reinterpret_cast<const TBuiltInResource*>(input->resource),
             input->default_version,
--- a/glslang/CInterface/glslang_c_interface.cpp
+++ b/glslang/CInterface/glslang_c_interface.cpp
@@ -431,15 +431,30 @@ GLSLANG_EXPORT int glslang_shader_preprocess(glslang_shader_t* shader, const gls
 
 GLSLANG_EXPORT int glslang_shader_parse(glslang_shader_t* shader, const glslang_input_t* input)
 {
-    const char* preprocessedCStr = shader->preprocessedGLSL.c_str();
-    shader->shader->setStrings(&preprocessedCStr, 1);
+    if (shader->preprocessedGLSL.length() > 0) {
+        const char* preprocessedCStr = shader->preprocessedGLSL.c_str();
+        shader->shader->setStrings(&preprocessedCStr, 1);
 
-    return shader->shader->parse(
-        reinterpret_cast<const TBuiltInResource*>(input->resource),
-        input->default_version,
-        input->forward_compatible != 0,
-        (EShMessages)c_shader_messages(input->messages)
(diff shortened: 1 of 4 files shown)


dnovillo (changes requested):
Is there a way to add this fragment shader as a test case?

antaalt:
I will try to look into it

dnovillo (changes requested):
I'm not sure this fix does what you meant. If the code is not pre-processed, we still fail to parse with:

```
ERROR: 0:4: 'gl_GlobalInvocationID' : undeclared identifier
ERROR: 0:4: '' : compilation terminated                                                                                 
```

Which I guess it's expected because gl_GlobalInvocationID is not expanded properly.

dnovillo:
@antaalt are you still working on this PR? Thanks.

antaalt:
Sorry, did not had the time to look into it, it compile fine for me if i specify the compute shader stage as gl_GlobalInvocationID is a compute only variable.

For me the issue is mostly due to the C API that ***requires*** to call preprocess before calling parse (it will crash if you dont), while with the C++ API, simply calling parse is enough. It looks parse is doing the preprocessing steps, calling preprocess before calling parse will then preprocess the code twice, which is the reason of the bug in this changelist.

Ideally, we should avoid preprocessing twice which looks like it it doing here if we call preprocess + parse, this is a small fix to just allow us to run parse without calling preprocess before.

I will try to look to add some test for this.

glslang/CInterface/glslang_c_interface.cpp
```diff
@@ -431,15 +431,34 @@ GLSLANG_EXPORT int glslang_shader_preprocess(glslang_shader_t* shader, const gls
...
-
-    return shader->shader->parse(
-        reinterpret_cast<const TBuiltInResource*>(input->resource),
-        input->default_version,
-        input->forward_compatible != 0,
-        (EShMessages)c_shader_messages(input->messages)
-    );
+    // Seems glslang is preprocessing files twice if we call preprocess then parse on the preprocessed string.
+    // As we are forcing the use of preprocess before parse, this could lead to some issues with line return being preprocessed twice for example.
+    // Should avoid preprocessing twice in this case ideally, but just allow using C Api without preprocess for now.
```
dnovillo, line 436: nit: s/C Api/C API/

gtests/TestFixture.h
```diff
@@ -708,6 +709,183 @@ class GlslangTest : public GT {
         }
     }
 
+    std::tuple<bool, std::string> preprocessAndCompileWithCAPI(const std::string& source)
+    {
+        const char * shaderStrings = source.data();
+        const int shaderLengths = static_cast<int>(source.size());
+
+        // Done by framework
+        //glslang_initialize_process();
```
dnovillo, line 718: No need to have this commented-out code here.

gtests/TestFixture.h
```diff
@@ -708,6 +709,183 @@ class GlslangTest : public GT {
...
+        };
+        shader_input.callbacks_ctx = NULL;
+
+        glslang_shader_t* shader = glslang_shader_create(&shader_input);
+        // Uncommenting following code will make code being preprocessed twice
+        /*if (glslang_shader_preprocess(shader, &shader_input) != 1) {
+            const std::string logs = std::string(glslang_shader_get_info_log(shader));
+            glslang_shader_delete(shader);
+            return std::make_tuple(false, logs);
+        }*/
```
dnovillo, line 864: Well, yes, but why have it here commented out?

dnovillo (commented):
Thanks for the new test. Would you consider adding a couple more tests that check for other escape sequences and macro expansion?

Also, if pre-processing has occurred, we should now just use the pre-processed string, right? Perhaps a test that validates this makes sense?

Thanks.

gtests/TestFixture.h
antaalt: Mostly for keeping track of the issue, but that might be overkill, I will remove it

antaalt:
Well from what I saw, I think the parser preprocess the input string without caring if we called preprocess before and if we passed a preprocess string or a non preprocessed one, which make sense as with the c++ api, we can just call parse without preprocess, and preprocess will still be done.

So the issue in the example is that the "/n" string is transformed into a '/n' character by the preprocess step. And the parse step is then outputting an error as it again try to preprocess the string, as it does not understand why the string has a '\n' character at the end. But it was not a '\n' character before the preprocessing step but a "\n" string.

So skipping the preprocess fix this because it does not change the string to a character in the input.

Logic is here https://github.com/KhronosGroup/glslang/blob/main/glslang%2FMachineIndependent%2Fpreprocessor%2FPpScanner.cpp#L1206

PR #3807, opened 2024-12-10
Fix Windows DLL test configuration

jeremy-lunarg:
Enable Windows DLL test configuration.
Configure tests to pull in DLL dependencies using `TARGET_RUNTIME_DLL_DIRS`.
Bump cmake version to 3.27 to use `TARGET_RUNTIME_DLL_DIRS`.
Remove Ubuntu 20.04 test configuration.

### DIFF
--- a/kokoro/linux-clang-cmake/build-docker.sh
+++ b/kokoro/linux-clang-cmake/build-docker.sh
@@ -39,7 +39,7 @@ set -e # Fail on any error.
 
 set -x # Display commands being run.
 
-using cmake-3.17.2
+using cmake-3.31.2
 using clang-10.0.0
 using ninja-1.10.0
 using python-3.12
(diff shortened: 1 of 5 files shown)


kokoro/linux-clang-cmake/build-docker.sh
```diff
@@ -39,7 +39,7 @@ set -e # Fail on any error.
 
 set -x # Display commands being run.
 
-using cmake-3.17.2
+using cmake-3.27
```
dneto0, line 42: This needs to correspond to a cmake that has been put into the kokoro image.
It should be `cmake-3.31.2`

dneto0 (changes requested):
Please use `cmake-3.31.2` in the kokoro scripts.

dneto0 (approved):
LGTM, thanks!

Issue #3814, opened 2024-12-16
16bit Modf is incorrectly depending on SPV_AMD_gpu_shader_half_float


eilswe-arm:
In #3757 a dependency was added for 16bit Modf on the SPV_AMD_gpu_shader_half_float extension. This is incorrect, and the dependency should rather be on the 16bit float capability.

This causes issues when compiling the resulting spirv on devices without SPV_AMD_gpu_shader_half_float support.

eilswe-arm:
Pull request #3815

PR #3815, opened 2024-12-16
Fix incorrect extension dependency for Modf

eilswe-arm:
See issue for details #3814


jeremy-lunarg:
Can you submit a test that exercises this code path? As it stands, none of the existing tests are affected.

jeremy-lunarg:
It seems this condition should be unnecessary. If we have a half-precision operand, then wouldn't this capability have already been generated? Can't we just eliminate these two lines?

eilswe-arm:
After a look, I agree that the check is unnecessary, I removed it from the patch. I guess this removes the need for a test? There is already a test for 16bit float modf in glslang/Test/spv.float16.frag.

jeremy-lunarg (approved):
LGTM. Thanks for the fix.

PR #3816, opened 2024-12-16
Re-add linux back compat test

jeremy-lunarg:
Enable building on older versions of CMake.
Add warning for Windows shared library test configuration.


arcady-lunarg (approved):
LGTM, thanks.

PR #3832, opened 2025-01-10
Support EXT_integer_dot_product

jeffbolznv:
Spec is at https://github.com/KhronosGroup/GLSL/pull/195.


glslang/MachineIndependent/Constant.cpp
```diff
@@ -1119,6 +1119,9 @@ TIntermTyped* TIntermediate::fold(TIntermAggregate* aggrNode)
             break;
         }
         case EOpDot:
+            if (!children[0]->getAsTyped()->isFloatingDomain()) {
```
arcady-lunarg, line 1122: I assume this means that integer dot products don't get treated as constant expressions and don't get constant folded?

jeffbolznv: Right. The tests are using zeros and I saw they were incorrectly folded, and this fixed it.

arcady-lunarg (approved):

Issue #3509, opened 2024-02-12
Issues using via CMake's FetchContent or add_subdirectory


AnyOldName3:
Not all platforms provide a new enough version of glslang that the proper CMake config files are present to just import the system installation. That means that some projects will still need to pull in the source and build it to support some platforms.

There are some problems with this, though.

### Warnings
glslang has code that triggers many commonly-used warnings, and by default will inherit the parent project's warnings. This could be resolved in a a couple of ways:
* Fix the warnings, like https://github.com/KhronosGroup/glslang/pull/3143 attempted for `-Wshadow`.
* Explicitly disable warnings on the glslang targets. This could be a hassle as just adding `-Wno-shadow` `-Wno-implicit-fallthrough` and `-Wno-missing-field-initializers` (plus equivalents for other compilers) depends on the order of arguments to work, so it would also need various CMake variables and properties to be scanned and have flags enabling these warnings replaced.

### Alias
When imported through `find_package`, glslang is available as `glslang::glslang`, and ancillary libraries are also namespaced like `glslang::SPIRV`. When included via `FetchContent` or `add_subdirectory`, the namespaced versions don't exist, just `glslang` and `SPIRV`. 

This can easily be resolved by adding a library alias with
```cmake
add_library(glslang::glslang ALIAS glslang)
```
which would make it always available in the namespaced form.

### Installing
Since https://github.com/KhronosGroup/glslang/pull/3439, glslang can only be installed if it's the top-level project.

This is okay for:
* applications using glslang with `BUILD_SHARED_LIBS` off - the static library will be linked so nothing from glslang will need to be installed.
* applications using glslang with `BUILD_SHARED_LIBS` on - (presumably - I haven't checked this, so could be wrong) one of the CMake features that installs runtime dependencies for a target would install the shared libraries.
* shared libraries - they'd work the same as applications.

This isn't okay for:
* static libraries using glslang with `BUILD_SHARED_LIBS` off - nothing's been linked, so the glslang static library needs to be installed, and CMake hasn't been told how.
* static libraries using glslang with `BUILD_SHARED_LIBS` on - nothing's been linked, so the glslang shared library (import library with MSVC) needs to be installed, and CMake hasn't been told how, although the runtime part could *presumably* be installed.

In cases where the headers are needed because a library uses glslang headers in its public headers, which could be the case for shared or static libraries, they simply aren't installed.

[...]

PR #3837, opened 2025-01-16
Move SPIRV into glslang

vikke1234:
This PR moves the SPIRV/ into glslang/.

Currently, when using CMake's FetchContent[^0] to fetch SPIRV as a dependency, the headers are only generated upon installation. This results in inconsistent paths for including the headers. For example, if you use `FIND_PACKAGE_ARGS` to first search for system packages and fall back to compiling from source, it complicates the compilation process

A sample application that shows this behaviour can be found here: https://github.com/vikke1234/glslang-demo
The `master` branch is the current behaviour, while `fix` cantains this change.

[^0]: https://cmake.org/cmake/help/latest/module/FetchContent.html


vikke1234:
I'll try to fix these ASAP, I'm in the middle of a move so might take a moment. Seems only a `-I` is missing

AnyOldName3:
This will fix part of https://github.com/KhronosGroup/glslang/issues/3509. Because the text of the original report got mangled when I edited it, I've lost track of whether this is the only remaining problem 3509 is about, but it's definitely one of the things it's about.

vikke1234:
Hi,

Sorry for the delay, should work now.

CMakeLists.txt
```diff
@@ -315,7 +315,7 @@ else()
 endif()
 
 if(ENABLE_SPIRV)
-    add_subdirectory(SPIRV)
+    add_subdirectory(glslang/SPIRV)
```
jeremy-lunarg, line 318: Can you move this to glslang/CMakeLists.txt?

jeremy-lunarg (changes requested):

PR #3845, opened 2025-01-22
Update cmake_minimum_required

jeremy-lunarg:
Track cmake_mimimum_required defined in SPIRV-Tools dependency.
Remove conditional dependent on older version of CMake.


arcady-lunarg (approved):
LGTM

PR #3846, opened 2025-01-23
Update spirv tools

alan-baker:
Update SPIRV-Tools 
* update test expectations


arcady-lunarg (approved):
Unfortunate that there are even more validation fails now but I guess we will have to fix them at some later point.

PR #3854, opened 2025-01-29
Add GLSL_EXT_nontemporal_keyword

HildarTheDorf:
This is an example implementation of KhronosGroup/GLSL#275


arcady-lunarg (changes requested):
Could you add a unit test exercising this new functionality? In this case, I think you can just add something to `Test/spv.memoryQualifier.frag` and then regenerate the expected results with `gtests/glslangtests --update-mode`.

JebScape:
Looks good. I’d just expand on the tests a bit more to test more of the changes made than just loads.

HildarTheDorf:
More tests added.

Note: There seems to be a problem with the recently added GL_NV_cooperative_vector tests. The output differs on my machine from the golden results that have been checked in.

jeremy-lunarg:
> Could you add a unit test exercising this new functionality? In this case, I think you can just add something to `Test/spv.memoryQualifier.frag` and then regenerate the expected results with `gtests/glslangtests --update-mode`.

@arcady-lunarg  can you mark this as resolved?

jeremy-lunarg:
> More tests added.
> 
> Note: There seems to be a problem with the recently added GL_NV_cooperative_vector tests. The output differs on my machine from the golden results that have been checked in.

I'm not seeing that at top of main.

HildarTheDorf:
Looks like the CI passes the tests, so I assume the test failures are something about my local machine.
The diff is here: https://github.com/KhronosGroup/glslang/compare/main...umbral-software:glslang:bug/coopvec_test_fails

arcady-lunarg (approved):

PR #3855, opened 2025-01-30
Implement GL_NV_cooperative_vector

jeffbolznv:


jeffbolznv:
Hi @arcady-lunarg, do you have an ETA for reviewing this?

jeremy-lunarg (approved):
@jeffbolznv LGTM. I was going to hold off on the merge until the spec landed but I see that slang has already landed it.

PR #3865, opened 2025-02-05
cmake: Cleanup CMake code

jpr42:
- Remove old style of ending functions/conditions
- MSVC RTTI flag /GR is not added to CMAKE_CXX_FLAGS by default anymore.
- Remove unneccessary pre 3.18 code path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2023 The Khronos Group Inc.
+# Copyright (C) 2020-2025 The Khronos Group Inc.
 #
 # All rights reserved.
 #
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -175,12 +175,7 @@ elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang" AND NOT MSVC)
     endif()
 elseif(MSVC)
     if(NOT ENABLE_RTTI)
-        string(FIND "${CMAKE_CXX_FLAGS}" "/GR" MSVC_HAS_GR)
-        if(MSVC_HAS_GR)
-            string(REGEX REPLACE "/GR" "/GR-" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
jpr42, line 180: See https://cmake.org/cmake/help/latest/policy/CMP0117.html

CMakeLists.txt
```diff
@@ -357,7 +352,7 @@ if(GLSLANG_TESTS)
             set_tests_properties(glslang-testsuite PROPERTIES ENVIRONMENT_MODIFICATION "PATH=path_list_prepend:$<JOIN:$<TARGET_RUNTIME_DLL_DIRS:spirv-remap>,\;>")
         endif()
     endif()
-endif(GLSLANG_TESTS)
```
jpr42, line 360: This is a really old way of doing things. Just works because of backwards compatibility.

https://cmake.org/cmake/help/latest/command/endif.html

jeremy-lunarg (approved):
LGTM

PR #3878, opened 2025-02-16
Throw error if using in/out with struct pointer

spencer-lunarg:
To address https://gitlab.khronos.org/spirv/SPIR-V/-/issues/779


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7061,6 +7061,15 @@ void TParseContext::layoutTypeCheck(const TSourceLoc& loc, const TType& type)
             error(loc, "Push constants blocks can't be an array", "push_constant", "");
     }
 
+    if (type.getBasicType() == EbtReference) {
+        if (qualifier.isPipeInput())
+            error(loc, "can not contain any structs with buffer_reference", "in",
```
arcady-lunarg, line 7066: I believe the correct spelling is 'cannot', and you should probably put a period at the end of the first string so the resulting error message reads better.

arcady-lunarg (changes requested):
Minor grammar nit, otherwise LGTM.

arcady-lunarg (approved):

PR #3879, opened 2025-02-19
kokoro: use streamlined license checker image

dneto0:
This image is only 85MB (32MB compressed), which is many gigabytes smaller than the old image.

It downloads and runs in about 5s.

BUG=397439052


arcady-lunarg (approved):
I'm all for faster CI.

PR #3900, opened 2025-03-11
Fix crash calling coopMatLoadTensorNV on an array element

jeffbolznv:
The array element is not a symbol, the call to getAsSymbolNode crashed. Without getAsSymbolNode, there was a crash in teardown while deleting the DAG. Just skip addOutputArgumentConversions for coopmat parameters, I don't think it's necessary.


jeffbolznv:
This is passing CI now, and I've also tested this against Vulkan CTS and llama.cpp.

jeremy-lunarg (approved):
LGTM. Thanks!

PR #3901, opened 2025-03-12
ci: Reduce usage of setup-python

jpr42:
Only keep it to test back compat for older versions of python.

Reduces CI cache misses which can be up to 1 minute when they happen.


arcady-lunarg (approved):
Good find, thanks.

PR #3902, opened 2025-03-17
Switch tests using Vulkan memory model to spir-v 1.3

alan-baker:
* Vulkan memory model requires SPIR-V 1.3
* Update tests using it to run in that mode

The validation is coming in https://github.com/KhronosGroup/SPIRV-Tools/pull/6042 and this is verified against that change. Want to land this first to avoid headaches.


alan-baker:
Not sure I should have moved spv.atomicFloat.comp, but also not sure how the test should be run instead.

jeremy-lunarg:
> Not sure I should have moved spv.atomicFloat.comp, but also not sure how the test should be run instead.

Yeah. Strange. The move looks appropriate to me.

I presume we need to update our SPIRV-Tool dependency before we update validation_fails.txt.

alan-baker:
Well the validation failure is correct. That test generates ImageRect dimensions for some image which Vulkan doesn't support, but OpenGL on gets checked against SPIR-V 1.0 so it feels like either the test needs split into two or we have to accept some validation failures.

PR #3908, opened 2025-03-24
LiveTraverser no longer traverses switch cases that are statically not accessed

daniel-story:
This PR improves LiveTraverser to no longer traverse switch cases that can be determined at compile time to not be live. This change is also a prerequisite for correct error reporting in cross-stage link-time IO validation, which necessarily depends on accurate information about IO variables' liveness.

A consequence of this change is that IO variables whose only access occurs in code paths which aren't live (now including switch cases) are no longer assigned bindings or locations by the IO mapper, which is how it already works now with if statements.

### DIFF
--- a/Test/baseResults/liveTraverser.switch.vert.out
+++ b/Test/baseResults/liveTraverser.switch.vert.out
@@ -1,0 +1,327 @@
+liveTraverser.switch.vert
+Shader version: 460
+0:? Sequence
+0:25  Function Definition: main( ( global void)
+0:25    Function Parameters: 
+0:27    Sequence
+0:27      switch
+0:27      condition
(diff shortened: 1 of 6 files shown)


arcady-lunarg:
@daniel-story You're going to need to get the author of the patch to sign the CLA. And also probably to look into the CI failures, I think you need to put the visibility annotation on the class rather than on an individual method.

daniel-story:
> @daniel-story You're going to need to get the author of the patch to sign the CLA. And also probably to look into the CI failures, I think you need to put the visibility annotation on the class rather than on an individual method.

Apologies, all issues should be fixed now.

arcady-lunarg (approved):

PR #3909, opened 2025-03-24
Fix issue preventing non-ASCII characters in strings from being properly represented in SPIR-V

daniel-story:
This PR fixes an issue in glslang's SPIR-V generation that was preventing strings containing non-ASCII characters (for example shaders containing comments in languages other than English) from being correctly represented in SPIR-V.

A new test has also been added to ensure support for non-ASCII characters in strings doesn't break in the future.


arcady-lunarg (approved):
That was definitely an unfortunate oversight, per the GLSL spec non-ASCII comments should have been supported (indeed comments are the only place where you can use non-ASCII characters).

PR #3910, opened 2025-03-24
Add support for DebugSourceContinued

daniel-story:
This PR fixes various issues that occur whenever glslang compiles a shader whose source length exceeds the maximum length of a single DebugSource instruction, for example when processing non-trivial shaders from real world applications. When compiling SPIR-V with non-semantic debug info, glslang's existing SPIR-V builder logic does not seem to account for the possibility of shaders that are too long to fit in DebugSource existing and consequently glslang emits malformed SPIR-V in such cases. When such malformed SPIR-V is passed to other parts of the SPIR-V ecosystem (for example, glslang's own disassembler) various unexepcted consequences including validation failures and crashes result.

This PR fixes the implementation of spv::Builder::makeDebugSource() to handle source strings of arbitrary lengths and emit an appropriate combination of DebugSource and DebugSourceContinued as required by the specification. It also adds a new debug info test with an artificially long shader to ensure that this functionality doesn't break in the future.


arcady-lunarg (approved):
LGTM, thanks for the patch.

PR #3911, opened 2025-03-25
Avoid reusing array types with stride when no stride is requested.

### LINKED ISSUE #3892: Struct Copy fails Explicit Layout
The following https://godbolt.org/z/bEn6b61Gd

```glsl
#version 450

struct Bar {
    uint x;
    uint y;
    uint z[2];
};

layout(set = 0, binding = 0, std430) buffer foo {
    uvec4 a;
    Bar b;
    uint c;
};

void main() {
    Bar new_bar;
    b = new_bar;
}
```

Will generate invalid SPIR-V (https://godbolt.org/z/qae3GKv15)

> Invalid explicit layout decorations on type for operand '16[%_ptr_Function__struct_15]'
> %23 = OpVariable %_ptr_Function__struct_15 Function

I think this occurs starting with https://github.com/KhronosGroup/SPIRV-Tools/pull/6012

jeffbolznv:
Newer SPIR-V versions forbid explicit layouts in function/private storage, so track which array types have an explicit layout and don't reuse them when stride is zero.

This bug has been around "forever", but validation was recently added.

Also update SPIRV-Tools/Headers commits, to pull in that validation rule.

Fixes #3892.


jeffbolznv:
I don't have access to the failing CI job, so I don't know if it's legit.

jeffbolznv:
Ping @arcady-lunarg @jeremy-lunarg

jeremy-lunarg:
Just missing an include. I pushed a commit on top.

jeremy-lunarg (approved):
LGTM

PR #3918, opened 2025-04-01
Fix several issues with link-time implicit array size merging

daniel-story:
This PR fixes several issues related to glslang's handling of unsized arrays during cross-stage linking and adds new test coverage to more rigorously exercise this functionality. As one example (but by no means the only issue), in the current mainline version of glslang, whenever:

- Multiple stages declare the same unsized arrayed uniforms, and
- Some but not all of the stages index into the array in a nonuniform and/or otherwise runtime-sized manner, and
- A user attempts to link a program containing the stages

then glslang's existing logic for adopting implicit array sizes causes such arrays to be correctly runtime sized in stages with nonuniform and/or runtime-sized accesses, but to be incorrectly implicitly sized in stages without such accesses. This causes an array dimensionality mismatch during linking which (incorrectly) causes linking to fail.

This PR builds upon past work to fix these kinds of issues and further improve glslang's linking logic in several ways:

First and foremost, implicit array size adoption that was previously happening at the end of TIntermediate::finalCheck() -- after the intermediates for each individual stage had been merged with each other but before they had been merged and cross-validated across stages -- has been moved into TProgram::crossStageCheck(). After all uniform objects for all stages in a program are checked for conflicts/inconsistencies and merged into a temporary intermediate by existing code, a new mergeImplicitArraySize() function is invoked on each active stage to copy array sizes of any uniforms that exist that stage back from the temporary merged intermediate into that stage's own intermediate.

Only after this merging is completed is each stage's intermediate traversed to adopt implicit array sizes. The actual array size adoption logic itself remains unchanged from before. For programs containing only a single stage the array size adoption happens just before crossStageCheck()'s early return, which produces effectively the same behavior as before the changes in this PR.

This allows unsized arrays that are present in multiple stages of a program but non-variably indexed in only some of them to be correctly runtime sized across all stages, fixing the erroneous link errors in the scenario described above.

As part of this work, existing implicit array size propagation logic to ensure that multiple stages with unsized arrays that are indexed with potentially different constant indices are implicitly sized to the largest implicit size across all stages is improved to also correctly work with unsized arrays on members of blocks (as opposed to only on blocks which are themselves arrayed).

Finally, error messages related to implicit array size conflicts are also improved to print the correct name of the stage(s) being linked instead of "unknown stage", as was incorrectly being printed in certain cases before now.

This PR implements a pair of new tests in order to exercise all of the above:

The first test (implicitArraySizeUniform) is a positive test to ensure that different stages with variously mismatching but valid unsized array inconsistencies can be linked successfully.

The second test (implicitArraySizeUniformContradict) is a negative test with several invalid cross-stage unsized array usage patterns. This test ensures that invalid unsized array inconsistencies across stages produce link errors as expected.

Because these tests need to validate the state of the program after cross-stage linking they depend on being able to dump the AST later in the linking process than it had been dumped before. In order to support tests that validate the state of the AST after cross-stage linking this PR moves the link-time AST dumping logic previously in TProgram::linkStage() up into its caller TProgram::link(), after all other linking steps (including cross-stage validation and optionally optimization) are completed.

As a consequence of that change, the AST dumped by existing tests in which linking intentionally fails now correctly reports unsized arrays that had previously already adopted their implicit sizes and/or "runtimeness" prior to dumping as being unsized, and this PR updates all affected base test results accordingly. As a bonus positive side effect of this reordering, it becomes possible to write gtest-based tests for more types of link-time functionality (for example the optional cross-stage optimization feature that was added recently).


arcady-lunarg (approved):
Looks good, thanks the patch and for the thorough testing.

Issue #3919, opened 2025-04-02
Linking multiple tese shaders with single layout, complains of contradictory layouts


amalon:
If multiple tessellation evaluation shaders are linked, where only one contains a layout qualifier, `glslangValidator` reports errors.

For example, if `test.tese` contains:
```glsl
#version 460 core

layout (quads, fractional_odd_spacing, ccw) in;

// lib.glsl
vec4 foo();

void main()
{
    gl_Position = foo();
}
```
and `lib.glsl` contains:
```glsl
#version 460 core

vec4 foo()
{
    return vec4(1.0);
}
```

Then:
```shell
$ glslangValidator -l -S tese test.tese lib.glsl
test.tese
lib.glsl
ERROR: Linking tessellation evaluation stage: Contradictory input vertex spacing
ERROR: Linking tessellation evaluation stage: Contradictory triangle ordering
```
If the layout qualifier is added to `lib.glsl`, the error goes away.

The glslang 4.6 spec says:
> Any or all of these identifiers may be specified one or more times in a single input layout
> declaration.
> If primitive mode, vertex spacing, or ordering is declared more than once in the
> tessellation evaluation shaders of a program, all such declarations must use the same identifier.
>
> At least one tessellation evaluation shader (compilation unit) in a program must declare a primitive
> mode in its input layout. Declaring vertex spacing, ordering, or point mode identifiers is optional. It
> is not required that all tessellation evaluation shaders in a program declare a primitive mode. If
> spacing or vertex ordering declarations are omitted, the tessellation primitive generator will use
> equal spacing or counter-clockwise vertex ordering, respectively. If a point mode declaration is
> omitted, the tessellation primitive generator will produce lines or triangles according to the
> primitive mode.

That doesn't specify that every single tese shader should have the layout qualifier, only that they should match if they have them, and in fact that would be an significant issue for shader stage agnostic library code to have to have the layout qualifier.

dnovillo:
@amalon please confirm whether #3920 addresses your issue. Thanks.

KarenGhavam-lunarG:
> Not sure what the schedule of the next LunarG release is. @jeremy-lunarg may know. 

Releases are typically done shortly before an SDK release. Next SDK is planned for early October.

Issue #1516, opened 2018-10-02
gl_FragDepth redeclaration only needed if there are a static assignment


infapi00:
Using the following shaders:

**01.frag**:
```
#version 450

layout(depth_less) out float gl_FragDepth;

void foo();
void main()
{
     gl_FragDepth = gl_FragCoord.z;
     foo();
}
```

**02.frag**:

```
#version 450

/* this shader contains no static write to gl_FragDepth,
 * so need not redeclare it
 */

out vec4 color;

void foo()
{
   color = vec4(1);
}
```

And executing glslang as:
`$glslangValidator -G --aml --amb /tmp/01.frag /tmp/02.frag`

It throws the following error:
> ERROR: Linking fragment stage: Contradictory depth layouts

It throws the error as the second fragment stage doesn't redeclare it with the previous depth layout. But as the comment say, from GLSL 4.6 spec, section 4.4.2 "Ouput Layout Qualifiers", "Fragment Outputs":

> If gl_FragDepth is redeclared in any fragment shader in a program, it must be redeclared in all
> fragment shaders in that program that have static assignments to gl_FragDepth

That redeclaration is only needed if gl_FragDepth is assigned on the shader.

Sh0g0-1758:
Hey, ran into this issue. Seeing this is an open issue,  Can I work on it?

arcady-lunarg:
Go for it, patches are most welcome.

infapi00:
> @infapi00 This may have been fixed by https://github.com/KhronosGroup/glslang/pull/3920. Please confirm so we can close this issue.

Just tested with the shaders pointed on the issue, and it works now fine. Thanks.

dnovillo:
> > [@infapi00](https://github.com/infapi00) This may have been fixed by [#3920](https://github.com/KhronosGroup/glslang/pull/3920). Please confirm so we can close this issue.
> 
> Just tested with the shaders pointed on the issue, and it works now fine. Thanks.

Thanks. Closing.

PR #3920, opened 2025-04-03
Fix incorrect "Contradictory layout" link errors

Valentin-Sarthou:
Hello,

I've come across issues similar to #1516 and #3919 where "Contradictory layout" link errors are generated when using multiple compilation units and declaring a layout in only 1 compilation unit.

A similar fix was done in 6f723ebbe3b0eba0c543fa615b53fbdc3e6e22c7 for tessellation input and output primitives.

But the issue was still present for the following layouts:
- Transform Feedback xfb_stride layout
- Tessellation Evaluation vertex spacing & triangle ordering layouts (#3919)
- Fragment Shader gl_FragDepth redeclaration with EXT_conservative_depth layouts (#1516)
- Mesh Shader max_primitives layout
- Vulkan Compute Shader local_size_x_id, local_size_y_id & local_size_z_id layouts

This PR attempts to fix all these remaining cases, and adds tests for each.


amalon:
FWIW I've tested that with this branch the errors are gone from flightgear's shader validation script (see https://gitlab.com/flightgear/fgdata/-/issues/45)

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -344,7 +344,7 @@ void TIntermediate::mergeModes(TInfoSink& infoSink, TIntermediate& unit)
     }
     if (primitives == TQualifier::layoutNotSet)
         primitives = unit.primitives;
-    else if (primitives != unit.primitives) {
+    else if (unit.primitives != TQualifier::layoutNotSet && primitives != unit.primitives) {
```
dnovillo, line 557: Could you add a comment at the start of these checks documenting how the GLSL spec allows layouts to be declared in a single CU?

Test/baseResults/150.tesc.out
```diff
@@ -962,8 +962,6 @@ ERROR: Linking tessellation evaluation stage: can't handle multiple entry points
 ERROR: Linking tessellation evaluation stage: Multiple function bodies in multiple compilation units for the same signature in the same stage:
     main(
 ERROR: Linking tessellation evaluation stage: can't handle multiple entry points per stage
```
dnovillo, line 964: The new tests are great. Thanks! They do, however, leave us with no/little coverage for the cases where _there are_ layout conflicts. Would you mind augmenting the new tests to cover that case?

dnovillo (changes requested):
Looks great. Thanks. A couple of suggestions before submitting.

Valentin-Sarthou:
@dnovillo Added missing entry in validation_fails.txt

Valentin-Sarthou:
Hi @dnovillo, do you think you could add issues #1516 and #3919 to the "Development" section of this PR so that users are notified that their issues may be closed ?
Thanks

PR #3922, opened 2025-04-08
Implement optional cross-stage check for input with no matching output

daniel-story:
glslang does not currently have logic to detect when cross-stage IO statically accessed in one stage is missing in the previous stage, and there is a longstanding TODO comment in the existing TIntermediate::checkStageIO() function suggesting that such functionality should be implemented. To that end, this PR implements the proposed cross-stage link time IO validation and introduces a new EShMessages flag to optionally enable it. If the user does not explicitly opt in to this functionality by supplying the new flag, the behavior of linking remains unchanged from how it worked prior to this PR.

When enabled, the new cross-stage IO validation functionality aims to validate that statically read cross-stage input locations have corresponding output declarations in the previous stage. It also aims to ensure that as many unusual but valid usage patterns that are known to exist in real-world applications' shaders (for example, input declarations that don't have matching outputs in the previous stage but that are also never read, or that are only read in branches that can statically be determined to not be live) can be linked without any errors. At a high level, this implementation adopts a philosophy that it's always more important for valid usage to be allowed than for invalid usage to be disallowed (though of course considerable effort is still made to detect as much invalid usage as is practical to detect).

It is worth noting that though the TODO in the existing code may make this problem appear straightforward at first glance, a sufficiently robust implementation of this functionality capable of handling all valid usage patterns permitted by the GLSL specification (particularly certain usage patterns commonly encountered in complex, real-world applications) can be deceptively complicated to get right, and the amount of test coverage needed to adequately validate its functionality is substantial. Versions of the implementation in this PR have existed in Nintendo's fork of glslang for approximately 5 years and have evolved and improved over the course of that time to be able to successfully validate and link complex shader corpora from numerous real-world applications. Despite this, and the fact that this PR does include fairly comprehensive test coverage, in order to minimize risk to real-world applications this PR's functionality is conservatively gated behind a new EShMsgValidateCrossStageIO flag (and corresponding --validate-io option in the standalone executable).

The bulk of the code changes in this PR are concentrated at the end of the existing TIntermediate::checkStageIO() function (where the aforementioned TODO existed previously).

At a high level, each pair of logically adjacent stages in a program is traversed in a liveness-aware manner to compare outputs declared in each earlier stage against inputs actually read/accessed in each corresponding later stage. Live inputs are matched to outputs using several methods, and any live inputs remaining at the end of this process that haven't been matched to any output produce linker errors. To elaborate:

First, all outputs declared in the earlier stage along with all inputs that are statically accessed in the later stage are collected into temporary sequences of symbols. Per the GLSL specification, inputs are only required to have matching outputs if they are statically used, so code paths that can be statically determined to not be live are ignored in the traversal of the later stage (but obviously not in the traversal of outputs in the earlier stage).

The sequence of output symbols in the earlier stage is then iterated (and output blocks parsed accordingly) in order to populate a set of all explicitly declared output locations, including locations comprising arrayed and/or block IO that spans ranges of multiple consecutive locations. Outputs without explicit locations (i.e., whose locations will be assigned later during IO mapping) will be matched by name later, so they are ignored during this step.

Next, the sequence of live inputs in the later stage is iterated: inputs with explicit locations are checked to determine whether their location(s) exist(s) in the set of declared outputs in the earlier stage, and inputs without explicit locations are checked against outputs in the earlier stage to try to find an output with a matching name and type. Inputs matched to outputs by either of these means are removed from the temporary sequence of live inputs.

Then, remaining non-block live input variables are matched against output block member names, and likewise removed from the temporary sequence of input symbols.

Finally, remaining live block inputs are checked for matching loose (non-block) outputs in the previous stage. Liveness of block members is not (yet) tracked at a per-member granularity (though this could be implemented in the future), so any member of an input block that matches any output in the previous stage causes all other members of that block to be considered matched. This can be overly permissive in certain cases, but redesigning common liveness traversal functionality to track per-member liveness would be well out of scope for this PR. As in the earlier steps, inputs successfully matched to outputs at this point are removed from the temporary input symbol sequence.

Any statically used input variables remaining in the temporary input symbol sequence after the above process is complete don't have matching outputs in the earlier stage. In this case, linking is considered to have failed and an appropriate link error is reported for each such input. This approach ensures that all instances of inputs without corresponding outputs are reported to the user rather than only the first one.

Due to the high-level architecture of linking in glslang these kinds of link errors will always be produced by the later stage, so this PR also necessarily modifies TProgram::crossStageCheck() to check for errors in both stages (as opposed to only the earlier stage as it was doing previously) when determining whether linking should be considered to have failed.

During internal development and real world use of this functionality it was determined that glslang's existing link-time error reporting logic would need to be improved for these kinds of errors to be able to provide users with enough information to be meaningfully helpful. As a result, this PR introduces new overloads of TIntermediate::error() and TIntermediate::warn() which include source locations (in this case, of the actual static accesses to live input variables and not merely the variable declarations). Other link error/warning messages could be improved in the future to leverage these new functions to include appropriate source location information, but this kind of change is obviously out of scope for this PR and is therefore left to the maintainers and/or original authors of those messages.

This cross-stage IO validation implementation has a few known limitations that mean it doesn't necessarily detect all possible invalid usage patterns, but it intentionally prioritizes never giving false errors for valid usage. These limitations include:

The GLSL specification permits (and arguably encourages) the compiler to generate link-time warnings (but never errors) for cases where the compiler can statically determine that an output in one stage is declared but never written to but whose corresponding input is statically read from in the next stage. Due to limitations in TLiveTraverser this is currently not implemented, but the high-level structure of the cross-stage IO validation functionality has been designed with future support for such functionality in mind. A new TODO comment has been added in the appropriate part of TIntermediate::checkStageIO() in order to encourage further work on this.

As mentioned earlier, liveness of blocks is not currently tracked on an individual block member variable basis. This means all members of an input block are treated as having matching outputs if any one or more members match any output. Similarly, all indices of arrayed inputs are treated as having matching outputs if any one (or more) index matches any output. This means that the validation implementation is overly permissive in certain circumstances, and missing output declarations may not be reported in all cases where they ideally would be. Future improvements to TIOTraverser could address this limitation by collecting additional context about block member and array index access during traversal.

The GL_NV_geometry_shader_passthrough extension is not fully supported. When linking geometry shaders that use this extension, cross-stage IO validation of the geometry stage (but not of any other stages) is skipped and a warning to that effect is emitted. This limitation is only relevant in cases where:

- the user explicitly explicitly opts in to cross-stage IO validation, and
- the GL_NV_geometry_shader_passthrough extension is used

It's possible to implement support for this extension in the future, but it would likely require substantial refactoring of TIntermediate::checkStageIO() which would be out of scope for this PR.

Finally, (as in other parts of glslang) liveness-aware traversal is performed using TLiveTraverser, which currently only considers code to be statically not accessed when it is part of an if/else branch that can be statically determined to not be accessed, or inside a function that can be statically determined to never be called. This liveness-aware traversal logic could be improved in the future to handle return, discard, goto, and other types of control flow. One such enhancement to add support for liveness-aware traversal of switch/case statements has already been implemented as part of https://github.com/KhronosGroup/glslang/pull/3908. Future improvements to the accuracy of TLiveTraverser's liveness-aware traversal will have the positive side effect of continuing to improve the quality of cross-stage IO validation "for free".

If in the future all of these limitations are addressed, it may become viable to enable link time cross-stage IO validation by default.

This PR adds multiple positive and negative tests under both OpenGL and Vulkan semantics to exercise the newly implemented cross-stage IO validation functionality.

The first test (crossStageIO.0) consists of a vertex shader and a fragment shader with variously mismatching but valid combinations block and non-block IO between stages, with a goal of ensuring that as many valid usage patterns as are practical to test do not produce errors.

The second test (crossStageIO.1) is similar to the first (crossStageIO.0) but also includes a geometry shader to ensure that the IO semantics unique to that stage are handled properly.

The last test (missingCrossStageIO.0) is a negative test similar to the first test, except none of the inputs in the later stage have corresponding outputs in the earlier stage. It aims to ensure that appropriate errors are reported for various cases in which they are required by the GLSL specification.

In addition to adding these new tests, the common framework code for link-time tests has been updated to unconditionally opt into cross-stage IO validation for all existing tests, and (as expected) this has not resulted in any changes to the outputs of any tests that already existed.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -112,6 +112,7 @@ enum TOptions : uint64_t {
     EOptionCompileOnly = (1ull << 32),
     EOptionDisplayErrorColumn = (1ull << 33),
     EOptionLinkTimeOptimization = (1ull << 34),
+    EOptionValidateCrossStageIO = (1ull << 35),
 };
 bool targetHlslFunctionality1 = false;
 bool SpvToolsDisassembler = false;
@@ -907,6 +908,8 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
                         Options |= EOptionDisplayErrorColumn;
                     } else if (lowerword == "lto") {
                         Options |= EOptionLinkTimeOptimization;
+                    } else if (lowerword == "validate-io") {
+                        Options |= EOptionValidateCrossStageIO;
                     } else if (lowerword == "help") {
                         usage();
                         break;
@@ -1095,6 +1098,10 @@ void ProcessArguments(std::vector<std::unique_ptr<glslang::TWorkItem>>& workItem
     if ((Options & EOptionLinkTimeOptimization) && !(Options & EOptionLinkProgram))
         Error("link time optimization requires -l for linking");
 
+    // cross stage IO validation makes no sense unless linking
+    if ((Options & EOptionValidateCrossStageIO) && !(Options & EOptionLinkProgram))
+        Error("cross stage IO validation requires -l for linking");
+
     // -o or -x makes no sense if there is no target binary
     if (binaryFileName && (Options & EOptionSpv) == 0)
         Error("no binary generation requested (e.g., -V)");
@@ -1185,6 +1192,8 @@ void SetMessageOptions(EShMessages& messages)
         messages = (EShMessages)(messages | EShMsgDisplayErrorColumn);
     if (Options & EOptionLinkTimeOptimization)
         messages = (EShMessages)(messages | EShMsgLinkTimeOptimization);
+    if (Options & EOptionValidateCrossStageIO)
+        messages = (EShMessages)(messages | EShMsgValidateCrossStageIO);
 }
 
 //
@@ -2154,7 +2163,8 @@ void usage()
            "  --no-link                         Only compile shader; do not link (GLSL-only)\n"
            "                                    NOTE: this option will set the export linkage\n"
            "                                          attribute on all functions\n"
-           "  --lto                             perform link time optimization\n");
+           "  --lto                             perform link time optimization\n"
+           "  --validate-io                     validate cross stage IO\n");
 
     exit(EFailUsage);
 }
--- a/Test/baseResults/link.crossStageIO.0.vert.out
+++ b/Test/baseResults/link.crossStageIO.0.vert.out
@@ -1,0 +1,1701 @@
+link.crossStageIO.0.vert
+Shader version: 460
+0:? Sequence
+0:66  Function Definition: main( ( global void)
+0:66    Function Parameters: 
+0:67    Sequence
+0:67      Test condition and select ( temp void)
+0:67        Condition
+0:67        Constant:
+0:67          false (const bool)
+0:67        true case
+0:68        Sequence
+0:68          move second child to first child ( temp 4-component vector of float)
+0:68            'a0' (layout( location=40) smooth out 4-component vector of float)
+0:68            Constant:
+0:68              1.000000
+0:68              1.000000
+0:68              1.000000
+0:68              1.000000
+0:70      move second child to first child ( temp 4-component vector of float)
+0:70        a1: direct index for structure ( out 4-component vector of float)
+0:70          'anon@0' (layout( location=1) out block{ out 4-component vector of float a1})
+0:70          Constant:
+0:70            0 (const uint)
+0:70        Constant:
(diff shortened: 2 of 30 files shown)


jeremy-lunarg:
@daniel-story Can you take a look at the merge conflicts?

daniel-story:
> @daniel-story Can you take a look at the merge conflicts?

Everything should be fixed now.

arcady-lunarg (approved):
LGTM

PR #3923, opened 2025-04-09
Apply `NoContraction` to `dot()`

### LINKED ISSUE #2708: Should `NoContraction` apply to `dot()`?
The code [looks like][1] it means to apply `NoContraction` on `dot()`, but doesn't actually.  Test case:

[1]: https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/propagateNoContraction.cpp#L178

```
#version 450 core

precise gl_Position;

float u;

void main()
{
    vec4 v = vec4(u);
    precise float f = dot(v, v);
    gl_Position = vec4(f);
}
```

The above doesn't generate any `NoContraction` decorations.

mjn33:
Implement the change in issue #2708 so `noContraction` is set for `dot()` in the AST.

Translate `EOpDot` using `createBinaryOperation()`, which handles setting the `NoContraction` decoration unlike `createMiscOperation()`.

Fixes #2708.


mjn33:
Updated to fix a new test added in ba1640446f3826a518721d1f083f3a8cca1120c3 that was failing.

arcady-lunarg (approved):

PR #3932, opened 2025-04-16
validation fails for spv.exportFunctions.comp when no linking performed

dneto0:
The generated SPIR-V uses OpCapability Linkage.
That's not allowed in the Vulkan environment, and the SPIRV-Tools validator issues a validation error.


jeremy-lunarg:
Rebased on #3936 to fix CI error.

jeremy-lunarg:
[This is the origin](https://github.com/KhronosGroup/glslang/pull/3319) of the suspect test. I spoke to @ncesario-lunarg and he believes this may no longer be used. VVL has moved to a different approach. Can @spencer-lunarg confirm?

jeremy-lunarg (approved):
LGTM. Thanks for the fix.

spencer-lunarg:
Until we move off of GLSL for something else, we still use `--no-link` and write our error checking logic in GLSL, from here we link it into the SPIR-V with our internal SPIR-V code, so we can easily adapt to any changes needed with how `--no-link` works

PR #3934, opened 2025-04-18
Fix exceptions being unconditionally disabled on GCC

jan-kelemen:
- Remove `-fno-exceptions` from unconditional compilation flags

It's added conditionally depending on the CMake option a few lines below.


arcady-lunarg (approved):
LGTM, this makes sense, I think it was added inadvertently in commit 8004d3652870b4dccf5fb1a1f266544a0a2a50e6

PR #3936, opened 2025-04-18
Update Ubuntu runners

### LINKED ISSUE #3933: Linux backcompat jobs fail: Ubuntu 20.04 runner has disappeared
The Linux Backcompat checks are now failing, as of April 15. The Ubuntu 20.04 runner has been retired.

See https://github.com/actions/runner-images/issues/11101 

The other runners use Ubuntu 22.04.    The options seem to be:
- delete it
- make it a no-op while the other runners are still 22.04
- update it to 22.04 and update other runners to a more recent Ubuntu.

jeremy-lunarg:
Fix #3933

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -37,7 +37,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [{genus: ubuntu-20.04, family: linux}]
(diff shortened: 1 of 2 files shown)


jeremy-lunarg:
@arcady-lunarg I tried to update the 22.04 runners to 24.04 but encountered warnings being treated as errors that I don't want to deal with ATM.

arcady-lunarg (approved):

PR #3937, opened 2025-04-19
Replace spirv.hpp with spirv.hpp11

jan-kelemen:
glslang uses `spirv.hpp` from SPIRV-Headers, while the SPIRV-Tools use `spirv.hpp11` . Linking glslang and SPIRV-Tools leads to an ODR violation that's actually being detected by GCC:
```
/home/runner/.conan2/p/b/glslade4429c0836b8/b/src/SPIRV/spirv.hpp:943: warning: type ‘spv::Scope’ violates the C++ One Definition Rule [-Wodr]
  943 | enum Scope {
/home/runner/.conan2/p/b/spirv0adda0305824e/p/include/spirv/unified1/spirv.hpp11:943: note: an enum with different value name is defined in another translation unit
  943 | enum class Scope : unsigned {
/home/runner/.conan2/p/b/glslade4429c0836b8/b/src/SPIRV/spirv.hpp:944: note: name ‘ScopeCrossDevice’ differs from name ‘CrossDevice’ defined in another translation unit
  944 |     ScopeCrossDevice = 0,
  ```
  
This PR ports the code in glslang to use `spirv.hpp11`, which should be OK according to https://github.com/KhronosGroup/glslang/pull/3644#pullrequestreview-2157015908

I tried to keep the diff as clean as possible and close to the original code to make it easier for review.

### CODE REVIEW

File: SPIRV/spirv.hpp11
Comment (jeremy-lunarg) [inline SPIRV/spirv.hpp11:1]: Can you preserve the line endings from spirv-headers? This file has carriage-return and line-feed. It's the only difference with the source.

**Comment by jeremy-lunarg** [comment]
The license check is complaining:

> 2 errors:
> * SPIRV/spirv.hpp11 has no license
> * SPIRV/spvUtil.h has no license

spvUtil.h is missing a license.

spirv.hpp11 certainly has one but it isn't being recognized? @dneto0?

**Comment by jan-kelemen** [comment]
> The license check is complaining:
> 
> > 2 errors:
> > 
> > * SPIRV/spirv.hpp11 has no license
> > * SPIRV/spvUtil.h has no license
> 
> spvUtil.h is missing a license.
> 
> spirv.hpp11 certainly has one but it isn't being recognized? @dneto0?

I can add a license to the `spvUtil.h`, just let me know which license text I should put. I wasn't sure of that as there are multiple ones in the repository.

**Comment by dneto0** [comment]
> > The license check is complaining:
> > > 2 errors:
> > > 
> > > * SPIRV/spirv.hpp11 has no license
> > > * SPIRV/spvUtil.h has no license
> > 
> > 
> > spvUtil.h is missing a license.
> > spirv.hpp11 certainly has one but it isn't being recognized? @dneto0?
> 
> I can add a license to the `spvUtil.h`, just let me know which license text I should put. I wasn't sure of that as there are multiple ones in the repository.

Sorry this is not obvious.....  Glslang has code from a bunch of different sources, with some generated code, and that means a lot of licensing complexity.    So we have a config file to specify the rules and exceptions.   


1. To fix spirv.hpp11, modify the line in `license-checker.cfg` in the root of the project to mark it as an exception.  change `spirv.hpp` to `spirv.hpp11`  [here](https://github.com/KhronosGroup/glslang/blob/4cf9d467c7c57679a0846b61a3d3b8d55574702c/license-checker.cfg#L40)

2. To fix spvUtil.h,  please use the 3-clause BSD license, for example copy it from from [Scan.h](https://github.com/KhronosGroup/glslang/blob/main/glslang/MachineIndependent/Scan.h)  then change the copyright line to name yourself.  For example, the top of the file should look like:

```
//
// Copyright (C) 2025 Your name here
//
// All rights reserved.
//
// Redistribution and use in source and binary forms, with or without
// modification, are permitted provided that the following conditions
// are met:
...
```
Where you put your own name there.


Thanks for your contribution and your patience!

**Comment by dneto0** [comment]
If you have Linux x86-64 and Docker, you can run the license checker yourself:

From the glslang project root directory, run:

```shell
docker run -v `pwd`:/src --workdir /src  us-east4-docker.pkg.dev/shaderc-build/radial-docker/ubuntu-24.04-amd64/license-checker
```

jan-kelemen:
The licence check should be OK now, checked it locally with the docker command too.

SPIRV/SpvBuilder.cpp
```diff
@@ -2953,13 +2954,22 @@ void Builder::createControlBarrier(Scope execution, Scope memory, MemorySemantic
 
 void Builder::createMemoryBarrier(unsigned executionScope, unsigned memorySemantics)
```
jeremy-lunarg, line 2955: Can this one be removed?

jan-kelemen: Yup

jeremy-lunarg:
Overall, this looks good. Could you please squash your commits?

jan-kelemen:
Squashed and rebased onto current main branch

jeremy-lunarg (approved):

jeremy-lunarg:
> Squashed and rebased onto current main branch

Fantastic. Thank you!

dneto0:
Thanks very much for this contribution!

PR #3938, opened 2025-04-20
Fix continuous deployment

jeremy-lunarg:
Remove package that was required when CI was ported from Travis but is no longer required.


arcady-lunarg (approved):
LGTM

PR #3940, opened 2025-04-21
Support emitting DebugSourceContinued instruction

qingyuanzNV:
This patch adds support for emitting DebugSourceContinued for extremely large source file that couldn't fit in one OpString. In addition to that, it also adds basic support for using std::string_view to query an OpString or a string operand to avoid unnecessary copying the string.


arcady-lunarg:
This was already implemented by an earlier PR.

PR #3949, opened 2025-04-29
HlslParseContext should respect global uniform block name override by intermediate

MohammadMDSA:
HlslParseContext should respect global uniform block name override by intermediate.
Currently global uniform block name resolved by hlsl parse context is always "$Global" (not even "$Globals" for compatibility with DXC). While the TShader accepts override name for the global block and even TParseContext respects it, it wouldn't work with hlsl. With this change, HlslParseContext would also respect the name override.

### DIFF
--- a/glslang/HLSL/hlslParseHelper.cpp
+++ b/glslang/HLSL/hlslParseHelper.cpp
@@ -10284,4 +10284,14 @@ void HlslParseContext::finish()
     TParseContextBase::finish();
 }
 
+
+const char* HlslParseContext::getGlobalUniformBlockName() const
+{
+    const char* name = intermediate.getGlobalUniformBlockName();
+    if (std::string(name) == "")
+        return "$Global";
+    else
+        return name;
+}
+
 } // end namespace glslang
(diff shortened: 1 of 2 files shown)


glslang/HLSL/hlslParseHelper.cpp
```diff
@@ -10291,4 +10284,14 @@ void HlslParseContext::finish()
     TParseContextBase::finish();
 }
 
+
+const char* HlslParseContext::getGlobalUniformBlockName() const
+{
+    const char* name = intermediate.getGlobalUniformBlockName();
+    if (std::string(name) == "")
```
dnovillo, line 10291: This is allocating a string just to compare against an empty literal. Please compare `name[0] == '\0'`.

dnovillo (changes requested):
There are no test exercising this new functionality. Could you add some?

MohammadMDSA:
Hey @dnovillo I'd love to, but because of the agreement I've signed with my workplace, I can no longer contribute to this repo. I am so sorry

dnovillo:
> Hey @dnovillo I'd love to, but because of the agreement I've signed with my workplace, I can no longer contribute to this repo. I am so sorry

OK, thanks. Absolutely fine. In that case, it's probably better to close this PR. If in the future you happen to be able to contribute again, you can re-open it. Thanks.

PR #3954, opened 2025-05-12
Remap discard to OpDemoteToHelperInvocation

daniel-schuermann:
(Moved from the internal repository)

The current behavior is:

if SPIR-V 1.6 is not supported, use (the deprecated) `OpKill`.
If SPIR-V 1.6 is supported and the source is HLSL, use `OpDemoteToHelperInvocation`, otherwise use `OpTerminateInvocation` (bad!).

As developers aren't generally aware of the subtle differences, this behavior blesses us with reoccurring application bugs.

The idea of `EXT_demote_to_helper_invocation` and `EXT_terminate_invocation` was to clarify the behavior and not to bring additional inconsistencies and confusion. Now, what happens is that graphics drivers are forced to bend backwards over to fix the weird mismatch between spec, compilers and application expectations. I am not aware of any OpenGL or Vulkan driver which still implements `discard`/`OpKill` as `terminate`, but the overall broad consensus seems to be to implement it as `demote`. I'm also not aware of any application which expects the `terminate` behavior, but there are countless of applications which would misrender with such an implementation.

Now what happens with glslang is that there are still new applications coming out which use `OpTerminateInvocation` and expect it to behave as `demote`. We went full circle with two extensions just to be in the exact same situation as before.

In SPIR-V `OpKill` is deprecated, and the semantics have been updated to allow for implementations with `demote` behavior (which is literally the default now). If the `discard` keyword is also deprecated in GLSL, I don't see any reason not map it to `OpKill` and let the drivers fix it. Otherwise, I highly recommend to just map it to `OpDemoteToHelperInvocation` and call it a day. Let's please eliminate this root cause of countless application bugs.

This MR implements the latter and remaps discard to demote, regardless of the source language.


HansKristian-Work:
Overall the ecosystem expects demote behavior rather than terminate. Given that discard; has meant OpKill which has effectively become OpDemote anyway, GLSL should never emit OpTerminate unless explicitly requested, since it's almost never what you want. Emitting terminate by default has caused real damage in the ecosystem already.

arcady-lunarg (approved):

daniel-schuermann:
ping. We have reports of another game with the same bug. Can we merge this asap? It already caused enough havoc :/

arcady-lunarg:
Given people are complaining about the absence of this and nobody complained about merging this, I'll just merge.

PR #3955, opened 2025-05-12
Fix logic in CMakeFiles for tests on Windows

philippemilink:
The logic in the condition to include tests on Windows in CMakeFiles was wrong: if one of `WIN32`, `BUILD_SHARED_LIBS` or `CMAKE_VERSION VERSION_LESS "3.27"` was false (eg having CMake version 3.25 on Linux with enabled shared libs) would try to add the test and to execute `$<TARGET_RUNTIME_DLL_DIRS:spirv-remap>` which was introduced only in CMake 3.27.

This PR fixes this (and I also removed a spurious `if (BUILD_SHARED_LIBS)`).

The problem was introduced by commit b3a6aa7b03c51ba976e4f4e96b1e31f77f43f312 and the CI didn't spot it because `BUILD_SHARED_LIBS` is set to `OFF` by default and the presence of the (not so spurious in this buggy case)  line `if (BUILD_SHARED_LIBS)`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,16 +338,16 @@ if(GLSLANG_TESTS)
         set(REMAP_PATH ${CMAKE_CURRENT_BINARY_DIR}/StandAlone/spirv-remap)
     endif()
 
-    # The TARGET_RUNTIME_DLL_DIRS feature requires CMake 3.27 or greater.
-    if(WIN32 AND BUILD_SHARED_LIBS AND CMAKE_VERSION VERSION_LESS "3.27")
-        message(WARNING "The Windows shared library test configuration requires CMake 3.27 or greater")
-    else()
-        add_test(NAME glslang-testsuite
-            COMMAND bash ${IGNORE_CR_FLAG} runtests ${RESULTS_PATH} ${VALIDATOR_PATH} ${REMAP_PATH}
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -338,16 +338,16 @@ if(GLSLANG_TESTS)
...
-            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Test/)
-
-        # Prepend paths to shared libraries.
-        if (BUILD_SHARED_LIBS)
+    if(WIN32 AND BUILD_SHARED_LIBS)
+        # The TARGET_RUNTIME_DLL_DIRS feature requires CMake 3.27 or greater.
+        if(CMAKE_VERSION VERSION_LESS "3.27")
+            message(WARNING "The Windows shared library test configuration requires CMake 3.27 or greater")
+        else()
+            add_test(NAME glslang-testsuite
```
svenvh, line 346: How is this `add_test` supposed to be run for non-WIN32 configurations now?  Did this change effectively disable testing for non-WIN32 platforms?

See e.g. https://github.com/KhronosGroup/glslang/actions/runs/15594520964/job/43921605480?pr=3973
```
Test project /home/runner/work/glslang/glslang/build
No tests were found!!!
```

dnovillo: @philippemilink perhaps the else() branch of the main if(WIN32... also needs add_test()

PR #3958, opened 2025-05-21
Add GLSLANG_ prefix to CMake options

ntsh-oni:
glslang has two CMake options: BUILD_EXTERNAL and BUILD_WERROR.

If glslang is linked as a library in a project, this is an issue as, on CMake GUI, these options appear as-is, which doesn't indicate where they come from. This can also conflict with other libraries which also do not prefix their CMake options.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,8 +46,8 @@ include(CMakePackageConfigHelpers)
 include(CMakeDependentOption)
 
 option(BUILD_SHARED_LIBS "Build Shared Libraries")
-option(BUILD_EXTERNAL "Build external dependencies in /External" ON)
-option(BUILD_WERROR "Enable warnings as errors (default is OFF)" OFF)
+option(GLSLANG_BUILD_EXTERNAL "Build external dependencies in /External" ON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -46,8 +46,8 @@ include(CMakePackageConfigHelpers)
 include(CMakeDependentOption)
 
 option(BUILD_SHARED_LIBS "Build Shared Libraries")
-option(BUILD_EXTERNAL "Build external dependencies in /External" ON)
-option(BUILD_WERROR "Enable warnings as errors (default is OFF)" OFF)
+option(GLSLANG_BUILD_EXTERNAL "Build external dependencies in /External" ON)
+option(GLSLANG_BUILD_WERROR "Enable warnings as errors (default is OFF)" OFF)
```
charles-lunarg, line 50: I would add a deprecation notice, since changing variable names like this is sure to catch developers off guard.

```cmake
if (DEFINED BUILD_EXTERNAL)
    message(DEPRECATION "The BUILD_EXTERNAL build options has been deprecated and replaced by GLSLANG_BUILD_EXTERNAL, please use the new variable")
endif()
```

Or a similar message to get the point across.

PR #3959, opened 2025-05-21
cmake: Remove BUILD_SHARED_LIBS option

charles-lunarg:
BUILD_SHARED_LIBS is a built in CMake variable. Because the code simply declares the varriable (ie, doesn't change the default value), it was redundant and caused confusion for users unfamiliar with CMake & its built in's.

CMake documentation for the curious:
https://github.com/KhronosGroup/glslang/compare/main...charles-lunarg:glslang:remove_built_in_cmake_option?expand=1


arcady-lunarg (approved):
And apparently has been this way at least since cmake 3.0.

PR #3962, opened 2025-05-28
Fix crash on unnamed function parameters under relaxed rules

sbourasse:
Old OpenGL-compatible code was allowed to contain variables named `sampler`.
This is not true under Vulkan semantics anymore, as this has become a reserved keyword.
Compiling shaders with function parameters named like this generates an unnamed parameter followed by a `SAMPLER` token.
Before this fix, such behaviour resulted in a segfault, as `vkRelaxedRemapFunctionParameter()` always expected the input variable to be named.
This PR addresses this issue by adding a function to check whether a type has a field name, and if not, `vkRelaxedRemapFunctionParameter()` generates all variables without prefixing them with anything.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -7964,12 +7964,15 @@ static void ForEachOpaque(const TType& type, const TString& path, Function callb
                  ++flatIndex)
             {
                 TString subscriptPath = path;
-                for (size_t dimIndex = 0; dimIndex < indices.size(); ++dimIndex)
+                if (path != "")
                 {
-                    int index = indices[dimIndex];
-                    subscriptPath.append("[");
-                    subscriptPath.append(String(index));
-                    subscriptPath.append("]");
+                    for (size_t dimIndex = 0; dimIndex < indices.size(); ++dimIndex)
+                    {
+                        int index = indices[dimIndex];
+                        subscriptPath.append("[");
+                        subscriptPath.append(String(index));
+                        subscriptPath.append("]");
+                    }
                 }
 
                 recursion(type, subscriptPath, true, recursion);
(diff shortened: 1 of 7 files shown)


dnovillo:
It seems that this is fixed on mainline already. Can you confirm? I'm getting:

```
./glslang -C -V vk.relaxed.syntaxerror.frag
vk.relaxed.syntaxerror.frag
ERROR: vk.relaxed.syntaxerror.frag:5: 'location' : SPIR-V requires location for user input/output
ERROR: vk.relaxed.syntaxerror.frag:20: '' :  syntax error, unexpected SAMPLER, expecting RIGHT_PAREN
```

sbourasse:
My apologies, I forgot about giving a repro case for the issue at hand.
The command to trigger the crash is `glslangValidator -C -V -R vk.relaxed.syntaxerror.frag`, you should be able to reproduce on mainline with this.
Thanks for checking in on this.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -8050,9 +8056,13 @@ void TParseContext::vkRelaxedRemapFunctionParameter(TFunction* function, TParame
...
 
-    ForEachOpaque(*param.type, (param.name ? *param.name : param.type->getFieldName()),
+    TString fieldName = param.name
+        ? *param.name
+        : param.type->hasFieldName() ? param.type->getFieldName() : "";
+
+    ForEachOpaque(*param.type, fieldName,
                   [function, param, newParams](const TType& type, const TString& path) {
-                      TString* memberName = NewPoolTString(path.c_str());
+                      TString* memberName = path != "" ? NewPoolTString(path.c_str()) : nullptr;
```
dnovillo, line 8065: I'm thinking it would be better to make `memberName` and empty or some other marker string here instead of `nil`. Further references to function parameters may get in similar troubles later on.

dnovillo (changes requested):
Largely LGTM. One potential issue with it.

glslang/MachineIndependent/ParseHelper.cpp
sbourasse: You make a good point.
I went and looked through the file to make sure storing an empty string wouldn't break anything, and it seems that anonymous parameters are already represented with a `nil`  name.
See the comment on this line for reference https://github.com/sbourasse/glslang/blob/d177400d02b1a3e82c83583ac718a6f01106905f/glslang/MachineIndependent/ParseHelper.cpp#L1290

dnovillo: Ah, yes. Thanks for confirming. LGTM.

dnovillo (approved):

PR #3968, opened 2025-06-04
Fix bug in emitting correct capability for LSS built-in functions.

alelenv:
Code was in wrong location for emitting capability / extension


alelenv:
Issue was discovered in internal testing with this small shader

```
#version 460
#extension GL_NV_shader_invocation_reorder : enable
#extension GL_NV_linear_swept_spheres : enable

void main()
{
        hitObjectNV hObj;
        bool temp = hitObjectIsSphereHitNV(hObj);
}
```

Since `hitObjectIsSphereHitNV` alone was used, correct capability `RayTracingSpheresGeometryNV` was not being added as per original change. Current change fixes this

Glslang Unit tests didnt catch this issue since they test all built-ins together in one file, and other built-in `rayQueryGetIntersectionSpherePositionNV` ended up adding this capability

dnovillo:
Thanks for the extra details. Is adding this to Test/ or gtests/ possible? It'd be nice to have a test that checks that we are emitting both the capability (`RayTracingSpheresGeometryNV`) and extension (`SPV_NV_linear_swept_spheres`). Though I think we just need to run the validator to confirm.

I'm not sure how feasible this is atm. If it's a hassle, let me know and I'll see about adding the option to have unit tests like this one.

diegonovillo:
Thanks! LGTM

dnovillo (approved):

PR #3973, opened 2025-06-11
Error for in/out structs containing booleans

rj123-nv:
The GLSL spec says:
It is a compile-time error to declare a fragment shader input with, or that contains, any of the following types:
 -a boolean type
 -an opaque type
and similar rules for other shader stages and 'out' variables.

Previously glslang would error for simple 'bool' IO variables, but not structs/blocks containing them.

This could lead to generating invalid spir-v, violating this validator rule:

"If OpTypeBool is stored in conjunction with OpVariable using Input or Output Storage Classes it requires a BuiltIn decoration"
- VUID 7290

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -4450,11 +4450,6 @@ void TParseContext::globalQualifierTypeCheck(const TSourceLoc& loc, const TQuali
 
     // now, knowing it is a shader in/out, do all the in/out semantic checks
 
-    if (publicType.basicType == EbtBool && !parsingBuiltins) {
-        error(loc, "cannot be bool", GetStorageQualifierString(qualifier.storage), "");
-        return;
-    }
-
     if (isTypeInt(publicType.basicType) || publicType.basicType == EbtDouble) {
         profileRequires(loc, EEsProfile, 300, nullptr, "non-float shader input/output");
         profileRequires(loc, ~EEsProfile, 130, nullptr, "non-float shader input/output");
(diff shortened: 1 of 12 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -6943,6 +6938,22 @@ void TParseContext::layoutObjectCheck(const TSourceLoc& loc, const TSymbol& symb
...
+    if (!parsingBuiltins && type.containsBasicType(EbtBool) && !builtInName(symbol.getName())) {
+        switch(qualifier.storage) {
+        case EvqVaryingIn:
+        case EvqVaryingOut:
+        {
+            const char *reason = type.getBasicType() == EbtBool ? "cannot be bool" : "cannot contain bool";
+            error(loc, reason, GetStorageQualifierString(qualifier.storage), "");
+            break;
+        }
+        default:
```
dnovillo, line 6953: This will make the compiler to silently reject without stating why. If this default case can never happen, I think it'd be better to either emit a generic error or an ICE.

rj123-nv: With the code in this PR, the default case here wont produce an error or silently fail, just continue. E.g. if there was a 'bool' in an SSBO or UBO, it wouldnt fail, at least not here. Im not clear on the rules for those, I was mainly focusing on the spec wrt to in and out variables.

dnovillo: Ah, yes. I see now. Thanks for the clarification. This is fine.

dnovillo (approved):

PR #3976, opened 2025-06-16
ShaderLang: use static local variable.

h-friederich:
These functions may be called in a static initialization context, i.e. `init_lock` may not have been initialized yet. To fix, use a static local variable, which gets initialized upon first access.


h-friederich:
> [...] the default initializer for std::mutex is constexpr so no code is actually run at initialization time

That's not what I'm observing when using MSVC in debug mode. Unlike release mode there is nontrivial initialization happening.

> Furthermore, your change introduces a race condition: what if two threads run the function at the same time?

initialization of static local variables is guaranteed to be thread safe by the C++ standard. Also on Windows...

PR #3983, opened 2025-06-22
fix handling of values around infinity for e4m3 constants

### LINKED ISSUE #3981: E4M3 constants don't consider extended range above 256.0
E4M3 goes up to 448.0 since there is no inf encoding. glslang's generic float code does not seem to take this into account and clamps floate4m3_t(448.0) to 256.0 in the SPIR-V.

jeffbolznv:
Fixes #3981.


jeremy-lunarg (approved):
LGTM

PR #3984, opened 2025-06-24
Add GL_ARM_tensors

svenvh:
Add support for the GL_ARM_tensors extension.  The corresponding specification can be found at https://github.com/KhronosGroup/GLSL/pull/287 .


jeremy-lunarg:
@svenvh I'll take a look. I think I can refactor that script to avoid that limit.

PR #3986, opened 2025-06-24
Workaround Windows command line limit

jeremy-lunarg:
Bash-like environments in Windows have a 32k command line limit. Eventually the number of tests exceeded that.

### DIFF
--- a/Test/runtests
+++ b/Test/runtests
@@ -82,14 +82,16 @@ diff -b $BASEDIR/hlsl.automap.frag.out "$TARGETDIR/hlsl.automap.frag.out" || HAS
 # multi-threaded test
 #
 echo Comparing single thread to multithread for all tests in current directory...
-run -i -C *.vert *.geom *.frag *.tesc *.tese *.comp > "$TARGETDIR/singleThread.out"
-run -i -C *.vert *.geom *.frag *.tesc *.tese *.comp -t > "$TARGETDIR/multiThread.out"
-diff "$TARGETDIR/singleThread.out" "$TARGETDIR/multiThread.out" || HASERROR=1
-if [ $HASERROR -eq 0 ]
-then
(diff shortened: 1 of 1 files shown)


jeremy-lunarg:
@dnovillo Can you do a quick review when you have a moment? This is blocking [another PR](#3984).

svenvh (commented):
LGTM, thanks!  Eventually this could hit the same limit again (if there are a large enough number of tests for a single stage), but this change should alleviate the problem for now.

Test/runtests
```diff
@@ -82,14 +82,16 @@ diff -b $BASEDIR/hlsl.automap.frag.out "$TARGETDIR/hlsl.automap.frag.out" || HAS
...
+for stage in vert frag geom tesc tese comp; do
+    run -i -C *.$stage > "$TARGETDIR/$stage.singleThread.out"
+    run -i -C *.$stage -t > "$TARGETDIR/$stage.multiThread.out"
+    diff "$TARGETDIR/$stage.singleThread.out" "$TARGETDIR/$stage.multiThread.out" || HASERROR=1
+    if [ $HASERROR -eq 0 ]
+    then
+        rm "$TARGETDIR/$stage.singleThread.out"
+        rm "$TARGETDIR/$stage.multiThread.out"
+    fi
+done
```
dnovillo, line 94: Is xargs an option in Windows?  I'm wondering if this will also run into problems eventually (though it's now much less likely).

jeremy-lunarg: I believe `xargs` will break up the input set into separate calls to `glslang` as necessary. I'm concerned this could lead to sporadic false positives in the output.

jeremy-lunarg: Also, you are correct that this could lead to the same problem in the future. See my comment above.

PR #3987, opened 2025-06-25
Export spv::Parameterize

Flakebi:
`spv::Parameterize()` needs to be called before working with SPIR-V. When using glslang as a library, not through the disassemble or remap entry-points, `Parameterize` needs to be called explicitly. Therefore the function needs to be exported.
As `doc.h` is used in spirv-remap, the `visibility.h` header cannot be included, so redefine the `GLSLANG_EXPORT` macro like it is done in `SPVRemapper.h`.

Fixes a build with `-DBUILD_SHARED_LIBS=ON`.

For reference, we call `spv::Parameterize()` here:
https://github.com/GPUOpen-Drivers/spvgen/blob/ba3bda984defc22aadb20f3e1fdfaf23972cf636/source/spvgen.cpp#L1748

### DIFF
--- a/SPIRV/doc.h
+++ b/SPIRV/doc.h
@@ -42,10 +42,25 @@
 
 #include <vector>
 
+#ifdef GLSLANG_IS_SHARED_LIBRARY
+    #ifdef _WIN32
+        #ifdef GLSLANG_EXPORTING
+            #define GLSLANG_EXPORT __declspec(dllexport)
+        #else
+            #define GLSLANG_EXPORT __declspec(dllimport)
+        #endif
+    #elif __GNUC__ >= 4
+        #define GLSLANG_EXPORT __attribute__((visibility("default")))
+    #endif
+#endif // GLSLANG_IS_SHARED_LIBRARY
+#ifndef GLSLANG_EXPORT
+#define GLSLANG_EXPORT
+#endif
+
 namespace spv {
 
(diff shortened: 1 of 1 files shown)


dnovillo:
Can you describe a bit more what you are trying to do but can't? May be we can have a better supported entry point that does the necessary SPIR-V metadata and operand info setup.  Duplicating the GLSLANG_EXPORT logic is undesirable.

SPIRV/doc.h
```diff
@@ -42,10 +42,25 @@
...
+#endif // GLSLANG_IS_SHARED_LIBRARY
+#ifndef GLSLANG_EXPORT
+#define GLSLANG_EXPORT
+#endif
+
 namespace spv {
 
 // Fill in all the parameters
-void Parameterize();
+GLSLANG_EXPORT void Parameterize();
```
dnovillo, line 63: This is not something we should need to export. We may be missing an entry point. The fact that this forced you to duplicate the GLSLANG_EXPORT machinery is a sign.

dnovillo (changes requested):

Flakebi:
Ok, in that case I’ll abandon this for now

arcady-lunarg:
@Flakebi What entry point are you calling that doesn't call `Parameterize()`? Or rather, what happens when you just remove the call to `Parameterize()`?

PR #3995, opened 2025-07-11
Add explicit type support for bitfieldReverse

gnl21:
This was missing for 8- and 16-bit types.


gnl21:
I tried to add a test for these functions to `Test/` but I couldn't find any evidence that the `glsl.*` tests were actually running in the test suite, so it had no effect. Do the tests need adding to an explicit list? If so, where?

dnovillo:
> I tried to add a test for these functions to `Test/` but I couldn't find any evidence that the `glsl.*` tests were actually running in the test suite, so it had no effect. Do the tests need adding to an explicit list? If so, where?

Perhaps look into how #3920 does it?

PR #3996, opened 2025-07-11
Fix erroneous parse error when declaring variables using struct names

### LINKED ISSUE #3931: Parse error when defining struct fields: unexpected TYPE_NAME, expecting IDENTIFIER
Hi,

This code leads to a parse error:
```glsl
#version 400
struct B
{
    vec3 t;
};
struct K
{
    float A, B; // parse error here
};
void main(){}
```

```
glslang.exe .\test.frag
ERROR: 0:8: '' :  syntax error, unexpected TYPE_NAME, expecting IDENTIFIER
```

To workaround this error, I can either:
- rename the field B (so that it's not the same name as the first struct;
- or split the line into `float A; float B;`
- or switch the order of fields: `float B, A;`

So the parse error happens when I define multiple fields of the same type, and one of them (except the first one) has the same name as a struct define above.

For the context, I'm working on a [shader minifier](https://github.com/laurentlb/shader-minifier), so using single-letter names is useful.

dnovillo:
This fixes #3931

The parser would reject a variable declaration like `float A, B;`  if struct named B existed in the same scope. It incorrectly treated B as a type name (TYPE_NAME) rather than as a variable identifier, resulting in a syntax error.

This updates the grammar to allow both `IDENTIFIER` and `TYPE_NAME` tokens for the first variable in a variable declaration (e.g., float A, B;), but only `IDENTIFIER` for subsequent variables after commas.

This is limited to variable declarations only.

Added some tests too.

### DIFF
--- a/Test/struct_name_as_array_member.frag
+++ b/Test/struct_name_as_array_member.frag
@@ -11,4 +11,4 @@ struct K
     float A[2], B[3]; // This should work - struct name B is used as an array member name.
 };
 
-void main(){} 
+void main(){}
--- a/Test/struct_name_as_struct_member.frag
+++ b/Test/struct_name_as_struct_member.frag
@@ -1,7 +1,7 @@
 #version 400
 
 // Original bug report from: https://github.com/KhronosGroup/glslang/issues/3931
-// Error was: "syntax error, unexpected TYPE_NAME, expecting IDENTIFIER" 
+// Error was: "syntax error, unexpected TYPE_NAME, expecting IDENTIFIER"
 // when defining a struct field with the same name as a previously defined
 // struct.
 struct B
@@ -17,4 +17,4 @@ struct K
 
 void main(){
   int x, B, y, K;
-} 
+}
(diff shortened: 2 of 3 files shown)


dnovillo:
@jeremy-lunarg I'm not overly confident that I'm using the correct approach here. Let me know if I'm missing anything obvious.

dnovillo:
> @dnovillo If we're going to change the grammar, should an issue be raised in the spec?

Perhaps. I was not completely sure if this was a bug in the implementation or the spec. I need to check the spec first.

dnovillo:
I think I found a way around the grammar changes. We were already handling name hiding in cases like this:

struct B {};
struct C { float B; };

In this case the field B inside C is already handled properly. So, I moved the fix to the scanner. We were getting confused in declarator lists. If instead we had

struct C { float x, B; };

the parser would get confused. The latest commit removes the grammar changes and adds better tracking of declarator lists in the scanner.

@jeremy-lunarg PTAL.

jeremy-lunarg (approved):
LGTM

Issue #4001, opened 2025-07-18
Syntax error from #3996


moritz-h:
I have observed "syntax error, unexpected IDENTIFIER" and assume the just merged PR #3996 is responsible for this. Using a struct as function parameter only works when the struct is the first parameter now. (CC @dnovillo @jeremy-lunarg)

```glsl
#version 400

struct B {
    int a;
};

void test1(B b, int c) {} // works
void test2(int c, B b) {} // does not work

void main(){}
```

dnovillo:
Thanks. I'll take a look.

ricoazzurro:
Causing mesa FTBFS.
```bash
[1392/1846] Generating src/vulkan/runtime/bvh/lbvh_generate_ir.spv.h with a custom command
FAILED: [code=2] src/vulkan/runtime/bvh/lbvh_generate_ir.spv.h 
/usr/bin/glslangValidator -V -I/blfs/x/mesa/src/vulkan/runtime/bvh --target-env spirv1.5 -x -o src/vulkan/runtime/bvh/lbvh_generate_ir.spv.h ../src/vulkan/runtime/bvh/lbvh_generate_ir.comp --quiet
ERROR: ../src/vulkan/runtime/bvh/vk_build_helpers.h:330: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking compute stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
[1394/1846] Generating src/vulkan/runtime/bvh/lbvh_main.spv.h with a custom command
FAILED: [code=2] src/vulkan/runtime/bvh/lbvh_main.spv.h 
/usr/bin/glslangValidator -V -I/blfs/x/mesa/src/vulkan/runtime/bvh --target-env spirv1.5 -x -o src/vulkan/runtime/bvh/lbvh_main.spv.h ../src/vulkan/runtime/bvh/lbvh_main.comp --quiet
ERROR: ../src/vulkan/runtime/bvh/vk_build_helpers.h:330: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking compute stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
[1395/1846] Generating src/vulkan/runtime/bvh/leaf.spv.h with a custom command
[...]

PR #4005, opened 2025-07-21
Fix erroneous parse errors in variable declarations using existing struct names

### LINKED ISSUE #3931: Parse error when defining struct fields: unexpected TYPE_NAME, expecting IDENTIFIER

dnovillo:
This fixes #3931 and #4001 (which was needed after the initial fix #3996).

The first commit is the direct re-application of the original fix #3996. The second commit fixes the case reported in #4001: struct types could not be used as function parameters when not in the first position. For example:

```
  void test1(B b, int c) {}  // worked
  void test2(int c, B b) {}  // failed with "unexpected IDENTIFIER"
```

The problem was that the declarator list logic was incorrectly being applied to function parameter lists. This adds parens depth tracking so function argument lists are not considered declarator lists (which were forcing IDENTIFIER recognition instead of TYPE).

Added the test from #4001
Adds reduced tests from the Tint failures reported in https://github.com/KhronosGroup/glslang/issues/4001#issuecomment-3097494774

Additionally, consolidated some existing tests for clarity.


dnovillo:
@Kangz, @ricoazzurro, @moritz-h could you please test this PR against the code that was failing to build for you? Thanks.

ricoazzurro:
@dnovillo 
All good for me now with this PR. Good job!
(My previous issue was mesa ftbfs.)

dnovillo:
> @dnovillo Yes works for me.
> 
> Looking at the diff, the two test files `Test/struct_parameter_after_non_struct.frag` and `Test/struct_name_in_function_args.frag` are identical, is this on purpose or unwanted?

Thanks. It was a side-effect of me re-factoring tests. I forgot to remove the duplicate.

Kangz:
@dnovillo this PR doesn't repro the previous failure so LGTM.

PR #4007, opened 2025-07-26
Fix bug in handling slice of array being passed cooperative matrix load/store functions and strengthen error checks.

alelenv:
Handle cases which were previously not working like
```
layout(binding = 0) buffer block
{ 
    float16_t f16arr[256][256];
};

coopMatLoadTensorNV( mat, f16arr[ 1 ], 0, t )
```
Also error out invalid uses for second argument buf[] to these intrinsics 
For ex

```
void main()
{
    float16_t f16arr[16];
    coopMatLoadTensorNV( mat, f16arr[ 1 ], 0, t )
}
```

Since spec states
_The "buf" arrays must be in either buffer storage or shared
    storage, and the array that is passed in can be sized or unsized._


alelenv:
@jeffbolznv to review

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3151,10 +3151,37 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
+
+    case EOpCooperativeMatrixLoad:
+    case EOpCooperativeMatrixLoadNV:
+    case EOpCooperativeMatrixLoadTensorNV:
+    case EOpCooperativeMatrixStore:
+    case EOpCooperativeMatrixStoreNV:
+    case EOpCooperativeMatrixStoreTensorNV:
+    {
+        const TIntermTyped *arg1 = (*argp)[1]->getAsTyped();
+        const TIntermTyped* base = TIntermediate::traverseLValueBase(arg1, true, false);
```
jeffbolznv: Why "false" for bufferReferenceOk?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -3151,10 +3151,37 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
+            error(loc, errMsg, fnCandidate.getName().c_str(), "");
+        }
+    }
+    break;
+
     default:
         break;
     }
 
+
```
jeffbolznv: remove these?

jeffbolznv (commented):
LGTM, just one question.

glslang/MachineIndependent/ParseHelper.cpp
alelenv: Mistake, good catch

glslang/MachineIndependent/ParseHelper.cpp
alelenv: Done

jeffbolznv (approved):

dnovillo (approved):
LGTM

PR #4022, opened 2025-08-19
Add checks to coopMatMulAdd

mmoult:
Validate properties of coopMatMulAdd such as:
* the inside dimension of the multiplication matrices must match
* the dimension of the addition matrix must match the product of the
  multiplication
* the usage property of all operands must match the expected value

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -1793,7 +1793,6 @@ void TParseContext::handleCoopMat2FunctionCall(const TSourceLoc& loc, const TFun
             auto cRows = cSize->getDimSize(1);
             auto cCols = cSize->getDimSize(2);
 
-            std::stringstream msg;
             if (aCols != bRows)
                 error(loc, "cannot multiply coop matrices with incompatible sizes",
                       sequence[0]->getAsSymbolNode()->getMangledName().c_str(),
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -1775,7 +1775,44 @@ void TParseContext::handleCoopMat2FunctionCall(const TSourceLoc& loc, const TFun
             // Set result type to match type of first parameter
             result->setType(result->getAsAggregate()->getSequence()[0]->getAsTyped()->getType());
         } else {
-            // For MulAdd, set result type to match type of C parameter
+            // The only remaining operation is MulAdd
+            assert(fnCandidate->getBuiltInOp() == EOpCooperativeMatrixMulAdd ||
+                   fnCandidate->getBuiltInOp() == EOpCooperativeMatrixMulAddNV);
+
+            // Validate that the matrix sizes are compatible for multiplication and addition
+            const auto &sequence = arguments->getAsAggregate()->getSequence();
+            const auto *aSize = sequence[0]->getAsTyped()->getType().getTypeParameters()->arraySizes;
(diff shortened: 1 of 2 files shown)


jeffbolznv:
Please do cleanup the commits. It doesn't need to be a new PR, you should be able to just force push.

Does this change handle both spec constant and literal dimensions correctly? I'd expect it to need to compare the TArraySize.

mmoult:
@jeffbolznv Done. Thank you for your feedback. There are several commits, but I assume that is ok since they will all be folded together when merged to the main branch.

I believe it is working with spec constants and with literals for the dimensions. I updated the example to demonstrate and verified locally. Did you have anything else in mind? If so, I would be happy to test it out if you give me an example.

jeffbolznv:
Hi, sure, I'll try to clarify a bit more. I think the following case should be invalid:

```
layout(constant_id = 0) const int M = 3;
layout(constant_id = 1) const int N = 3;

coopmat<float16_t, gl_ScopeSubgroup, M, M, gl_MatrixUseA> A;
coopmat<float16_t, gl_ScopeSubgroup, N, N, gl_MatrixUseB> B;
coopmat<float16_t, gl_ScopeSubgroup, M, N, gl_MatrixUseAccumulator> C;
coopMatMulAdd(A, B, C);
```

because A's cols and B's rows use different spec constants, even though they are initialized to the same value. I doubt the GL_KHR_cooperative_matrix spec is clear enough about this, but IMO that's the correct behavior based on how spec constant-sized array types are defined here: https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_vulkan_glsl.txt#L1207. I think this would be considered valid with your change because I think comparing getDimSize will compare the initializer values, not the spec constants (the 'nodes').

I think the right way to check for this is to use TArraySize::operator==. You can make helpers like:
```
// in TSmallArrayVector
    bool elementEqual(const TSmallArrayVector& rhs, int thisIdx, int rhsIdx) const {
        return sizes && rhs.sizes && thisIdx < sizes->size() && rhsIdx < rhs.sizes->size() &&
               (*sizes)[thisIdx] == (*rhs.sizes)[rhsIdx];
    }

// in TArraySizes
    bool elementEqual(const TArraySizes& rhs, int thisIdx, int rhsIdx) const { return sizes.elementEqual(rhs.sizes, thisIdx, rhsIdx); }
```

and then the checks would be something like:

```
ATypeParameters->arraySizes->elementEqual(*BTypeParameters->arraySizes, /*cols*/2, /*rows*/1);
```

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1777,7 +1778,75 @@ void TParseContext::handleCoopMat2FunctionCall(const TSourceLoc& loc, const TFun
...
+                    buf << "spec_const";
+                return buf.str();
+            };
+
+            auto aRowsStr = toString(aRows);
+            auto aColsStr = toString(aCols);
+            auto bRowsStr = toString(bRows);
+            auto bColsStr = toString(bCols);
+            auto cRowsStr = toString(cRows);
+            auto cColsStr = toString(cCols);
```
jeffbolznv: It would be nice to only construct the strings once the error occurs. Also, please add curly braces to all if/else. Otherwise, I think this looks good.

mmoult: I think this is fixed now. Thanks for your reviews and your help overall!

jeffbolznv (approved):

arcady-lunarg (changes requested):
You will need to add the new test file into the list of tests in one of the files under `gtests/` and generate an expected result file under `Test/out` using `glslangtests --update-mode`, because currently your new test isn't being run.

PR #4028, opened 2025-08-26
Fix inconsistent labeling of capability strings

### LINKED ISSUE #4018: A possible typo in SPIRV/doc.cpp ?
It seems "Capability" in the return value is not needed at L1117.

1117     case (int)Capability::ReplicatedCompositesEXT:             return "CapabilityReplicatedCompositesEXT";
1118 
1119     case (int)Capability::DotProductKHR:                       return "DotProductKHR";
1120     case (int)Capability::DotProductInputAllKHR:               return "DotProductInputAllKHR";
1121     case (int)Capability::DotProductInput4x8BitKHR:            return "DotProductInput4x8BitKHR";
1122     case (int)Capability::DotProductInput4x8BitPackedKHR:      return "DotProductInput4x8BitPackedKHR";

dnovillo:
This fixes #4018

Capability names do not include the "Capability" prefix, but some capabilities where being labeled with it. This PR addresses the inconsistency.

Tests adjusted to match the new labels.

### DIFF
--- a/SPIRV/doc.cpp
+++ b/SPIRV/doc.cpp
@@ -977,7 +977,7 @@ const char* CapabilityString(int info)
     case (int)Capability::SubgroupBallotKHR: return "SubgroupBallotKHR";
     case (int)Capability::DrawParameters:    return "DrawParameters";
     case (int)Capability::SubgroupVoteKHR:   return "SubgroupVoteKHR";
-    case (int)Capability::GroupNonUniformRotateKHR: return "CapabilityGroupNonUniformRotateKHR";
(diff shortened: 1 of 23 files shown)


dnovillo:
@jeremy-lunarg I'm not sure whether this will produce grief downstream. It does seem like we should be consistent in these labels.

jeremy-lunarg (approved):
LGTM

jeremy-lunarg:
> @jeremy-lunarg I'm not sure whether this will produce grief downstream. It does seem like we should be consistent in these labels.

I'm not sure either. I agree that it should be fixed.

PR #4029, opened 2025-09-01
Hide DefaultTBuiltInResource / Resources

timblechmann:
DefaultTBuiltInResource and Resources had global visibility and could potentially name clash with symbols of the same name when building as static libraries. Localizing them to that translation unit by making them static


arcady-lunarg (approved):
Doesn't hurt since they're not supposed to be publicly exposed.

PR #4040, opened 2025-09-19
Remove spirv-remapper

jeremy-lunarg:
The spirv-remap utility from glslang has been ported to the SPIRV-Tools repository as a new optimization pass called canonicalize-ids, available in spirv-opt. See spirv-opt --help for usage details.


dnovillo (approved):
LGTM. Thanks for pushing this through!

jeremy-lunarg:
@jeffbolznv I did not. I was not aware CTS was using it. I will investigate. Thanks Jeff.

jeffbolznv:
OK. My biggest concern is that CTS doesn't build against the latest glslang. I don't know why CTS needs to strip debug info. I've worked around it locally for now.

PR #4041, opened 2025-09-22
Fix issue where TShader VTable is not exported in shader library.

Kwizatz:
This seems to affect MingW and MacOS.

This also includes a change in Common.h to remove a MinGW deprecated security API.


Kwizatz:
See discussion at https://github.com/msys2/MINGW-packages/issues/25598

Kwizatz:
This should supersede #4032

glslang/Include/Common.h
```diff
@@ -67,13 +67,7 @@ std::string to_string(const T& val) {
 }
 #endif
 
-#if defined(MINGW_HAS_SECURE_API) && MINGW_HAS_SECURE_API
```
dnovillo, line 70: I'm not sure I understand this part. It doesn't seem to be related to the symbol export issue you are fixing here.

dnovillo (changes requested):

Kwizatz:
I've removed the extra unrelated change, please review. @arcady-lunarg @dnovillo 

Thanks!

Kwizatz:
Hi @dnovillo , can we kickstart the workflow again on this one? thanks!

dnovillo:
> Hi @dnovillo , can we kickstart the workflow again on this one? thanks!

Done. Thanks for the heads up. I missed it.

PR #4050, opened 2025-09-26
Fix flat interpolation check on fs inputs qualified with perPrimitiveNV

dreamelon:
Spec allows any interpolation qualifier on fragment shader inputs qualified with "perprimitiveNV".
https://github.com/KhronosGroup/GLSL/blob/d368fe808d7e11476945901120510edbe490f961/extensions/nv/GLSL_NV_mesh_shader.txt#L1119


dnovillo (changes requested):
The fix itself LGTM, but this needs a test. Please add a test that exercises this path. Maybe something like

```
#version 450
#extension GL_NV_mesh_shader : require
layout(location = 0) perprimitiveNV in int idx;
layout(location = 0) out vec4 fragColor;
void main() { fragColor = vec4(float(idx)); }
```

dnovillo:
@dreamelon are you still working on this PR?

Issue #3739, opened 2024-09-23
Targeting spv1.6 with MeshShader generates LocalSizeId instead of LocalSize


felix-ri:
Compiling this shader

```
#version 450
#extension GL_EXT_mesh_shader : enable
#pragma shader_stage(mesh)

layout(local_size_x = 128, local_size_y = 1, local_size_z = 1) in;

layout(triangles) out;
layout(max_vertices = 256, max_primitives = 256) out;

void main()
{
}
```

with spv1.5 (in glslc or glslvalidator) results in `OpExecutionMode LocalSize 128 1 1` being emitted, which seems fine.

However compiling it with spv1.6, it results in `OpExecutionModeId LocalSizeId 128 1 1`.

If I understand correctly, LocalSizeId is used, when you want to use specialization constants to specify the Workgroup Size. The numbers behind LocalSizeId are the Ids used to identify the specialization constants. This doesn't make any sense, because I explicitely specified `local_size_x` with a constant size (not an Id) instead of `local_size_x_id`.

If I understood the comments, LocalSize with a constant was to be deprecated in 1.6 (https://github.com/KhronosGroup/glslang/issues/2934#issuecomment-1116386068), but later was not deprecated (https://github.com/KhronosGroup/glslang/pull/3351#issue-1930982762). This pullrequest then fixed LocalSizeId to be only created, when needed, in *compute* shaders. Other shaders were not touched.

This issue (https://github.com/google/shaderc/issues/1414) might also be related to this (and I also commented there).

spencer-lunarg:
Same with Task shaders (which always requires `VK_KHR_maintenance4` now)

Simple reproduce case

https://godbolt.org/z/PvK8oWc4W
https://godbolt.org/z/oxE3M9j6P

> If I understand correctly, LocalSizeId is used

I have a write up on it in https://docs.vulkan.org/guide/latest/compute_shaders.html

But yes, `LocalSizeId` was added to "allow" the user to use spec constants... which `glslang` supports if you go

```
layout(local_size_x_id = 1, local_size_y_id = 2, local_size_z_id = 3) in;
```

but this is just a bug IMO

Issue #3510, opened 2024-02-13
OpExecutionModeId not used unless targeting SPIRV-1.6


castano:
It appears that [OpExecutionModeId](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpExecutionModeId) was introduced in SPIR-V 1.2. However, it is not used unless we target SPIR-V 1.6.

Here's a minimal example:

```
#version 450
layout (constant_id = 0) const uint local_size_x = 128;
layout (local_size_x_id = 0, local_size_y = 1, local_size_z = 1) in;
void main() {}
```
When compiled with:
```
glslangValidator.exe -V100 --target-env spirv1.2  -e main repro.comp.glsl -o spirv/repro.comp.spv
```
It produces:
```
; SPIR-V
; Version: 1.2
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 12
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 1
               OpSource GLSL 450
               OpName %main "main"
               OpDecorate %8 SpecId 0
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
          %8 = OpSpecConstant %uint 1
     %uint_1 = OpConstant %uint 1
     %v3uint = OpTypeVector %uint 3
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %8 %uint_1 %uint_1
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpReturn
               OpFunctionEnd
```
But when compiled with:
```
glslangValidator.exe -V100 --target-env spirv1.6  -e main repro.comp.glsl -o spirv/repro.comp.spv
```
Produces:
```
; SPIR-V
; Version: 1.6
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 13
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionModeId %main LocalSizeId %7 %uint_1 %uint_1
               OpSource GLSL 450
               OpName %main "main"
[...]

PR #4052, opened 2025-09-30
`local_size_*_id` and SPIR-V 1.2

pobrn:
The first commit adds a diagnostic when `local_size_*_id` is used when targeting < SPIR-V 1.2. I think it would be nice to get a diagnostic if the qualifier will be ignored in the end; I was surprised by this behaviour.

The second commit essentially applies #3351 to mesh and task shaders (#3739).

The third commit, I'm not entirely sure about, but according to https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpExecutionModeId `local_size_*_id` should be available when targeting at least SPIR-V 1.2, so that commit lowers the required versions.

The fourth is a trivial cleanup.

Related:
  * #3351
  * #3510
  * #3739


SPIRV/GlslangToSpv.cpp
```diff
@@ -1836,7 +1836,7 @@ TGlslangToSpvTraverser::TGlslangToSpvTraverser(unsigned int spvVersion,
                 break;
             }
         }
-        if (glslangIntermediate->getSpv().spv >= glslang::EShTargetSpv_1_6 && needSizeId) {
+        if (glslangIntermediate->getSpv().spv >= glslang::EShTargetSpv_1_2 && needSizeId) {
```
dnovillo, line 2052: Could you please add a test for this? The one in #3510 seems fine. Also, please split it into its own PR for clarity.

dnovillo (changes requested):

SPIRV/GlslangToSpv.cpp
pobrn: I suppose I should add it to `Spv.FromFile.cpp`? And I imagine I can just add two new test suites `CompileToSpirv11Test` and `CompileToSpirv12Test`, and then use the same input file for both?

dnovillo:
@pobrn we need you to sign the CLA before proceeding. Are you still interested in this?

pobrn:
> @pobrn we need you to sign the CLA before proceeding. Are you still interested in this?

Done. Yes.

dnovillo:
> > @pobrn we need you to sign the CLA before proceeding. Are you still interested in this?
> 
> Done. Yes.

Thanks. When you have a chance, please update the PR with the review feedback above.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7684,17 +7684,21 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
...
+                    "local_size_y_id",
+                    "local_size_z_id",
+                };
+
+                for (std::size_t i = 0; i < std::size(specIds); i++) {
+                    if (id == specIds[i]) {
+                        if (spvVersion.spv >= glslang::EShTargetSpv_1_2)
+                            publicType.shaderQualifiers.localSizeSpecId[i] = value;
+                        else
+                            error(loc, "requires SPIR-V 1.2", id.c_str(), "");
```
dnovillo, line 7758: This changes the output of every test that uses `local_size...`. Could you regenerate the failing golden files?

SPIRV/GlslangToSpv.cpp
dnovillo: yes, you can add both to `Spv.FromFile.cpp` as two new fixtures (`CompileToSpirv11Test` and `CompileToSpirv12Test`). Pairing each with `EShTargetVulkan_1_0` is fine.

dnovillo (changes requested):

SPIRV/GlslangToSpv.cpp
pobrn: I have made an attempt in the last commit. Please advise if it's ok. And I can create a separate PR if you want: should it be merged before this with an expected failing test or after this one? Or what did you have in mind?

glslang/MachineIndependent/ParseHelper.cpp
pobrn: Should I not change them instead to avoid using `local_size_{xyz}_id = ...`? Otherwise I imagine the compilation failures will cause the tests to not really test what they are supposed to?

SPIRV/GlslangToSpv.cpp
dnovillo: Yeah, the test approach is good.  No need to split into a separate PR.

glslang/MachineIndependent/ParseHelper.cpp
dnovillo: You can update the baselines to include the new error message.  Do not remove `local_size_*_id` from existing tests. Tests targeting SpvV < 1.2 now correctly emit a "requires SPIR-V 1.2" error. Tests targeting SpvV >= 1.2 now emit `LocalSizeId` instead of `LocalSize`. 

You can regenerate by running the test suite with `--update-mode`.

dnovillo (changes requested):
The CLA bot is still failing to find your signature. Can you try again?

glslang/MachineIndependent/ParseHelper.cpp
pobrn: Okay, done. I suppose I'm having a hard time understanding why this update won't cause issues, specifically, previously if the test successfully compiled, then the compiled code was compared, but now we only really have an error message, so whatever else was in the shader code is now not getting compiled and compared to the expected result.

dnovillo:
@pobrn can you please re-base so the CI can run at head? The changes look fine. Thanks.

dnovillo (approved):

dnovillo:
@pobrn  can you re-check the SPIRV-Tools you used for the `--update-mode` run? You only need to regenerate the files that have `local_size_*_id`. There are a bunch of new failures now (see the CI results).

dnovillo:
@pobrn still some CI failures. PTAL.

dnovillo:
@pobrn one more CI error: https://github.com/KhronosGroup/glslang/actions/runs/32526516858/job/97549365249?pr=4052

pobrn:
Huh, what kind of validation is that? And why does e.g. `spv.OpExecutionModeId.12.comp` fail it?

dnovillo:
> Huh, what kind of validation is that? And why does e.g. `spv.OpExecutionModeId.12.comp` fail it?

You had a push in-flight when the tests were running. I've triggered the CI again. I think it should be gone now.

pobrn:
Ahh, note that I have force pushed before that comment with the diff from the CI errors.

jeffbolznv:
> The third commit, I'm not entirely sure about, but according to https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpExecutionModeId `local_size_*_id` should be available when targeting at least SPIR-V 1.2, so that commit lowers the required versions.

I think the third commit is causing failures in llama.cpp, see discussion at https://github.com/ggml-org/llama.cpp/pull/27900. ggml-vulkan targets vulkan1.2 and enables maintenance4 which makes local_size_id valid in vulkan, but there is no shaderc option to inform spirv-val that local_size_id should be allowed, so it rejects it.

I think this commit should be reverted, at least until some other option can be sorted out.

pobrn:
So shaderc's `SpirvToolsOptimize()` has no option to call `SetAllowLocalSizeId(true)` on the used `spvtools::ValidatorOptions` object. That seems unfortunate, but I still hold that this is better than silently ignoring the `local_size_*_id` annotations. Is this issue already tracked on the shaderc side?

pobrn:
Ahh, the same thing also happens if one uses the `--spirv-val` option of `glslang`.

pobrn:
If that change is reverted, and `LocalSizeId` is only used above SPIR-V 1.6, then again, I have to wonder... how can people actually use it when targeting vulkan 1.2 (+ `VK_KHR_maintenance4`)? Because this was my entire motivation for this change: https://github.com/KhronosGroup/glslang/issues/3510#issuecomment-3239578387. And the fact that `LocalSizeId` would again be silently demoted to `LocalSize` just doesn't seem right.

jeffbolznv:
What's wrong with using the WorkgroupSize builtin when using vulkan 1.2?

pobrn:
(It's getting a bit off-topic, but could you elaborate what you mean? Can that address the `local_size_*_id` use case?)

jeffbolznv:
There are multiple ways in SPIR-V to have a spec constant workgroup size. One is to decorate a spec constant with the workgroupsize builtin. Another is to use the localsizeid builtin. Both of them work, but the workgroupsize builtin is deprecated in newer versions of SPIR-V. But what's wrong with continuing to use it for older versions?

pobrn:
Can you please refer to an example? Because this must be most obscure feature of vulkan, I cannot find an actual example for the life of me.

> But what's wrong with continuing to use it for older versions?

I suppose nothing, but then what's point of `VK_KHR_maintenance4` allowing `LocalSizeId`? If one is targeting vulkan1.2 + `VK_KHR_maintenance4`, vulkan1.3 and above; then one would reasonably expect that shaders could be compiled targeting vulkan1.2, and things will work, but they don't because `local_size_*_id` would be silently ignored by glslang, complete opposite of what one wanted by requiring `VK_KHR_maintenance4` for vulkan1.2.

For example, I took a quick look at `llama.cpp`. Consider `ggml/src/ggml-vulkan/vulkan-shaders/argmax.comp`, which states

```glsl
layout(local_size_x_id = 0, local_size_y = 1, local_size_z = 1) in;
```

but based on `vulkan-shaders-gen.cpp`, it will get compiled targeting `vulkan1.2`, so a previous `glslang` would silently compile it as if

```glsl
layout(local_size_id = 1, local_size_y = 1, local_size_z = 1) in;
```

was used:

```
$ spirv-dis build/ggml/src/ggml-vulkan/vulkan-shaders.spv/argmax_f32.spv | grep OpExecutionMode
               OpExecutionMode %4 LocalSize 1 1 1
```

Which means that this shader will never use specialization constant overridden workgroup sizes. This seems like pretty much the opposite of what was intended, and it happens without any indication or warning.

jeffbolznv:
When compiled with an earlier version of glslang, the argmax shader has:
```
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
...
               OpDecorate %190 SpecId 0
        %190 = OpSpecConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %190 %uint_1 %uint_1
```

So gl_WorkGroupSize.x comes from spec id 0. It works fine, it's just a different way of expressing it in SPIR-V.

pobrn:
Ahh, right, now I understand. Sorry, this all has been a grave misunderstanding on my part. This PR should indeed be reverted. Sorry again.

spencer-lunarg:
@jeffbolznv @pobrn what is the context on this?

```glsl
#version 450
layout(local_size_x_id = 3, local_size_y_id = 4) in;
void main(){}
```

with `--target-env 1.0` use to compile to valid SPIR-V

```
               OpExecutionMode %main LocalSize 1 1 1
               OpDecorate %8 SpecId 3
               OpDecorate %9 SpecId 4
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
       %uint = OpTypeInt 32 0
          %8 = OpSpecConstant %uint 1               
          %9 = OpSpecConstant %uint 1               
     %uint_1 = OpConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %8 %9 %uint_1   
```

now it is getting hit with `ERROR: a.comp:2: 'local_size_x_id' : requires SPIR-V 1.2` which I think it wrong here

spencer-lunarg:
sorry. I just Realize @jeffbolznv also noticed this (misread)

@arcady-lunarg @dnovillo I think we might need to revert this as I am not sure this is the correct fix and going to give many false warning to people when they update glslang

dnovillo:
> sorry. I just Realize @jeffbolznv also noticed this (misread)
> 
> @arcady-lunarg @dnovillo I think we might need to revert this as I am not sure this is the correct fix and going to give many false warning to people when they update glslang

Yeah. I totally missed this in my review. Apologies. I'm OK with reverting it.

spencer-lunarg:
@pobrn the Workgroup size this is... unfun, here is some information around it https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/compute_shaders.adoc#workgroup-size

PR #4057, opened 2025-10-07
Fix debug struct member names

### LINKED ISSUE #3861: Shader DebugInfo doesn't give correct string name for same type
(see with `-gVS` https://godbolt.org/z/9o3TdvYTn)

```glsl
#version 450
#extension GL_EXT_shader_8bit_storage: enable
#extension GL_EXT_shader_explicit_arithmetic_types_int8: enable

struct Foo {
    uint a;
};

layout(push_constant) uniform PushConstant {
    uint b;
    Foo foo;
    vec4 c;
} data;

void main(){
    gl_Position = data.c;
}
```

There is not a `OpString` for both `"a"` and `"b"` so this is generating

```swift
%16 = OpString "a"
%17 = OpString "Foo"
%20 = OpString "PushConstant"

%member_a = OpExtInst %26 %1 DebugTypeMember %16 %32 %35 %30 %64 %31 %31 %33
%66 = OpExtInst %26 %1 DebugTypeComposite %17 %37 %35 %55 %31 %40 %17 %31 %33 %member_a

%member_b = OpExtInst %26 %1 DebugTypeMember %16 %32 %35 %30 %64 %31 %31 %33
%73 = OpExtInst %26 %1 DebugTypeComposite %20 %37 %35 %55 %31 %40 %20 %31 %33 %member_b %70 %72
```

and results and showing a VVL error message of `data.a` when it should be `data.b`

qingyuanzNV:
Fixes #3861 

This commit should fix broken debug info of composite type member names. We wire debug info of struct members through `makeStructType` and `makeCompositeDebugType` so that correct names could be emitted. Note that although this commit also passes debug location info to debug info builder, we don't have accurate line/column in AST yet. Therefore, it doesn't fix line/column info. That requires an additional fix for #3403.

 In addition, we add a pseudo wrapper to separate building opaque type debug info from normal structure type.

### DIFF
--- a/SPIRV/SpvBuilder.h
+++ b/SPIRV/SpvBuilder.h
@@ -236,7 +236,6 @@ public:
     Id makeGenericType(spv::Op opcode, std::vector<spv::IdImmediate>& operands);
 
     // SPIR-V NonSemantic Shader DebugInfo Instructions
-    std::unordered_map<Id, DebugTypeLoc> debugTypeLocs;
     Id makeDebugInfoNone();
     Id makeBoolDebugType(int const size);
     Id makeIntegerDebugType(int const width, bool const hasSign);


SPIRV/GlslangToSpv.cpp
```diff
@@ -5646,15 +5647,17 @@ spv::Id TGlslangToSpvTraverser::convertGlslangStructToSpvType(const glslang::TTy
             //  + Not as clean as desired. Traverser queries/sets persistent state. This is fragile.
             //  + Table lookup during creation of composite debug types. This really shouldn't be necessary.
             if(options.emitNonSemanticShaderDebugInfo) {
-                builder.debugTypeLocs[spvMember].name = glslangMember.type->getFieldName().c_str();
```
dnovillo, line 5649: You can now remove `debugTypeLocs` from `SpvBuilder.h`

Issue #2407, opened 2020-09-27
Stringify preprocessor operator doesn't seem to work


vaspoul:
Not sure if I'm missing something obvious here, but the following produces an error:

#define TOSTRING(x) #x
TOSTRING(Hello)

I get this error:
'#' : preprocessor directive cannot be preceded by another token

(I'm parsing HLSL if that makes a difference)

johnkslang:
It has never been implemented.

PR #4068, opened 2025-10-20
Implement stringify (#) for preprocessor macros

tnibler:
Stumbled upon #2407, so I though I'd give it a shot. 

I tried touching as little as possible so nothing should break hopefully.

The rules sort of follow what GCC does, barring the most complex stuff that would probably need a complete restructuring to implement. A notable difference is that [this indirection](https://stackoverflow.com/a/2653351) isn't needed, arguments are expanded and stringified in one step. That's because the actual stringifying is done higher up in the preprocessing loop with marker tokens, which avoids having to thread more state through the entire machinery like for pasting. I feel that's very error prone (in fact a `currentPos = savePos` was missing in the pasting logic AFAICT) for not much benefit here. 

Draft because:
 - One `tokenPaste.vert` test broke, something up with recursive macros or something
 - Nested stringification is not handled, they are just flattened into one string. The nesting depth is available, so nothing stops us from building the quote/backslash escapes. Just wasn't sure if you'd actually want that, because ad-hoc escaping is (almost) always incorrect and this is just a debug feature so whatever.
 - No tests that work within the current setup, because it's tricky to test all the edge cases and produce legal code at the same time ;)

But this is what it produces:
```
#version 450
#extension GL_EXT_debug_printf : enable

#define PASTE(arg) pasted_##arg
#define STRINGIFY(arg) #arg
#define STRINGIFY_TWICE(arg, regular) before #arg #arg after regular
#define STRINGIFY_WITH_SPACE(arg) before # arg
#define PASTE_WITH_SPACE(arg) pasted_## arg
#define BOTH(arg1,    arg2) pasted_##arg1 and     #arg2
#define NOARGS(arg) 1 1 1 1
#define REGULAR(arg1, arg2) arg1    : arg2 :arg1
#define STR(a) STRHELP(a)
#define STRHELP(a) #a // this isn't necessary right now, as would be in C
#define LEGAL(arg) a######arg

void main() {
    REGULAR(8,9);
    STRINGIFY(12 3);
    STRINGIFY_TWICE(letters_too,       1);
    STRINGIFY_WITH_SPACE(aaa1);
    PASTE(symbol);
    PASTE_WITH_SPACE(symbol);
    BOTH(symbol, string);
    STRHELP(REGULAR(3, 4));
    STR(REGULAR(3, 4));
    STR(aaa     );
    STR(STR(hello));

```
```
#version 450
#extension GL_EXT_debug_printf : enable
[...]
void main() {
               8 : 9 : 8;
                  "12 3";
                                        before "letters_too" "letters_too" after 1;
                             before "aaa1";
                pasted_symbol;
                           pasted_symbol;
                       pasted_symbol and "string";
                         "3 : 4 :3";
                     "3 : 4 :3";
                "aaa";
                  "hello";
}
```

### DIFF
--- a/Test/baseResults/preprocessor.paste_stringify.vert.out
+++ b/Test/baseResults/preprocessor.paste_stringify.vert.out
@@ -1,0 +1,36 @@
+#version 450
+#extension GL_EXT_debug_printf : enable
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+bool before, after;
+
+void main() {
+               8 == 9 ? 9 : 8;
+                  "12 3";
+    debugPrintfEXT("%d", (before, "letters_too", "letters_too", after, 1));
+                             debugPrintfEXT("aaa1");
+
+    int pasted_symbol;
+    float pasted_symbol2;
+                    debugPrintfEXT("string");
+                         "3 == 4 ? 4 :3";
+                     "3 == 4 ? 4 :3";
+                "aaa";
+                  "hello";
+    debugPrintfEXT("regular string");
+    debugPrintfEXT("");
+}
+
--- a/Test/baseResults/preprocessor.stringify_invalid.vert.err
+++ b/Test/baseResults/preprocessor.stringify_invalid.vert.err
@@ -1,0 +1,4 @@
+ERROR: 0:4: '#' : Illegal sequence of paste (##) and stringify (#). 
+ERROR: 1 compilation errors.  No code generated.
+
+
--- a/Test/preprocessor.paste_stringify.vert
+++ b/Test/preprocessor.paste_stringify.vert
@@ -1,0 +1,35 @@
+#version 450
+#extension GL_EXT_debug_printf : enable
+
+#define PASTE(arg) pasted_##arg
+#define STRINGIFY(arg) #arg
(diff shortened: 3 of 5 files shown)


glslang/MachineIndependent/preprocessor/Pp.cpp
```diff
@@ -1119,7 +1164,7 @@ int TPpContext::tMacroInput::scan(TPpToken* ppToken)
     }
 
     // see if are preceding a ##
-    if (mac->body.peekUntokenizedPasting()) {
```
tnibler, line 1122: UntokenizedPasting doesn't exist anymore, because all `#` handling is now already done at macro parsing time.

glslang/MachineIndependent/preprocessor/PpTokens.cpp
```diff
@@ -146,8 +136,10 @@ bool TPpContext::TokenStream::peekTokenizedPasting(bool lastTokenPastes)
 
     // 2. last token and we've been told after this there will be a ##
 
-    if (! lastTokenPastes)
+    if (! lastTokenPastes) {
```
tnibler, line 139: I think this was missing, independently of the new feature

PR #4070, opened 2025-10-24
Fix a call site that had an incorrect number of arguments.

sajjadmirzanv:
This call to compileAndLink is missing two arguments, which causes the `automap` variable to be passed as `enableDebug`. Fix is just to add the arguments.

As far as I can tell, commit [5d43c4a](https://github.com/KhronosGroup/glslang/commit/5d43c4aac776f1504dc5c364bae49251eadb8111#diff-6d4f47ad1b3ad05e9053cb43f6d237d8dd317ace06fbf7e53a8f9147ede35692R218) added a new parameter to the `compileAndLink` function. Then later on the same day, commit [bd1c183](https://github.com/KhronosGroup/glslang/commit/bd1c1831d5d1a81bdd1c595250199babca37a938#diff-6d4f47ad1b3ad05e9053cb43f6d237d8dd317ace06fbf7e53a8f9147ede35692R482) added a call to `compileAndLink`, but did not account for the extra argument. The compiler did not throw an error because the last few arguments to `compileAndLink` are all `bool` with a default value.


arcady-lunarg:
Yes, enableDebug and enableNonSemanticShaderDebugInfo should only be true in the tests that are explicitly testing debugging info features.

PR #4071, opened 2025-10-24
Add semantic check for cooperative vector loads/stores.

dnovillo:
For cooperative vector loads and stores, the buffer arrays cannot be local variables. The compiler was not validating this and when it went to generate debug information, it would hit an assertion because the ID could not be retrieved.

This adds a semantic check that fails compilation when the array is neither in buffer nor shared storage.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -1484,6 +1484,17 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
                         i == 1 && arg->getAsTyped()->getType().getBasicType() != aggregate->getSequence()[0]->getAsTyped()->getType().getBasicType())
                         error(arguments->getLoc(), "cooperative vector basic types must match", fnCandidate->getName().c_str(), "");
 
+                    // Check that coopVecLoadNV and coopVecStoreNV buffer parameter is in buffer or shared storage
+                    if (builtIn && (fnCandidate->getBuiltInOp() == EOpCooperativeVectorLoadNV ||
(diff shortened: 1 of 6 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1484,6 +1484,17 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
                         i == 1 && arg->getAsTyped()->getType().getBasicType() != aggregate->getSequence()[0]->getAsTyped()->getType().getBasicType())
                         error(arguments->getLoc(), "cooperative vector basic types must match", fnCandidate->getName().c_str(), "");
 
+                    // Check that coopVecLoadNV and coopVecStoreNV buffer parameter is in buffer or shared storage
+                    if (builtIn && (fnCandidate->getBuiltInOp() == EOpCooperativeVectorLoadNV ||
+                                    fnCandidate->getBuiltInOp() == EOpCooperativeVectorStoreNV) &&
+                        i == 1) {
+                        TStorageQualifier storage = arg->getAsTyped()->getType().getQualifier().storage;
+                        if (storage != EvqBuffer && storage != EvqShared) {
```
jeffbolznv, line 1492: Is buffer_reference still considered to be EvqBuffer? If so, this looks fine. But it would be good to have a positive test for buffer_reference (or maybe one already exists in other tests)?

dnovillo: Yes, buffer_reference are EvqBuffer. There is already a test for buffer_reference in `Test/spv.coopvecloadstore.comp`.

PR #4074, opened 2025-10-28
Add AST node for variable declarators for better quality debug info

qingyuanzNV:
At the moment, GLSLANG drops variable declaration info immediately after making the symbol in the symbol table. This leads to difficulty in emitting debug info, presenting that 1) debug info for constant variable is missing 2) debug line of variables points to its first use instead of declaration.

This patch introduces a new `TIntermVariableDecl` node in the AST, consisting of an `TIntermSymbol* declSymbol` and an `TIntermNode* initNode`. `declSymbol` is an imaginary access to the declared variable, being used to instantiate the basic variable info and debug information at the point of declaration. It doesn't participate AST traversal, such that code generation is still operating on the same AST as before. `TIntermNode* initNode` is the AST node for the initializer, which used to be emitted directly in the OpSequence. It's now wrapped within a TIntermVariableDecl node. For example,

- When a declaration "int ^x;" is seen, a new tree node is inserted to the AST pointing to ^.
- When a declaration "int ^x = 42;" is seen, the initializer node in the AST is replace with the new tree node pointing to ^.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -195,7 +195,6 @@ protected:
     void makeGlobalInitializers(const glslang::TIntermSequence&);
     void collectRayTracingLinkerObjects();
     void visitFunctions(const glslang::TIntermSequence&);
-    void handleFunctionEntry(const glslang::TIntermAggregate* node);
     void translateArguments(const glslang::TIntermAggregate& node, std::vector<spv::Id>& arguments,
         spv::Builder::AccessChain::CoherentFlags &lvalueCoherentFlags);
     void translateArguments(glslang::TIntermUnary& node, std::vector<spv::Id>& arguments);
@@ -3130,7 +3129,12 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
                 builder.enterFunction(shaderEntry);
                 currentFunction = shaderEntry;
             } else {
-                handleFunctionEntry(node);
+                // SPIR-V functions should already be in the functionMap from the prepass
+                // that called makeFunctions().
+                currentFunction = functionMap[node->getName().c_str()];
+                spv::Block* functionBlock = currentFunction->getEntryBlock();
+                builder.setBuildPoint(functionBlock);
+                builder.enterFunction(currentFunction);
             }
             if (options.generateDebugInfo && !options.emitNonSemanticShaderDebugInfo) {
                 const auto& loc = node->getLoc();
@@ -3139,7 +3143,7 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
                 currentFunction->setDebugLineInfo(sourceFileId, loc.line, loc.column);
             }
         } else {
-            currentFunction = nullptr;
+            // Here we have finished visiting the function (post-visit). Finalize it.
             if (options.generateDebugInfo) {
                 if (glslangIntermediate->getSource() == glslang::EShSourceGlsl && node->getSequence().size() > 1) {
                     auto endLoc = node->getSequence()[1]->getAsAggregate()->getEndLoc();
@@ -3150,6 +3154,7 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
                 entryPointTerminated = true;
             builder.leaveFunction();
             inEntryPoint = false;
+            currentFunction = nullptr;
         }
 
         return true;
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -10323,6 +10338,17 @@ void TParseContext::declareBlock(const TSourceLoc& loc, TTypeList& typeList, con
 
     // Save it in the AST for linker use.
     trackLinkage(variable);
+
+    TIntermAggregate* declNode = nullptr;
+    if (intermediate.getDebugInfo()) {
+        auto blockDeclNode = new TIntermVariableDecl(intermediate.addSymbol(variable, loc), nullptr);
+        blockDeclNode->setLoc(loc);
+
+        // We have to wrap the decalaration with a sequence to fit the same processing logic with variables.
```
dnovillo: nit: s/decalaration/declaration/

glslang/Include/intermediate.h
```diff
@@ -1092,6 +1095,37 @@ struct TIntermNodePair {
...
+
+    TIntermSymbol* getDeclSymbol() { return declSymbol; }
+    const TIntermSymbol* getDeclSymbol() const { return declSymbol; }
+
+    TIntermNode* getInitNode() { return initNode; }
+    const TIntermNode* getInitNode() const { return initNode; }
+
+private:
+    // This symbol is an imaginary access of the variable defined, which isn't an AST node and 
+    // doesn't participate tree traversal.
```
dnovillo: nit: s/participate tree/participate in tree/

qingyuanzNV:
> Some initial comments (still going through it). Is it feasible to add tests for some edge cases like re-declarations, nested scopes with multiple declarations and arrays? I don't think I've seen those.

Done.

> I'm still trying to understand the different handling for declarations in the grammar. regular variables, block declarations and condition declarations emit different patterns. Do we need grammar changes here? It doesn't seem like we do, but I think I'm missing something.

Yes, we have grammar change in glslang.y and glslang_tab.cpp

---

I have rebased and and modified the traverser to visit the imaginary symbol if wanted. This makes the corner case to be handled more elegantly. @dnovillo please help review.

glslang/MachineIndependent/glslang.y
```diff
@@ -3990,11 +3990,20 @@ condition
...
-        TIntermNode* initNode = parseContext.declareVariable($2.loc, *$2.string, $1, 0, $4);
-        if (initNode)
-            $$ = initNode->getAsTyped();
-        else
+        TIntermNode* declNode = parseContext.declareVariable($2.loc, *$2.string, $1, 0, $4);
+        if (declNode) {
+            if (declNode->getAsVariableDecl()) {
+                // FIXME: could we support decl AST in this case?
+                $$ = declNode->getAsVariableDecl()->getInitNode()->getAsTyped();
+                delete declNode;
```
dnovillo: I'm concerned about double-free issues or memory corruption in the pool allocator. I'm not sure, though. Would it be possible to delay creation beforehand in `declareVariable()`?

SPIRV/GlslangToSpv.cpp
```diff
@@ -3138,6 +3139,7 @@ bool TGlslangToSpvTraverser::visitAggregate(glslang::TVisit visit, glslang::TInt
                 currentFunction->setDebugLineInfo(sourceFileId, loc.line, loc.column);
             }
         } else {
+            currentFunction = nullptr;
```
dnovillo: This is in the middle of processing `currentFunction`, could you add a comment on why this is needed? Won't this confuse downstream processing?

glslang/Include/intermediate.h
```diff
@@ -1092,6 +1095,37 @@ struct TIntermNodePair {
...
+    const TIntermVariableDecl* getAsVariableDecl() const override { return this; }
+
+    TIntermSymbol* getDeclSymbol() { return declSymbol; }
+    const TIntermSymbol* getDeclSymbol() const { return declSymbol; }
+
+    TIntermNode* getInitNode() { return initNode; }
+    const TIntermNode* getInitNode() const { return initNode; }
+
+private:
+    // This symbol is an imaginary access of the variable defined, which isn't an AST node and
```
dnovillo: Maybe substitute "This symbol is an imaginary access" with something like "This symbol represents the declared variable at its declaration point. It's not traversed by default. To traverse it, the visitor needs to have `includeDeclSymbol` enabled." Also, add some comment on what type of traversals will need this and why (I suppose this is only a debug info requirement for now).

SPIRV/GlslangToSpv.cpp
qingyuanzNV: This is actually quite natural here to reset `currentFunction` to `nullptr` as we are at the end of processing a function definition (post-visit). This handles the following case:
```
void foo() {}

int x; // <- we should have currentFunction to be nullptr while visiting AST of this

void bar() {}
```

Adding this reset logic allows us to quickly identify if we are in a local scope.

dnovillo: Thanks. Please add a comment here explaining this. It tripped me at first :)

glslang/Include/intermediate.h
qingyuanzNV: Done

glslang/MachineIndependent/glslang.y
qingyuanzNV: Instead of freeing this node (this was old workaround as this couldn't be done easily), I figured out an approach to store the declaration node within `TIntermLoop`. Please help review.

dnovillo (approved):
LGTM. Thanks for all the fixes.

Issue #4077, opened 2025-10-31
Segfault on string literal in ternary operator (GLSL)


tnibler:
```
#version 450
#extension GL_EXT_debug_printf : enable

bool b;

void main() {
    b ? "a" : "b";
}
```

`glslang -V .segfault.comp` on current main.

jeremy-lunarg:
Fixed by #4079.

PR #4079, opened 2025-10-31
Reject string operands in binary and select ops (#4077)

tnibler:
This would previously cause a segfault during SPIR-V codegen


glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -121,6 +121,9 @@ TIntermTyped* TIntermediate::addBinaryMath(TOperator op, TIntermTyped* left, TIn
     if (left->getType().getBasicType() == EbtBlock || right->getType().getBasicType() == EbtBlock)
         return nullptr;
 
+    if (left->getType().getBasicType() == EbtString || right->getType().getBasicType() == EbtString)
```
dnovillo: Could you factor this together with the check above? It will read more concisely. Thanks.

dnovillo (changes requested):

glslang/MachineIndependent/Intermediate.cpp
dnovillo: Also, please add a compile test showing the error message instead of the crash.

PR #4082, opened 2025-11-03
Implement GL_EXT_shader_64bit_indexing

jeffbolznv:
Spec is at https://github.com/KhronosGroup/GLSL/pull/292.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -638,6 +638,7 @@ TIntermTyped* TParseContext::handleBracketDereference(const TSourceLoc& loc, TIn
                     TType& leftType = binaryNode->getLeft()->getWritableType();
                     TArraySizes& arraySizes = *leftType.getArraySizes();
                     assert(arraySizes.getNumDims() == 2);
+                    assert(indexValue < std::numeric_limits<int>::max());
                     arraySizes.setDimSize(1, std::max(arraySizes.getDimSize(1), (int)indexValue + 1));
                 }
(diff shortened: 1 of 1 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -624,7 +638,7 @@ TIntermTyped* TParseContext::handleBracketDereference(const TSourceLoc& loc, TIn
                     TType& leftType = binaryNode->getLeft()->getWritableType();
                     TArraySizes& arraySizes = *leftType.getArraySizes();
                     assert(arraySizes.getNumDims() == 2);
-                    arraySizes.setDimSize(1, std::max(arraySizes.getDimSize(1), indexValue + 1));
+                    arraySizes.setDimSize(1, std::max(arraySizes.getDimSize(1), (int)indexValue + 1));
```
dnovillo, line 642: This looks like it may overflow. Add a check here.

jeffbolznv: This is for per-view input/output storage, the array sizes and indices should be very small. What kind of check do you want?

dnovillo: Just asserting that `indexValue` cannot overflow a 32-bit int, but if it's impossible then just a comment.

jeffbolznv: OK, I added an assert.

dnovillo: Thanks!

Issue #4010, opened 2025-07-29
Massive performance penalty using non semantic debug info


Axel-Reactor:
We are compiling pretty large GLSL compute shaders (~22k lines). Usually glslang takes a little slow but reasonable ~600ms to do this on my Zen5 desktop machine, but when enabling non semantic debug info this shoots up to >10s.

arcady-lunarg:
If you could attach a sample shader, that would make it much easier for us to reproduce the issue. It might also be helpful to specify which platform you're running this on. If you want to do some profiling yourself to try to see where it's spending its time, that would be even better.

Novum:
Here is one of the shaders (I had to obfuscate the symbols). Not the worst case, but still clearly shows the problem:

```
time glslangValidator -V --target-env vulkan1.4 test.comp 
test.comp

real	0m0.199s
user	0m0.152s
sys	0m0.047s

time glslangValidator -gVS -V --target-env vulkan1.4 test.comp 
test.comp

real	0m2.284s
user	0m2.226s
sys	0m0.049s
```

[test.zip](https://github.com/user-attachments/files/21517404/test.zip)

arcady-lunarg:
Thanks, I can reproduce it with the shader you gave. Running some profiles, I think I understand what's happening. Because nonsemantic instructions only have `Id` operands, and because line tracking debuginfo is emitted pretty much for each line in your shader, it needs to create an `OpConstant` instruction for every line number. However, the routine that creates these constant instructions first checks whether one already exists for the requested type and constant, and does so using a simple linear search. This results in a quadratic slowdown that basically depends on the number of unique constants created.

arcady-lunarg:
This has been fixed by #4083.

PR #4083, opened 2025-11-03
Fix linear time lookups in Builder::findScalarConstant

Axel-Reactor:
As discussed in https://github.com/KhronosGroup/glslang/issues/4010
Replaces the linear lookup with a separate hash map to map scalar constants to result IDs.
The old hash map is now only used for composites so I renamed it accordingly.

### DIFF
--- a/SPIRV/SpvBuilder.cpp
+++ b/SPIRV/SpvBuilder.cpp
@@ -1716,32 +1716,17 @@ bool Builder::containsPhysicalStorageBufferOrArray(Id typeId) const
 // can be reused rather than duplicated.  (Required by the specification).
 Id Builder::findScalarConstant(Op typeClass, Op opcode, Id typeId, unsigned value)
 {
-    Instruction* constant;
-    for (int i = 0; i < (int)groupedConstants[enumCast(typeClass)].size(); ++i) {
-        constant = groupedConstants[enumCast(typeClass)][i];
-        if (constant->getOpCode() == opcode &&
-            constant->getTypeId() == typeId &&
-            constant->getImmediateOperand(0) == value)
-            return constant->getResultId();
-    }
-
-    return 0;
+    ScalarConstantKey key{ enumCast(typeClass), enumCast(opcode), typeId, value, 0 };
+    auto it = groupedScalarConstantResultIDs.find(key);
+    return (it != groupedScalarConstantResultIDs.end()) ? it->second : 0;
 }
 
 // Version of findScalarConstant (see above) for scalars that take two operands (e.g. a 'double' or 'int64').
 Id Builder::findScalarConstant(Op typeClass, Op opcode, Id typeId, unsigned v1, unsigned v2)
 {
-    Instruction* constant;
-    for (int i = 0; i < (int)groupedConstants[enumCast(typeClass)].size(); ++i) {
-        constant = groupedConstants[enumCast(typeClass)][i];
-        if (constant->getOpCode() == opcode &&
-            constant->getTypeId() == typeId &&
-            constant->getImmediateOperand(0) == v1 &&
-            constant->getImmediateOperand(1) == v2)
-            return constant->getResultId();
-    }
-
-    return 0;
+    ScalarConstantKey key{ enumCast(typeClass), enumCast(opcode), typeId, v1, v2 };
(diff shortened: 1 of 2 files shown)


Axel-Reactor:
Timings. First two are my branch, last one is glslangvalidator from SDK. Still 18% slower with non-semantic debug info, but not 10x anymore.
```
time ./glslangValidator -V --target-env vulkan1.4 test.comp
________________________________________________________
Executed in  192.91 millis    fish           external
   usr time  162.08 millis    0.00 micros  162.08 millis
   sys time   30.77 millis  754.00 micros   30.02 millis

time ./glslangValidator -gVS -V --target-env vulkan1.4 test.comp
________________________________________________________
Executed in  248.66 millis    fish           external
   usr time  192.96 millis    0.00 micros  192.96 millis
   sys time   55.89 millis  898.00 micros   54.99 millis
   
time glslangValidator -gVS -V --target-env vulkan1.4 test.comp
________________________________________________________
Executed in    2.30 secs    fish           external
   usr time    2.24 secs  445.00 micros    2.24 secs
   sys time    0.06 secs  342.00 micros    0.06 secs
```

SPIRV/SpvBuilder.h
```diff
@@ -1019,8 +1019,35 @@ class Builder {
...
+        }
+    };
+
+    struct ScalarConstantKeyHash {
+        std::size_t operator()(const ScalarConstantKey& k) const {
+            return std::hash<unsigned>{}(k.typeClass) ^
+                   (std::hash<unsigned>{}(k.opcode)) ^
+                   (std::hash<Id>{}(k.typeId)) ^
+                   (std::hash<unsigned>{}(k.value1)) ^
+                   (std::hash<unsigned>{}(k.value2));
```
dnovillo: I've seen xor hashing run into more collisions than other polynomial hashes. But this is just a nit. Any specific reason you chose XORing the keys?

SPIRV/SpvBuilder.cpp
```diff
@@ -1813,30 +1798,29 @@ Id Builder::makeNullConstant(Id typeId)
...
 
-    return c->getResultId();
+    Id resultId = c->getResultId();
+
+    if (!specConstant) {
+        ScalarConstantKey key{enumCast(Op::OpTypeBool), enumCast(opcode), typeId, 0, 0};
+        groupedScalarConstantResultIDs[key] = resultId;
+    }
+
+    return resultId;
```
dnovillo, line 1821: This is repeated a bit. Perhaps re-factor it into a private helper?

dnovillo (changes requested):
Thanks for the speedup! LGTM with some minor requests.

SPIRV/SpvBuilder.h
Axel-Reactor: I did actually try to do something better, but:
1. C++ to this day does not have a built in hash-combine function even in C++23 which is insanity.
2. glslang still needs to compile for both 32 and 64 bit. Now the problem is that the C++ committee in its infinite wisdom decided to make std::hash return `size_t`. That means the hash combine prime/shift constants should be different depending on 32 bit vs 64 bit which is very annoying to do.
3. I believe std::hash in some implementations will just pass through ints without scrambling them, so this is the actual problem. If the bits would be properly distributed the XOR would be fine.

I'll take another look. After writing this I realized that especially 3. is bad.

SPIRV/SpvBuilder.h
```diff
@@ -1023,8 +1019,58 @@ class Builder {
...
+                return h;
+            }
+        }
+
+        // Hash combine from boost
+        inline std::size_t hash_combine(std::size_t seed, std::size_t v) const {
+            return hash_mix(seed + 0x9e3779b9 + v);
+        }
+        
+        std::size_t operator()(const ScalarConstantKey& k) const {
```
Axel-Reactor: I verified that this didn't change timings. Neither faster nor slower, but it should be more robust against outliers.

SPIRV/SpvBuilder.cpp
jeremy-lunarg: This is still unresolved but I'm not going to hold up the PR for this.

arcady-lunarg: You can do a followup cleanup change.

Issue #4024, opened 2025-08-20
Pointers to another PhysicalStorageBuffer is not recognized as 8 bytes


spencer-lunarg:
https://godbolt.org/z/xzoEvG14M
and
https://godbolt.org/z/oEM8f5fdn (the reserve direction)

These are both invalid as the `Aligned` should be `8` (not `4`) because the PhysicalStorageBuffer is a 8 bytes  scalar and would violate the `06314` VU in Vulkan (which I'm working on spirv-val for currently)

arcady-lunarg:
I don't think glslang currently checks the alignment value at all other than making sure it's a power of two. It probably should though, as the spec says that it's an error to have an alignment that is smaller than that of the most-aligned member.

Issue #4084, opened 2025-11-05
Setting the wrong alignment value when loading a Buffer Reference through a Buffer Reference


spencer-lunarg:
https://godbolt.org/z/j8jGvsnMh

```glsl
layout(buffer_reference) buffer S4;

layout(buffer_reference, buffer_reference_align = 8, std430) readonly buffer S3 {
    S4 e;
};

// Set buffer_reference_align to 8 and it generates valid code
layout(buffer_reference, buffer_reference_align = 4, std430) readonly buffer S4 {
    uint f;
};

void main() {    
    S3 x;
    S4 y = x.e;
}
```

Seems currently glslang sets the `x.e` as an `OpLoad` with a result type of `OpTypePointer` but the `Aligned 4` is wrong and should be `8` and produces a spirv-val error (https://godbolt.org/z/sxT1asEdv)

spencer-lunarg:
this might be a duplicate of https://github.com/KhronosGroup/glslang/issues/4024

GabeRundlett:
A potentially more clear example could be the following, as it demonstrates that it's not getting the alignment from what is being assigned into, but instead incorrectly aligning via S4's alignment. In godbolt, this uses `Align 1`
```glsl
#version 450
#extension GL_EXT_scalar_block_layout : enable
#extension GL_EXT_buffer_reference : enable
#extension GL_EXT_shader_explicit_arithmetic_types_int64 : enable

layout(buffer_reference, buffer_reference_align = 1) readonly buffer S4;

layout(buffer_reference, buffer_reference_align = 8) readonly buffer S3 {
    S4 e;
};

void main() {    
    S3 x;
    uint64_t y = uint64_t(x.e);
}
```

Issue #4058, opened 2025-10-07
Generated `OpLoad` does not respect the minimum type alignment


maxime-modulopi:
glslang version: 16.0.0 (SDK 1.4.328.0)

Given this shader:

<details>
<summary>GLSL code</summary>

```glsl
#version 460 core

#extension GL_EXT_buffer_reference: require
#extension GL_EXT_scalar_block_layout: require
#extension GL_EXT_nonuniform_qualifier: require

struct Vertex
{
	vec3 position;
	vec2 texCoord;
	vec3 normal;
};

layout (buffer_reference, buffer_reference_align = 4, scalar) readonly buffer VertexBuffer
{
	Vertex vertex;
};

layout (buffer_reference, buffer_reference_align = 8, scalar) readonly buffer ParametersBuffer
{
	VertexBuffer vertexBuffer;
};

layout(push_constant) uniform constants
{
	ParametersBuffer u_params;
};

void main()
{
	gl_Position = vec4(u_params.vertexBuffer.vertex.position, 1.0);
}
```
</details>

<details>
<summary>SPIR-V code</summary>

```spirv
OpCapability Shader
OpCapability PhysicalStorageBufferAddresses
%1 = OpExtInstImport "GLSL.std.450"
OpMemoryModel PhysicalStorageBuffer64 GLSL450
OpEntryPoint Vertex %4 "main" %13 %25
OpSource GLSL 460
OpSourceExtension "GL_EXT_buffer_reference"
OpSourceExtension "GL_EXT_nonuniform_qualifier"
OpSourceExtension "GL_EXT_scalar_block_layout"
OpName %4 "main"
OpName %11 "gl_PerVertex"
OpMemberName %11 0 "gl_Position"
OpMemberName %11 1 "gl_PointSize"
OpMemberName %11 2 "gl_ClipDistance"
OpMemberName %11 3 "gl_CullDistance"
OpName %13 ""
OpName %17 "constants"
OpMemberName %17 0 "u_params"
OpName %19 "ParametersBuffer"
OpMemberName %19 0 "vertexBuffer"
OpName %22 "Vertex"
OpMemberName %22 0 "position"
OpMemberName %22 1 "texCoord"
OpMemberName %22 2 "normal"
OpName %23 "VertexBuffer"
OpMemberName %23 0 "vertex"
OpName %25 ""
OpDecorate %11 Block
OpMemberDecorate %11 0 BuiltIn Position
[...]

PR #4089, opened 2025-11-11
Fix Buffer Reference having wrong alignment

spencer-lunarg:
closes https://github.com/KhronosGroup/glslang/issues/4024 and https://github.com/KhronosGroup/glslang/issues/4084 and https://github.com/KhronosGroup/glslang/issues/4058


SPIRV/SpvPostProcess.cpp
```diff
@@ -341,15 +341,30 @@ void Builder::postProcess(Instruction& inst)
...
+
                 // Compute the index of the alignment operand.
                 int alignmentIdx = 2;
-                if (inst.getOpCode() == Op::OpStore)
+                if (is_store)
                     alignmentIdx++;
                 // Merge new and old (mis)alignment
                 alignment |= inst.getImmediateOperand(alignmentIdx);
+
+                if (!is_store) {
```
arcady-lunarg, line 356: Do we not need this logic for stores?

spencer-lunarg: no, because I couldn't find a case where you can store of `OpTypePointer`

arcady-lunarg (approved):

SPIRV/SpvPostProcess.cpp
spencer-lunarg: ... Eating my words now https://github.com/KhronosGroup/glslang/issues/4058#issuecomment-3642479028

PR #4090, opened 2025-11-12
Fix spv.coopmat_armlayout.comp test

kpet:
The test was generating invalid SPIR-V.


kpet:
This is currently blocking the introduction of the validation checks that flagged this test as invalid (https://github.com/KhronosGroup/SPIRV-Tools/pull/6408).

arcady-lunarg:
Shouldn't there be some error produced by the compiler here rather than just generating invalid code?

Issue #4095, opened 2025-11-17
Support depth layout qualifier for OpenGL ES profile


beaufortfrancois:
My understanding from reading https://registry.khronos.org/OpenGL/extensions/EXT/EXT_conservative_depth.txt is that the `EXT_conservative_depth` would allow "depth layout qualifier" in OpenGL ES. Right now, the following code line says it requires OpenGL 4.20.

https://github.com/KhronosGroup/glslang/blob/7099c123729e02f81d70559e79ee4360096fdfe5/glslang/MachineIndependent/ParseHelper.cpp#L6803

Would you be willing to accept a PR that adds `EXT_conservative_depth` check for OpenGL ES?

PR #4096, opened 2025-11-17
Add ES support for depth layout qualifier

beaufortfrancois:
FIX https://github.com/KhronosGroup/glslang/issues/4095

### DIFF
--- a/Test/baseResults/320.frag.out
+++ b/Test/baseResults/320.frag.out
@@ -10,6 +10,7 @@ ERROR: 0:29: 'vAnon' : redefinition
 ERROR: 0:43: 'origin_upper_left' : not supported with this profile: es
 ERROR: 0:43: 'pixel_center_integer' : not supported with this profile: es
 ERROR: 0:43: 'redeclaration' : cannot redeclare with different qualification: gl_FragCoord
+ERROR: 0:47: 'depth layout qualifier' : not supported for this version or the enabled extensions 
 ERROR: 0:51: 'assign' :  l-value required "gl_FragDepth" (can't modify gl_FragDepth if using early_fragment_tests)
 ERROR: 0:79: 'textureGatherOffsets(...)' : must be a compile-time constant: offsets argument
 ERROR: 0:124: 'rgba32f' : format requires readonly or writeonly memory qualifier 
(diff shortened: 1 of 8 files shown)


Test/baseResults/320.frag.out
```diff
@@ -10,7 +10,6 @@ ERROR: 0:29: 'vAnon' : redefinition
 ERROR: 0:43: 'origin_upper_left' : not supported with this profile: es
 ERROR: 0:43: 'pixel_center_integer' : not supported with this profile: es
 ERROR: 0:43: 'redeclaration' : cannot redeclare with different qualification: gl_FragCoord
-ERROR: 0:47: 'depth layout qualifier' : not supported with this profile: es
```
arcady-lunarg, line 13: This doesn't seem quite right, shouldn't it still complain here but about the extension not being enabled?

glslang/MachineIndependent/Versions.cpp
arcady-lunarg, line 1: You also need to add the `"#define GL_EXT_conservative_depth 1"` to `TParseVersions::getPreamble`

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -6799,8 +6799,10 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
         }
         for (TLayoutDepth depth = (TLayoutDepth)(EldNone + 1); depth < EldCount; depth = (TLayoutDepth)(depth+1)) {
             if (id == TQualifier::getLayoutDepthString(depth)) {
-                requireProfile(loc, ECoreProfile | ECompatibilityProfile, "depth layout qualifier");
-                profileRequires(loc, ECoreProfile | ECompatibilityProfile, 420, nullptr, "depth layout qualifier");
+                const char* feature = "depth layout qualifier";
+                requireProfile(loc, ECoreProfile | ECompatibilityProfile | EEsProfile, feature);
+                profileRequires(loc, ECoreProfile | ECompatibilityProfile, 420, nullptr, feature);
+                profileRequires(loc, EEsProfile, 300, E_GL_EXT_conservative_depth, feature);
```
arcady-lunarg: You need `profileRequires(loc, EEsProfile, 0, ...)` here because the feature isn't core in any ES version but requires the extension.

Test/baseResults/320.frag.out
beaufortfrancois: Fixed. Thanks!

glslang/MachineIndependent/Versions.cpp
beaufortfrancois: Done

glslang/MachineIndependent/ParseHelper.cpp
beaufortfrancois: Good catch!

beaufortfrancois (commented):
I've addressed your feedback and added a test file.

arcady-lunarg (approved):
LGTM, thanks.

PR #4103, opened 2025-11-19
Fix type and remove condition

jeremy-lunarg:
Change int to uint and remove condition that can never be true since coopMatKHRuse is unsigned and CM_MatrixUseA is a contant zero.


jeremy-lunarg:
@wooyoungqcom This was pointed out to me by @pmistryNV. Can you take a look? Thanks.

jeremy-lunarg:
> LGTM
> 
> Technically, though, int will be promoted to unsigned before comparison is performed, and so "int" is OK.

Yeah, I don't really think this is a bug per se, but I prefer the unsigned type.

PR #4104, opened 2025-11-20
Fix Debug Info for Retargeted Builtins

### LINKED ISSUE #3951: Debug Info for Aliased Builtin is Missing/Incorrect
In GLSL (extension) specs, some builtins may be just an alias to the other. These are mapped using `TSymbolTable::retargetSymbol` in glslang to a canonical form. The list of such builtins includes:

- gl_InstanceID -> gl_InstanceIndex
- gl_VertexID -> gl_VertexIndex
- gl_HitTNV -> gl_RayTmaxNV
- gl_HitTEXT -> gl_RayTmaxEXT

However, such mapping causes loss of information of the original used builtin and therefore leads to missing/incorrect debug info. For example, if "gl_HitTNV" is used in a shader, one variable decorated with RayTmaxKHR will be emitted and associated with a DebugGlobalVariable "RayTmaxNV", which is off expectation.

Considering only four special cases exist, I'm thinking that we special case those four builtins so that whenever we emit debug info for "RayTmaxNV", we emit another debug global variable instruction "gl_HitTNV" pointing to the same variable (same handling for other cases). Would you guys consider that as a valid solution instead of a full-blown implementation that tracks all variables that is being retargeted throughout the compilation process? Please help share your opinion :)

qingyuanzNV:
Fix #3951 

This patch wires retargeted symbol mappings from symbol table to root intermediate so that we may emit extra DebugGlobalVariable for builtins that has alias names. Because we now may have one variable id that maps to multiple debug global variable ids, we also removes debug id mapping between these in case of unexpected bugs. Such mapping isn't being used right now.


glslang/MachineIndependent/SymbolTable.h
```diff
@@ -662,9 +668,10 @@ class TSymbolTable {
     //
 protected:
     static const uint32_t LevelFlagBitOffset = 56;
+    static const int builtinLevel = 2;
```
dnovillo: This will need a definition for this in a .cpp file. This is probably causing a linker error. Maybe use a `constexpr`? Or just add a definition in a .cpp file.

SPIRV/GlslangToSpv.cpp
```diff
@@ -5167,6 +5167,16 @@ spv::Id TGlslangToSpvTraverser::createSpvVariable(const glslang::TIntermSymbol*
     }
 
     spv::Id var = builder.createVariable(spv::NoPrecision, storageClass, spvType, name, initializer, false);
+
+    if (options.emitNonSemanticShaderDebugInfo && storageClass != spv::StorageClass::Function) {
+        // Create variable alias for retargeted symbols if any.
+        // Notably, this is only applicable to built-in variables so that it is okay to only use name the key.
```
dnovillo: nit: s/use name the key/use the name as the key/

glslang/MachineIndependent/SymbolTable.h
```diff
@@ -505,6 +506,11 @@ class TSymbolTableLevel {
         retargetedSymbols.push_back({from, to});
     }
 
+    void collectRetargetedSymbols(std::unordered_multimap<std::string, std::string> &out) const {
+        for (const auto &[fromName, toName] : retargetedSymbols)
+            out.insert({std::string{toName}, std::string{fromName}});
+    }
```
dnovillo, line 512: fromName and toName tripped me a bit. Maybe `canonicalName` and `aliasName`?

PR #4105, opened 2025-11-23
Fix #3828: --glsl-version doesn't work

### LINKED ISSUE #3828: --glsl-version flag for glslangValidator.exe does not work
Repro:
1. create shader file 'test.vert'
```
#version 300 es
layout (location = 0) uniform vec4 test;
void main()
{
}
```
2. test file using glslangValidator.exe
`glslangValidator.exe -S vert --glsl-version 310es test.vert`
3. result:
`ERROR: 0:1: 'location qualifier on uniform or buffer' : not supported for this version or the enabled extensions
`
4. expected behavior:
`code compiles`

After changing `#version 300 es` in source code to `#version 310 es` error is gone.

Glslang Version: 11:15.1.0

Jaso5:
This fixes the bug based on my use case.

### DIFF
--- a/StandAlone/StandAlone.cpp
+++ b/StandAlone/StandAlone.cpp
@@ -1932,11 +1932,15 @@ void CompileFile(const char* fileName, ShHandle compiler)
     if (UserPreamble.isSet())
         Error("-D, -U and -P options require -l (linking)\n");
 
+    int defaultVersion = (GlslVersion == 0) ?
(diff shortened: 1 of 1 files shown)


jeremy-lunarg:
Did you consider exposing the `overrideVersion` parameter of `CompileDeferred` via `ShCompile`? I'm concerned about the usage of `defaultVersion` here.

PR #4116, opened 2025-12-03
Allow gl_FragDepth identifier to be redeclared

beaufortfrancois:


arcady-lunarg:
Yeah, you will need to dig a bit deeper in `TParseContext::redeclareBuiltinVariable` which controls whether redeclaration is allowed at all and it prohibits it entirely for ES < 3.2. I think it would be reasonable to special-case this one variable with this one extension.

PR #4122, opened 2025-12-11
Remove invalid reading from MeshEXT output

spencer-lunarg:
From https://gitlab.khronos.org/vulkan/vulkan/-/issues/4587 and https://github.com/KhronosGroup/SPIRV-Tools/pull/6426

You are not allowed to read from the Output in a MeshEXT shader... seems the tests were copied-and-pasted form MeshNV where it is allowed

image


arcady-lunarg (approved):

arcady-lunarg (approved):
LGTM, though in the future I would prefer you (and all contributors) put more of the helpful explanation in the commit message because that gets replicated to everyone whereas PR comments stay on github.

spencer-lunarg:
@arcady-lunarg ... let me update my other PR right now :runner:

PR #4132, opened 2025-12-18
Implement GL_EXT_long_vector

jeffbolznv:
GLSL spec: https://github.com/KhronosGroup/GLSL/pull/297
SPIR-V spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/377

### DIFF
--- a/known_good.json
+++ b/known_good.json
@@ -3,16 +3,16 @@
     {
       "name" : "spirv-tools",
       "site" : "github",
-      "subrepo" : "jeffbolznv/SPIRV-Tools",
+      "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit": "a614bf498a943e68a81a2489638860225f08c828"
+      "commit": "5f4b62de998d161ee01e6f7706da2d855f8fcdc8"
     },
     {
       "name" : "spirv-tools/external/spirv-headers",
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Headers",
       "subdir" : "External/spirv-tools/external/spirv-headers",
-      "commit" : "0a7f626a6ae86284a413d105b47a6fb413bf6c92"
+      "commit" : "babee77020ff82b571d723ce2c0262e2ec0ee3f1"
     },
     {
       "name": "googletest",


known_good.json
```diff
@@ -3,16 +3,16 @@
     {
       "name" : "spirv-tools",
       "site" : "github",
-      "subrepo" : "KhronosGroup/SPIRV-Tools",
+      "subrepo" : "jeffbolznv/SPIRV-Tools",
```
jeffbolznv: Temporary until https://github.com/KhronosGroup/SPIRV-Tools/pull/6471 is merged.

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -394,6 +394,81 @@ void AddTabledBuiltin(TString& decls, const BuiltInFunction& function)
     }
 }
 
+// Add long vector prototype for the builtin function
```
dnovillo: Could you add a high-level comment here? This one is kind of difficult to follow. This skips over ClassV1 and ClassV3, right?

glslang/MachineIndependent/Initialize.cpp
```diff
@@ -11548,4 +11741,407 @@ void TBuiltIns::identifyBuiltIns(int version, EProfile profile, const SpvVersion
...
+        }
+    };
+
+    const auto &checkSameType = [&](TIntermTyped *t0, TIntermTyped *t1) {
+        if (t0->getType() != t1->getType()) {
+            valid = false;
+        }
+    };
+
+    auto const &checkArgsMatch = [&](uint32_t argStart, uint32_t argEnd, uint32_t types) {
```
dnovillo: nit: 'const auto' (like the other lambdas).

glslang/MachineIndependent/linkValidate.cpp
```diff
@@ -2573,6 +2573,22 @@ int TIntermediate::getBaseAlignment(const TType& type, int& size, int& stride, T
         }
     }
 
+    // rules 2 and 3
+    if (type.isLongVector()) {
+        int scalarAlign = getBaseAlignmentScalar(type, size);
+        uint32_t vectorSize = type.getTypeParameters()->arraySizes->getDimSize(0);
+        switch (vectorSize) {
+        case 1: // HLSL has this, GLSL does not
+            return scalarAlign;
```
dnovillo, line 2582: But GLSL does not actually forbid 1-component vectors, right? Not sure if we are actually validating for this.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1574,7 +1585,7 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
 
             handleVector2CoopMatConversionCall(loc, fnCandidate, result, arguments);
 
-            if (result->getAsTyped()->getType().isCoopVecNV() &&
+            if ((result->getAsTyped()->getType().isCoopVecNV() || result->getAsTyped()->getType().isLongVector()) &&
```
dnovillo: This shows up a few times. Use isCoopmatOrVec()?

SPIRV/SpvBuilder.cpp
```diff
@@ -3344,12 +3344,20 @@ Id Builder::createLvalueSwizzle(Id typeId, Id target, Id source, const std::vect
 // Comments in header
 void Builder::promoteScalar(Decoration precision, Id& left, Id& right)
 {
-    int direction = getNumComponents(right) - getNumComponents(left);
+    int direction = !isScalar(right) - !isScalar(left);
```
dnovillo, line 3348: Comment this one? It made more immediate sense before when we were subtracting number of components, but this one is more subtle.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -2179,6 +2190,12 @@ void TParseContext::handleVector2CoopMatConversionCall(const TSourceLoc& loc, co
   }
 }
 
+void TParseContext::handleLongVectorBuiltin(const TSourceLoc& loc, const TFunction* fnCandidate, TType* resultType, TIntermNode* arguments)
+{
+    if (!IsSupportedLongVectorBuiltin(fnCandidate, resultType, arguments)) {
+        error(loc, "can't resolve types", fnCandidate->getName().c_str(), "");
```
dnovillo: Make the error more specific. This is for long vector types only, right?

glslang/MachineIndependent/linkValidate.cpp
jeffbolznv: HLSL has "float1", GLSL does not have "vec1". I think for GLSL this is rejected in the parser (it's not a valid token) but it is a valid token in the HLSL parser.

long vector supports vectors with one component.

glslang/MachineIndependent/Initialize.cpp
jeffbolznv: done.

glslang/MachineIndependent/ParseHelper.cpp
jeffbolznv: done.

known_good.json
jeffbolznv: The SPIRV-Tools change has landed, I've fixed this.

PR #4137, opened 2025-12-31
fix alignment of guard blocks

jamesdolan:
the `GUARD_BLOCKS` feature that is enabled when `NDEBUG` is not specified can cause unaligned memory allocations. This was causing Emscripten debug builds to throw alignment faults when `-sSAFE_HEAP=2`, and causing Safari to leak large amounts of memory when safe heap was disabled.

this change simply fixes the guard block size logic to ensure the combined header+guard size is 16-byte aligned which I believe was the intention.

Also, I realize the leak claim sounds crazy, and if true is clearly a safari bug, but fixing alignment issue is still a good idea. Also here is a [simple repro](https://github.com/jamesdolan/safari_leak).

### DIFF
--- a/glslang/Include/PoolAlloc.h
+++ b/glslang/Include/PoolAlloc.h
@@ -127,9 +127,13 @@ private:
 #   ifdef GUARD_BLOCKS
     inline static constexpr size_t headerSize() { return sizeof(TAllocation); }
     inline static constexpr size_t guardBlockSize() {
-        constexpr size_t kMinGuard = 16;
-        constexpr size_t kAlignment = 16;
-        return ((kMinGuard + sizeof(TAllocation) + kAlignment - 1) & ~(kAlignment - 1)) - sizeof(TAllocation);
+        constexpr size_t minGuardSize = 16;
+        constexpr size_t alignmentSize = 16;
+        constexpr size_t guardLayoutSize =
+            (minGuardSize + sizeof(TAllocation) + alignmentSize - 1) & ~(alignmentSize - 1);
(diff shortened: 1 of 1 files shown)


glslang/Include/PoolAlloc.h
```diff
@@ -125,15 +125,15 @@ class TAllocation {
...
-#   else
-    static inline constexpr size_t guardBlockSize = 0;
-#   endif
-
-#   ifdef GUARD_BLOCKS
-    inline static size_t headerSize() { return sizeof(TAllocation); }
+    inline static constexpr size_t headerSize() { return sizeof(TAllocation); }
+    inline static constexpr size_t guardBlockSize() {
+        constexpr size_t kMinGuard = 16;
+        constexpr size_t kAlignment = 16;
```
dnovillo: glslang does not use `k` prefixes, just camel case is fine. Perhaps `minGuardSize` and `alignmentSize`?

glslang/Include/PoolAlloc.h
```diff
@@ -125,15 +125,15 @@ class TAllocation {
...
+        return ((kMinGuard + sizeof(TAllocation) + kAlignment - 1) & ~(kAlignment - 1)) - sizeof(TAllocation);
+    }
```
dnovillo, line 137: To prevent future regressions, how about adding an assert here?

```
static_assert((TAllocation::headerSize() + TAllocation::guardBlockSize()) % 16 == 0,
              "Guard block layout is not 16-byte aligned.");
```

PR #4138, opened 2026-01-03
Print compile/link logs to stderr

emersion:
When glslang is integrated into a build system (like Meson), stdout is stored in an output file and stderr is printed to the user. This hides error messages from the user. See: https://github.com/mesonbuild/meson/issues/11506

stderr is the standard stream errors should get printed to, always use that.


dnovillo (changes requested):
This may be confusing tests expecting output on stdout, though. They will need to be adapted. I think you may find failures in the testsuite with this.

dnovillo (approved):
LGTM

emersion:
Thanks for reviewing!

PR #4150, opened 2026-01-16
cmake: Enable more MSVC warnings

arcady-lunarg:
These warnings are automatically enabled by MSVC when doing LTO so we may as well enable them all the time.
These came up in #4065


jeremy-lunarg (approved):
LGTM

PR #4151, opened 2026-01-17
Add missing preamble #define for GL_EXT_nontemporal_keyword

### LINKED ISSUE #4141: GL_EXT_nontemporal_keyword not defined when the extension is enabled
The following code compiles fine when it should error out:

``` GLSL
#version 460
#pragma shader_stage(compute)
#extension GL_EXT_nontemporal_keyword : require 

#ifdef GL_EXT_nontemporal_keyword
#error Success!
#endif

void main(void) {}
```

Replacing `GL_EXT_nontemporal_keyword` with any other extension causes the error to be printed and the compilation halted.

arcady-lunarg:
Fixes #4141

### DIFF
--- a/glslang/MachineIndependent/Versions.cpp
+++ b/glslang/MachineIndependent/Versions.cpp
@@ -663,6 +663,11 @@ void TParseVersions::getPreamble(std::string& preamble)
         }
     }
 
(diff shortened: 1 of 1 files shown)


jeremy-lunarg (approved):
LGTM

PR #4154, opened 2026-01-19
Implement <system> include for the standalone wrapper

mmmhj2:
This pull request adds `#include <system>` directive support for the standalone wrapper as specified in the `ARB_shading_language_include` extension.

The current spec seems to give the implementation a lot of liberties on how to resolve the inclusion paths, similar to the C/C++ standard. This PR implements this directive similar to the [GCC preprocessor](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):
- "local" inclusions will be searched first in the current source directory recursively, and then external directories specified by `-I` flag, in left-to-right order;
- \<system\> inclusions will be searched in external directories in left-to-right order only.

### DIFF
--- a/Test/baseResults/hlsl.include.system.unresolved.vert.out
+++ b/Test/baseResults/hlsl.include.system.unresolved.vert.out
@@ -1,0 +1,8 @@
+hlsl.include.system.unresolved.vert
+ERROR: hlsl.include.system.unresolved.vert:2: '#include' : Could not process include directive for header name: foo.h
+ERROR: 1 compilation errors.  No code generated.
+
+
+WARNING: Linking vertex stage: Entry point not found
+
+SPIR-V is not generated for failed compile or link
--- a/Test/hlsl.include.system.unresolved.vert
+++ b/Test/hlsl.include.system.unresolved.vert
@@ -1,0 +1,6 @@
+// Should fail even though "./foo.h" exists
+#include <foo.h>
+
+int main() {
+#error in main
+}
--- a/Test/runtests
+++ b/Test/runtests
@@ -189,6 +189,8 @@ run -D -Od -e main -H -Od hlsl.includeNegative.vert > "$TARGETDIR/hlsl.includeNe
 diff -b $BASEDIR/hlsl.includeNegative.vert.out "$TARGETDIR/hlsl.includeNegative.vert.out" || HASERROR=1
 run -D -Od -e main -H -Od hlsl.include.system.vert -Iinc1 -Iinc1/path1 -Iinc1/path2 > "$TARGETDIR/hlsl.include.system.vert.out"
 diff -b $BASEDIR/hlsl.include.system.vert.out "$TARGETDIR/hlsl.include.system.vert.out" || HASERROR=1
+run -D -Od -e main -H -Od hlsl.include.system.unresolved.vert > "$TARGETDIR/hlsl.include.system.unresolved.vert.out"
+diff -b $BASEDIR/hlsl.include.system.unresolved.vert.out "$TARGETDIR/hlsl.include.system.unresolved.vert.out" || HASERROR=1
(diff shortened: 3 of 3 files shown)


StandAlone/DirStackFileIncluder.h
```diff
@@ -94,27 +97,31 @@ class DirStackFileIncluder : public glslang::TShader::Includer {
...
+            localDirectoryStack.back() = getDirectory(includerName);
+    }
 
+    // Search for a valid "local" path based on combining the stack of include
+    // directories and the nominal name of the header.
+    virtual IncludeResult* readLocalPath(const char* headerName)
+    {
         // Find a directory that works, using a reverse search of the include stack.
-        for (auto it = directoryStack.rbegin(); it != directoryStack.rend(); ++it) {
+        for (auto it = localDirectoryStack.rbegin(); it != localDirectoryStack.rend(); ++it) {
```
dnovillo, line 119: Before, this would go through `directoryStack`, which interleaves external and local directories. By splitting the two, this one only looks at `localDirectoryStack`. This seems to contradict the description (local inclusions are searched first followed by external directories).

I'm half expecting something like `Test/hlsl.dashI.vert` to fail. Shouldn't this go on to test `externalDirectoryList`?

StandAlone/DirStackFileIncluder.h
```diff
@@ -94,27 +97,31 @@ class DirStackFileIncluder : public glslang::TShader::Includer {
...
+    std::vector<std::string> externalDirectoryList;
     std::set<std::string> includedFiles;
 
-    // Search for a valid "local" path based on combining the stack of include
-    // directories and the nominal name of the header.
-    virtual IncludeResult* readLocalPath(const char* headerName, const char* includerName, int depth)
+    // Record local path for future local inclusion.
+    // Discard popped include directories, and
+    // initialize when at parse-time first level.
+    void RecordLocalPath(const char * includerName, size_t depth)
```
dnovillo: Methods should use camelCase. s/RecordLocalPath/recordLocalPath/

StandAlone/DirStackFileIncluder.h
```diff
@@ -124,9 +131,19 @@ class DirStackFileIncluder : public glslang::TShader::Includer {
     }
 
     // Search for a valid <system> path.
-    // Not implemented yet; returning nullptr signals failure to find.
-    virtual IncludeResult* readSystemPath(const char* /*headerName*/) const
+    virtual IncludeResult* readSystemPath(const char* headerName)
     {
+        // Search for external diectories only.
```
dnovillo: s/diectories/directories/

StandAlone/DirStackFileIncluder.h
mmmhj2: It is [documented](https://github.com/KhronosGroup/glslang/blob/f1437db5fbf51161b5de0eca18121ef359c5dedd/glslang/Public/ShaderLang.h#L652) that when `includeLocal()` fails by returning a `nullptr`, `includeSystem()` method will be called instead. So this method searchs only `localDirectoryStack` and delegate `externalDirectoryList` to the `includeSystem()` method, and therefore local inclusions are searched (by `includeLocal()`) and then system inclusions (by `includeSystem()`).

dnovillo: > It is [documented](https://github.com/KhronosGroup/glslang/blob/f1437db5fbf51161b5de0eca18121ef359c5dedd/glslang/Public/ShaderLang.h#L652) that when `includeLocal()` fails by returning a `nullptr`, `includeSystem()` method will be called instead. So this method searchs only `localDirectoryStack` and delegate `externalDirectoryList` to the `includeSystem()` method, and therefore local inclusions are searched (by `includeLocal()`) and then system inclusions (by `includeSystem()`).

Thanks. I had missed that connection.

PR #4157, opened 2026-01-21
Implement support for GL_NV_push_constant_bank

pdaniell-nv:


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7204,6 +7204,21 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
         return;
     }
 
+    if (id == "bank") {
+        requireExtensions(loc, 1, &E_GL_NV_push_constant_bank, "push constant bank");
+        if (nonLiteral)
+            error(loc, "needs a literal integer", id.c_str(), "");
+        else
```
dnovillo, line 7213: Minor nit: should this validate that the bank is in the range 0 <= value <= TQualifier::layoutBankEnd?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7204,6 +7204,21 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
...
+            publicType.qualifier.layoutBank = value;
+        return;
+    }
+    if (id == "member_offset") {
```
dnovillo, line 7217: This should also have an extension check.

Issue #3903, opened 2025-03-18
NULL Pointer Dereference in glslang::TIntermediate::isConversionAllowed


sae-as-me:
Note: the [security advisory](https://github.com/KhronosGroup/glslang/security/advisories/new) will redirect to a 404 page.
So I report it in this.

## Affected Projects
glslang 15.1.0 ([https://github.com/KhronosGroup/glslang](https://github.com/KhronosGroup/glslang))
## Problem Type
CWE-476: NULL Pointer Dereference
### Summary
A null pointer read vulnerability was discovered in the `glslang::TIntermediate::isConversionAllowed` function within glslang. This issue occurs when processing certain files, leading to a potential application crash.
### Details
The vulnerability arises in the `glslang::TIntermediate::isConversionAllowed` function defined in `glslang/MachineIndependent/Intermediate.cpp` at line `541`. 
The reason is that the function `H5O__cache_chk_serialize` did not check the validity of `node` before calling `node->getBasicType()`, resulting in a null pointer dereference read.
```c
bool TIntermediate::isConversionAllowed(TOperator op, TIntermTyped* node) const
{
    //
    // Does the base type even allow the operation?
    //
    switch (node->getBasicType()) {    //null pointer read
    case EbtVoid:
        return false;
    case EbtAtomicUint:
    case EbtSampler:
    case EbtAccStruct:
        // opaque types can be passed to functions
        if (op == EOpFunction)
            break;

        // HLSL can assign samplers directly (no constructor)
        if (getSource() == EShSourceHlsl && node->getBasicType() == EbtSampler)
            break;

        // samplers can get assigned via a sampler constructor
        // (well, not yet, but code in the rest of this function is ready for it)
        if (node->getBasicType() == EbtSampler && op == EOpAssign &&
            node->getAsOperator() != nullptr && node->getAsOperator()->getOp() == EOpConstructTextureSampler)
            break;

        // otherwise, opaque types can't even be operated on, let alone converted
        return false;
    default:
        break;
    }

    return true;
}
```

### PoC
**Steps to reproduce:**
1. Clone the glslang repository and build it using the following commands : 
```sh
export CC='clang'
export CXX='clang++'
export CFLAGS='-fsanitize=address -O0 -g'
export CXXFLAGS='-fsanitize=address -O0 -g'

mkdir -p build
cd build
[...]

PR #4169, opened 2026-02-02
Fix CVE-2025-3010: Null pointer dereference in isConversionAllowed

cnDelbert:
Added a null check for 'node' in TIntermediate::isConversionAllowed to prevent a null pointer dereference when processing malformed HLSL conditional expressions.

This addresses the vulnerability reported in CVE-2025-3010 where a crafted shader could cause a crash due to missing validation before accessing node->getBasicType().

References:
- CVE-2025-3010
- https://github.com/KhronosGroup/glslang/issues/3903


pmistryNV:
Seems like ToSpirv/HlslCompileTest.FromFile/hlsl_cve_2025_3010_frag is failing. Can you please check? Let me know if you need help with upstreaming

glslang/MachineIndependent/Intermediate.cpp
```diff
@@ -546,6 +546,9 @@ TIntermTyped* TIntermediate::setAggregateOperator(TIntermNode* node, TOperator o
 
 bool TIntermediate::isConversionAllowed(TOperator op, TIntermTyped* node) const
 {
+    if (node == nullptr)
+        return false;
```
dnovillo, line 550: This seems like it's papering over the issue. The HLSL frontend is generating a `nil` `node` and passing it down. While this prevents the crash, it will silently supresses a potential semantic error. Where is this `nil` pointer generated? That is probably the spot to fix.

Test/hlsl.cve-2025-3010.frag
dnovillo, line 1: This file seems like a valid HLSL shader? The CVE should produce a `nil` `TIntermTyped*`. Do you have the original invalid shader?

dnovillo (changes requested):
The golden output generated starts with `Test\`. It should not have the path, just the shader name. That seems to be the failure you're getting.

PR #4173, opened 2026-02-19
Raise the maximum limit for specialization constant IDs

### LINKED ISSUE #3852: Raise Specialization Constant Limit
As discussed in https://github.com/KhronosGroup/Vulkan-Docs/issues/2484, there is no defined limit on specialization constants. However, glslang will not allow more than `0x7FF` constants to be defined.

Is there a specific reason for this, or can this limit be raised or removed?

arcady-lunarg:
This was capped at 0x7ff for unclear reasons, now it can be any 32-bit value up to INT_MAX.

Fixes #3852

### DIFF
--- a/Test/baseResults/vulkan.vert.out
+++ b/Test/baseResults/vulkan.vert.out
@@ -17,7 +17,7 @@ ERROR: 0:13: 'constant_id' : specialization-constant id already used
 ERROR: 0:13: 'binding' : sampler/texture/image requires layout(binding=X) 
 ERROR: 0:13: 'constant_id' : can only be applied to 'const'-qualified scalar 
 ERROR: 0:13: 'constant_id' : cannot be applied to this type 
-ERROR: 0:14: 'constant_id' : specialization-constant id is too large 
+ERROR: 0:14: 'layout-id value' : cannot be negative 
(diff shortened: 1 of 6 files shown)


Test/baseResults/vulkan.vert.out
```diff
@@ -17,7 +17,7 @@ ERROR: 0:13: 'constant_id' : specialization-constant id already used
 ERROR: 0:13: 'binding' : sampler/texture/image requires layout(binding=X) 
 ERROR: 0:13: 'constant_id' : can only be applied to 'const'-qualified scalar 
 ERROR: 0:13: 'constant_id' : cannot be applied to this type 
-ERROR: 0:14: 'constant_id' : specialization-constant id is too large 
+ERROR: 0:14: 'layout-id value' : cannot be negative
```
jeremy-lunarg, line 20: Is this error expected?

jeremy-lunarg (approved):
Just the one question. Otherwise LGTM.

Test/baseResults/vulkan.vert.out
arcady-lunarg: Kind of? It's out of range for INT_MAX, but maybe it should be phrased better.

PR #4192, opened 2026-03-20
Add support for bfloat16 and float8 tensors

einarhov-arm:
Add the following extensions:

GL_ARM_tensors_bfloat16
GL_ARM_tensors_float_e5m2
GL_ARM_tensors_float_e4m3


svenvh (approved):

arcady-lunarg (changes requested):
You need to add the GLSL preprocessor `#define`s for the new extension names to the preamble as per point 3 of the instructions in `glslang/MachineIndependent/Versions.cpp` (and actually also to do this for the original GL_ARM_tensors symbol as well).

arcady-lunarg (approved):

PR #4198, opened 2026-03-24
Fix for issue #4197

ErikHogeman:
Update access chain handling to be compatible with raw descriptor heap accesses.


arcady-lunarg (changes requested):
@ErikHogeman You will need to add your new test to one of the lists in `gtests/Spv.FromFile.cpp` so that it actually gets run.

ErikHogeman:
> @ErikHogeman You will need to add your new test to one of the lists in `gtests/Spv.FromFile.cpp` so that it actually gets run.

Thanks, not sure how I missed that. Fixed now.

arcady-lunarg (approved):

PR #4200, opened 2026-03-27
Check for bad parameters on long vector type

dneto0:
Bug: crbug.com/486238211


dneto0:
cc: @dnovillo @jeffbolznv

glslang/Include/Types.h
```diff
@@ -1567,6 +1567,9 @@ class TPublicType {
     bool isTensorLayoutNV() const { return basicType == EbtTensorLayoutNV; }
     bool isTensorViewNV() const { return basicType == EbtTensorViewNV; }
 
+    const TTypeParameters* getTypeParameters() const { return typeParameters; }
+    const TArraySizes* getArraySizes() const { return arraySizes; }
```
dnovillo: This is not used anywhere. Needed?

Test/spv.longVectorBadParamsError.comp
```diff
@@ -0,0 +1,20 @@
...
+
+void main()
+{
+    vector<5, 5> noElemType;
+    vector<void, 5> badElemType;
+    vector<float> missingCount;
+    vector<float, 5, 7> excessCount;
+
+    vector() yep;
+    vector<void,5>() oof;
```
dnovillo: I can't access the original crbug (is there a way to change that?), but I don't think this will exercise the null-dereference. The `vector() yep` will produce a syntax error and it supresses any other diagnostics. Maybe something like this?

```
float a = vector<float>(0);
float b = vector<float,5,7>(0);
```

glslang/Include/Types.h
dnovillo: `getArraySizes` does not seem to be called anywhere. Remove?

dnovillo (changes requested):

glslang/Include/Types.h
dnovillo: ugh. sorry for the duplicate comment. I had a stale connection.

Test/spv.longVectorBadParamsError.comp
dneto0: Sorry, I can't change the visibility on the bug just yet.

The original bug is actually exactly this code pattern:
```
  vector() yep;
```
Prior to the fix, you never actually get as far as emitting a diagnostic for the syntax error.

But point taken, I will split these tests into those that are passably syntactically correct, and those that are not (which would include this bad scenario).

dnovillo: > Sorry, I can't change the visibility on the bug just yet.

Yeah, I figured. Not a big deal if the root issue is represented in a test here.

> But point taken, I will split these tests into those that are passably syntactically correct, and those that are not (which would include this bad scenario).

Thanks!

dneto0: tests are now split.

Separating them actually found another bug, so thanks!  There's three files now.

glslang/Include/Types.h
dneto0: Sharp eyes.
Done.

dnovillo: Thanks. Merging as soon as the bot is done.

PR #4220, opened 2026-04-16
Initial support for NonSemantic.Shader.DebugInfo 101

dnovillo:
This implements NSDI.101 (https://github.com/KhronosGroup/SPIRV-Registry/pull/390) in glslang. It updates SPIRV-Tools and SPIRV-Headers dependencies to bring in the new headers for NSDI.101.

### Version promotion

To minimize disruption with existing tools, glslang will always generate an import of `NonSemantic.Shader.DebugInfo.100`. If the module uses an opcode that requires NSDI.101 (cooperative types and new floating point types), it then changes the import string to `.101`.  The promotion is done by patching the `OpExtInstImport` instruction's string operand.

The entry point is `Builder::requireNonSemanticShaderDebugInfoVersion`.  Each of the three new emitter functions
calls it unconditionally before emitting the instruction.

### New emitters

- `Builder::makeVectorIdDebugType`: emits `DebugTypeVectorIdEXT`.  Called from `makeCooperativeVectorTypeNV` when `emitNonSemanticShaderDebugInfo` is set.
- `Builder::makeCooperativeMatrixDebugTypeKHR`: emits `DebugTypeCooperativeMatrixKHR`.  Replaces the former opaque-composite workaround in `makeCooperativeMatrixTypeKHR`.
- `Builder::makeFloatDebugType`: extended with an optional `fpEncoding` operand.  The three float-8 and bfloat16 type builders (`makeBFloat16Type`, `makeFloatE5M2Type`, `makeFloatE4M3Type`) now handle the new floating point types.

### Header rename

This needed to import the new `NonSemanticShaderDebugInfo.h` from SPIRV-Headers. This means that all the previous C identifiers that had `100` embedded in them needed to change. Those changes are mechanical, there is no functional change there.

I also added a section in `README.md` to document how to update this headers in future revisions of NSDI.

### Tests

I replaced the previous golden output tests `spv.debuginfo.coopmatKHR.comp` with a new unit test file `gtests/SpvDebugInfoTest.cpp`. All the NSDI.101 opcodes are tested there.


dnovillo:
For reference, these are the PRs adding NSDI.101 support:

- SPIRV-Headers: https://github.com/KhronosGroup/SPIRV-Headers/pull/579
- SPIRV-Tools: https://github.com/KhronosGroup/SPIRV-Tools/pull/6635

gtests/Spv.FromFile.cpp
```diff
@@ -1157,14 +1149,6 @@ INSTANTIATE_TEST_SUITE_P(
     FileNameAsCustomTestSuffix
 );
 
-INSTANTIATE_TEST_SUITE_P(
-    Glsl, GlslNonSemanticShaderDebugInfoSpirv13Test,
-    ::testing::ValuesIn(std::vector<std::string>({
-        "spv.debuginfo.coopmatKHR.comp",
```
jeremy-lunarg, line 1166: Can this test be removed?

arcady-lunarg:
LGTM aside from Jeremy's comment. Before I merge, could you squash this into just two or three commits? One for the rename of the header file and associated enumeration values and the other one for the "initial implementation"? And the tests can go in their own commit or together with the implementation.

gtests/Spv.FromFile.cpp
dnovillo: Done.

dnovillo:
> LGTM aside from Jeremy's comment. Before I merge, could you squash this into just two or three commits? One for the rename of the header file and associated enumeration values and the other one for the "initial implementation"? And the tests can go in their own commit or together with the implementation.

Sure. I'll re-do the commits so they are split that way.

dnovillo:
> > LGTM aside from Jeremy's comment. Before I merge, could you squash this into just two or three commits? One for the rename of the header file and associated enumeration values and the other one for the "initial implementation"? And the tests can go in their own commit or together with the implementation.
> 
> Sure. I'll re-do the commits so they are split that way.

Done. I split it in two commits: the rename with no functional changes and the new implementation with tests. I also realized that the tests were not covering the coopmat test I removed from Test/ (spec constants with all three use types simultaneously). I added it to the new gtests/ implementation.

PTAL. Thanks.

SPIRV/disassemble.cpp
```diff
@@ -547,8 +547,8 @@ void SpirvStream::disassembleInstruction(Id resultId, Id /*typeId*/, Op opCode,
                     extInstSet = NonSemanticDebugPrintfExtInst;
                 } else if (strcmp("NonSemantic.DebugBreak", name) == 0) {
                     extInstSet = NonSemanticDebugBreakExtInst;
-                } else if (strcmp("NonSemantic.Shader.DebugInfo.100", name) == 0) {
-                    extInstSet = NonSemanticShaderDebugInfo100;
+                } else if (strncmp("NonSemantic.Shader.DebugInfo.", name, 29) == 0) {
```
jeffbolznv: I'd rather use strlen or sizeof than "29", but it's not a huge deal.

jeffbolznv (approved):
I primarily reviewed the spvbuilder changes, I didn't go through the test code.

SPIRV/disassemble.cpp
dnovillo: Sure. Done.

arcady-lunarg (approved):

PR #4229, opened 2026-04-24
Add small type capabilities for GLSL.STD.450

GinShio:
Typically, GLSL.STD.450 extension instructions are used with other arithmetic operations or conversions for small types, which works fine. However, if the shader only uses GLSL.STD.450 functions, we fail to add the capabilities for small types. This causes a validation error without the corresponding capabilities.


arcady-lunarg (changes requested):
You will need to add your new test to the list in `gtests/Spv.FromFile.cpp` in order for it to actually get run as part of the test suite.

arcady-lunarg (approved):

PR #4230, opened 2026-04-24
Miss spv extension for derivative functions with float16

GinShio:
AMD_gpu_shader_half_float allows use derivative functions with float16, it's similar to interpolation functions.

Extension update: https://github.com/KhronosGroup/SPIRV-Registry/pull/403


arcady-lunarg:
@GinShio Sounds like you are also working on a change to spirv-val to allow this case? Once that lands you will be able to update the known_good.json file with the appropriate git revision of SPIRV-Tools and the validation failure will go away.

GinShio:
> Sounds like you are also working on a change to spirv-val to allow this case?

Yep.

> Once that lands you will be able to update the known_good.json file with the appropriate git revision of SPIRV-Tools and the validation failure will go away.

Thanks!

GinShio:
> Once that lands you will be able to update the known_good.json file with the appropriate git revision of SPIRV-Tools and the validation failure will go away.

Change 716f9503264d539cc05503cae562e6949eada8f5 updates it. So I rebased onto branch main.

arcady-lunarg (approved):

PR #4231, opened 2026-04-25
kokoro: test with clang-18

dneto0:
Google-internal bug 393437270


dneto0:
PTAL @dnovillo

dneto0:
This is part of routine updates to Google's test images.

arcady-lunarg (approved):

PR #4234, opened 2026-04-27
Add template deduction guide for Defer

dneto0:
This avoids a Clang -Wctad-maybe-unsupported in some compiler configurations.


dneto0:
PTAL @dnovillo

dnovillo (approved):

PR #4235, opened 2026-04-28
kokoro: use GCC 15

dneto0:
Google internal bug 506281797


dneto0:
This is anothe routine tooling upgrade, in preparation for moving our bots to Ubuntu 26.04

arcady-lunarg (approved):

PR #4244, opened 2026-05-02
Fix generated decorate string from old UTF8EncodedKHR to new UTFEncodedKHR

ShchchowAMD:
Fix enum string name to keep same with current SPIRV-Header's implementation:

https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/unified1/spirv.hpp


ShchchowAMD:
@arcady-lunarg @Samsung-CraigG

arcady-lunarg (approved):

PR #4247, opened 2026-05-04
GL_KHR_compute_shader_derivatives bugfixes

### LINKED ISSUE #4239: 91bfd96ae494a9f1c503cacb0dfceef6d3075cb2 breaking GLSL_NV_compute_shader_derivatives
The validation added in 91bfd96ae494a9f1c503cacb0dfceef6d3075cb2 requires that there is a layout qualifier in the shader but that's not a thing in the NV variant of the extension : 

```
    If neither layout qualifier is specified, derivatives in compute shaders
    return zero, which is consistent with the handling of built-in texture
    functions like texture() in GLSL 4.50 compute shaders.
```

https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_compute_shader_derivatives.txt

/cc @janharaldfredriksen-arm

janharaldfredriksen-arm:
Update implementation of GL_KHR_compute_shader_derivatives to avoid regressions on existing shaders using the GL_NV variant.

In particular:
 - in the NV variant, it is allowed to specify neither of derivative_group_linearNV and derivative_group_quadsNV
 - the order of the layout qualifiers for work group size and derivative_group_* should not matter
 - add a negative test for the NV variant in mesh shaders

Should fix https://github.com/KhronosGroup/glslang/issues/4239


Test/baseResults/spv.computeShaderDerivativesNoLayout.comp.out
```diff
@@ -0,0 +1,60 @@
+spv.computeShaderDerivativesNoLayout.comp
+Validation failed
```
arcady-lunarg, line 2: I don't really like the idea of glslang generating invalid SPIR-V. Given the spec says "If neither layout qualifier is specified, derivatives in compute shaders return zero", a possible way to fix this would be a postprocessing pass (in `SPIRV/SpvPostProcess.cpp`) that would detect if a compute shader didn't have a layout declared and replace the derivative instructions with a constant 0 of the appropriate type.

janharaldfredriksen-arm: I agree it's not great.
The underlying problem is that the KHR SPIR-V extension aliases the token from the NV extension - even though the rules are subtly different. This means that spirv-val can't tell the two extensions apart and is now effectively validating this shader against the KHR spec.

Fixing it in a postprocessing pass sounds plausible for the explicit derivative instructions (like here), but not sure I can do that for the texture built-ins with implicit derivatives.

I suspect the proper way to fix this is to modify spirv-val.

arcady-lunarg: If you open a spirv-tools issue for this, I can accept the validation failure as a provisional state of affairs until the question is resolved one way or the other.

janharaldfredriksen-arm: Thanks, will do once I've had a little bit of a look at how hard it would be to fix on the spirv-tools side.

(I'm also hoping Lionel has the chance to test if these patches fixes the reported regressions before we merge - just to avoid churn.)

arcady-lunarg:
@llandwerlin-intel does this fix your issue?

llandwerlin-intel:
> @llandwerlin-intel does this fix your issue?

Yes, it does, thanks a lot!

arcady-lunarg (approved):

Test/baseResults/spv.computeShaderDerivativesNoLayout.comp.out
janharaldfredriksen-arm: @arcady-lunarg - thanks for merging. As a follow up - attempted spirv-tools fix here: https://github.com/KhronosGroup/SPIRV-Tools/pull/6678

PR #4248, opened 2026-05-05
kokoro: use NDK r29

dneto0:
Upgrade kokoro testing from NDK r27c to r29

Bug: crbug.com/507146491


dneto0:
cc: @arcady-lunarg @dnovillo

dnovillo (approved):

PR #4250, opened 2026-05-06
GLSL_QCOM_multiple_wait_queues

wooyoungqcom:


glslang/MachineIndependent/glslang.y
```diff
@@ -119,7 +119,7 @@ extern int yylex(YYSTYPE*, TParseContext&);
 
 %parse-param {glslang::TParseContext* pParseContext}
 %lex-param {parseContext}
-%pure-parser  // enable thread safety
+%define api.pure  // enable thread safety
```
arcady-lunarg, line 122: The bison documentation says this should be `%define api.pure full` but apparently it doesn't make a difference with the current version of bison.

kimwooyoung: What do you want me to do? My local bison complained about it.

SPIRV/GlslangToSpv.cpp
```diff
@@ -1393,6 +1393,10 @@ spv::LoopControlMask TGlslangToSpvTraverser::TranslateLoopControl(const glslang:
             operands.push_back(loopNode.getPartialCount());
         }
     }
+    if (loopNode.getMultipleWaitQueuesQCOM() != loopNode./*TIntermLoop::*/noMultipleQaitQueues) {
```
arcady-lunarg: Commented out code, you should remove it.

glslang/MachineIndependent/attribute.cpp
```diff
@@ -338,6 +340,14 @@ void TParseContext::handleLoopAttributes(const TAttributes& attributes, TIntermN
             if (unsignedArgument("partial_count", uiValue))
                 loop->setPartialCount(uiValue);
             break;
+        case EatMultipleWaitQueuesQCOM:
+            value = -1u;
+            if (it->size() == 0)
+                loop->setMultipleWaitQueuesQCOM(0);
+            else
+                if (positiveSignedArgument("multiple_wait_queuesQCOM", value))
```
arcady-lunarg, line 352: I think this will silently fail for non-positive arguments. You need to include handling for that case and tests to exercise it.

arcady-lunarg (changes requested):

glslang/MachineIndependent/glslang.y
arcady-lunarg: I think the change you have in the PR is fine.

glslang/MachineIndependent/attribute.cpp
wooyoungqcom: Many of the other attributes do not have codes to handle the non-positive (or negative) cases.

arcady-lunarg: Fair enough. I can try to fix that in a separate change.

arcady-lunarg (approved):

arcady-lunarg:
@wooyoungqcom Looks like you'll need to re-sign the CLA because it was updated to a new version.

wooyoungqcom:
let me talk to James Riordan again.
He said my name was added to the new list and it should be signed automatically.

wooyoungqcom:
now, the page shows CLA signed

arcady-lunarg:
@wooyoungqcom could you rebase?

wooyoungqcom:
done

PR #4251, opened 2026-05-06
Only use cooperative matrix conversions when needed

gorazd-sumkovski-arm:
Conversions should only be used when needed, such as when the matrix use differs. When creating an array of KHR cooperative matrices it is not needed.


kpet:
@jeffbolznv You might want to take a look at this PR.

jeffbolznv (approved):

arcady-lunarg (approved):

PR #4253, opened 2026-05-06
Ignore .jj tree for .gitignore and license check

dneto0:
The jj version control system uses .jj/** for storing its data, like git uses .git/**


dneto0:
This is a non-functional change

arcady-lunarg (approved):

PR #4256, opened 2026-05-06
Add REUSE compliance infrastructure and contributing guidelines

KhronosWebservices:
Establishes REUSE-compliant license tracking for the repository. Adds REUSE.toml mapping all files to their correct SPDX identifiers, full license texts for all licenses in use (Apache-2.0, BSD-2-Clause, BSD-3-Clause, MIT, MIT-Khronos-old, CC-BY-4.0, GPL-3.0-or-later with Bison exception, and LicenseRef-NVIDIA-MIT), and a GitHub Actions workflow that runs the FSFE REUSE compliance checker on every push and pull request.                                                                                         
                                         
Also adds CONTRIBUTING.md documenting how to submit contributions, SPDX licensing requirements, the Apache 2.0 CLA terms, and the policy on AI-assisted contributions.       

This was as a fairly extensive undertaking, and to the best of my knowledge nothing was missed. Please leave a comment if something doesn't make sense to you. 

The CLA  for this repository should also be updated from MIT to Apache 2.0 once these changes are made.


KhronosWebservices:
How should this repository get around the Kokoro checks with these errors:

5 errors:
* LICENSES/Bison-exception-2.2.txt has no license
* LICENSES/GPL-3.0-or-later.txt uses unsupported license 'GPL-3.0'
* LICENSES/LicenseRef-NVIDIA-MIT.txt has no license
* LICENSES/MIT-Khronos-old.txt has no license
* REUSE.toml has no license

Can the Kokoro config be set to ignore these files? Or do we need to disable reuse on this repository?

oddhack:
> How should this repository get around the Kokoro checks with these errors:
> 
> 5 errors:
> 
> * LICENSES/Bison-exception-2.2.txt has no license
> * LICENSES/GPL-3.0-or-later.txt uses unsupported license 'GPL-3.0'
> * LICENSES/LicenseRef-NVIDIA-MIT.txt has no license
> * LICENSES/MIT-Khronos-old.txt has no license
> * REUSE.toml has no license
> 
> Can the Kokoro config be set to ignore these files? Or do we need to disable reuse on this repository?

The REUSE bits of this look fine. I guess the Kokoro checks are using another license checker that I know nothing about, but hopefully it can be configured to ignore the same REUSE metadata that REUSE itself ignores (e.g. REUSE.toml and the LICENSES/ directory). Perhaps the Kokoro checks could be removed, but that seems kinda intrusive on the repo and maybe there's a need for them.

N.b. there are process advantages to implementing REUSE in all our repos, in that we can generate manifests easily if that ever comes up (for example).

oddhack (approved):
I can't speak to the Kokoro aspects but the REUSE stuff LGTM.

KhronosWebservices:
Thanks @oddhack. As we are moving towards all Khronos repos using reuse, it would be good to figure how to make it work here too.

oddhack:
> Thanks @oddhack. As we are moving towards all Khronos repos using reuse, it would be good to figure how to make it work here too.

It looks like there's a license-checker.cfg file that controls the license-check kokoro stage based on casual inspection of the build logs and kokoro repo: https://github.com/google/shaderc/blob/main/license-checker.cfg

Whether that's in this repository and can be modified, I don't know.

arcady-lunarg:
There does appear to be a license_checker.cfg in the glslang repository, you can probably just update it, there's an "exclude" section in there. It does seem a bit redundant to be doing two separate license checks though.

KhronosWebservices:
Is there way to disable the licensing checks in Kokoro? If not, trying to exclude should be fine.

arcady-lunarg:
@dneto0 Do you think it makes sense to keep the kokoro license check given that this is going to be checked  on the github/Khronos side now? Otherwise I suppose we should just add these files to the exclude list.

dneto0:
> @dneto0 Do you think it makes sense to keep the kokoro license check given that this is going to be checked on the github/Khronos side now? Otherwise I suppose we should just add these files to the exclude list.

I have to go through some bureaucracy on our side when licensing metadata is updated.  So I'd prefer the early warning via the Kokoro bot.  So I prefer adding the exclusions.

I'll try suggesting a patch.

dneto0:
FYI, for the CI-android-ndk-check , that should resolve itself after you rebase onto main.
(The github-side scripts assume an older NDK that I've since removed from the testing image)

dneto0:
Modify the license-checker.cfg files with this patch:

```
diff --git a/license-checker.cfg b/license-checker.cfg
index 8298857eb..58df087b8 100644
--- a/license-checker.cfg
+++ b/license-checker.cfg
@@ -15,6 +15,7 @@
 
                     "_config.yml",
                     ".*",
+                    ".*/**",^M
                     ".github/workflows/*.yml",
                     ".github/workflows/*.js",
                     "CMakeSettings.json",
@@ -22,6 +23,10 @@
                     "known_good.json",
                     "LICENSE.txt",
                     "make-revision",
+                    "standalone.gclient",^M
+                    "standalone.gclient_entries",^M
+                    "LICENSES/*",^M
+                    "REUSE.toml",^M
                     "WORKSPACE",
 
                     "glslang/OSDependent/Web/glslang.*.js",
```

arcady-lunarg:
@KhronosWebservices could you rebase this and update the license-checker.cfg file as suggested above?

arcady-lunarg (approved):

PR #4257, opened 2026-05-06
Adjust tests for the ENABLE_HLSL=OFF case

dneto0:
Issue: #4210


dneto0:
cc: @dnovillo

dnovillo (approved):
Thanks for fixing my oversight! LGTM. I think we should have a CI that tests ENABLE_HLSL=0

PR #4262, opened 2026-05-07
Omit SPV NonUniform decoration with descriptor heaps

### LINKED ISSUE #4258: Uniformity decorations are backwards when using descriptor heaps
When GLSL_EXT_descriptor_heap is used, it generates SPIR-V with the DescriptorHeapEXT capability, which allows the SPIR-V NonUniform decoration to be dropped, and instead asks for the Uniform decoration to be applied as a guarantee to the compiler.

Device compilers already do uniformity analysis so glslang doesn't necessarily need to add the Uniform hint anywhere (though an extension to do that would be nice), but it shouldn't add the NonUniform decoration to the output.

Tobski:
This is not required by the SPIR-V spec when descriptor heaps are used. This change makes it so that if the application enables GL_EXT_descriptor_heap, non-uniform decorations in GLSL are still required, but are not output in the SPIR-V, instead requiring SPV_EXT_descriptor_heap and the DescriptorHeapEXT capability.

Uniform decorations are _not_ added when non-uniformity is detected, as this is something that should be added by users, if we ever add this ability to GLSL, rather than auto-detected (as drivers can do that themselves).

Fixes #4258


Tobski:
cc @spencer-lunarg

arcady-lunarg (approved):

PR #4263, opened 2026-05-07
DO NOT SUBMIT kokoro: linux gn flow support for no HLSL. DO NOT SUBMIT

dneto0:


dneto0:
oh, hilariously, the 'gn' build doesn't have an active kokoro configuration.

arcady-lunarg:
Having some CI coverage for the gn build would be helpful for us not breaking it.

dneto0:
> Having some CI coverage for the gn build would be helpful for us not breaking it.

Agreed.  The Github-side files exist but it also needs an inside-google configuration.  I assumed it existed but it does not. 

As you can probably tell I'm in the middle of setting up HLSL-disabled flows.  But got stuck because my local testing of the gn config is not working.  Stay tuned.

If this goes very long without a resolution, I'll set up a dumm always-pass config on the GitHub side, so that I can configure the inside-google side.

dneto0:
See #4268 instead

PR #4267, opened 2026-05-11
GL_KHR_compute_shader_derivatives - Update expected result for a test that previously (incorrectly) failed validation

janharaldfredriksen-arm:
A follow up to #4247 that gets rid of the expected SPIRV validation failure after a fix in that repo.

I updated known good to the SPIRV-Tools revision that introduce the fix. Let me know if you want to that bit differently.

(I left the known good khr / gitlab file alone since it doesn't look that's been updated for years.)


arcady-lunarg (approved):

arcady-lunarg:
Thanks for following up on this.

PR #4268, opened 2026-05-12
Make GN builds work again

dneto0:
- Create or update files as needed to make a working standalone GN build. It assumes sources for dependencies have already been downloaded, as is done by ./update_glslang_sources.py

  See DEPS for build instructions.

- Add Kokoro configurations for Linux with GCC that exercise the default build, and one that disables the HLSL frontend. The GCC is the one taken from the 'sysroot' downloaded by gclient.

Issue: #4210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ CMakeUserPresets.json
 third_party/
 buildtools/
 tools/
+.gclient_previous_sync_commits
+.gcs_entries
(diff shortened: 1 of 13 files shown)


dneto0:
cc @dnovillo  This needs a Google-side config change for the kokoro configs to take effect

arcady-lunarg (approved):

PR #4269, opened 2026-05-12
.gitignore: ignore ndk_test/obj

dneto0:
This is generated during local runs of the NDK test build


arcady-lunarg:
Looks like the links to CI results use non-global hostnames so I can't see what went wrong here

dneto0:
> Looks like the links to CI results use non-global hostnames so I can't see what went wrong here

Yeah. I fixed that with https://github.com/KhronosGroup/glslang/pull/4271  

I've rebased so it should be all clean.

This  patch was my way of checking everything was working with the new kokoro flows.

arcady-lunarg (approved):

PR #4275, opened 2026-05-15
Implement GL_NV_cooperative_matrix_decode_vector

jeffbolznv:
See https://github.com/KhronosGroup/GLSL/pull/327.


SPIRV/spirv.hpp11
```diff
@@ -1218,6 +1218,7 @@ enum class Capability : unsigned {
     CooperativeMatrixPerElementOperationsNV = 5432,
     CooperativeMatrixTensorAddressingNV = 5433,
     CooperativeMatrixBlockLoadsNV = 5434,
+    CooperativeMatrixDecodeVectorNV = 5447,
```
dnovillo: nit: move it further down so it is properly collated?

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1723,6 +1705,95 @@ void TParseContext::handleCoopMat2FunctionCall(const TSourceLoc& loc, const TFun
...
+                }
+                if (!(*fn)[1].type->isArray() || !(*fn)[2].type->isArray()) {
+                    error(loc, "coordinate parameters must be uint32_t", name, "");
+                }
+            };
+
+            const TType& matType = sequence[0]->getAsTyped()->getType();
+
+            if (funcArgs.size() > 2) {
+                error(loc, "too many function arguments (at most decodeFunc and decodeVectorFunc)",
```
dnovillo: Isn't this dead code? From what I understand here, no call with 3 or more args will have a matching overload so it will never reach this point.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -1723,6 +1705,95 @@ void TParseContext::handleCoopMat2FunctionCall(const TSourceLoc& loc, const TFun
...
+                checkBaseShape(vectorFn, vectorName);
+
+                const TType& retType = vectorFn->getType();
+                const bool retIsVector = retType.isVector() || retType.isLongVector();
+                if (!retIsVector) {
+                    error(loc, "decodeVectorFunc return type must be a vector of the matrix component type",
+                          vectorName, "");
+                } else {
+                    if (retType.getBasicType() != matType.getBasicType()) {
+                        error(loc, "decodeVectorFunc return type component must match cooperative matrix component type",
```
dnovillo, line 1778: nit: I don't think I saw a test for this error message.

dnovillo (changes requested):

SPIRV/spirv.hpp11
jeffbolznv: done

glslang/MachineIndependent/ParseHelper.cpp
jeffbolznv: changed to an assert.

glslang/MachineIndependent/ParseHelper.cpp
jeffbolznv: It's there:
```
spv.coopmat2_decode_vector_error.comp.out(4):ERROR: 0:55: 'decodeF32x2_badType(1;u1[2];u1[2];' : decodeVectorFunc return type component must match cooperative matrix component type 
```

dnovillo: Sigh. So it is. Thanks!

dnovillo:
@jeffbolznv the no-hlsl bots seem to have troubles with this PR. Could you take a look?

jeffbolznv:
> @jeffbolznv the no-hlsl bots seem to have troubles with this PR. Could you take a look?

I think it's some kind of preexisting spirv-tools failure. I've reproduced it, not sure what's going on yet.

jeffbolznv:
https://github.com/KhronosGroup/SPIRV-Tools/pull/6706 should fix the CI error.

dnovillo (approved):

PR #4276, opened 2026-05-16
Refine abortKHR code in glslang

ShchchowAMD:
Making following changes:
1. User defined abort message string won't be split automatically in glslang (confirmed from spec author)
2. According to first point, now each message string would only has one padding terminator part.
3. If a message string has format specifiers, it will still be treated as a spec constant data for developers to easily change its real data or length.


arcady-lunarg:
The CI failures seem spurious, I will need to fix them in the main branch and then you can rebase.

arcady-lunarg (approved):

PR #4278, opened 2026-05-18
glsl: implement Y-inversion for GLSL to SPIR-V

### LINKED ISSUE #2936: --invert-y does nothing for SPIRV backend
This is referenced in an older issue: https://github.com/KhronosGroup/glslang/issues/2002, but the gist of the issue is that the `--invert-y` option does not work for the SPIRV backend. I think that this should be implemented, or if that is out of the question, it should atleast be documented and/or given a warning at runtime that the option does nothing when compiling for the SPIRV backend.

shadeslayer:
setInvertY() recorded the "invert-y" process string but had no effect on GLSL shaders - the actual position transformation only existed in the HLSL parse path (HlslParseContext::assignPosition()).

Add a post-parse AST traversal pass (TInvertYTraverser) that rewrites gl_Position assignments as: temp=rhs; temp.y=-temp.y; gl_Position=temp. The pass runs in postProcess() for GLSL vertex, geometry, and tessellation-evaluation stages only, so HLSL shaders (which already invert at parse time) are unaffected.

Fixes #2936

### DIFF
--- a/glslang/MachineIndependent/Intermediate.cpp
+++ b/glslang/MachineIndependent/Intermediate.cpp
@@ -2725,11 +2725,92 @@ void TIntermBranch::updatePrecision(TPrecisionQualifier parentPrecision)
     }
 }
 
+struct TInvertYTraverser : public TIntermTraverser {
+    TInvertYTraverser(TIntermediate& inter)
+        : TIntermTraverser(true, false, false),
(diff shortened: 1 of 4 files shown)


arcady-lunarg (approved):
You will need to add the new

Issue #4165, opened 2026-01-28
Add mode to allow unbounded layout(set = N) values


HansKristian-Work:
With EXT_descriptor_heap, set decoration is completely virtual and can take any value. However, when compiling this shader

```
#version 450

layout(set = 300, binding = 400) uniform sampler2D blah;
layout(location = 0) out vec4 FragColor;

void main()
{
        FragColor = texture(blah, vec2(0.5));
}
```

```
/tmp/test.frag
ERROR: /tmp/test.frag:3: 'set' : set is too large 
ERROR: /tmp/test.frag:3: '' : compilation terminated 
ERROR: 2 compilation errors.  No code generated.


ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point
```

It would be good to have some way to bypass this limit from the CLI and API. This is causing problems in vkd3d-proton since we use rather large set values to parcel out the set space to tables vs root desc vs root constant, etc.

HansKristian-Work:
A command line flag would be necessary here. Gating behind descriptor heap extension is not correct since set/binding is supposed to be useable with older shader styles.

Guang-035:
Hello @HansKristian-Work , I have pushed [PR#4290](https://github.com/KhronosGroup/glslang/pull/4290) for fixing this issue, could you help me review?

arcady-lunarg:
Fixed by #4290

PR #4290, opened 2026-05-29
Support relaxing set and binding limits

Guang-035:
**Motivation** [ISSUE#4165](https://github.com/KhronosGroup/glslang/issues/4165):
vkd3d-proton uses large descriptor set values to partition the set space across descriptor tables, root descriptors, root constants, and related binding categories. The previous fixed-width `layout(set)` / `layout(binding)` limits caused these shaders to fail even though the larger values are intentional for descriptor heap style workflows.

**Solution**:
- Add `--relax-set-binding-limits` / `EShMsgRelaxSetBindingLimits` as an opt-in mode for larger `layout(set)` and `layout(binding)` values.
- This updates qualifier storage and unset sentinels so large values can round-trip through parsing and mapIO, while preserving the existing limits unless the flag is enabled. The option is not gated on `GL_EXT_descriptor_heap`.

**Test**:
- Added `Test/spv.relaxSetBindingLimits.frag` to cover large GLSL `layout(set)` and `layout(binding)` values with `EShMsgRelaxSetBindingLimits`.

**NOTE**:
- When this option is enabled, the maximum accepted `layout(set)` and `layout(binding)` values are relaxed to `INT_MAX`.


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7268,7 +7268,7 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
             error(loc, "needs a literal integer", "location", "");
         return;
     } else if (id == "set") {
-        if ((unsigned int)value >= TQualifier::layoutSetEnd)
+        if (!relaxSetBindingLimits() && value >= static_cast<int>(TQualifier::layoutSetEnd))
```
dnovillo, line 7271: This will now silently accept negative values. Guard against them? Also, please add a test that expects an error for negative values.

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -7280,7 +7280,7 @@ void TParseContext::setLayoutQualifier(const TSourceLoc& loc, TPublicType& publi
     } else if (id == "binding") {
         profileRequires(loc, ~EEsProfile, 420, E_GL_ARB_shading_language_420pack, "binding");
         profileRequires(loc, EEsProfile, 310, nullptr, "binding");
-        if ((unsigned int)value >= TQualifier::layoutBindingEnd)
+        if (!relaxSetBindingLimits() && value >= static_cast<int>(TQualifier::layoutBindingEnd))
```
dnovillo, line 7283: LIkewise here.

glslang/MachineIndependent/iomapper.cpp
```diff
@@ -61,6 +61,16 @@
...
+static int getLayoutSetLimit(bool relaxSetBindingLimits)
+{
+    return relaxSetBindingLimits ? INT_MAX : int(TQualifier::layoutSetEnd);
+}
+
+static int getLayoutBindingLimit(bool relaxSetBindingLimits)
+{
+    return relaxSetBindingLimits ? INT_MAX : int(TQualifier::layoutBindingEnd);
+}
+
```
dnovillo, line 73: These are used on auto-map, right? Could you add a test that exercises this code?

dnovillo (changes requested):

glslang/MachineIndependent/ParseHelper.cpp
Guang-035: Fixed

glslang/MachineIndependent/iomapper.cpp
Guang-035: Done. I added a GLSL IoMap auto-map test with sampler/texture uniforms that do not specify layout(binding), so the IoMap auto-map path assigns large bindings (70000/70001) with relaxSetBindingLimits enabled.

glslang/MachineIndependent/ParseHelper.cpp
Guang-035: Done. The shared guard at `ParseHelper.cpp:7221-7224` rejects negative layout-id values, and I added a test covering both `set = -1` and `binding = -1` under relaxSetBindingLimits.

arcady-lunarg (approved):

Issue #4081, opened 2025-11-03
GLSL discard behavior


Valentin-Sarthou:
Hello,

Since 93231001597dad1149a5d035af30eda50b9e6b6c (merged in [15.4.0](https://github.com/KhronosGroup/glslang/tree/15.4.0)), GLSL's **discard** now maps to **OpDemoteToHelperInvocation** instead of **OpTerminateInvocation** when targeting SPIR-V >= 1.6.

This change has already been discussed in #3954, but I would like to add a few more elements.

From the [GLSL 4.60 specification](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf), section 6.4. Jumps, about the **discard** keyword:
> Control flow exits the shader, and subsequent implicit or explicit derivatives are undefined when this control flow is non-uniform

This sentence seems to describe the behavior of **OpTerminateInvocation**, as opposed to **OpDemoteToHelperInvocation**.

Also, from the specification of [VK_KHR_shader_terminate_invocation](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VK_KHR_shader_terminate_invocation.html):
> OpTerminateInvocation provides the behavior required by the GLSL discard statement, and should be used when available by GLSL compilers and applications that need the GLSL discard behavior.

So from what I understand, mapping **discard** to **OpTerminateInvocation** was indeed the expected behavior, at least according to the specification.

I understand that the ambiguity of discard may cause bugs in applications, but isn't that why [GLSL_EXT_demote_to_helper_invocation](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_demote_to_helper_invocation.txt) and [GL_EXT_terminate_invocation](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_terminate_invocation.txt) were introduced ? The latter states:
> It is recommended that programmers use demote [...] or terminateInvocation instead of discard.

Is it reasonable for glslang to deviate from the GLSL, VK and EXTs specifications on this point? Am I missing something?

PR #4291, opened 2026-05-31
Add --discard-is-terminate option (#4081)

V-Sarthou:
Adds an option to map discard to OpTerminateInvocation instead of OpDemoteToHelperInvocation.
(see issue #4081 for context)

Only affects GLSL input and SPIR-V >= 1.6.
The current behavior (emitting OpDemoteToHelperInvocation) is not changed, and remains the default.

I am not 100% convinced about the naming of the option and new public API method, so suggestions are welcome.

Also, was I supposed to add any documentation in the public header for the new API? (Similar methods do not seem to have any doc)

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -5283,8 +5283,12 @@ bool TGlslangToSpvTraverser::visitBranch(glslang::TVisit /* visit */, glslang::T
     switch (node->getFlowOp()) {
     case glslang::EOpKill:
         if (glslangIntermediate->getSpv().spv >= glslang::EShTargetSpv_1_6) {
-            builder.addCapability(spv::Capability::DemoteToHelperInvocation);
-            builder.createNoResultOp(spv::Op::OpDemoteToHelperInvocationEXT);
+            if (glslangIntermediate->getDiscardIsTerminate() && glslangIntermediate->getSource() == glslang::EShSourceGlsl) {
+                builder.makeStatementTerminator(spv::Op::OpTerminateInvocation, "post-terminate-invocation");
+            } else {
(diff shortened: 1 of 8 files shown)


glslang/MachineIndependent/localintermediate.h
```diff
@@ -1247,6 +1256,7 @@ class TIntermediate {
     bool useStorageBuffer;
     bool invariantAll;
     bool nanMinMaxClamp;            // true if desiring min/max/clamp to favor non-NaN over NaN
+    bool discardIsTerminate; // true if discard should be emitted as OpTerminateInvocation instead of OpDemoteAsHelperInvocation
```
dnovillo: it's `OpDemoteToHelperInvocation` (instead of `As`).

SPIRV/GlslangToSpv.cpp
```diff
@@ -5274,8 +5274,12 @@ bool TGlslangToSpvTraverser::visitBranch(glslang::TVisit /* visit */, glslang::T
     switch (node->getFlowOp()) {
     case glslang::EOpKill:
         if (glslangIntermediate->getSpv().spv >= glslang::EShTargetSpv_1_6) {
-            builder.addCapability(spv::Capability::DemoteToHelperInvocation);
-            builder.createNoResultOp(spv::Op::OpDemoteToHelperInvocationEXT);
+            if (!glslangIntermediate->getDiscardIsTerminate() || glslangIntermediate->getSource() == glslang::EShSourceHlsl) {
```
dnovillo: This is changing HLSL semantics. Regardless of the flag value, it will choose demote-to-helper. I'd rather keep HLSL semantics unaltered.

dnovillo (changes requested):

arcady-lunarg (approved):

Issue #4281, opened 2026-05-20
Use of uninitialized stack memory in GLSL swizzle parser


Carmel0:
I found a use of uninitialized stack memory in the GLSL swizzle parser. This is not related to the deprecated HLSL front-end.

The issue is in `TParseContextBase::parseSwizzleSelector()` in `glslang/MachineIndependent/ParseContextBase.cpp`.

`fieldSet` is a local stack array:

```c++
enum {
    exyzw,
    ergba,
    estpq,
} fieldSet[MaxSwizzleSelectors];
```

Legal swizzle characters initialize `fieldSet[i]`, but the `default` case for an invalid character only reports an error:

```c++
default:
    error(loc, "unknown swizzle selection", compString.c_str(), "");
    break;
```

The later consistency check reads `fieldSet[i]`:

```c++
if (i > 0 && fieldSet[i] != fieldSet[i-1]) {
    error(loc, "vector swizzle selectors not from the same set", compString.c_str(), "");
    selector.resize(i);
    break;
}
```

For example, with `in_v.xmy`, `x` initializes `fieldSet[0]`, `m` takes the `default` path and leaves `fieldSet[1]` uninitialized, and `y` pushes the second valid selector entry.

The second loop is indexed by `selector.size()`, so it reaches `i == 1` and reads the uninitialized `fieldSet[1]`.

Affected versions:

- first affected commit found: `c142c889` from 2017-01-13;
- first affected public tag found: `vulkan-1.1-rc1` from 2017-10-04;
- first affected stable release tag found: `5.0` from 2018-03-06;
- still affected: `16.3.0` / `vulkan-sdk-1.4.350.0` from 2026-05-01 and current source `605c7f6` from 2026-04-23;
- older public tags `2.1` through `3.0` do not appear affected by this specific uninitialized read.

Memcheck reports:

```text
Conditional jump or move depends on uninitialised value(s)
    at glslang::TParseContextBase::parseSwizzleSelector(...) (ParseContextBase.cpp:612)
Uninitialised value was created by a stack allocation
    at glslang::TParseContextBase::parseSwizzleSelector(...) (ParseContextBase.cpp:530)
```

The test shader is:

```glsl
#version 450
layout(location = 0) in vec4 in_v;
layout(location = 0) out vec4 out_color;

void main() {
  float a = length(in_v.xmy);
  out_color = vec4(a, 0.0, 0.0, 1.0);
}
```

[...]

PR #4296, opened 2026-06-09
Don't use uninitialized stack values when parsing invalid swizzles.

arcady-lunarg:
Related to issue #4281, this at least removes the use of uninitialized stack memory.


dneto0 (approved):
drive-by: This looks like a good fix to me. Thanks!

PR #4249, opened 2026-05-05
Support for GLSL_EXT_opacity_micromap

wooyoungqcom:
with GLSL_EXT_opacity_micromap_ray_query_mode

PR #4297, opened 2026-06-09
SPV: emit opacity-micromap capability for ray-query use

lordalcol:
GL_EXT_opacity_micromap declares gl_RayFlagsForceOpacityMicromap2StateEXT only "in all ray tracing shader stages", but its usage note says the flag is valid for the rayflags argument of traceRayEXT() or rayQueryInitializeEXT(), and ray queries run in any stage (compute, vertex, ...). So the constant's declared scope omits ray-query use.

The flag is the SPIR-V RayFlags bit ForceOpacityMicromap2StateKHR (0x400), which requires the RayTracingOpacityMicromapKHR capability. glslang declares that capability (and SPV_EXT_opacity_micromap) only for the ray-tracing- pipeline stages (the EShLangRayGen..Callable block in GlslangToSpv.cpp). A ray-query shader in any other stage emits the flag without the capability, producing SPIR-V that fails validation once the spirv-val check from KhronosGroup/SPIRV-Tools#6670 is present.

Also emit RayTracingOpacityMicromapEXT + SPV_EXT_opacity_micromap on the ray-query path when GL_EXT_opacity_micromap is requested. Adds a compute ray-query regression test.


arcady-lunarg:
Looking at the SPIR-V specification, this capability is only needed when the flag in question is actually used, I think it would be better to only enable it if the `ForceOpacityMicromap2StateEXT` flag is actually used. Also, it looks like #4249 also tries to implement this same change (in the "only if the flag is actually used" manner) but looks like it fails to actually carry it through all the way. I think for ease of merging I would prefer to merge that PR first, then rebase this one on top of it to either finish the work or else at least add the new test case.

lordalcol:
> Looking at the SPIR-V specification, this capability is only needed when the flag in question is actually used, I think it would be better to only enable it if the `ForceOpacityMicromap2StateEXT` flag is actually used. Also, it looks like #4249 also tries to implement this same change (in the "only if the flag is actually used" manner) but looks like it fails to actually carry it through all the way. I think for ease of merging I would prefer to merge that PR first, then rebase this one on top of it to either finish the work or else at least add the new test case.

That makes sense, but I'm a bit worried that I don't know when https://github.com/KhronosGroup/glslang/pull/4249 will be merged, and we cannot update the glslang external sources of the CTS until this problem is resolved.

lordalcol:
Thanks for the review.

Reworked the patch standalone rather than waiting on #4249, which is on a different axis (it adds the new `GLSL_EXT_opacity_micromap_ray_query_mode` execution mode plus enums in spirv.hpp11, while this PR is just about the existing `gl_RayFlagsForceOpacityMicromap2StateEXT` flag and the SPIR-V capability it implies). If #4249 lands first I will reconcile.

PS2 moves capability emission from the per-stage and per-call-site code to `visitSymbol`, triggered by reference to `gl_RayFlagsForceOpacityMicromap2StateEXT`. The two prior unconditional sites (the rt-pipeline stage block in `EShLangRayGen..Callable` and the ray-query site added in PS1) are removed. Net effect: the capability is emitted only when the flag is actually referenced, uniformly across rt-pipeline and ray-query, in any stage. The only two existing tests that `#extension` the extension (`spv.ext.RayGenShader.rgen`, `rayQuery-opacityMicromap.comp`) both reference the symbol, so their expected outputs are unchanged.

arcady-lunarg (approved):

PR #4300, opened 2026-06-12
[fuchsia] Use resolved git HEAD path for build info

aidanwolter3:
Update BUILD.gn to redirect glslang_build_info to the default toolchain
when building in the Fuchsia tree, and use the resolved git HEAD path
from git_revisions.gni. This ensures the action reruns when the HEAD
changes, even in a git worktree.


dneto0 (approved):
LGTM.  Thanks for fixing this for me

PR #4301, opened 2026-06-13
fix use after free on vector resize

maxbachmann:
This fixes a potential use after free when reading the arguments[i-1] while the arguments vector is resized on a push back.


arcady-lunarg:
I'm pretty sure the C++ spec specifies that the arguments of a function are evaluated before the function call is made, so the `arguments[i - 1]` will be evaluated and the result produced before `push_back` is called and this shouldn't cause any issues. Do you have any specific evidence that there is a real problem here?

maxbachmann:
The use after free occurs because `arguments[i-1]` returns a reference. So the use after free is essentially
```python
spv::Id& id = arguments[i-1];
arguments.resize(arguments.size() + 1);
// use after free reading the id reference
arguments[arguments.size() - 1] = id;
```

arcady-lunarg:
Ah, I think you're right. In that case, is there any more systematic way we can find these cases, or is this likely to be the only case of this pattern in the codebase? Is there some compiler warning we can enable or a sanitizer we can run or something?

arcady-lunarg (approved):

maxbachmann:
I am not aware of any tooling for this purpose. Tooling would also need to detect cases in which the vector can't trigger a resize because we purposefully reserved the size.
I think pushing back elements into the same vector again is no terribly common operation.

PR #4302, opened 2026-06-15
kokoro: update license-check.cfg to allow REUSE license updates

dneto0:
See changes in PR #4256

This updates the license check configuration file to ignore the license check for the license file metadata itself. Also exclude local temporary files in .jj and .cipd, which helps me with my local testing.

### DIFF
--- a/license-checker.cfg
+++ b/license-checker.cfg
@@ -15,6 +15,7 @@
 
                     "_config.yml",
                     ".*",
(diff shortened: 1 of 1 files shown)


dneto0:
PTAL @arcady-lunarg 

This is safe to land now and will unblock #4256

arcady-lunarg (approved):

PR #4306, opened 2026-06-18
fix out-of-bounds read on empty block in layoutObjectCheck

saddamr3e:
ASan, Vulkan/SPIR-V target, tessellation-control stage:

    ParseHelper.cpp:7770: container-overflow ... layoutObjectCheck
    READ of size 8 ... (*type.getStruct())[0]

Redeclaring a built-in block with an instance name and members that don't match
the originals erases every original member (the instance-name erase in
redeclareBuiltinBlock), so the block is left with an empty member list.
layoutObjectCheck then reads member [0] to inspect its location qualifier, past
the end of an empty list. Skip the member check when the block has no members.

Repro:

    #version 450
    in gl_PerVertex {
        float bad;
    } gl_in[];
    void main() {}

compiled for a tesc stage with a Vulkan target.


saddamr3e:
gentle ping

arcady-lunarg (approved):

PR #4307, opened 2026-06-18
Relax compute shader derivative error checking

janharaldfredriksen-arm:
After landing the support for GL_KHR_compute_shader_derivatives, a couple of people have contacted me about regressions. 

Prior to my changes, texture operations with implicit derivatives worked in compute shaders used to work - and got automatically converted to their explicit LoD equivalents with LoD = 0.

It is ambiguous whether this is legal or undefined behavior in GLSL / ESSL (I'll get clarification on that separately). But in any case, there's a chance there are shaders in the wild that rely on this - and I'd rather not break them.

The key change here is the logic in ParseHelper.cpp. The rest is adding test coverage.


arcady-lunarg (approved):

ShabbyX (approved):

nvpbrown:
Somebody pointed me here in a bug report related to this issue.  Regarding this topic:

> It is ambiguous whether this is legal or undefined behavior in GLSL / ESSL (I'll get clarification on that separately). But in any case, there's a chance there are shaders in the wild that rely on this - and I'd rather not break them.

I didn't see any update on the "get clarification" issue here.  Not being currently involved in Khronos-related work, I have no idea if it's been covered elsewhere.  In any event, the following language indicating that the old (and now restored) behavior was correct can be found in the GLSL 4.60 and ESSL 3.20 specifications[:

> 8.9. Texture Functions
> Texture lookup functions are available in all shading stages. However, level-of-detail is implicitly
> computed only for fragment shaders. Other shaders operate as though the base level-of-detail were
> computed as zero.

Thanks,
Pat

janharaldfredriksen-arm:
Thanks Pat - it' been a while :)

The ambiguity I see is that later on in the same Texture Function chapter, we say that:

> Some texture functions (non-“Lod” and non-“Grad” versions) may require implicit derivatives.
Implicit derivatives are undefined within non-uniform control flow and for non-fragment shader
texture fetches.

Filed an internal issue to resolve: https://gitlab.khronos.org/GLSL/GLSL/-/issues/92

PR #4316, opened 2026-06-26
cmake: Install glslang.pc

### LINKED ISSUE #1715: Add pkgconfig file in the sources

arcady-lunarg:
Installs glslang.pc on all platforms except for Windows.

See Test/find_package/CMakeLists.txt for full explanation.

closes #1715

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,5 +454,5 @@ if (GLSLANG_ENABLE_INSTALL)
         "${CMAKE_CURRENT_BINARY_DIR}/glslang.pc"
         @ONLY
     )
-    install(CODE "configure_file(\"${CMAKE_CURRENT_BINARY_DIR}/glslang.pc\" \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/pkgconfig/glslang.pc\"  @ONLY) ")
+    install(CODE "cmake_path(ABSOLUTE_PATH CMAKE_INSTALL_PREFIX NORMALIZE OUTPUT_VARIABLE ABS_INSTALL_PREFIX)\nconfigure_file(\"${CMAKE_CURRENT_BINARY_DIR}/glslang.pc\" \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/pkgconfig/glslang.pc\"  @ONLY) ")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,6 +396,9 @@ if (GLSLANG_ENABLE_INSTALL)
         @INSTALL_CONFIG_UNIX@
         include("@PACKAGE_PATH_EXPORT_TARGETS@")
     ]=])
+    if (ENABLE_OPT)
+        set(GLSLANG_PRIVATE_LIBS "-lSPIRV-Tools-opt")
+    endif()
 
     set(PATH_EXPORT_TARGETS "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/glslang-targets.cmake")
     if(UNIX OR "${CMAKE_SYSTEM_NAME}" STREQUAL "Fuchsia")
@@ -403,6 +406,7 @@ if (GLSLANG_ENABLE_INSTALL)
             set(THREADS_PREFER_PTHREAD_FLAG ON)
             find_dependency(Threads)
         ]=])
+        set(GLSLANG_PRIVATE_LIBS "${GLSLANG_PRIVATE_LIBS} -lpthread")
     endif()
     configure_package_config_file(
         "${CMAKE_CURRENT_BINARY_DIR}/glslang-config.cmake.in"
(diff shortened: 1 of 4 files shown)


jpr42 (changes requested):
I'm just casting my opinion. You should just adopt CPS instead:
https://www.kitware.com/common-package-specification-is-out-the-gate/

Supporting *.pc files with CMake is never going to be well supported and you are just going to get more issues.

CMakeLists.txt
```diff
@@ -430,4 +434,25 @@ if (GLSLANG_ENABLE_INSTALL)
...
+    # for most of the variables, and the second at install time once the actual
+    # CMAKE_INSTALL_PREFIX is known. To protect the latter we need something that expands
+    # a literal @ sign.
+    set(AT "@")
+    configure_file(
+        "${CMAKE_CURRENT_SOURCE_DIR}/glslang/glslang.pc.in"
+        "${CMAKE_CURRENT_BINARY_DIR}/glslang.pc"
+        @ONLY
+    )
+    install(CODE "cmake_path(ABSOLUTE_PATH CMAKE_INSTALL_PREFIX NORMALIZE OUTPUT_VARIABLE ABS_INSTALL_PREFIX)\nconfigure_file(\"${CMAKE_CURRENT_BINARY_DIR}/glslang.pc\" \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/pkgconfig/glslang.pc\"  @ONLY) ")
```
jpr42, line 457: Doesn't properly handle `DESTDIR`: https://cmake.org/cmake/help/latest/envvar/DESTDIR.html

This will cause problems for package managers in some scenarios.

glslang already has some code that handles the trickiness with `DESTDIR` you can use for reference:

```
if(GLSLANG_ENABLE_INSTALL)
    install(TARGETS glslang-standalone EXPORT glslang-targets)

    install(CODE "\
        message(STATUS \"Installing (${link_method}): \$<TARGET_FILE_NAME:glslang-standalone> -> \$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}\")
        execute_process(
            COMMAND \"\${CMAKE_COMMAND}\" -E ${link_method} [=[\$<TARGET_FILE_NAME:glslang-standalone>]=] [=[${legacy_glslang_name}]=]
            WORKING_DIRECTORY \"\$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}\"
        )
    ")

endif()
```

Test/find_package/CMakeLists.txt
```diff
@@ -75,3 +75,46 @@ if (target_type STREQUAL STATIC_LIBRARY)
...
+    if (NOT (CMAKE_PKG_CONFIG_VERSION VERSION_EQUAL glslang_VERSION))
+        message(FATAL_ERROR "${CMAKE_PKG_CONFIG_VERSION} != ${glslang_VERSION}")
+    endif()
+endif()
+
+# Additional sanity check the glslang.pc is correct.
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(glslang REQUIRED IMPORTED_TARGET glslang)
+
+add_library(stub_pc STATIC stub.cpp)
```
jpr42, line 117: A static library build never triggers a real link step.

IE this integration test isn't good enough.

So: swap add_library(... STATIC ...) for add_executable(...), give it a source that actually uses glslang (not the empty stub.cpp), and link it only against PkgConfig::glslang — not shared library, executable.

Same goes for the other stubs.

dneto0:
The license check failure is:
```
* glslang/glslang.pc.in has no license
```
The checker doesn't recognize the SPDX annotation.

Please add that path to the exclusion list at https://github.com/KhronosGroup/glslang/blob/main/license-checker.cfg#L31 or thereabouts.

jpr42 (approved):

arcady-lunarg:
@kaniini as the pkg-config expert, could you take a look?

Test/find_package/CMakeLists.txt
```diff
@@ -75,3 +75,51 @@ if (target_type STREQUAL STATIC_LIBRARY)
...
+pkg_check_modules(glslang REQUIRED IMPORTED_TARGET glslang)
+pkg_check_modules(SPIRV-Tools REQUIRED IMPORTED_TARGET SPIRV-Tools)
+
+add_executable(example_pc example.c)
+set_target_properties(example_pc PROPERTIES LINKER_LANGUAGE CXX)
+
+target_link_libraries(example_pc PRIVATE
+    PkgConfig::SPIRV-Tools
+    PkgConfig::glslang
+    glslang::glslang-default-resource-limits
```
dnovillo, line 124: I think this breaks static-library consumers of `glslang.pc`, because `example_pc` never queries `Libs.private`. 

Add a `pkg_check_modules(glslang REQUIRED IMPORTED_TARGET glslang)` earlier?

dnovillo (changes requested):

PR #4321, opened 2026-06-30
Implement GL_EXT_spirv_intrinsics_variadic and GL_EXT_spirv_intrinsics_string

rdb:
These implement the extensions discussed in KhronosGroup/GLSL#331 that extend `GL_EXT_spirv_intrinsics`. These are necessary to be able to represent functions like `debugPrintfEXT`.

This is one PR with separate commits per extension; let me know if I should split this into two PRs instead, but I figured this might be easier since they both touch the parser.

### Example

```glsl
#version 460

#extension GL_EXT_spirv_intrinsics : enable
#extension GL_EXT_spirv_intrinsics_variadic : enable
#extension GL_EXT_spirv_intrinsics_string : enable

spirv_instruction(extensions = ["SPV_KHR_non_semantic_info"],
                  set = "NonSemantic.DebugPrintf", id = 1)
void debugPrintf(spirv_string format, ...);

void main()
{
    debugPrintf("x = %f\n", 1.0);
}
```

### Variadics support

Implements `GL_EXT_spirv_intrinsics_variadic` (KhronosGroup/GLSL#332). This builds upon the existing variadic parsing that was used for builtins, notably:
- `...` now lexes as a single ELLIPSIS token (so `. . .` is no longer accepted; doesn't affect existing uses)
- Zero fixed parameters, like `f(...)` are now allowed
- There's a overload tie-break between equivalent candidates if one is variadic and the other isn't, see issue 5
- Implemented for the 120-330 case as well, which defers to selectFunction (probably worth a separate refactor so that all cases here use selectFunction, but that's out of scope)
- Accepts `spirv_literal` and `spirv_by_reference` qualifiers on the variadic tail

### String support

Implements `GL_EXT_spirv_intrinsics_string` (KhronosGroup/GLSL#333). Adds a new `spirv_string` type that is only valid in a parameter type of a `spirv_instruction` function declaration, which is mapped to EbtString.

It should be noted that glslang currently parses strings as a primary-expression, which is laxer than the specified grammar prescribes (which just allows it in a function argument). This basically allows parenthesized strings and the comma operator in relaxation of the spec. This is a pre-existing "issue" that my implementation inherits, because it also applies to `GL_EXT_debug_printf`, `GL_EXT_abort`, etc., none of which suggest that strings can be expressions or that such expressions are accepted by these calls.

Embedded NUL characters now cause an error across the board, not scoped to this extension; the previous behavior was silent truncation.


Tobski:
I should've read more emails - missed that you already implemented, and with tests! @arcady-lunarg I think is the one ultimately signing off here, but IMO this looks like a faithful implementation of the two new extensions. May need changes based on spec discussion, but so far so good IMO.

arcady-lunarg:
@rdb Could you rebase to resolve the conflicts?

Test/baseResults/spv.intrinsicsSpirvInstructionVariadic330.vert.out
```diff
@@ -0,0 +1,91 @@
...
+           35(i):     34(ptr) Variable Function
+           37(f):     26(ptr) Variable Function
+              12:    9(fvec4) Load 11(vec4In)
+              13:    6(float) CompositeExtract 12 2
+                              Store 8(floatOut) 13
+                              Store 16(base) 17
+                              Store 18(arg) 19
+              20:     14(int) Load 16(base)
+              21:     14(int) Load 18(arg)
+              23:           2 ExtInst 22(Fictional.TestExtInstSet) 1(Round) 20 21
```
arcady-lunarg, line 59: It's interesting that the disassembler is printing out opcode names corresponding to the `GLSL.std.450`, looks like the built-in disassembler can't handle unknown extended instruction sets. That's a bug that you don't necessarily have to fix though.

rdb:
I am swamped at the moment, I will try to get to it soon.

PR #4322, opened 2026-06-30
Generate error when trying to use literal string in constructor

### LINKED ISSUE #4293: `uint(literal)` compiles to random integer constant
This shader code:
```glsl
#version 460

#extension GL_EXT_spirv_intrinsics : enable

spirv_instruction(set = "NonSemantic.Test", id = 1)
void _test(uint expr);

void main() {
  _test(uint("test"));
}
```

Produces the following, except with OpConstant randomized every time:
```
$ glslangValidator -H test.vert 
test.vert
// Module Version 10000
// Id's are bound by 10

                              Capability Shader
               8:             ExtInstImport  "NonSemantic.Test"
                              MemoryModel Logical GLSL450
                              EntryPoint Vertex 4  "main"
                              Source GLSL 460
                              SourceExtension  "GL_EXT_spirv_intrinsics"
                              Name 4  "main"
               2:             TypeVoid
               3:             TypeFunction 2
               6:             TypeInt 32 0
               7:      6(int) Constant 1039692592
               5:             Label
               9:           2 ExtInst 8(NonSemantic.Test) 1(Round) 7
                              Return
                              FunctionEnd
```

I expected either an error message (if this is unsupported on purpose) or a meaningful OpString identifier, not a random constant.

rdb:
Trying to use a literal string in a constructor (eg. `uint("blah")`) would not cause an error, but cause undefined behaviour / random constants to be generated. This displays an error message instead.

The fix should be uncontroversial; as far as I'm aware, there are no extensions permitting strings to be used as expressions, much less as constructor arguments.

This was the only path that suffered from this issue that I identified. Assignments, invalid use in expressions, etc. of literal strings already resulted in an error.

Fixes #4293

### DIFF
--- a/Test/baseResults/spv.intrinsicsStringConstructorError.vert.out
+++ b/Test/baseResults/spv.intrinsicsStringConstructorError.vert.out
@@ -1,0 +1,12 @@
+spv.intrinsicsStringConstructorError.vert
+ERROR: 0:8: 'constructor' : cannot convert a string 
+ERROR: 0:8: '=' :  cannot convert from ' const float' to ' temp highp uint'
+ERROR: 0:9: 'constructor' : cannot convert a string 
+ERROR: 0:9: '=' :  cannot convert from ' const float' to ' temp highp int'
+ERROR: 0:10: 'constructor' : cannot convert a string 
+ERROR: 0:11: 'constructor' : cannot convert a string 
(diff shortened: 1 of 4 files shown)


arcady-lunarg (approved):
Thanks for the fix.

PR #4324, opened 2026-07-01
fix out-of-bounds localSize write on over-long numthreads

saddamr3e:
The HLSL numthreads attribute handler in handleEntryPointAttributes walks the attribute's argument list and calls setLocalSize(lid, ...) once per argument, with lid running up to the argument count. localSize is a three element array for the x, y and z workgroup dimensions, and setLocalSize does not range check its dim argument. The attribute grammar puts no limit on how many comma separated expressions a shader can pass, so a compute entry point declared with [numthreads(1,1,1,1)] or more begins writing at localSize[3] and keeps going. With a long enough argument list it runs off the end of the TIntermediate allocation entirely; asan reports a heap-buffer-overflow read and write from setLocalSize while parsing such a shader. numthreads only ever has three dimensions, so I clamp the loop to three and let any extra arguments fall away. Valid three argument shaders produce identical SPIR-V and the HLSL test suite is unchanged.

### DIFF
--- a/Test/baseResults/hlsl.numthreads.negative.comp.out
+++ b/Test/baseResults/hlsl.numthreads.negative.comp.out
@@ -1,0 +1,44 @@
+hlsl.numthreads.negative.comp
+ERROR: 0:3: 'numthreads' : expected at most three arguments 
+ERROR: 1 compilation errors.  No code generated.
+
+
+Shader version: 500
+local_size = (1, 1, 1)
+ERROR: node is still EOpNull!
+0:3  Function Definition: @main(vu3; ( temp void)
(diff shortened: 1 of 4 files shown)


arcady-lunarg (changes requested):
I think it would be better to add an actual error check here rather than silently truncating, and it would probably be good to add a test for this case as well.

saddamr3e:
Makes sense. It now errors out when numthreads has more than three arguments instead of quietly dropping the extras, and I added hlsl.numthreads.negative.comp covering the [numthreads(1,1,1,1)] case.

arcady-lunarg (approved):

PR #4325, opened 2026-07-02
Implement GL_EXT_ocp_microscaling_types

jeffbolznv:
https://github.com/KhronosGroup/GLSL/pull/334


glslang/MachineIndependent/Versions.cpp
```diff
@@ -1379,6 +1389,56 @@ void TParseVersions::floate4m3ScalarVectorCheck(const TSourceLoc& loc, const cha
     }
 }
 
+void TParseVersions::floate2m1ScalarVectorCheck(const TSourceLoc& loc, const char* op, bool builtIn)
+{
+    if (!builtIn) {
+        const char* const extensions[] = {
+                                           E_GL_EXT_float_e2m1,
+                                         };
+        requireExtensions(loc, sizeof(extensions)/sizeof(extensions[0]), extensions, op);
```
arcady-lunarg: You can use std::size for this

gtests/SpvDebugInfoTest.cpp
```diff
@@ -317,6 +317,44 @@ void main() {
...
+        "floate2m1_t",
+        "floate3m2_t",
+        "floate2m3_t",
+        "floatue8m0_t",
+        "floatmxint8_t",
+    };
+    for (const char* typeName : expectedTypeNames) {
+        EXPECT_NE(spirv.find(std::string("\"") + typeName + "\""), std::string::npos)
+            << "Expected \"" << typeName << "\" string in debug type.\nSPIR-V:\n" << spirv;
+    }
```
arcady-lunarg, line 505: This feels like a pretty weak test but I don't think our framework is quite powerful enough to make it easy to check the thing we actually want to check here (that we get a DebugTypeBasic with the right format). Maybe it's something for a follow-up change, or else something for the new standalone SPIR-V test suite.

arcady-lunarg (changes requested):
Just one nitpick. I don't think it's worth worrying too much about the debuginfo test at this point, so that one is just more of a comment.

glslang/MachineIndependent/Versions.cpp
jeffbolznv: done.

gtests/SpvDebugInfoTest.cpp
jeffbolznv: Codex loves writing tests ;-). It made something that goes through the SPIR-V and makes sure the debuginfo instructions are correct. PTAL.

arcady-lunarg (approved):

PR #4326, opened 2026-07-03
fix oob write in memberRemapper for nested buffer_reference blocks

### LINKED ISSUE #3977: Crash/UB in Glslang::GlslangToSpv caused by OOB vector subscript
# The Problem

In my app, I am experiencing a crash when calling Glslang::GlslangToSpv, but only when compiling two particular shaders. The crash is an OOB vector subscript that's being caught by an assert in the MS STL.

Image

Where `_Pos == 8` and `_My_data._MyLast - _My_data._MyFirst == 1`.

The call stack:

Image

# Affected Versions

The issue is present in the latest release (15.3.0) as well as the latest commit (05cfcc1613c28c1274036f53616d66324f7cd383).

# Environment

Windows 10. MSVC (_MSC_VER=1944) and Clang-CL both exhibit the crash, but only if debug iterators are enabled.

When debug iterators are disabled, the crash disappears and the affected shaders *appear* to work correctly.

# Repro

The problem can be reproduced by compiling and running [this application](https://github.com/juanDiegoMontoya/Viator).

Build instructions: recursively clone the repo, make a build folder, then run `cmake ..` in it. All dependencies are downloaded with FetchContent.

The affected shaders are: [DownsampleProbeDepth.comp.glsl](https://github.com/JuanDiegoMontoya/Viator/blob/voxels/data/shaders/ddgi/DownsampleProbeDepth.comp.glsl) and [DebugProbes.vert.glsl](https://github.com/JuanDiegoMontoya/Viator/blob/voxels/data/shaders/ddgi/DebugProbes.vert.glsl).

The app will compile the affected shaders on startup and a crash should immediately follow (if debug iterators are enabled).

# Misc.

In `glslang::SpvOptions`, disabling the debug options (`generateDebugInfo=false`, `stripDebugInfo=true`) does not affect the crash.

The Glslang CLI does not seem to suffer from this issue, but probably only because I have a version of it that was compiled without debug iterators. This command should be equivalent to what my code does and does not induce a crash:
`glslang .\DownsampleProbeDepth.comp.glsl --target-env vulkan1.3 --P "#extension GL_GOOGLE_include_directive: enable" --glsl-version 460 -o test.spv -gVS`

A minidump can be provided upon request.

saddamr3e:
ASAN, compiling a push-constant `buffer_reference` block whose first member is a struct wrapping another `buffer_reference` (repro from #3977):

    ERROR: AddressSanitizer: heap-buffer-overflow
    WRITE of size 4 at 0x...
    #0 convertGlslangStructToSpvType  GlslangToSpv.cpp:6288
    #1 convertGlslangToSpvType        GlslangToSpv.cpp:5867
    freed/resized by memberRemapper[...].resize() at GlslangToSpv.cpp:5866

`memberRemapper` was keyed through `glslangTypeToIdMap`, whose `operator[]` returns id 0 for any struct member list never registered from a visited symbol. `buffer_reference` blocks reached only while building types are never registered, so distinct blocks all land on `memberRemapper[0]`. Converting the inner one-member block resizes that shared vector down to 1, then the outer two-member block's next member write runs past the end. In release builds the write just corrupts the heap silently, which is why it only shows up under a sanitizer or MSVC debug iterators.

`glslangTypeToIdMap` existed only to produce this key, so keying `memberRemapper` by the member-list pointer directly removes the collision and the map. Added `spv.bufferReferenceNestedStruct.comp` for the case; it trips the sanitizer before the fix and passes after.

Fixes #3977.


dnovillo (approved):
LGTM. Thanks for fixing this.

dnovillo:
@arcady-lunarg I think this is safe to merge. I'll defer merging this you since you're marked as the reviewer. The PR seems safe to merge. Thanks.

PR #4327, opened 2026-07-04
Implement GL_EXT_function_control_attributes

Karpov-Ruslan:
Implementation of https://github.com/KhronosGroup/GLSL/pull/335


glslang/MachineIndependent/Scan.cpp
```diff
@@ -795,8 +795,6 @@ const std::unordered_set<const char*, str_hash, str_eq> ReservedSet {
     "template",
     "this",
     "goto",
-    "inline",
-    "noinline",
```
arcady-lunarg, line 824: I think it's incorrect to remove these from the list, because it unconditionally allows people to start calling their functions or variables "inline" regardless of whether they're using your extension. Is there a way you could either make this conditional on your extension or else to make the stuff inside the `[[]]` of an attribute declaration not check against reserved words?

arcady-lunarg (changes requested):
I have some concerns about the reserved words issue, which is actually going to be an issue with the spec extension itself too.

glslang/MachineIndependent/Scan.cpp
Karpov-Ruslan: Fixed. Now if the extension is not used, then `inline` and `noinline` remain reserved words. Additionally, when using the extension, it is possible to use `inline` and `noinline` only in attributes. To do this, I had to slightly change the extension specification.

arcady-lunarg (approved):

Issue #4238, opened 2026-04-30
Various EXT_descriptor_heap bugs uncovered in SPIRV-Cross bringup


HansKristian-Work:
I ran into a myriad of bugs while doing bringup on commit a3a83d095e358aae4d63ece088dcca71828e2de0.

## atomics are broken

```
#version 460
#extension GL_EXT_descriptor_heap : require
#extension GL_EXT_nonuniform_qualifier : require
#extension GL_EXT_samplerless_texture_functions : require

layout(descriptor_heap) buffer Blah { uint a, b; } Buffers[];
layout(location = 0) out vec4 FragColor;

void main()
{
        FragColor = vec4(atomicAdd(Buffers[int(gl_FragCoord.x)].b, 1));
}
```

Notice how it's doing an atomic straight on the SSBO instead of access chaining properly into b member here:

```
              24:     10(ptr) UntypedAccessChainKHR 27 11(resource_heap) 20
              28:     23(ptr) BufferPointerEXT 24
              30:     14(int) AtomicIAdd 28 29 15 29
```

## readonly / writeonly qualifiers for images are dropped

```
#version 460
#extension GL_EXT_descriptor_heap : require
#extension GL_EXT_nonuniform_qualifier : require
#extension GL_EXT_samplerless_texture_functions : require

layout(descriptor_heap, r32f) writeonly uniform image2D Writes[];
layout(descriptor_heap, r32f) readonly uniform image2D Reads[];

void main()
{
        imageStore(Writes[10], ivec2(0), imageLoad(Reads[20], ivec2(0)));
}
```

There are no NonWritable or NonReadable decorations anywhere in the SPIR-V.

```
// Module Version 10000
// Generated by (magic number): 8000b
// Id's are bound by 27

                              Capability Shader
                              Capability UntypedPointersKHR
                              Capability DescriptorHeapEXT
                              Extension  "SPV_EXT_descriptor_heap"
                              Extension  "SPV_KHR_untyped_pointers"
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main"
                              ExecutionMode 4 OriginUpperLeft
                              Source GLSL 460
                              SourceExtension  "GL_EXT_descriptor_heap"
                              SourceExtension  "GL_EXT_nonuniform_qualifier"
                              SourceExtension  "GL_EXT_samplerless_texture_functions"
                              Name 4  "main"
[...]

PR #4329, opened 2026-07-07
Remove Block decoration from descriptor heap image wrappers

Guang-035:
**Issue**: [glslang#4238](https://github.com/KhronosGroup/glslang/issues/4238)  
**Follow-up to**: https://github.com/Guang-035/glslang/commit/90a07568d79b17a968b9a9cd4bb48334df5fc1d2

**Description**

For `layout(descriptor_heap)` image arrays, for example:

```glsl
layout(descriptor_heap, r32f) writeonly uniform image2D Writes[];
layout(descriptor_heap, r32f) readonly uniform image2D Reads[];
```

- glslang creates wrapper struct types to carry the `NonReadable` / `NonWritable` member decorations. This change removes the incorrect `Block` decoration from those wrappers because they are not shader interface blocks.
- Remove Block decorations from descriptor heap layout and payload structs

**Note**: This currently requires a SPIRV-Tools validation update. I opened [SPIRV-Tools#6723](https://github.com/KhronosGroup/SPIRV-Tools/pull/6723) for that fix.

### DIFF
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -6382,18 +6382,6 @@ spv::Id TGlslangToSpvTraverser::convertGlslangStructToSpvType(const glslang::TTy
 
     // Decorate it
     if (useDescHeapIdDecorations) {
-        // Only the top-level heap layout struct represents a resourceheap/samplerheap block.
-        // If its last member is a runtime array, Vulkan still requires the struct type to
-        // carry Block/BufferBlock; nested heap member structs should not get this decoration.
-        const bool isTopLevelHeapStruct =
-            qualifier.storage == glslang::EvqResourceHeap ||
-            qualifier.storage == glslang::EvqSamplerHeap;
-        if (isTopLevelHeapStruct && glslangIntermediate->getSpv().vulkan > 0 && !spvMembers.empty() &&
-            builder.getOpCode(spvMembers.back()) == spv::Op::OpTypeRuntimeArray) {
-            builder.addDecoration(spvType, TranslateBlockDecoration(qualifier.storage,
-                                  glslangIntermediate->usingStorageBuffer()));
-        }
-
         assert(descHeapMemberOffsets.size() == glslangMembers->size());
         for (int member = 0; member < (int)glslangMembers->size() && member < (int)spvMembers.size(); ++member) {
             glslang::TType& glslangMember = *(*glslangMembers)[member].type;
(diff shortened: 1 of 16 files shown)


Guang-035:
Hello @arcady-lunarg, based on an offline discussion with @Tobski, the 'Block' decoration should not be added to the heap struct layout or to any SSBO/UBO loaded from the heap.

This PR needs more time to be polished, so please do not review it yet.

Guang-035:
Hi @Tobski, Descriptor heap layout and payload structs do not require a Block decoration, but they can still contain runtime arrays, which currently triggers  [VUID-StandaloneSpirv-OpTypeRuntimeArray-04680](https://gitlab.khronos.org/vulkan/vulkan/-/blob/main/appendices/spirvenv.adoc?ref_type=heads&plain=1#L558 ).

Do you think this validation should be relaxed for this case, or could someone help clarify the intended behavior?

Tobski:
> Hi @Tobski, Descriptor heap layout and payload structs do not require a Block decoration, but they can still contain runtime arrays, which currently triggers [VUID-StandaloneSpirv-OpTypeRuntimeArray-04680](https://gitlab.khronos.org/vulkan/vulkan/-/blob/main/appendices/spirvenv.adoc?ref_type=heads&plain=1#L558).
> 
> Do you think this validation should be relaxed for this case, or could someone help clarify the intended behavior?

Ooh, I missed this VU, but yes this should be modified. I'll raise an internal MR to fix this, but please do feel free to go ahead and update spirv-val in preparation for that.

Guang-035:
> > Hi @Tobski, Descriptor heap layout and payload structs do not require a Block decoration, but they can still contain runtime arrays, which currently triggers [VUID-StandaloneSpirv-OpTypeRuntimeArray-04680](https://gitlab.khronos.org/vulkan/vulkan/-/blob/main/appendices/spirvenv.adoc?ref_type=heads&plain=1#L558).
> > Do you think this validation should be relaxed for this case, or could someone help clarify the intended behavior?
> 
> Ooh, I missed this VU, but yes this should be modified. I'll raise an internal MR to fix this, but please do feel free to go ahead and update spirv-val in preparation for that.

I’d like to fix this too, but I’m not sure how we should relax this check. Figuring out whether a struct is used through the descriptor heap path seems a bit complicated. Do you have any suggestions?

Tobski:
> I’d like to fix this too, but I’m not sure how we should relax this check. Figuring out whether a struct is used through the descriptor heap path seems a bit complicated. Do you have any suggestions?

I've put what I _think_ has to be the correct fix in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8263, which is to say the struct must be the direct result of OpBufferPointer. Tracking that might be tricky with untyped pointers though.

Guang-035:
hello @Tobski , Thanks for your [clarification](https://github.com/KhronosGroup/SPIRV-Tools/pull/6723#discussion_r3803504653):
```
layout(descriptor_heap) uniform struct A {
    writeonly image2D a;
} myImages[];
```
Based on your feedback, I’ve refined the image qualifier wrapper implementation in the latest commit. When you have a chance, could you please take another look? I’d really appreciate it.

Tobski:
> hello @Tobski , Thanks for your [clarification](https://github.com/KhronosGroup/SPIRV-Tools/pull/6723#discussion_r3803504653):
> 
> ```
> layout(descriptor_heap) uniform struct A {
>     writeonly image2D a;
> } myImages[];
> ```
> 
> Based on your feedback, I’ve refined the image qualifier wrapper implementation in the latest commit. When you have a chance, could you please take another look? I’d really appreciate it.

Looks correct as best I can tell 👍

arcady-lunarg:
@Guang-035 What's needed to get this moving forward? A review on the spirv-tools change that it depends on?

PR #4334, opened 2026-07-11
Guard std::filesystem so InfoSink.h builds without <filesystem>

andy-emerson:
`InfoSink.h` includes `<filesystem>` and calls `std::filesystem::absolute()` in `TInfoSinkBase::location()`. This does not compile if the standard library was built without a working `<filesystem>`. One example is LLVM libc++ built with `LIBCXX_ENABLE_FILESYSTEM=OFF`, which happens on wasm32-wasi and some embedded targets where the C library has no `statvfs`. A lot of files include `InfoSink.h`, so the whole build fails, not just the diagnostics.

I wrapped the include in `#if __has_include(<filesystem>)` and wrapped the two `absolute()` calls in `#if defined(__cpp_lib_filesystem)` to check whether `<filesystem>` is usable. When it is not usable, the code appends the plain path instead of the absolute one. On platforms that have `<filesystem>`, nothing changes.

### DIFF
--- a/glslang/Include/InfoSink.h
+++ b/glslang/Include/InfoSink.h
@@ -36,7 +36,9 @@
 #define _INFOSINK_INCLUDED_
 
 #include "../Include/Common.h"
+#if __has_include(<filesystem>)
 #include <filesystem>
+#endif
 #include <cmath>
 
 namespace glslang {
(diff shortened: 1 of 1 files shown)


arcady-lunarg:
Do you think it would be feasible to add a WASI build to CI? Ideally both building and running the regression tests? Otherwise I fear that we'll keep inadvertently breaking this configuration.

arcady-lunarg (approved):

PR #4337, opened 2026-07-16
Make TVarEntryInfo part of the public API in ShaderLang.h

### LINKED ISSUE #3904: TVarEntryInfo is private which makes custom TIoMapResolver impossible
We have a program that generates SPIR-V shaders in run time using the following workflow. This used to work before, but now because TVarEntryInfo was moved internally into iomapper.cpp which makes implementing a custom TIoMapResolver impossible because we cannot access the TVarEntryInfo variable in resolveBinding and other virtual function callbacks.

Since TProgram::mapIO and TIoMapResolver are public APIs, I think this is probably not intended.

```
class MyIoMapResolver : public glslang::TIoMapResolver
{
...
};
MyIoMapResolver ioMapResolver;

program = new glslang::TProgram();
program->addShader(...);
program->addShader(...);
succeeded = program->link(EShMsgDefault);
if (succeeded)
	succeeded = program->mapIO(&ioMapResolver);
if (succeeded)
	call glslang::GlslangToSpv(*program->getIntermediate(lang), spirv) etc
```

arcady-lunarg:
Fixes #3904

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -71,83 +71,43 @@ static int getLayoutBindingLimit(bool relaxSetBindingLimits)
     return relaxSetBindingLimits ? INT_MAX : int(TQualifier::layoutBindingEnd);
 }
 
-struct TVarEntryInfo {
-    long long id;
-    TIntermSymbol* symbol;
-    bool live;
-    TLayoutPacking upgradedToPushConstantPacking; // ElpNone means it hasn't been upgraded
-    int newBinding;
-    int newSet;
-    int newLocation;
-    int newComponent;
-    int newIndex;
-    EShLanguage stage;
+bool TVarEntryInfo::TOrderByPriority::operator()(const TVarEntryInfo& l, const TVarEntryInfo& r)
+{
+    const TQualifier& lq = l.symbol->getQualifier();
+    const TQualifier& rq = r.symbol->getQualifier();
 
-    void clearNewAssignments() {
-        upgradedToPushConstantPacking = ElpNone;
-        newBinding = -1;
-        newSet = -1;
-        newLocation = -1;
-        newComponent = -1;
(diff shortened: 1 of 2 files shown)


glslang/MachineIndependent/iomapper.cpp
```diff
@@ -71,83 +71,41 @@ static int getLayoutBindingLimit(bool relaxSetBindingLimits)
...
-    bool live;
-    TLayoutPacking upgradedToPushConstantPacking; // ElpNone means it hasn't been upgraded
-    int newBinding;
-    int newSet;
-    int newLocation;
-    int newComponent;
-    int newIndex;
-    EShLanguage stage;
+bool TVarEntryInfo::TOrderByPriority::operator()(const TVarEntryInfo& l, const TVarEntryInfo& r) {
+const TQualifier& lq = l.symbol->getQualifier();
```
dnovillo: clang-format this? the first line is unindented.

glslang/Public/ShaderLang.h
```diff
@@ -783,9 +783,55 @@ class GLSLANG_EXPORT TObjectReflection {
...
+        inline bool operator()(const TVarEntryInfo& l, const TVarEntryInfo& r) { return l.id < r.id; }
+    };
+
+    struct TOrderByPriority {
+        // ordering:
+        // 1) has both binding and set
+        // 2) has binding but no set
+        // 3) has no binding but set
+        // 4) has no binding and no set
+        inline bool operator()(const TVarEntryInfo& l, const TVarEntryInfo& r);
```
dnovillo: no need to have `inline` here.

dnovillo (approved):
LGTM. Just a couple of nits that don't need re-review.

glslang/Public/ShaderLang.h
arcady-lunarg: In fact, I think there's a need not to, thanks.

PR #4341, opened 2026-07-17
preprocessor: guard elseSeen/ifdepth against negative elsetracker

ashketchumwashere:
A malformed #if expression (e.g. '#if %%%') causes eval() to return err=true. CPPif increments ifdepth and elsetracker but does not invoke CPPelse, leaving the preprocessor tracking a phantom conditional block. Subsequent excess #endif directives are consumed inside CPPelse: the depth > 0 path at Pp.cpp:328 runs --ifdepth without any guard, driving ifdepth negative. Once ifdepth < 0 the 'if (ifdepth == 0)' guard in readCPPline is permanently bypassed (only catches exactly zero). Every further #endif then executes elseSeen[elsetracker] = false with elsetracker running arbitrarily negative, writing zero-bytes backward through TPpContext into the lastLineTokenLocs vector control block, corrupting its internal pointers and causing a deterministic SEGV.

Fix with three one-line guards:
- Pp.cpp:319 (CPPelse): check elsetracker >= 0 before writing elseSeen
- Pp.cpp:328 (CPPelse depth>0 path): check ifdepth > 0 before decrement
- Pp.cpp:1011 (readCPPline): check elsetracker >= 0 before writing elseSeen

Add test preprocessor.elseseen.oob.vert that reproduces the crash path: a bad #if desynchronises the counters, excess #endifs then drive elsetracker to -10. With the fix all excess #endifs produce 'mismatched statements' errors cleanly; without it the process crashes.

### DIFF
--- a/Test/baseResults/preprocessor.elseseen.oob.vert.err
+++ b/Test/baseResults/preprocessor.elseseen.oob.vert.err
@@ -1,0 +1,15 @@
+ERROR: 0:18: 'preprocessor evaluation' : bad expression 
+ERROR: 0:18: '#if' : unexpected tokens following directive 
+ERROR: 0:31: '#endif' : mismatched statements 
+ERROR: 0:32: '#endif' : mismatched statements 
+ERROR: 0:33: '#endif' : mismatched statements 
+ERROR: 0:34: '#endif' : mismatched statements 
+ERROR: 0:35: '#endif' : mismatched statements 
+ERROR: 0:36: '#endif' : mismatched statements 
+ERROR: 0:37: '#endif' : mismatched statements 
+ERROR: 0:38: '#endif' : mismatched statements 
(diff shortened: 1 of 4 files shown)


ashketchumwashere:
Hi @arcady-lunarg @jeremy-lunarg @saddamr3e — this PR guards `elseSeen[elsetracker]` and `ifdepth` against negative `elsetracker` values in the preprocessor (the OOB write that causes heap vector corruption). No conflicts with main, CLA signed. Would appreciate a review when you get a chance. Thanks!

dneto0 (approved):
LGTM.

There's an intended invariant in the CPPelse processing that `ifdepth >= depth` but your malformed `#if` violates that. Nice find.

Additionally, defensive coding would check the indices with `<= 0` instead of `== 0`.  But this change is sufficient.

arcady-lunarg (approved):

PR #4342, opened 2026-07-18
preprocessor: guard against OOB read in handlePragma for #pragma STDGL

ashketchumwashere:
## Summary

Fixes an out-of-bounds heap read (SEGV) in `glslang::TParseContext::handlePragma()` at `ParseHelper.cpp:422` when a shader contains `#pragma STDGL` with fewer than 4 tokens.

**Confirmed crash in production (non-ASan) builds** — not a sanitizer-only artifact. The `_M_p` field of the out-of-bounds `std::string` object always resolves to a non-readable address (NULL under standard allocator; code-segment address under the pool allocator), causing `memcmp()` to SEGV without any sanitizer involvement.

Also reproducible via the OSS-Fuzz target `shaderc_general_fuzzer` (google/shaderc) with a 34-byte input.

## Root Cause

The condition at line 422 accesses `tokens[1]` and `tokens[3]` without first checking `tokens.size()`:

```cpp
// BEFORE (vulnerable):
} else if (spvVersion.spv > 0 && tokens[0].compare("STDGL") == 0 &&
           tokens[1].compare("invariant") == 0 && tokens[3].compare("all") == 0) {
```

Every other `tokens[N]` access in `handlePragma()` has an explicit size check (lines 340, 366). This branch was the only exception.

For input `#pragma STDGL` with no trailing tokens:
- `tokens` = `{"STDGL"}` → `tokens.size() == 1`
- `tokens[1]` → OOB heap read (32 bytes past vector end)
- `tokens[3]` → OOB heap read (96 bytes past vector end)

## Fix

```cpp
// AFTER (fixed):
} else if (spvVersion.spv > 0 && tokens[0].compare("STDGL") == 0 &&
           tokens.size() >= 4 &&
           tokens[1].compare("invariant") == 0 && tokens[3].compare("all") == 0) {
```

### Via glslang directly (requires `setEnvClient(EShClientVulkan)`)

```glsl
#pragma STDGL
```

### Via shaderc / OSS-Fuzz `shaderc_general_fuzzer` (34-byte input)

```glsl
#version 450
#pragma STDGL
void main(){}
```

```bash
# Reproduce via OSS-Fuzz:
git clone https://github.com/google/oss-fuzz /tmp/oss-fuzz-repro
cd /tmp/oss-fuzz-repro
python3 infra/helper.py build_image shaderc
python3 infra/helper.py build_fuzzers --sanitizer address shaderc
printf '#version 450
#pragma STDGL
void main(){}' > /tmp/crash.glsl
python3 infra/helper.py reproduce shaderc shaderc_general_fuzzer /tmp/crash.glsl
```

### ASan output

```
AddressSanitizer: SEGV on unknown address (READ)
  #0 memcmp
  #1 std::basic_string::compare(char const*)
  #2 glslang::TParseContext::handlePragma()  ParseHelper.cpp:422
  #3 glslang::TPpContext::CPPpragma()        Pp.cpp:884
  #4 glslang::TShader::preprocess()
```

### Production crash (no sanitizer)

```
Scenario: standard malloc  →  tokens[1]._M_p = NULL
→ memcmp(NULL, "invariant", 57633) → SEGV

Scenario: pool allocator   →  tokens[1]._M_p = non-readable addr
→ memcmp(<bad ptr>, "invariant", N) → SEGV
```

## Test

Added `Test/pragma.stdgl.oob.vert` covering the three short-token forms that previously triggered the crash.


ashketchumwashere:
Hi @arcady-lunarg @jeremy-lunarg @saddamr3e — this PR adds a bounds check in `handlePragma` to guard against an OOB read when `#pragma STDGL` appears without a following token. No conflicts with main, CLA signed. Would appreciate a review when you get a chance. Thanks!

dneto0 (approved):
LGTM!

arcady-lunarg (changes requested):
You're going to need to add your test to the appropriate list of tests and add an expected result file so the test is actually run.

ashketchumwashere:
Hi @arcady-lunarg — addressed your feedback:

- Renamed `Test/pragma.stdgl.oob.vert` → `Test/spv.pragma.stdgl.oob.vert` (the bug requires Vulkan mode / `spvVersion.spv > 0`, so it belongs in the SPV compile test suite rather than the preprocessor suite)
- Added `"spv.pragma.stdgl.oob.vert"` to the `CompileVulkanToSpirvTest` list in `gtests/Spv.FromFile.cpp`
- Added `Test/baseResults/spv.pragma.stdgl.oob.vert.out` — generated from a build with the fix applied; confirms clean SPIR-V output with no crash

Without the fix, the test crashes (SIGSEGV / exit 139) under Vulkan semantics. With it, it compiles to a minimal SPIR-V module cleanly. Force-pushed to update the PR.

arcady-lunarg (approved):
The parsing of this pragma syntax is really not very robust but that's a pre-existing problem that should probably be a separate fix, so I'm inclined to merge this

PR #4344, opened 2026-07-18
iomapper: guard uniformVarMap before dereferencing it in uniform validation

### LINKED ISSUE #4313: Possible usage of null pointer, due to validating `outVarMaps[i]` when using `uniformVarMap[i]`
Issue found by AI, validated by me - full report can be found here - [https://github.com/qarmin/ClaudeReports/blob/master/glslang_20260622.html](https://github.com/qarmin/ClaudeReports/blob/master/glslang_20260622.html)

https://github.com/KhronosGroup/glslang/blob/20960a4872f681e4213312b06b48cc4ddae3c73d/glslang/MachineIndependent/iomapper.cpp#L689-L692

**Description:** In TSymbolValidater's uniform-type validation loop, the guard checks outVarMaps[i] != nullptr but the body dereferences uniformVarMap[i] (uniformVarMap[i]->find(name) and uniformVarMap[i]->end()). The guard tests the wrong array. If for some stage i outVarMaps[i] is non-null while uniformVarMap[i] is null, this is a null-pointer dereference; conversely if uniformVarMap[i] is non-null but outVarMaps[i] is null, cross-stage uniform validation is silently skipped. The maps usually happen to be allocated/null together (iomapper.cpp:1722), which masks the bug, but the guard is logically incorrect.

**Locations:**

```
glslang/MachineIndependent/iomapper.cpp:689-692
  689 |             for (int i = 0; i < EShLangCount; i++) {
  690 |                 if (i != currentStage && outVarMaps[i] != nullptr) {
  691 |                     auto ent2 = uniformVarMap[i]->find(name);
  692 |                     if (ent2 != uniformVarMap[i]->end()) {
```

**Fix:** Change the guard to check the array actually used: 'if (i != currentStage && uniformVarMap[i] != nullptr)'.

SAY-5:
In `TSymbolValidater`'s uniform-type validation loop, the null guard tests `outVarMaps[i]` but the body dereferences `uniformVarMap[i]` (`uniformVarMap[i]->find(name)` and `->end()`). The two arrays are usually allocated or left null together, which masks the problem, but the guard is checking the wrong array: if a stage has `outVarMaps[i]` non-null while `uniformVarMap[i]` is null the loop dereferences a null pointer, and in the reverse case cross-stage uniform validation is silently skipped.

Check `uniformVarMap[i]` instead, matching how the in/out branches above guard `inVarMaps`/`outVarMaps` against the array they actually use.

Fixes #4313.

### DIFF
--- a/glslang/MachineIndependent/iomapper.cpp
+++ b/glslang/MachineIndependent/iomapper.cpp
@@ -647,7 +647,7 @@ struct TSymbolValidater
         } else if (base->getQualifier().isUniformOrBuffer() && !base->getQualifier().isPushConstant()) {
             // validate uniform type;
             for (int i = 0; i < EShLangCount; i++) {
-                if (i != currentStage && outVarMaps[i] != nullptr) {
+                if (i != currentStage && uniformVarMap[i] != nullptr) {
                     auto ent2 = uniformVarMap[i]->find(name);
                     if (ent2 != uniformVarMap[i]->end()) {
                         ent2->second.symbol->getType().appendMangledName(mangleName2);


arcady-lunarg (approved):
While I don't think there's necessarily any way to actually trigger this bug, there's no harm in updating the logic to be more correct.

PR #4350, opened 2026-07-23
fix oob read on empty array initializer in convertInitializerList

saddamr3e:
ASan, on a shader with a nested empty initializer list:

    SEGV on unknown address 0x000000000000
    ParseHelper.cpp:9987 in glslang::TParseContext::convertInitializerList

The array branch reads getSequence()[0] to derive inner array dimensions but never checks the initializer list is non-empty. A nested {} (an empty EOpNull aggregate, legal under GL_EXT_null_initializer) that lands on an array element type reaches it with a zero-length sequence, so the index runs past the end. The struct/matrix/vector branches already validate the count first, and the HLSL convertInitializerList guards the same read with size() > 0. Reject the empty list with an error to match.

Repro: `int a[1][3] = { {} };` with GL_EXT_null_initializer. Test added under gtests AST (glsl.nullInitializer.error.vert).

### DIFF
--- a/Test/baseResults/glsl.nullInitializer.error.vert.out
+++ b/Test/baseResults/glsl.nullInitializer.error.vert.out
@@ -1,0 +1,27 @@
+glsl.nullInitializer.error.vert
+ERROR: 0:11: 'initializer list' : array initializer must be non-empty 
+ERROR: 1 compilation errors.  No code generated.
+
+
+Shader version: 460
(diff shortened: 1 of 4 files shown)


arcady-lunarg (approved):

arcady-lunarg:
@saddamr3e I think you'll need to re-sign the CLA because they changed to a different CLA.

PR #4352, opened 2026-07-23
Increase input_attachment_index limit

rg3igalia:
The new limit is INT_MAX, as opposed to UINT_MAX, because INT_MAX is large enough and there are other places in the code where the value of the limit is cast to int.

The need to increase this comes from an upcoming Vulkan extension that will allow the indices to be much larger in practice, and I'm creating GLSL tests in CTS. I would very much like not to have to write SPIR-V assembly for this.


arcady-lunarg (approved):

arcady-lunarg:
Apparently there's supposed to be a `gl_MaxInputAttachments` constant available in GLSL but that was never implemented in glslang.

PR #4353, opened 2026-07-26
Add performance optimization and update size optimization

Karpov-Ruslan:
Add `-O` optimization flag for performance. Thus, it is no longer necessary to call spirv-opt with the `-O` flag.
Update size optimization and HLSL legalization.

Performance optimization, size optimization and HLSL legalization will now be the same as in spirv-opt. The advantage is that the sequence of passes is now completely dependent on SPIR-V Tools, so optimizations will always be up-to-date.


arcady-lunarg (commented):
From a maintenance perspective I think this would be a good idea, and it looks like the recipes in spirv-opt have been improved somewhat and do more optimization. Most of the difference between what glslang does here and what spirv-opt does is that spirv-opt does a few more optimizations, the one item that kind of stands out is the loop-unroll pass but that says "Fully unrolls loops marked with the Unroll flag" so I think it shouldn't be a problem.

arcady-lunarg:
Thinking about this more, I think the only way we're going to get feedback on it is by merging it and then if people complain we can roll it back.

PR #4357, opened 2026-07-27
GLSL_EXT_split_barrier / SPV_EXT_split_barrier

wooyoungqcom:


glslang/MachineIndependent/Initialize.cpp
```diff
@@ -4896,6 +4896,11 @@ void TBuiltIns::initialize(int version, EProfile profile, const SpvVersion& spvV
             "void barrier();"
             );
     }
+    if ((profile != EEsProfile && version >= 420) || esBarrier)
```
arcady-lunarg, line 4899: The extension spec says "This extension can be applied to OpenGL GLSL versions 4.60 and" but should probably say "4.20 and higher"

arcady-lunarg (approved):
This PR itself looks fine, but you really should make a public GLSL spec PR as well, with the change mentioned about the required GLSL version.

glslang/MachineIndependent/Initialize.cpp
wooyoungqcom: I fixed it to have 4.20.

And, created a pull request https://github.com/KhronosGroup/GLSL/pull/337

PR #4358, opened 2026-07-28
avoid undefined shifts and signed overflow in #if expression operators

saddamr3e:
UBSan, clang -fsanitize=undefined:

    Pp.cpp:417:41: runtime error: shift exponent 100 is too large for 32-bit type 'int'
    Pp.cpp:418:41: runtime error: shift exponent -1 is negative
    Pp.cpp:419:41: runtime error: signed integer overflow: 2147483647 + 1 cannot be represented in type 'int'

from `#if (1 << 100)`, `#if (1 >> -1)` and `#if (2147483647 + 1)`. The `#if` operator table already guards INT_MIN/-1 in `op_div`, `op_mod` and `op_neg`; `op_shl`, `op_shr`, `op_add`, `op_sub` and `op_mul` were missed. Shift counts are now bounded by the int width and the wrapping ops go through unsigned, so only the cases that were undefined before change behavior.

`preprocessor.shift_out_of_range.vert` in the preprocessing suite asserts the now-defined results for the new guards.


gtests/AST.FromFile.cpp
```diff
@@ -90,6 +90,7 @@ INSTANTIATE_TEST_SUITE_P(
         "cppSimple.vert",
         "cppIndent.vert",
         "cppIntMinOverNegativeOne.frag",
+        "cppShiftOutOfRange.frag",
```
arcady-lunarg: Maybe this would be a better fit for `Pp.FromFile.cpp`?

arcady-lunarg (changes requested):

gtests/AST.FromFile.cpp
saddamr3e: Done, moved it to Pp.FromFile.cpp and renamed to preprocessor.shift_out_of_range.vert to match that suite. Still fails without the Pp.cpp change (the #error output lands in the .err baseline), passes with it.

arcady-lunarg (approved):

PR #4361, opened 2026-07-30
Add support for GL_EXT_optional_input_attachment_index

rg3igalia:
Input attachment indices are optional for depth/stencil attachments with VK_KHR_dynamic_rendering_local_read, so it makes sense to allow omitting the qualifier from GLSL.


arcady-lunarg:
And this validity isn't something that can be determined from the SPIR-V side, but rather depends on how you're using the shader from Vulkan? I think it might be a good idea to also update the text in `GL_KHR_vulkan_glsl` to explicitly allow this.

rg3igalia:
Yes, that's correct as far as I understand.  An `OpTypeImage` variable with `Dim` set to `SubpassData` may or may not have an `InputAttachmentIndex` decoration. The resulting SPIR-V module can only be used from Vulkan if the `dynamicRenderingLocalRead` feature is enabled, which is a runtime check (see VUID-RuntimeSpirv-None-09558), but it's not invalid by itself.

rg3igalia:
I've updated the PR to implement support for GL_EXT_optional_input_attachment_index as proposed in https://github.com/KhronosGroup/GLSL/pull/339

arcady-lunarg (changes requested):
You also need to add the `#define GL_EXT_optional_input_attachment_index 1` to `TParseVersions::getPreamble()`, as described in the step by step instructions in the big comment at the top of glslang/MachineIndependent/Versions.cpp.

rg3igalia:
> You also need to add the `#define GL_EXT_optional_input_attachment_index 1` to `TParseVersions::getPreamble()`, as described in the step by step instructions in the big comment at the top of glslang/MachineIndependent/Versions.cpp.

This should be done now.

arcady-lunarg (approved):

rg3igalia:
Should we revert this and wait till the GLSL extension is approved?

arcady-lunarg:
I'm not sure what the approval process for GLSL extensions is exactly, but if there are any changes as a result of that you can just make another PR here.

PR #4363, opened 2026-07-31
Remove en dash which fails the build with cl.exe

daniel-story:
Depending on the system locale, this unnecessary use of an en dash causes the build to fail.


arcady-lunarg (approved):

arcady-lunarg:
I don't think this is a failure mode I've seen before, thanks for fixing it. cc @jeffbolznv who introduced the offending line.

jeffbolznv:
Claude had a habit of using non-standard characters in comments. I try to remove them before pushing but I must have missed this one.

PR #4364, opened 2026-08-01
SPIR-V Builder: Fix quadratic runtime used for creation of composite and struct constants.

mbeck-nv:
- the original code uses a simple linear search over all already created constants to figure out if the constant already exists
- replaced by a hash table approach now
- this leads to very long runtime for bigger constant arrays

### DIFF
--- a/SPIRV/SpvBuilder.h
+++ b/SPIRV/SpvBuilder.h
@@ -91,6 +91,230 @@ struct StructMemberDebugInfo {
 };
 
 class Builder {
+    class CompositeConstantKey {
+    public:
+        enum Kind {
+            Key_Instruction,
+            Key_CompositeConstant,
+        };
+
+        CompositeConstantKey(
+            Op typeClass, Instruction* instruction)
+        : kind(Key_Instruction)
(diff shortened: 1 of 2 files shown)


SPIRV/SpvBuilder.h
```diff
@@ -91,6 +91,230 @@ struct StructMemberDebugInfo {
...
+                }
+                return false;
+            }
+        };
+
+    private:
+        Kind kind;
+        Instruction* instruction;
+        Id typeId;
+        size_t numMembers;
```
arcady-lunarg, line 314: Is there any reason to keep a separate numMembers here when the vector below already knows its own size?

mbeck-nv: Yes, this is on purpose and follows the original API. The code first collects all operands and checks then for all equal. If yes, the opcode is changed and numMembers is set to 1. So, either the numNumbers is passed, OR the vector must be recreated with length one. This solution follows the original implementation. Of course it would be possible the change that.

arcady-lunarg (approved):

PR #4383, opened 2026-08-15
Fix string escaping in preprocessed output

### LINKED ISSUE #4375: `glslang_shader_preprocess` turns '\n' inside a string into real newlines
```glsl
#version 460
#extension GL_EXT_debug_printf : enable
void main() {
  debugPrintfEXT("hi\n");
  debugPrintfEXT("v={\n  %f\n}", 1.0);
}
```
after glslang_shader_preprocess, the debugprintf looks like this
```glsl
debugPrintfEXT("v={
 %f
}", 1.0);
```
thus resulting in compile errors

jvron:
Fixes #4375

When preprocessing GLSL source, escape sequences in string literals were decoded during tokenization and then emitted directly into the preprocessed output.

For example:
```glsl
debugPrintfEXT("newline: \n");
```
was emitted as:
```
debugPrintfEXT("newline: 
");
```
This produces invalid GLSL and can cause compilation errors when the preprocessed output is compiled again.

This change re-escapes characters when emitting `PpAtomConstString` tokens during preprocessing, preserving valid escape sequences in the preprocessed output.

The fix handles the standard escape sequences, including:
```
\"
\\
\a
\b
\f
\n
\r
\t
\v
```

I tested the standard escape sequences with `glslangValidator -E` and verified that the existing test suite passes

### DIFF
--- a/Test/baseResults/preprocessor.string_escaping.frag.out
+++ b/Test/baseResults/preprocessor.string_escaping.frag.out
@@ -1,0 +1,16 @@
+#version 450
+#extension GL_EXT_debug_printf : enable
+
+void main()
+{
+    debugPrintfEXT("double quote: \"");
+    debugPrintfEXT("backslash: \\");
+    debugPrintfEXT("bell: \a");
+    debugPrintfEXT("backspace: \b");
+    debugPrintfEXT("form feed: \f");
+    debugPrintfEXT("newline: \n");
+    debugPrintfEXT("carriage return: \r");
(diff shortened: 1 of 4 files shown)


arcady-lunarg (changes requested):
Please add a unit test that exercises this new functionality. You will need to add a file under `Test/` then the expected results under `Test/baseResults`, and also add the file to the list in `gtests/Pp.FromFile.cpp`.

arcady-lunarg (approved):
LGTM, thanks.

PR #4387, opened 2026-08-20
fix oob read on short arg list to struct buffer methods

saddamr3e:
UBSan/ASan on `buf.Store(0)` (missing value argument) in a fragment shader:

    hlslParseHelper.cpp:3652:70: runtime error: member call on misaligned address 0xfefefefefefefefe for type 'TIntermNode'
    AddressSanitizer: SEGV ... READ, decomposeStructBufferMethods -> handleFunctionCall

Struct/byte-address buffer methods (Load/Store/GetDimensions/Interlocked*/Append) resolve by name against a zero-parameter builtin prototype in `handleFunctionCall`, so overload resolution never checks the caller's argument count. `decomposeStructBufferMethods` then indexes the argument aggregate directly (`getSequence()[1]`, `getSequence()[2]`), so a call with too few arguments reads one past the end of the sequence and uses it as a node pointer; a bare `buf.GetDimensions()` with no arguments leaves the aggregate null and dereferences it.

Validate the count against the minimum each method actually reads before indexing, and error out otherwise.

### DIFF
--- a/Test/baseResults/hlsl.structbuffer.store.args.error.frag.out
+++ b/Test/baseResults/hlsl.structbuffer.store.args.error.frag.out
@@ -1,0 +1,64 @@
+hlsl.structbuffer.store.args.error.frag
+ERROR: 0:6: '' : too few arguments to buffer method 
+ERROR: 1 compilation errors.  No code generated.
+
+
+Shader version: 500
+gl_FragCoord origin is upper left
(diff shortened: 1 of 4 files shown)


arcady-lunarg (approved):
I don't know that this is necessarily the cleanest way to do this, but it's good enough.

PR #4388, opened 2026-08-21
Improve parsing of STDGL pragma

arcady-lunarg:
Actually check that it is of the expected form consisting of exactly the tokens `#pragma` `STDGL` `invariant` `(` `all` `)`


jeremy-lunarg (approved):
LGTM

PR #4397, opened 2026-09-01
Add ArrayStride decoration on shader abort message strings

### LINKED ISSUE #4369: OpAbortKHR must use an explicit layout
It seems the implementation of OpAbortKHR doesn't use an explicit layout and it then fails to be parsed correctly in Mesa. VVL support is also missing but I reported it here https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12298.

This has been found with the KHR_shader_abort VKCTS tests here https://gerrit.khronos.org/c/vk-gl-cts/+/16088.

I think it should emit the ArrayStride decoration to match what the SPIR-V spec is saying: "Message Type is the type of the data to be written. Message Type must be a concrete type. If the Shader capability is declared, Message Type must have an explicit layout."

arcady-lunarg:
Fixes #4369

### DIFF
--- a/known_good.json
+++ b/known_good.json
@@ -5,7 +5,7 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit": "f589ef005c49f6f19c8e78eb5269104ba293beb4"
+      "commit": "ecea80f176e17a7afbea8c626b5adc6b5a4cbefb"
     },
     {
       "name" : "spirv-tools/external/spirv-headers",
--- a/SPIRV/GlslangToSpv.cpp
+++ b/SPIRV/GlslangToSpv.cpp
@@ -3429,6 +3429,7 @@ void TGlslangToSpvTraverser::createAbortEXT(const glslang::TIntermSequence &glsl
     // 2.4 Add decoration for this string.
     builder.addDecoration(msgArrType, spv::Decoration::UTFEncodedKHR);
     builder.addDecoration(msgLoadArrType, spv::Decoration::UTFEncodedKHR);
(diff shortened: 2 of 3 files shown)


spencer-lunarg (approved):
(we should 100% get this in ASAP for CTS)

cc @r-potter I know this was the "easy" case, you said you were also playing around with various other inputs?

I am also curious if glslang will just break/fail when trying to deal with struct-of-structs, array-of-structs, or matrixs

r-potter:
@arcady-lunarg @spencer-lunarg You might want to look at the PR I just put up, which is much more comprehensive (https://github.com/KhronosGroup/glslang/pull/4398)

known_good.json
```diff
@@ -5,7 +5,7 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit": "f589ef005c49f6f19c8e78eb5269104ba293beb4"
+      "commit": "ecea80f176e17a7afbea8c626b5adc6b5a4cbefb"
```
spencer-lunarg, line 8: @r-potter this is the part your PR is missing, this will ensure your tests are ran through the new `spirv-val` in CI

r-potter: Does that commit include the SPIR-Tools change I suggested? If not, CI will fail due to layout issues (We should probably have this conversation on the other PR)

spencer-lunarg: didn't see that, no it is not, let me get that spirv-val change in ASAP

spencer-lunarg: https://github.com/KhronosGroup/SPIRV-Tools/pull/6859

arcady-lunarg:
I'm going to close this in favor of #4398.

PR #4398, opened 2026-09-01
Improve EXT_abort support

### LINKED ISSUE #4369: OpAbortKHR must use an explicit layout

r-potter:
@spencer-lunarg 

closes https://github.com/KhronosGroup/glslang/issues/4369

* Addresses lack of explicitly laid out types. 
* Adds coverage for all SPIR-V concrete types

Will need a SPIR-Tools change. In discussion with @Tobski, we concluded that the message type parameter should always be scalar layout, rather than conditional. 

I think this is what you need on the SPIRV-Tools side:
```
    } else if (inst->opcode() == spv::Op::OpAbortKHR) {
      reference->type_id = inst->GetOperandAs<uint32_t>(0u);
      // Abort messages doesn't have a storage class
      reference->storage_class = spv::StorageClass::Max;
      // The abort message is always packed with scalar rules, independently of whether
      // the scalarBlockLayout feature is enabled for the rest of the module.
      // https://gitlab.khronos.org/spirv/SPIR-V/-/work_items/962
      reference->layout = LayoutMode::kScalar;
      reference->requirement = LayoutRequirement::kRequired;
      return true;
    }
```


r-potter:
@spencer-lunarg Pushed a commit with your SPIRV-Tools hash but the Khronos SPIRV-Tools repo. It'll fail CI until your PR lands, but **should** be good after

known_good.json
```diff
@@ -5,7 +5,7 @@
       "site" : "github",
       "subrepo" : "KhronosGroup/SPIRV-Tools",
       "subdir" : "External/spirv-tools",
-      "commit": "f589ef005c49f6f19c8e78eb5269104ba293beb4"
+      "commit": "0f34a8ce16db7c0293b4760a05ed7e6086abf94e"
```
r-potter:
```suggestion
      "commit": "0db14571ad2dccdbcbec4f05e57a820e3a15c815"
```

Updated to match SPIRV-Tools hash

Test/spv.shader.abort.type.float16.comp
```diff
@@ -0,0 +1,24 @@
...
+
+layout(local_size_x = 1) in;
+
+layout(set = 0, binding = 0) buffer Data {
+    float seed;
+} data;
+
+// 16-bit floats, scalar, vector and matrix. The matrices are built from explicit
+// columns because f16mat2(scalar) and friends currently miscompile, emitting a
+// 32-bit zero for the off-diagonal entries.
```
r-potter, line 14: This is fixed in https://github.com/KhronosGroup/glslang/pull/4405, so the test can subsequently be simplified

spencer-lunarg (approved):

lordalcol:
This fixes many of the issues I'm seeing when bumping SPIR-V and glslang in the CTS

dneto0:
This works and passes all gtests when combined with top-of-tree SPIRV-Tools.  Thanks!

glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4128,6 +4128,15 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
...
 
+    case EOpAbortEXT:
+        // GL_EXT_abort allows abortEXT() with no arguments at all. Otherwise the message
+        // must be a string literal, not merely an expression of string type: isLiteral()
+        // is false for e.g. ("msg"). createAbortEXT() assumes this.
+        if (arg0 != nullptr) {
+            const TIntermConstantUnion* message = arg0->getAsConstantUnion();
+            if (message == nullptr || message->getBasicType() != EbtString || ! message->isLiteral())
+                error(loc, "first argument must be a string literal", fnCandidate.getName().c_str(), "");
+        }
```
dneto0, line 4139: The most recent commit was a merge, and it accidentally dropped the `break;` fortunately GCC caught it.

dneto0 (changes requested):

arcady-lunarg (approved):

lordalcol:
This breaks against a newer spirv-tools check: ["Validate ArrayStride on OpConstantDataKHR result types" (KhronosGroup/SPIRV-Tools#6870)](https://github.com/KhronosGroup/SPIRV-Tools/pull/6870). `msgArrType` here is both the message string's `OpConstantDataKHR` result type and (via `structLoadMemberType`) the abort message struct's member type, decorated with `ArrayStride`. `SPV_KHR_constant_data` requires `OpConstantDataKHR`'s result type to carry no explicit layout, since its data is tightly packed - so decorating it with `ArrayStride` is invalid the moment that check lands (confirmed locally: fails validation against spirv-tools 14521db9).

The spec's own example actually uses two distinct types for exactly this reason - `%string1_t` (the constant-data result type, no layout) vs `%string1_x` (the message struct member, with `ArrayStride`). Suggest splitting them back out:

```suggestion
    auto msgArrType = builder.makeArrayType(charType, constLen, 1);
    auto msgLoadArrType = builder.makeArrayType(charType, constLen, 1);
```

and further down:

```suggestion
    builder.addDecoration(msgArrType, spv::Decoration::UTFEncodedKHR);
    builder.addDecoration(msgLoadArrType, spv::Decoration::UTFEncodedKHR);
    // Array stride for char is 1 byte per element for explicit layout. This goes on the
    // load-type array only: OpConstantDataKHR's result type (msgArrType) must not have one.
    builder.addDecoration(msgLoadArrType, spv::Decoration::ArrayStride, 1);
    // 2.5 Collect data and type for construct an internal message structure member.
    structMemberType.push_back(msgArrType);
    structLoadMemberType.push_back(msgLoadArrType);
```

`createCompositeConstruct`'s existing array-mismatch handling copies the constant across into the differently-typed struct member without any further changes needed.

Verified locally: passes `vk-gl-cts`'s `postmortem.device_fault.*` tests against both the old and new spirv-tools. Full fix on a branch here if useful: https://github.com/lordalcol/glslang/commit/3c2fa78160b50ed9b084ddcc3012158c1b0edce1

r-potter:
@lordalcol That should be resolved by https://github.com/KhronosGroup/glslang/pull/4412

PR #4404, opened 2026-09-02
printf requires that the first parameter be a string literal

r-potter:
@arcady-lunarg @spencer-lunarg @jeffbolznv @rdb Spinning off a question on the EXT_abort GLSL MR. I believe this is what debugPrintfEXT was specced to do, and I plan to make abortEXT consistent if we agree this is the correct approach.


spencer-lunarg (approved):
Since `NonSemantic.DebugPrintf` only takes a single `OpString` anything else here is nonsense IMO

and these all look like things I would hope are caught (since you can't produce meaningful SPIR-V of it)

jeffbolznv (approved):
LGTM. Minor nitpick: maybe move the positive test case to the positive test.

r-potter:
> Since `NonSemantic.DebugPrintf` only takes a single `OpString` anything else here is nonsense IMO
> 
> and these all look like things I would hope are caught (since you can't produce meaningful SPIR-V of it)

I believe they are not currently caught. I wrote a simple shader that did `debugPrintfEXT(1)` and glslang just took it and generated a SPIR-V shader that passed that into the SPIR-V Op as the first parameter without complaint i.e. I suspect there is also a spirv-val hole here.

r-potter:
> LGTM. Minor nitpick: maybe move the positive test case to the positive test.

Good catch. I somehow missed the positive tests

spencer-lunarg:
> I believe they are not currently caught. I wrote a simple shader that did debugPrintfEXT(1) and glslang just took it and generated a SPIR-V shader that passed that into the SPIR-V Op as the first parameter without complaint i.e. I suspect there is also a spirv-val hole here.

tested it, https://godbolt.org/z/Mnseqa1WT, you are right... will make a `spirv-val` fix today then

arcady-lunarg (commented):
LGTM modulo the moving of the positive test to the positive test file.

spencer-lunarg:
@r-potter for completeness https://github.com/KhronosGroup/SPIRV-Tools/pull/6861

jeffbolznv (approved):

arcady-lunarg (approved):

PR #4406, opened 2026-09-03
Fix missing ArrayStride on OpAbortKHR message string

lordalcol:
createAbortEXT() decorates the message struct's Offset members but
never gives the string array member an ArrayStride, so any shader
using abortEXT()/OpAbortKHR fails SPIR-V validation with:

    Array must be explicitly laid out with ArrayStride or
    ArrayStrideIdEXT decorations.

Per the SPV_KHR_abort registry spec's own example, ArrayStride goes
on the array type itself (OpDecorate, not OpMemberDecorate). Add it
to msgLoadArrType, matching that example exactly.


r-potter:
@lordalcol Could you try https://github.com/KhronosGroup/glslang/pull/4398 instead? It's much more comprehensive

lordalcol:
Superseded by #4398, which already includes this exact fix (ArrayStride on the message array type in createAbortEXT) as part of a broader rewrite covering matrices, structs and arrays. Closing in favor of that PR.

PR #4411, opened 2026-09-08
Implement DebugEntryPoint instruction in NSDI 102

qingyuanzNV:
Add support for emitting DebugEntryPoint instructions from NonSemantic.Shader.DebugInfo, allowing generated SPIR-V to record information needed to reproduce a compilation. This allows tooling to recover compilation environment of the SPIR-V and re-compile updated shader.

In glslang standalone, we produce "glslang" compiler signature and copy the exact compiler CLI arguments. For programmatic invocation, caller should provide these parameters.


arcady-lunarg:
@qingyuanzNV Could you write a test for this new functionality? I think the best thing would be a `runtests` test that runs the standalone compiler with nonsemantic debuginfo enabled and produces a SPIR-V output which is then compared against a known good output.

arcady-lunarg (approved):

PR #4412, opened 2026-09-08
Give the abort message separate value and laid out types

r-potter:
Give the abort message separate value and laid out types
The message string's array type was doing two contradictory jobs at once:
it was the result type of the OpConstantDataKHR holding the string, and it
was a member of the explicitly laid out abort message struct.  The first
role forbids ArrayStride, since the constant's data literals are tightly
packed, and the second requires it.  spirv-val now rejects the former, so
every abort test failed validation.

Split the string into a tightly packed type for the constant and a laid out
type for the struct member, and split the message struct the same way: one
struct typed after the values it is built from, one explicitly laid out to
describe where each member sits.  OpAbortKHR only requires its message
operand to match the message type operand logically, which disregards
decorations, so the value needs no conversion to reach the laid out form.

That also removes the OpCopyLogical instructions the aggregate arguments
used to need, since the constructed value now matches its own type exactly.


known_good.json
```diff
@@ -3,16 +3,16 @@
     {
       "name" : "spirv-tools",
       "site" : "github",
-      "subrepo" : "KhronosGroup/SPIRV-Tools",
+      "subrepo" : "r-potter/SPIRV-Tools",
```
r-potter:
```suggestion
      "subrepo" : "KhronosGroup/SPIRV-Tools",
```

r-potter:
@spencer-lunarg @arcady-lunarg @lordalcol This addresses the fact that ArrayStride cannot appear on constant data arrays, which requires splitting the types used for the OpAbortKHR message type parameter and the type of the payload struct.

~~Will require: https://github.com/KhronosGroup/SPIRV-Tools/pull/6870 to land first~~ Removed the spirv-val dependency.

arcady-lunarg (approved):
This looks like it's undoing one of the changes your chatbot made earlier, which I was somewhat suspicious about but figured you knew what you were doing.

Issue #4030, opened 2025-09-02
Crash with the `--no-link` option


glebov-andrey:
Hi,

I'm trying to compile two parts of a single shader separately with the `--no-link` option, and then link them together using `spirv-link`, but `glslang` crashes with a null pointer dereference (or emits an error, depending on the presence of `main()`).

**UPD:** *I made a mistake when testing the first compiler invocation (`main.frag`) which suppressed the error message. If needed, I'll create a separate issue for this, but it probably makes sense to figure out if this usage pattern is even valid first.*

The general idea is to prepare a compiled (but not linked) shader offline, with one function missing. Then at runtime, generate and compile that missing function (based on a material definition), and link the SPIR-V files together. `--no-link` seemed like the correct option for the job.

Or perhaps there's a different way to achieve such a result?

Here's the code:
```glsl
// interface.glsl
#ifndef INTERFACE_GLSL
#define INTERFACE_GLSL
vec4 do_some_math(vec4 src); // implemented in main.frag
vec4 calculate_color(vec4 arg); // implemented in impl.glsl
#endif  // INTERFACE_GLSL
```

```glsl
// main.frag
#extension GL_GOOGLE_include_directive : require
#include "interface.glsl"
layout(location = 0) in vec4 frag_arg;
layout(location = 0) out vec4 out_color;
vec4 do_some_math(vec4 arg) {
    return arg + 1.0;
}
void main() {
    out_color = calculate_color(frag_arg);
}
```

```glsl
// impl.glsl
#extension GL_GOOGLE_include_directive : require
#include "interface.glsl"
vec4 calculate_color(vec4 arg) {
    return clamp(do_some_math(arg), 0.0, 2.0) * 0.5;
}
```

This invocation fails with `ERROR: main.frag:14: '' : main function cannot be exported`:
```shell
glslang -o main.frag.spv --glsl-version 460 --target-env vulkan1.3 --no-link -S frag main.frag
```

This invocation crashes:
```shell
glslang -o impl.glsl.spv --glsl-version 460 --target-env vulkan1.3 --no-link -S frag impl.glsl
```

The crash stack trace (from 3289b1d61b69a6c66c4b7cd2c6d3ab2a6df031e5):
```
[glslang.exe] spv::Instruction::getOpCode() spvIR.h:217
[glslang.exe] spv::Module::getStorageClass(unsigned int) spvIR.h:577 // idToInstruction[typeId] is NULL
[...]

PR #4414, opened 2026-09-09
Don't crash on a call to an undefined function under --no-link

Chang-Jin-Lee:
Fixes the segfault in #4030.

A call to a function the unit only declared reaches SPIR-V generation with no callee. `handleUserFunctionCall()` returns 0, `EOpFunctionCall` reports that and stops, and the enclosing expression then loads through whatever the previous traversal left behind. For a bare `f(x)` that is an empty access chain, so `accessChainLoad()` asks the module for the storage class of id 0 and dereferences a null instruction, which matches the trace in the issue.

`checkCallGraphBodies()` already catches a missing body, but it runs from `finalCheck()` during link, and `--no-link` (`EOptionCompileOnly`) never links — StandAlone skips `program.link()`. Compile-only is the one mode where nothing reports it, so code generation keeps going and reaches the null.

**Rewritten after review: this now fails the compile instead of generating code for the call.** The first version set the access chain to an `OpUndef`. @arcady-lunarg said that just deletes the call, and checking the tree confirmed it — the caller came out with `LinkageAttributes "calculate_color" Export` and the callee with nothing at all, so the module exported a value computed from undef and `spirv-link` had no symbol to resolve.

There is also no way to write that call correctly. `LinkageType::Import` appears once in the whole repository, in `spirv.hpp11`'s enum-to-string table, and nothing under `SPIRV/` ever emits it; the one site that decorates a function, `GlslangToSpv.cpp:7684`, goes through `convertGlslangLinkageToSpv()`, which maps `ELinkExport` and sends everything else to `LinkageType::Max`. No glslang-produced module can name a function defined in another module, so no later stage could resolve such a call and rejecting it takes nothing away.

The check now runs at the end of parse, through the parse context, so the compile fails and no module is written. `findUndefinedCallees()` walks the call graph and skips any callee defined in the same unit, which keeps a prototype resolved further down from firing. `spv.exportFunctions.comp` has no calls at all and would not have caught that, so `spv.noLinkForwardDeclaration.frag` covers it; its baseline keeps the `OpFunctionCall` and carries no `OpUndef`.

One change had to come with it. `gtests/TestFixture.h` handed a failed compile-only compile to `GlslangToSpv` unconditionally, where the linked path directly above returns early. Left alone, the new diagnostic is emitted and then ignored, and the test still segfaults.

`glslang --no-link -S frag impl.glsl` goes from exit 139 to `ERROR: 0:16: 'do_some_math(vf4;' : no function definition (body) found:` with no SPIR-V. 2062 of 2062 tests pass and no other baseline changed.

The reporter's underlying request is separate and still unmet: the calling side would need an `Import` declaration for their `spirv-link` plan to work, and whether `--no-link` should emit one is a design question they raised themselves.


arcady-lunarg (changes requested):
I think this is fundamentally the wrong approach to the problem here. While not crashing is good, it also just deletes the function call entirely. I think it would probably be better to error out entirely in this case, as I am pretty sure this isn't even valid Vulkan GLSL and there's no real mechanism in GLSL to support imported functions yet.

Chang-Jin-Lee:
You're right, and it's worse than deleting the call. I compiled the report's shader with the old patch and looked at what came out: the caller got `LinkageAttributes "calculate_color" Export`, and the callee got nothing at all — no declaration, no `Import`, no `OpFunctionCall`. `spirv-link` has no symbol to resolve there, so the module just exports a value computed from undef.

That matches what the tree says. `TLinkType` is `ELinkNone` or `ELinkExport`, and `convertGlslangLinkageToSpv()` sends everything that isn't Export to `LinkageType::Max`. There is no way to name the callee, so this can only be an error.

Rewritten to do that. `checkCallGraphBodies()` already produces "No function definition (body) found", but it runs from `finalCheck()` during link, and a compile-only unit never gets there — StandAlone skips `program.link()` under `--no-link`. So the check is reported at the end of parse instead, through the parse context, which fails the compile and writes no module. The codegen change is gone.

One thing that had to come with it: `gtests/TestFixture.h` passed a failed compile-only compile to `GlslangToSpv` unconditionally, where the linked path right above it returns early. With that left alone the new diagnostic is emitted and then ignored, and the test still segfaults.

Before: exit 139. After: `ERROR: 0:16: 'do_some_math(vf4;' : no function definition (body) found:` and no SPIR-V. 2061 of 2061 tests pass.

Chang-Jin-Lee:
Went and checked the "no mechanism for imported functions" part properly, since the whole rewrite rests on it.

`LinkageType::Import` occurs exactly once in the tree, in `spirv.hpp11`'s enum-to-string table. Nothing in `SPIRV/` ever emits it. The one call site that decorates a function is `GlslangToSpv.cpp:7684`, and it goes through `convertGlslangLinkageToSpv()`, which maps `ELinkExport` and sends everything else to `LinkageType::Max`. So no glslang-produced module can reference a function defined in a different module — not just this shader, any of them.

Which also settles what the issue was asking for. Compiling the definition side gives `LinkageAttributes "do_some_math" Export`, so `spirv-link` has an export to offer; the calling side has no way to state that it wants one. That workflow was never going to work, so failing the compile is not taking something away.

One gap in what I sent yesterday, now fixed. The check reports callees with no body in the unit, and nothing guarded the case it must *not* fire on — a prototype defined further down the same unit. `spv.exportFunctions.comp` has no calls at all. Added `spv.noLinkForwardDeclaration.frag`, whose baseline keeps the `OpFunctionCall` and carries no `OpUndef`. 2062 of 2062 pass.

Issue #1314, opened 2018-03-27
Unknown crash in glslang when compiling my hlsl shader


mmostajab:
I am having a crash in glslangValidator when compiling 
[TerrainVS.hlsl.txt](https://github.com/KhronosGroup/glslang/files/1850314/TerrainVS.hlsl.txt) (Please rename it to TerrainVS.hlsl). The command line I use to compile the shader is as follows:

```
.\glslangValidator.exe -D -fhlsl_functionality1 .\TerrainVS.hlsl -o TerrainVS.spv --target-env vulkan1.0 -S vert -e TerrainVS -V100
```

johnkslang:
It's triggered by using the `m._m03_m13_m23` syntax to extract a `float3` out of `float4x4`.  It can handle that syntax for individual components or full columns, but not odder subsets than those.

Instead, you can use `float3(m.[0][3], m.[1][3], m.[2][3])` or whatever the correct equivalent really is.

(Glslang cannot do that itself, because the expression could turn into an l-value, but the constructor can only make an r-value. So, glslang needs full support for an "l-value subset of matrix" operator.)

There is an existing issue for it, but good to know also a new case of bumping into it.

Glslang tries to report the missing functionality, but something downstream crashes it before the message comes out. That could perhaps be somewhat improved, especially by modifying the `SpvBuildLogger`, which you might be able to customize for your environment.

johnkslang:
See also #1240 and #844.

Chang-Jin-Lee:
The crash in the title no longer reproduces on main. What happens instead, with

```hlsl
float4x4 m;

float3 main() : SV_Target
{
    return m._m03_m13_m23;
}
```

is that glslang prints `Missing functionality: matrix swizzle`, exits 0, and still writes a module that `spirv-val` rejects:

```
error: line 20: OpStore Pointer <id> '27[%_entryPointOutput]'s type does not match Object <id> '18[%int_3]'s type.
  OpStore %_entryPointOutput %int_3
```

`%int_3` is the swizzle's last index. `EOpMatrixSwizzle` logs the message and returns `true`, so the traverser goes on to visit the index list and one of those constants is left in the access chain as the value of the swizzle. `Test/baseResults/hlsl.matrixSwizzle.vert.out` has the same thing recorded in it as `Store 42(f3) 34`, an `int` constant stored into a `float3`.

#4416 keeps the emitted module well formed. Matrix swizzle itself is still unimplemented, so the feature request in this issue stands.

PR #4415, opened 2026-09-09
Only encode a real bool when storing to a storage bool

### LINKED ISSUE #4086: Copying a struct with a bool produces invalid select statement
https://godbolt.org/z/srbW3T8Tc

```glsl
#version 460
#extension GL_EXT_shader_explicit_arithmetic_types_int64 : require
#extension GL_EXT_buffer_reference2 : require

struct S1 { bool x; };

layout(set = 0, binding = 1) restrict readonly buffer SSBO {
	S1 s1;
} InBuffer;

layout(buffer_reference) buffer MyBuffer {
	S1 s1;
};

void main() {
	MyBuffer payload;
	payload.s1 = InBuffer.s1;
}
```

generates invalid SPIR-V

```
Expected bool scalar or vector type as condition: Select
  %29 = OpSelect %uint %28 %uint_1 %uint_0
```

Chang-Jin-Lee:
Fixes #4086.

`accessChainStore()` encodes a bool into its storage form with `OpSelect(rvalue, 1u, 0u)`, using `rvalue` as the condition. A member-wise struct copy extracts the source member in its storage form, so `rvalue` already has the nominal integer type and the emitted instruction is

```
%31 = OpSelect %uint %27 %uint_1 %uint_0
```

which `spirv-val` rejects with "Expected bool scalar or vector type as condition: Select".

The `else if` arm in the same function already guards on the operand type (`getTypeId(rvalue) != boolType`); the `if` arm did not. This gives the `if` arm the same guard, in both the scalar and the vector branch — a struct with a `bool` member emitted `Select %uint`, one with a `bvec3` member emitted `Select %ivec3`. For the repro the store becomes a plain copy of two storage bools of the same width.

It only reproduces below Vulkan 1.2: from SPIR-V 1.4 the struct copy goes through `OpCopyLogical` and never reaches this conversion. On the test shader, `vulkan1.0` and `vulkan1.1` are invalid before the change and valid after; `vulkan1.2` and `vulkan1.3` are valid either way.

`Test/spv.bufferReferenceStorageBool.comp` copies a bool-member struct and a bvec3-member struct from an SSBO into a `buffer_reference`, registered on `CompileVulkanToSpirvTest` because that fixture targets SPIR-V 1.0. Without the change its baseline carries `Validation failed` and the two `Select` instructions. Full suite green, and no other baseline moved — `accessChainStore()` is on the store path for every shader in the suite, and this only skips a conversion when the value was already converted.


arcady-lunarg:
@Chang-Jin-Lee Thanks for your interest in fixing outstanding issues. I would like to note however that you talking to a human here and copy-pasting walls of text that you haven't written and may not have even read from a chatbot is not the most effective or polite way to communicate. Have you reviewed the code generated by your chatbot in detail and are you sure it is a minimal fix? The same caveat about walls of text applies to comments in the code, because I find the Claude style of exceedingly verbose comments with EMPHASIS to be grating, and also counterproductive in terms of their purpose of communicating author intent to future readers of the code.

Chang-Jin-Lee:
Fair. The descriptions were too long, and I'll keep them short from here.

On minimality: in both the scalar and the vector branch, the `else if` arm already guarded on the rvalue's type (`getTypeId(rvalue) != boolType`). The `if` arm did not, and that is where the invalid `OpSelect %uint %uint` came from — a member-wise struct copy hands `accessChainStore` a value that is already in the nominal type. The change gives the `if` arm the same guard the `else if` arm has, so the two are symmetric. Nothing is restructured, no existing conversion path changes, and the 2060 tests already in the suite are untouched; the one added test is the reproducer from #4086.

I did consider folding both guards into a single check at the top of the bool block. It reads a little better but re-indents the whole block for no semantic gain, so I left it alone.

Comments trimmed here and on #4414 and #4416.

KarenGhavam-lunarG:
> Have you reviewed the code generated by your chatbot in detail and are you sure it is a minimal fix? 

Also wanted to make sure you have seen this in the CONTRIBUTING.MD about AI generated PRs:

" By submitting a Contribution to this repository, you additionally represent that, to the extent any of Your Contributions were developed with the assistance of artificial intelligence tools or AI-generated code, You have exercised sufficient review, judgment, and creative direction over such tools and resulting material to reasonably consider it Your original creation, and You are not aware of any third-party license, intellectual property claim, or other restriction arising from such use that is associated with any part of Your Contribution or use thereof."

Chang-Jin-Lee:
I've read it. That representation holds for all three PRs.

They were written with AI assistance, on my machine and under my direction, and I didn't take any of them on trust. For each one I reverted just the changed source file and re-ran the tests to watch them fail first: without the fix, #4030's gtest exits 139, spirv-val rejects #4086's module under vulkan1.0 and 1.1, and #1314 stores an int constant into a float3 in two separate baselines. None of the diffs is copied from another project, and I know of no third-party license or IP claim on any of them.

If you'd still rather not take them on those terms, say so and I'll close all three.

Chang-Jin-Lee:
Worth flagging so the empty check list is not read as a failure: CI last ran green on this branch on 09-09 at 13:43, and the push at 18:09 went to `action_required`. The current head has not been tested, and the runs need approving before anything shows up here.

PR #4416, opened 2026-09-09
Keep the module valid when a matrix swizzle is unimplemented

Chang-Jin-Lee:
Refs #1314. The crash that issue reports no longer reproduces, but the module glslang writes for that shader is invalid.

`EOpMatrixSwizzle` in `TGlslangToSpvTraverser::visitBinary` logs missing functionality and returns `true`, which asks the traverser to visit the node's children. The right child is the swizzle's index list, so its last integer constant stays in the access chain and becomes the value of the swizzle expression. For

```hlsl
float4x4 m;

float3 main() : SV_Target
{
    return m._m03_m13_m23;
}
```

glslang exits 0, reports `Missing functionality: matrix swizzle`, and writes a module that `spirv-val` rejects:

```
error: line 20: OpStore Pointer <id> '27[%_entryPointOutput]'s type does not match Object <id> '18[%int_3]'s type.
  OpStore %_entryPointOutput %int_3
```

This traverses only the operand, so side effects in it are still generated, then clears the access chain and hands back an `OpUndef` of the swizzle's own type — the same shape as the `EOpFunctionCall` fallback further down.

`Test/baseResults/hlsl.matrixSwizzle.vert.out` already had the defect recorded in it as `Store 42(f3) 34`, where 34 is `25(int) Constant 0` and `f3` is a `float3`. The baseline never caught it because `outputResultToStream` decides its `Validation failed` line from `logger.getAllMessages()` rather than from validating the module. That baseline now reads `79: 12(fvec3) Undef` followed by `Store 42(f3) 79`.

`Test/hlsl.matrixSwizzle.frag` is new and covers the shape from the issue, where the bad value reaches the entry point's return rather than a local. Reverting only `SPIRV/GlslangToSpv.cpp` turns both baselines red; with the change the suite is 2061 of 2061. The rest of the vert baseline diff is id renumbering from the one added instruction.

Implementing HLSL matrix swizzle is still out of scope. The missing-functionality message is unchanged; only the emitted module differs.


Chang-Jin-Lee:
Worth separating this from #4414 before it reads the same way. There the input is invalid — a call to a function with no definition and no way to import one — so erroring out is the only honest outcome, and that PR now does.

Matrix swizzle is the opposite case. `hlslParseHelper.cpp:1038` builds `EOpMatrixSwizzle` from HLSL the front end accepts, and `Test/hlsl.matrixSwizzle.vert` compiles today and writes a module, logging only `Missing functionality: matrix swizzle`. Making it an error would reject HLSL that glslang deliberately parses.

So this keeps to something narrower: the module that already gets written stops storing an `int` constant into a `float3`. The message and the exit status are untouched. If you would rather the missing-functionality path be fatal in general, that is a larger call than this PR and I will close it.

Chang-Jin-Lee:
Worth flagging so the empty check list is not read as a failure: every workflow run on this branch is sitting at `action_required`, so CI has not executed on this PR at all yet. It needs approving before any result appears here.

arcady-lunarg (changes requested):
If the goal here is to keep the module valid, why do the baseline test results still have a "Validation failed" message?

Chang-Jin-Lee:
That line is not a `spirv-val` result. `gtests/TestFixture.h:335` derives it from the logger instead:

```cpp
bool validation_result = !options().validate || logger.getAllMessages().empty();
```

and line 419 prints `Validation failed` whenever that is false. So any logger message at all produces it, and this shader still produces one — `Missing functionality: matrix swizzle` — because the PR does not implement the swizzle. It only stops the traverser from leaving the index list's last constant in the access chain, so the module it writes is well typed rather than correct.

That is also why the baseline never caught the original defect: `Store 42(f3) 34` with 34 an `int` constant was already recorded in `hlsl.matrixSwizzle.vert.out`, and the harness reported the same `Validation failed` line for the missing-functionality note rather than for the bad store.

The line would only clear once matrix swizzle is actually implemented and nothing is logged. If it is worth separating the two meanings, having the harness run `spirv-val` rather than reading the logger looks like its own change, and I would rather not fold it into this one.

Chang-Jin-Lee:
Following up on my own answer, because the PR's CI was red for exactly this and I had not looked.

The workflow runs

```
grep -l 'Validation failed' Test/baseResults/* | sort -fd | diff -u Test/baseResults/validation_fails.txt -
```

so the baselines carrying that line are tracked in a checked-in manifest. `hlsl.matrixSwizzle.frag.out` is new here and was never added to it, and all four linux jobs failed on the diff. Pushed the one line; regenerating the manifest with that same command produces exactly it and nothing else, and `ctest` is 2/2 locally.

That manifest is also a more concrete answer than the one I gave. glslang already keeps an explicit list of which baselines are expected to log something, and this shader belongs on it for as long as matrix swizzle is unimplemented, since the line is the `Missing functionality` note surfacing through the logger rather than a `spirv-val` verdict. It clears when the swizzle is implemented, not when this PR lands.

PR #4423, opened 2026-09-11
Fix xfb size computation losing 32-bit-ness through arrays

### LINKED ISSUE #4314: Duplicated `contains16BitType` passed into function, instead correct `contains16BitType` var
Issue found by AI, validated by me - full report can be found here - [https://github.com/qarmin/ClaudeReports/blob/master/glslang_20260622.html](https://github.com/qarmin/ClaudeReports/blob/master/glslang_20260622.html)

https://github.com/KhronosGroup/glslang/blob/20960a4872f681e4213312b06b48cc4ddae3c73d/glslang/MachineIndependent/linkValidate.cpp#L2334-L2339

**Description:** In computeTypeXfbSize the recursive call for a sized array passes the wrong out-parameters: it passes contains16BitType for both the 32-bit and 16-bit tracking arguments (the second-to-last argument should be contains32BitType). As a result an array element that contains a 32-bit type never propagates contains32BitType up to the array; instead contains16BitType is written twice. This corrupts xfb stride alignment computation (the 4-byte vs 2-byte rounding in finalCheck depends on these flags), and can lead to wrong xfb_stride alignment / multiple-of checks for arrays of structs/scalars.

**Locations:**

```
glslang/MachineIndependent/linkValidate.cpp:2334-2339
 2334 |     if (type.isSizedArray()) {
 2335 |         // TODO: perf: this can be flattened by using getCumulativeArraySize(), and a deref that discards all arrayness
 2336 |         // Unsized array use to xfb should be a compile error.
 2337 |         TType elementType(type, 0);
 2338 |         return type.getOuterArraySize() * computeTypeXfbSize(elementType, contains64BitType, contains16BitType, contains16BitType);
 2339 |     }
```

**Fix:** Change the call to computeTypeXfbSize(elementType, contains64BitType, contains32BitType, contains16BitType).

Wint3rNight:
Fixes #4314.

`computeTypeXfbSize` passes `contains16BitType` where it means `contains32BitType` when recursing into array element types, so a 32-bit type reached through an array never sets the 32-bit flag.

That flag is read in two places, so the typo has two effects. The stride is checked against the 16-bit rule instead of the 32-bit one, which lets an invalid `xfb_stride` link silently. It also makes `RoundToPow2` round the implicit stride to a multiple of 2 instead of 4, which changes the emitted `XfbStride`: a valid shader that states no stride of its own compiles to a stride of 6 where it should be 8.

Tests cover invalid strides through arrays for 16-bit, 32-bit and 64-bit components, plus the implicit-stride case. Both fail without the fix.


arcady-lunarg (changes requested):
Thanks, the fix itself looks good. Could you also add cases to the unit test for invalid strides for 16-bit and 64-bit components? Also this is really just fixing an obvious typo, I think you can simplify your commit description to explain the fix as such.

Wint3rNight:
> Thanks, the fix itself looks good. Could you also add cases to the unit test for invalid strides for 16-bit and 64-bit components? Also this is really just fixing an obvious typo, I think you can simplify your commit description to explain the fix as such.

Both done, thanks.

arcady-lunarg (approved):

PR #4425, opened 2026-09-12
validate saturatedConvertEXT argument count

saddamr3e:
ASAN, `saturatedConvertEXT(x)` with one argument:

    ERROR: AddressSanitizer: SEGV on unknown address
    #0 glslang::TParseContext::builtInOpCheck(...) ParseHelper.cpp:4099

`saturatedConvertEXT` (GL_EXT_float_e4m3 / GL_EXT_float_e5m2) has an empty prototype, so overload resolution accepts any argument list. The `EOpConstructSaturated` case reads `sequence[0]`/`sequence[1]` after only warning on `size() != 2` without returning, so a one-argument call reads one element past a size-1 aggregate and dereferences a garbage `TIntermNode*`. A zero-argument call folds to a constant that is not an operator, so it skips `builtInOpCheck` and reaches `result->getAsAggregate()->getSequence()[0]` in `handleFunctionCall` on an absent aggregate, then asserts in `createSpvConstantFromConstUnionArray` during SPIR-V generation.

Validate the count in both spots before indexing: `builtInOpCheck` null-checks the aggregate and requires two operands, and `handleFunctionCall` only takes the result type from operand 0 when two operands are present, erroring otherwise so no SPIR-V is generated. Added Test/spv.floate4m3.saturatedConvert.error.comp.

### DIFF
--- a/glslang/MachineIndependent/ParseHelper.cpp
+++ b/glslang/MachineIndependent/ParseHelper.cpp
@@ -1648,7 +1648,13 @@ TIntermTyped* TParseContext::handleFunctionCall(const TSourceLoc& loc, TFunction
 
             if (fnCandidate->getBuiltInOp() == EOpConstructSaturated) {
                 // result type is taken from the first parameter
-                result->setType(result->getAsAggregate()->getSequence()[0]->getAsTyped()->getType());
+                TIntermAggregate* aggregate = result->getAsAggregate();
+                if (aggregate != nullptr && aggregate->getSequence().size() == 2)
+                    result->setType(aggregate->getSequence()[0]->getAsTyped()->getType());
+                else if (aggregate == nullptr)
+                    // a call with the wrong argument count folds away from an aggregate and
+                    // never reaches builtInOpCheck; reject it here before SPIR-V generation
+                    error(loc, "requires exactly two parameters", "", "");
             }
         }
     }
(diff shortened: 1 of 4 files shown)


glslang/MachineIndependent/ParseHelper.cpp
```diff
@@ -4091,10 +4026,12 @@ void TParseContext::builtInOpCheck(const TSourceLoc& loc, const TFunction& fnCan
 
     case EOpConstructSaturated:
     {
-        auto &sequence = callNode.getAsAggregate()->getSequence();
-        if (sequence.size() != 2) {
+        const TIntermAggregate* aggregate = callNode.getAsAggregate();
+        if (aggregate == nullptr || aggregate->getSequence().size() != 2) {
             error(loc, "requires exactly two parameters", "", "");
```
arcady-lunarg, line 4031: This is a pretty common pattern of bug in glslang unfortunately, because in normal operation `error` is treated as an abort but it doesn't always work like that, especially given "cascading errors" mode.

arcady-lunarg (approved):
LGTM, and I'm going to merge this as is. I think the better thing would have been to have the prototypes checked in the normal way, but that results in a huge combinatorial explosion of prototypes that would need to be added and that is probably more prone to bugs. What we would really need here is some kind of type generics even if only for internal use now that there are potentially 8 different floating point types but that's well outside the scope of a small bugfix.

PR #4426, opened 2026-09-12
Mangle buffer reference types by block name

### LINKED ISSUE #3997: No compilation error on function parameter type mismatch for buffer device addresses, obtuse validation layer error
After refactoring my push constant BDAs from, roughly:

```C
(PushConstants) { 
    .previousFrame = previousFrameDeviceAddr, 
    .currentFrame = currentFrameDeviceAddr 
};
```
to
```C
(GameStorageBuffer) { VkDeviceAddress simFrames[2], ... };

(PushConstants) { 
    .GAME = gameDeviceAddr,
    .PREVIOUS_FRAME_INDEX = ...,
    .CURRENT_FRAME_INDEX = ...
};
```

`vkCreateShaderModule` validation of an affected shader gives the following validation error:

```glsl
Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x55f76945d6e0, name = Device : NVIDIA GeForce GTX 1650, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: OpStore Pointer <id> '1424[%1424]'s type does not match Object <id> '1425[%1425]'s type.
  OpStore %param_240 %1425
```

I can't yet rule out an error in my refactoring, but at worst this looks like compiler bug, and at best an unhelpful error.

I've included the GLSL and generated SPV below (compiled via `glslang -H -t -DGAME_CONSTS_HEADER='"[redacted]/Consts.h"' --target-env vulkan1.2 [redacted]/PolyGrid.vert -o [redacted]/PolyGrid.vert.spv`)

[Gist containing GLSL and SPV](https://gist.github.com/floopfloopfloopfloopfloop/e9ea6213cdc431b3f5e564b3fdc6a25d)

Wint3rNight:
buildMangledName has no EbtReference case, so every buffer_reference type mangles to the same string. Overload resolution can't tell two reference types apart, so passing a BlockB where a BlockA is declared compiles with no diagnostic and emits SPIR-V that spirv-val rejects with `OpStore Pointer's type does not match Object's type`. It also means you can't overload a function on two different reference types at all, you just get `function already has a body`.

I mangled by block name rather than structurally. The parser rejects a second block with the same name, so the name already identifies the type. Doing it structurally like the EbtStruct case wouldn't terminate: a reference block can legally name itself or another block that points back at it, which a struct can't do, and that's the only reason the struct case can recurse into members safely.

Two new tests. bufferhandle27 covers both overloads, a reference reached through a struct member, and a self-referential block. bufferhandle28_Errors is the miscompile from the issue. Both fail without the change.

Six existing baselines move. It's all OpName strings for functions taking reference parameters, plus the id column padding shifting where the name got longer. No instructions changed.

Fixes #3997


arcady-lunarg (approved):
LGTM. I think the comment is a bit verbose for my taste but I think it's reasonable to have it and that's more a note for future PRs, I'll merge this as is.

Wint3rNight:
Thanks for merging, and sorry about the long comment. I'll keep them shorter from now on.

PR #4430, opened 2026-09-15
Check error message formats with the printf format attribute

mattst88:
error(), warn(), ppError() and ppWarn() take a printf format, but the
compiler never checked it. That let an int64_t index printed with '%d'
go unnoticed until runtests failed on 32-bit ARM and PowerPC (#4180,
fixed in b44e6091).

This adds the format attribute so GCC and Clang catch mismatches like
that at build time. Reverting b44e6091 on top of this branch makes GCC
warn at all nine call sites it fixed.

The first two commits clean up the call sites the attribute would
otherwise warn about:

 - Runtime strings (identifiers, extension names, type strings) passed
   directly as the format now go through "%s". Clang's
   -Wformat-security flags these. glslang_tab.cpp is regenerated with
   Bison 3.8.2.
 - A few calls passed an unused "" after an empty format.

Many callers pass "" as the format when there is nothing to add, which
GCC's -Wformat-zero-length warns about at over 1100 call sites. The
last commit turns that warning off in CMake and in BUILD.gn for GCC
builds, rather than rewriting every call as "%s", "".

The library builds with no warnings under GCC and Clang.


glslang/MachineIndependent/Versions.cpp
```diff
@@ -944,7 +944,7 @@ void TParseVersions::requireNotRemoved(const TSourceLoc& loc, int profileMask, i
             const int maxSize = 60;
             char buf[maxSize];
             snprintf(buf, maxSize, "%s profile; removed in version %d", ProfileName(profile), removedVersion);
-            error(loc, "no longer supported in", featureDesc, buf);
+            error(loc, "no longer supported in", featureDesc, "%s", buf);
```
arcady-lunarg: You can combine this with the snprintf above and get rid of `buf` and `maxSize` entirely.

arcady-lunarg (changes requested):
Thanks for putting this in. Just one minor extra fix you can do since since you're already touching the code.

glslang/MachineIndependent/Versions.cpp
mattst88: Thanks! Done.

arcady-lunarg (approved):

PR #4432, opened 2026-09-17
Don't emit OpUConvert for spec constants before SPIR-V 1.4

### LINKED ISSUE #3506: OpUConvert for specialization constant should not generate spirv-opt error.
When I compile the following program:

```
#version 450
#extension GL_EXT_shader_explicit_arithmetic_types_int16 : require

layout (constant_id = 0) const uint local_size_x = 8;
layout (constant_id = 1) const uint local_size_y = 8;
layout (local_size_x_id = 0, local_size_y_id = 1, local_size_z = 1) in;

void main() {
    u16vec2 tid = u16vec2(gl_LocalInvocationID.xy);
    uint16_t idx = tid.y * uint16_t(gl_WorkGroupSize.x) + tid.x;
}
```

As follows:

```
glslangValidator.exe -V100 -e main repro.comp.glsl -o spirv/repro.comp.spv
spirv-opt.exe -O spirv/repro.comp.spv -o spirv/repro.comp.opt.spv
```

spirv-opt produces the following error:

```
error: line 38: Prior to SPIR-V 1.4, specialization constant operation UConvert requires Kernel capability or extension SPV_AMD_gpu_shader_int16
  %28 = OpSpecConstantOp %ushort UConvert %27
```

According to https://github.com/KhronosGroup/glslang/issues/848 the spec has been updated to include OpUConvert in the list of permitted opcodes. Is this a spirv-opt error?

I'm using the most recent builds that are publicly available:
  
```
$ glslangValidator.exe --version
Glslang Version: 11:14.0.0
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 14.0.0
GLSL Version: 4.60 glslang Khronos. 14.0.0
SPIR-V Version 0x00010600, Revision 1
Khronos Tool ID 8
SPIR-V Generator Version 11
ARB_GL_gl_spirv version 100

$ spirv-opt.exe --version
SPIRV-Tools v2023.6 v2023.6.rc1-0-gf0cc85ef
```

Wint3rNight:
OpUConvert only became a valid specialization constant operation in SPIR-V 1.4, so converting a specialization constant to a different integer width, for example uint64_t(su), produced a module that fails validation on SPIR-V 1.0 through 1.3 targets. Below 1.4 this emits OpSConvert instead, which is valid in every version, and masks off the sign-extended bits when widening so the result is unchanged.

SPV_AMD_gpu_shader_int16 also legalizes the op, but it is added during post-processing, after code generation, so it is not used to gate this.

Two existing baselines change, spv.int64.frag and spv.int16.amd.frag, both SPIR-V 1.0 modules that were emitting the op.

Fixes #3506


Test/baseResults/spv.int64.frag.out
```diff
@@ -2,7 +2,7 @@ spv.int64.frag
 Validation failed
```
arcady-lunarg, line 2: Still a validation fail but for a different reason, in case you are looking for more things to fix.

arcady-lunarg (approved):

Test/baseResults/spv.int64.frag.out
Wint3rNight: Will fix it separately.

PR #4435, opened 2026-09-19
Reject constant_id on a specialization constant computed from others

### LINKED ISSUE #3748: Expressions in specialization constants
Both glslang and glslangValidator compile the following code without any diagnostics
```glsl
#version 450 core

layout(constant_id = 0) const int baz = 1;
layout(constant_id = 1) const int buz = baz + 1;

void main () { }
```

And produce
```
; SPIR-V
; Version: 1.6
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 10
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpSource GLSL 450
               OpName %main "main"
               OpName %baz "baz"
               OpName %buz "buz"
               OpDecorate %baz SpecId 0
               OpDecorate %buz SpecId 1
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
        %int = OpTypeInt 32 1
        %baz = OpSpecConstant %int 1
      %int_1 = OpConstant %int 1
        %buz = OpSpecConstantOp %int IAdd %baz %int_1
          %5 = OpLabel
               OpReturn
               OpFunctionEnd
```

but spirv-val produces the following error:
```
error: line 10: SpecId decoration on target <id> '9[%buz]' must be a scalar specialization constant
  OpDecorate %buz SpecId 1
```

The spec (2.12) does seem to indicate that spirv-val is correct and glslanValidator is producing incorrect code.

Wint3rNight:
A specialization constant whose initializer is an expression over other specialization constants is emitted as an `OpSpecConstantOp`, which SPIR-V does not allow a `SpecId` decoration on, so the module fails validation.

`Test/spv.int64.frag` declared one and its baseline recorded `Validation failed`. The declaration keeps its initializer and loses the `constant_id`, since a derived specialization constant cannot be independently specialized anyway.

Fixes #3748


arcady-lunarg (approved):
I think this is a reasonable interpretation of the GLSL spec, and the fix looks good.

PR #4437, opened 2026-09-19
fix null deref on non-symbol l-value base in l/rValueErrorCheck

saddamr3e:
ASan/UBSan, reading a member of a `writeonly` buffer_reference through pointer arithmetic (`int v = (root.r + 1).a;`):

    ParseContextBase.cpp:247:65: runtime error: member call on null pointer of type 'const glslang::TIntermSymbol'
    AddressSanitizer: SEGV on unknown address 0x000000000000
      #0 TParseContextBase::rValueErrorCheck(...) ParseContextBase.cpp:247
      #1 TParseContext::declareVariable(...)       ParseHelper.cpp:9976

`rValueErrorCheck` (and the sibling `lValueErrorCheck`) name the offending base with `leftMostTypeNode->getAsSymbolNode()->getName()`, where `leftMostTypeNode = traverseLValueBase(node, true)`. `traverseLValueBase` returns null as soon as an inner op in the access chain is not an index/swizzle (buffer_reference math bottoms out on an `EOpAdd`, not a symbol), and it can also return a non-symbol node, so `getAsSymbolNode()` is null and `getName()` derefs it.

Gate the symbol-name branch on `leftMostTypeNode` resolving to a `TIntermSymbol` at both sites; a non-symbol or null base falls through to the generic diagnostic that is already there. The other two `traverseLValueBase` callers in ParseHelper.cpp already null-check the result before use.


arcady-lunarg (approved):
Nice as it would be to have the diagnostic include the name of the writeonly/readonly type, not crashing is more important.

